PR #3, opened 2026-02-17
refactor: extract shared bitcoind startup and data dir logic into common module

erickcestari:


erickcestari:
> Concept ACK, but perhaps we pause this refactor until all 4 LN implementations are supported. I suspect we will need some special config for Eclair.

Agree!

morehouse:
@erickcestari: All 4 LN implementations are now supported if you're interested in updating this PR.

PR #4, opened 2026-03-16
Select active scenario via cargo feature flags

NishantBansal2003:
These commits add support for building the target binaries with different fuzzing scenarios selected at compile time. This allows us to add more scenarios while keeping the bin the same for all targets. So, the new commits use feature flags to build a specific scenario, for example `encrypted-bytes, noise`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -36,11 +36,17 @@ jobs:
       - name: Run Clippy
         run: cargo clippy --all-targets --all-features -- -D warnings
 
-      - name: Build
-        run: cargo build --verbose
+      - name: Build encrypted-bytes scenario
+        run: cargo build --verbose --features encrypted-bytes
 
-      - name: Build with nyx feature
-        run: cargo build --verbose --features nyx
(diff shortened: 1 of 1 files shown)


morehouse (changes requested):
NACK.  This change seems net-negative.

Pros:

- Fewer driver files (4 instead of N*M)
- Slight simplification to Dockerfiles.

Cons:

- Introduces mutually-exclusive features, which are [strongly discouraged](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification) in Rust.  We cannot enable both the `noise` and the `encrypted-bytes` features or we get a compile error (see CI failure).
- The mutually-exclusive features break the default rust-analyzer config, which runs with all features enabled.
- Hides the scenario that is built.  If all we have is a binary called "cln" or "ldk", we don't immediately know which scenario the binary will fuzz.


Yes, it would be nice to not have N*M driver files.  But they're trivial boilerplate (6 lines of code each) in a single location, while this change requires *other* boilerplate in different locations: 4 lines CI + 2 lines in scenarios.rs.

PR #6, opened 2026-03-21
smite: implement shutdown codec

erickcestari:


erickcestari:
I have addressed all the comments.

PR #7, opened 2026-03-21
smite: implement open channel codec

NishantBansal2003:
Added support for the `open_channel` BOLT message. Also refactored the code and introduced the `WireFormat` trait for serializing and deserializing BOLT messages for `u8`, `u16`, `u32`, `u64`, fixed-size byte arrays, and `secp256k1::PublicKey`. This wire format abstraction should make it easier to build additional messages in the future.

### DIFF
--- a/smite/src/bolt/open_channel.rs
+++ b/smite/src/bolt/open_channel.rs
@@ -1,0 +1,543 @@
+//! BOLT 2 open channel message.
+
+use super::BoltError;
+use super::tlv::TlvStream;
+use super::types::{CHAIN_HASH_SIZE, ChannelId};
+use super::wire::WireFormat;
+use secp256k1::PublicKey;
+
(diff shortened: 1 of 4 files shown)


smite/src/bolt/open_channel.rs
```diff
@@ -0,0 +1,545 @@
...
+use secp256k1::PublicKey;
+
+/// TLV type for upfront shutdown script.
+const TLV_UPFRONT_SHUTDOWN_SCRIPT: u64 = 0;
+
+/// TLV type for channel type.
+const TLV_CHANNEL_TYPE: u64 = 1;
+
+/// Size of a chain hash (SHA256).
+const CHAIN_HASH_SIZE: usize = 32;
```
morehouse: This constant is used for other messages too.  Perhaps we should move it to `types.rs`.

Issue #5, opened 2026-03-20
IR Design and Implementation Plan


morehouse:
### Problem

The current `EncryptedBytesScenario` encrypts and sends raw bytes as BOLT messages.  This mostly tests message parsing, never reaching deeper protocol states.

To fuzz deeper, we need the fuzzer to become structure aware, able to send and receive sequences of valid (or semi-valid) messages that get past the initial parsing code and exercise deeper protocol logic.  Ideally the fuzzer would be smart enough to generate the common message sequences to open channels, send or receive HTLCs, close channels, etc.  Many of these message sequences have dependencies between the messages -- e.g., `commitment_signed` must contain a channel ID that matches a previously-opened channel, as well as signatures generated from the previously negotiated keys and commitment states.  We want the fuzzer to be able to satisfy these dependencies as well.

### Solution

We can use an intermediate representation (IR) to capture the type and structure knowledge needed to fuzz deeper.  The fuzzer can then use this IR to generate and mutate short *programs* to be executed in the Nyx VM.

The following design is inspired by ideas from both [syzkaller](https://github.com/google/syzkaller) and [fuzzamoto](https://github.com/dergoegge/fuzzamoto).

## Architecture

We add a new crate `smite-ir/` to contain the IR, custom mutators, and program generators, and a new scenario `IrScenario<T, S>` under `smite-scenarios/` to execute IR programs in the Nyx VM.

![Image](https://github.com/user-attachments/assets/8f278252-b429-45eb-b32e-ee622eb2647a)

The `smite-ir` crate is intended to be *engine-agnostic* -- no dependency on AFL++ or LibAFL.  We intend to use AFL++ at first with a thin custom mutator wrapper library `libsmite_ir_mutator.so` loaded via `AFL_CUSTOM_MUTATOR_LIBRARY`.  In the future, we may migrate to LibAFL by simply replacing the wrapper.  See "Appendix: Fuzzing Engine Tradeoffs" for details.

Our custom mutators and generators create new `Program`s and serialize them for the fuzzing engine, which then sends each `Program` to the Nyx VM via shared memory.  Inside the VM, our `IrScenario` deserializes the `Program` and executes it line-by-line.  `IrScenario` then checks for and reports any crashes or hangs and resets the VM snapshot before processing the next `Program`.

## Example IR Program

This program executes the channel funding flow up to the point where the target sends `funding_signed`.  We will refer to this example in subsequent sections.

```
# Generate channel keys (6 key pairs)
[...]

PR #8, opened 2026-03-23
smite: implement gossip_timestamp_filter codec

ekzyis:
I saw https://github.com/morehouse/smite/issues/5#issuecomment-4101143436 and decided to add [`gossip_timestamp_filter`](https://github.com/lightning/bolts/blob/master/07-routing-gossip.md#the-gossip_timestamp_filter-message) as a simple message to get started.

Things worth mentioning:

* Added helper functions for u32 similar to the existing ones for u16. When #7 is merged, I can rebase this to use the new `WireFormat` trait.
* Moved `CHAIN_HASH_SIZE` and `BITCOIN_MAINNET` from init.rs to types.rs so I can reuse them in gossip_timestamp_filter.rs. This creates a [small conflict](https://github.com/morehouse/smite/pull/7/changes#diff-f45231f82f4b67291619ee291d90dd387b8a293a56596aaf3d2a7368ab989556R16) with #7.
* I'm new to rust, and I've let [Composer 1.5](https://cursor.com/docs/models/cursor-composer-1-5) help me write this code. I made sure to keep the changes consistent with the existing code, and to only make small improvements like the move to types.rs (instead of letting it run wild).

### DIFF
--- a/smite/src/bolt/gossip_timestamp_filter.rs
+++ b/smite/src/bolt/gossip_timestamp_filter.rs
@@ -1,0 +1,158 @@
+//! BOLT 7 `gossip_timestamp_filter` message.
+
+use super::BoltError;
+use super::types::CHAIN_HASH_SIZE;
+use super::wire::WireFormat;
+
+/// BOLT 7 `gossip_timestamp_filter` message (type 265).
+///
+/// Allows a node to constrain future gossip messages to a specific range.
+/// A node which wants any gossip messages has to send this, otherwise no
+/// gossip messages would be received.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct GossipTimestampFilter {
+    /// The 32-byte hash that uniquely identifies the chain for the gossip.
+    pub chain_hash: [u8; CHAIN_HASH_SIZE],
+    /// Lower bound (inclusive) of the timestamp range.
+    pub first_timestamp: u32,
+    /// Number of seconds in the range (exclusive upper bound).
+    pub timestamp_range: u32,
+}
+
+impl GossipTimestampFilter {
+    /// Creates a gossip timestamp filter.
+    #[must_use]
+    pub fn new(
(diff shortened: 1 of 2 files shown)


smite/src/bolt/gossip_timestamp_filter.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// Returns `Truncated` if the payload is too short.
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        const FILTER_SIZE: usize = CHAIN_HASH_SIZE + 4 + 4;
+        if payload.len() < FILTER_SIZE {
+            return Err(BoltError::Truncated {
+                expected: FILTER_SIZE,
+                actual: payload.len(),
+            });
+        }
+        let mut cursor = payload;
```
morehouse: For consistency with other codecs, let's move the cursor definition to the top of `decode` and then only reference `cursor` in the rest of the function.

smite/src/bolt/gossip_timestamp_filter.rs
```diff
@@ -0,0 +1,137 @@
...
+        let filter = GossipTimestampFilter::no_gossip(BITCOIN_MAINNET);
+        let encoded = filter.encode();
+        assert_eq!(encoded.len(), 40);
+        assert_eq!(encoded[..CHAIN_HASH_SIZE], BITCOIN_MAINNET);
+        assert_eq!(encoded[32..36], u32::MAX.to_be_bytes());
+        assert_eq!(encoded[36..40], 0u32.to_be_bytes());
+    }
+
+    #[test]
+    fn decode_truncated() {
```
morehouse: We should also add tests for messages that are truncated into the `first_timestamp` and `timestamp_range` fields.

ekzyis:
Addressed all feedback and rebased on master

morehouse (approved):
LGTM, thanks

Issue #10, opened 2026-03-25
Building afl++ with nyx mode on NixOS


ekzyis:
AFL++ needs to be built with [nyx mode](https://github.com/AFLplusplus/AFLplusplus/tree/e5a8ba39ecf97d05e286fdd4e01da96554dbf64f/nyx_mode) to follow the quick start guide in the README and run [setup-nyx.sh](https://github.com/morehouse/smite/blob/813c2415cb7e9f9da196340087b3b36aea735ea8/scripts/setup-nyx.sh).

I'm on NixOS. Unfortunately, [aflplusplus](https://github.com/NixOS/nixpkgs/blob/nixos-25.11/pkgs/tools/security/aflplusplus/default.nix) in nixpkgs does not include nyx mode.

The [instructions](https://github.com/AFLplusplus/AFLplusplus/tree/e5a8ba39ecf97d05e286fdd4e01da96554dbf64f/nyx_mode#building-nyx-mode) to build AFL++ with nyx mode run [build_nyx_support.sh](https://github.com/AFLplusplus/AFLplusplus/blob/e5a8ba39ecf97d05e286fdd4e01da96554dbf64f/nyx_mode/build_nyx_support.sh), which calls nyx-fuzz/packer/init_ramfs/pack.sh, which [expects shared libraries to be at specific locations](https://github.com/nyx-fuzz/packer/blob/7967b432c1994059a02013c9bc8fdd22230df48a/linux_initramfs/pack.sh#L42-L48).

I'm creating this issue to mention that I'm working on this. Maybe there are (or will be) others on NixOS who run into this.

ekzyis:
I cleaned up my nix derivation and created a nixpkgs PR, see https://github.com/NixOS/nixpkgs/pull/514406.

To use my derivation to build AFL++ with nyx mode:

```
$ git clone --depth=1 --branch=aflplusplus-nyx-mode git@github.com:ekzyis/nixpkgs
$ cd nixpkgs/pkgs/tools/security/aflplusplus
$ nix-build -E "with import <nixpkgs> {}; callPackage ./default.nix { withNyx = true; wine = null; }"
```

This will print a /nix/store path. Before you can use it with setup-nyx.sh, you need to apply this patch:

```diff
diff --git a/scripts/setup-nyx.sh b/scripts/setup-nyx.sh
index 7f8c5ef..a14a422 100755
--- a/scripts/setup-nyx.sh
+++ b/scripts/setup-nyx.sh
[...]

PR #13, opened 2026-03-26
smite-scenarios: deconstruct `connect_to_target`

morehouse:
Rename `connect_to_target` to `handshake_with_target`, and no longer echo back the received `init` message.  This change has two motivations:

- Enabling future scenarios to customize the `init` message sent to the target.
- Enabling the future `IrScenario` to save the target's `init` features in the `ProgramContext`.

### DIFF
--- a/smite-scenarios/src/scenarios.rs
+++ b/smite-scenarios/src/scenarios.rs
@@ -28,16 +28,19 @@ const EPHEMERAL_KEY: [u8; 32] = [
     0x12, 0x12, 0x12, 0x12, 0x12, 0x12, 0x12, 0x12, 0x12, 0x12, 0x12, 0x12, 0x12, 0x12, 0x12, 0x12,
 ];
 
-/// Connect to a target and perform the init handshake.
+/// Perform a Noise handshake with a target and receive its `Init` message.
+///
+/// Returns the encrypted connection and the target's `Init`. The caller is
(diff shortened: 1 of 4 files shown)


erickcestari (changes requested):
It's looking good for me. There is only a minor enhancement to cut down on some lines of code.

ekzyis (approved):
Couldn't run this in the fuzzer because of #10, so I only ran the tests and reviewed the code.

nit: could make sense to keep `connect_to_target` but make it call `handshake_with_target` internally. This would avoid repeating this where `connect_to_target` was used previously:

```rust
let echo = Message::Init(Init::echo(&target_init)).encode();
conn.send_message(&echo).map_err(|e| e.to_string())?;
```

But I think this would conflict with this motivation if all scenarios are meant to do this via `IrScenario` in the future, including the existing ones:

> - Enabling the future `IrScenario` to save the target's `init` features in the `ProgramContext`.

---

Unrelated to this change: I was curious about the decision to have `ScenarioError` but not use it in the function signature of `Scenario::new` for errors (`String` is used instead). Looks to me like `ScenarioError` is meant to be reserved for functions in smite-scenarios/src. Is that right? (The term "scenario" seems to be used interchangeably, making it a bit confusing. I don't have a suggestion to improve this though.)

PR #18, opened 2026-04-01
smite-ir: implement `OpenChannelGenerator`

morehouse:
Adds the functionality needed to generate IR programs that exercise the `open_channel -> accept_channel` flow.

Key pieces:
- `ProgramBuilder` is the toolkit for building type-correct programs.
- `Generator` is the trait defining the generator interface.
- `OpenChannelGenerator` generates IR programs that build and sends `open_channel` messages, then wait for the `accept_channel` reply.

Ref: #5 (Milestone 1)

### DIFF
--- a/smite-ir/src/builder.rs
+++ b/smite-ir/src/builder.rs
@@ -44,7 +44,35 @@ impl ProgramBuilder {
 
     /// Appends an instruction and registers its output variable (if any) as a
     /// candidate. Returns the instruction index.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the inputs have the wrong count, reference out-of-bounds or
+    /// void instructions, or have mismatched types.
     pub fn append(&mut self, operation: Operation, inputs: &[usize]) -> usize {
+        let expected = operation.input_types();
+        assert_eq!(
+            inputs.len(),
+            expected.len(),
+            "{operation:?}: expected {} inputs, got {}",
+            expected.len(),
+            inputs.len(),
+        );
+        let len = self.instructions.len();
+        for (i, (&input_idx, &expected_type)) in inputs.iter().zip(expected.iter()).enumerate() {
+            let actual_type = self
+                .instructions
+                .get(input_idx)
+                .and_then(|instr| instr.operation.output_type())
+                .unwrap_or_else(|| {
+                    panic!("{operation:?} input {i}: index {input_idx} out of bounds ({len})")
+                });
+            assert_eq!(
+                actual_type, expected_type,
+                "{operation:?} input {i}: expected {expected_type:?}, got {actual_type:?}",
+            );
+        }
+
         let idx = self.instructions.len();
 
         if let Some(out_type) = operation.output_type() {
--- a/smite-ir/src/builder.rs
+++ b/smite-ir/src/builder.rs
@@ -1,0 +1,171 @@
+//! Program builder for generators.
+//!
+//! `ProgramBuilder` maintains an instruction list and a type-indexed variable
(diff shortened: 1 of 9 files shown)


devvaansh (commented):
took a deep look at the IR Builder and Operation logic, have a few suggestions regarding determinism for reproducibility and some potential hot-path optimizations for fuzzing throughput.

ekzyis (approved):
lgtm, just one question

erickcestari (approved):
LGTM!

devvaansh:
Also LGTM👍

harsh04044 (approved):
The separation of generate_fresh() for keys vs pick_variable() for scalar parameters is a clean design. For the interactive-tx generator, serial_ids have a similar constraint -- parity is protocol-mandated (even for initiator, odd for non-initiator), so they'd need to be generated fresh with explicit parity enforcement rather than picked from the candidate pool.

Is the intent that generators are responsible for enforcing protocol constraints like this, or is there a planned mechanism in ProgramBuilder to support constrained variable generation?

smite-ir/src/generators.rs
```diff
@@ -0,0 +1,20 @@
...
+pub use open_channel::OpenChannelGenerator;
+
+use rand::Rng;
+
+use super::builder::ProgramBuilder;
+
+/// A generator that emits instructions into a `ProgramBuilder`.
+pub trait Generator {
+    /// Emits instructions for this generator's protocol interaction.
+    fn generate(&self, builder: &mut ProgramBuilder, rng: &mut impl Rng);
```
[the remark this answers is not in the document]
morehouse: I think it makes sense to have `rng` in the `generate` interface.  Currently generators don't use it directly, but in the future they probably will.  For example, an interactive-tx generator may randomly choose what transaction inputs and outputs to construct, or randomly arrange the `tx_add_*` and `tx_remove_*` messages to send.

morehouse:
> The separation of generate_fresh() for keys vs pick_variable() for scalar parameters is a clean design. For the interactive-tx generator, serial_ids have a similar constraint -- parity is protocol-mandated (even for initiator, odd for non-initiator), so they'd need to be generated fresh with explicit parity enforcement rather than picked from the candidate pool.
> 
> Is the intent that generators are responsible for enforcing protocol constraints like this, or is there a planned mechanism in ProgramBuilder to support constrained variable generation?

I would lean towards managing those kinds of constraints in the generators.  If we have a recurring need for the same functionality we could consider moving it to `ProgramBuilder`.

smite-ir/src/builder.rs
```diff
@@ -0,0 +1,199 @@
...
+            inputs: inputs.to_vec(),
+        });
+
+        idx
+    }
+
+    /// Selects or creates a variable of the given type using probabilistic
+    /// variable selection (75% most recent, 15% any existing, 10% fresh).
+    #[allow(clippy::missing_panics_doc)] // candidates is always non-empty
+    pub fn pick_variable(&mut self, var_type: VariableType, rng: &mut impl Rng) -> usize {
```
Chand-ra, line 107: Nit: I think a `TODO` comment explaining that we'd like to replace these hardcoded probabilities with an adaptive mutation scheduler (like `MOpt`) in the future would make it easy to circle back here in the future.

Other than this, LGTM.

morehouse: Maybe.  This is something I want data on to support the effectiveness, rather than just doing it.  Most fuzzers just do hardcoded probabilities (e.g., libFuzzer, syzkaller) and work fine.

I think this experiment would be better described in an issue than a TODO.

PR #23, opened 2026-04-03
bolt: implement tx_add_input message codec

harsh04044:
Implements the `tx_add_input` message codec (type 0x0042) for BOLT 2 interactive transaction construction.

The `prevtx` field carries a full consensus-encoded Bitcoin transaction, decoded using `bitcoin::consensus::deserialize` and length-prefixed on the wire via the `WireFormat` trait.

Includes encode/decode, `Message` enum integration, and tests covering roundtrip, truncation, and trailing bytes.

Part of #5

### DIFF
--- a/smite/src/bolt.rs
+++ b/smite/src/bolt.rs
@@ -20,6 +20,7 @@ mod shutdown;
 mod tlv;
 mod tx_abort;
 mod tx_ack_rbf;
+mod tx_add_input;
 mod tx_complete;
 mod tx_init_rbf;
 mod tx_remove_input;
@@ -48,6 +49,7 @@ pub use shutdown::Shutdown;
 pub use tlv::{TlvRecord, TlvStream};
 pub use tx_abort::TxAbort;
 pub use tx_ack_rbf::{TxAckRbf, TxAckRbfTlvs};
+pub use tx_add_input::{TxAddInput, TxAddInputTlvs};
 pub use tx_complete::TxComplete;
 pub use tx_init_rbf::{TxInitRbf, TxInitRbfTlvs};
 pub use tx_remove_input::TxRemoveInput;
@@ -127,6 +129,8 @@ pub mod msg_type {
     pub const OPEN_CHANNEL2: u16 = 64;
     /// `accept_channel2` message (BOLT 2).
     pub const ACCEPT_CHANNEL2: u16 = 65;
+    /// `tx_add_input` message (BOLT 2).
+    pub const TX_ADD_INPUT: u16 = 66;
     /// `tx_remove_input` message (BOLT 2).
     pub const TX_REMOVE_INPUT: u16 = 68;
     /// `tx_remove_output` message (BOLT 2).
@@ -179,6 +183,8 @@ pub enum Message {
     OpenChannel2(OpenChannel2),
     /// `accept_channel2` message (type 65).
     AcceptChannel2(AcceptChannel2),
+    /// `tx_add_input` message (type 66).
+    TxAddInput(TxAddInput),
     /// `tx_remove_input` message (type 68).
     TxRemoveInput(TxRemoveInput),
     /// `tx_remove_output` message (type 69).
@@ -229,6 +235,7 @@ impl Message {
             Self::Shutdown(_) => msg_type::SHUTDOWN,
             Self::OpenChannel2(_) => msg_type::OPEN_CHANNEL2,
             Self::AcceptChannel2(_) => msg_type::ACCEPT_CHANNEL2,
+            Self::TxAddInput(_) => msg_type::TX_ADD_INPUT,
             Self::TxRemoveInput(_) => msg_type::TX_REMOVE_INPUT,
             Self::TxRemoveOutput(_) => msg_type::TX_REMOVE_OUTPUT,
             Self::TxComplete(_) => msg_type::TX_COMPLETE,
@@ -262,6 +269,7 @@ impl Message {
             Self::Shutdown(m) => out.extend(m.encode()),
             Self::OpenChannel2(m) => out.extend(m.encode()),
             Self::AcceptChannel2(m) => out.extend(m.encode()),
+            Self::TxAddInput(m) => out.extend(m.encode()),
             Self::TxRemoveInput(m) => out.extend(m.encode()),
             Self::TxRemoveOutput(m) => out.extend(m.encode()),
             Self::TxComplete(m) => out.extend(m.encode()),
@@ -302,6 +310,7 @@ impl Message {
             msg_type::SHUTDOWN => Ok(Self::Shutdown(Shutdown::decode(cursor)?)),
             msg_type::OPEN_CHANNEL2 => Ok(Self::OpenChannel2(OpenChannel2::decode(cursor)?)),
             msg_type::ACCEPT_CHANNEL2 => Ok(Self::AcceptChannel2(AcceptChannel2::decode(cursor)?)),
+            msg_type::TX_ADD_INPUT => Ok(Self::TxAddInput(TxAddInput::decode(cursor)?)),
             msg_type::TX_REMOVE_INPUT => Ok(Self::TxRemoveInput(TxRemoveInput::decode(cursor)?)),
             msg_type::TX_REMOVE_OUTPUT => Ok(Self::TxRemoveOutput(TxRemoveOutput::decode(cursor)?)),
             msg_type::TX_COMPLETE => Ok(Self::TxComplete(TxComplete::decode(cursor)?)),
@@ -630,6 +639,27 @@ mod tests {
         assert_eq!(decoded, Message::AcceptChannel2(accept2));
     }
 
+    /// Valid `TxAddInput` message for testing.
+    fn sample_tx_add_input() -> TxAddInput {
+        TxAddInput {
+            channel_id: ChannelId::new([0xab; CHANNEL_ID_SIZE]),
+            serial_id: 42,
+            prevtx: vec![0xde, 0xad, 0xbe, 0xef],
+            prevtx_vout: 0,
+            sequence: 0xffff_fffd,
+            tlvs: TxAddInputTlvs::default(),
+        }
+    }
+
+    #[test]
+    fn message_tx_add_input_roundtrip() {
+        let tx_add_input = sample_tx_add_input();
+        let msg = Message::TxAddInput(tx_add_input.clone());
+        let encoded = msg.encode();
+        let decoded = Message::decode(&encoded).unwrap();
+        assert_eq!(decoded, Message::TxAddInput(tx_add_input));
+    }
+
     #[test]
     fn message_tx_remove_input_roundtrip() {
         let tx_remove_input = TxRemoveInput {
@@ -760,8 +790,8 @@ mod tests {
         assert_eq!(decoded, msg);
     }
 
-    #[test]
     #[allow(clippy::too_many_lines)]
+    #[test]
     fn message_type_values() {
         assert_eq!(
             Message::Warning(Warning::all_channels("")).msg_type(),
@@ -805,6 +835,10 @@ mod tests {
         assert_eq!(
             Message::AcceptChannel2(sample_accept_channel2(None)).msg_type(),
             msg_type::ACCEPT_CHANNEL2
+        );
+        assert_eq!(
+            Message::TxAddInput(sample_tx_add_input()).msg_type(),
+            msg_type::TX_ADD_INPUT
         );
         assert_eq!(
             Message::TxRemoveInput(TxRemoveInput {
--- a/smite/src/bolt/tx_add_input.rs
+++ b/smite/src/bolt/tx_add_input.rs
@@ -1,0 +1,263 @@
+//! BOLT 2 `tx_add_input` message.
+
+use secp256k1::hashes::Hash;
+
+use super::BoltError;
+use super::tlv::TlvStream;
+use super::types::{ChannelId, Txid};
+use super::wire::WireFormat;
+
+/// BOLT 2 `tx_add_input` message (type 66).
+///
+/// Sent during interactive transaction construction to propose adding an
+/// input to the shared transaction.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TxAddInput {
+    /// The channel this message pertains to.
+    pub channel_id: ChannelId,
+    /// Serial ID for this input. Must be even if sent by the initiator,
+    /// odd if sent by the non-initiator (BOLT 2 parity rule).
+    pub serial_id: u64,
+    /// Previous transaction being spent (consensus-encoded bytes).
+    pub prevtx: Vec<u8>,
+    /// The output index within `prevtx` being spent.
+    pub prevtx_vout: u32,
+    /// The sequence number for this input.
+    pub sequence: u32,
+    /// Optional TLV extensions.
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -19,5 +19,6 @@ warnings = "deny"
 log = "0.4"
 simple_logger = { version = "5", default-features = false }
 postcard = { version = "1.1", default-features = false, features = ["alloc"] }
+bitcoin = { version = "0.32", default-features = false, features = ["std"] }
```
ekzyis: I think you forgot to push the changes to Cargo.lock

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,138 @@
...
+
+impl TxAddInput {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.serial_id.write(&mut out);
+        let tx_bytes = bitcoin::consensus::serialize(&self.prevtx);
+        tx_bytes.write(&mut out);
```
ekzyis: I think we'll want serialization of `Transaction` to be implemented in `WireFormat`, so this can just be

```rust
self.prevtx.write(&mut out)
```

like the other properties

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Returns `Truncated` if the payload is too short, or
+    /// `InvalidTransaction` if `prevtx` is not a valid consensus-encoded
+    /// Bitcoin transaction.
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        let mut cursor = payload;
+        let channel_id = WireFormat::read(&mut cursor)?;
+        let serial_id = WireFormat::read(&mut cursor)?;
+        let tx_bytes: Vec<u8> = WireFormat::read(&mut cursor)?;
+        let prevtx = bitcoin::consensus::deserialize(&tx_bytes)
+            .map_err(|_| BoltError::InvalidTransaction)?;
```
ekzyis: same here

smite/src/bolt.rs
```diff
@@ -118,6 +124,8 @@ pub mod msg_type {
     pub const FUNDING_SIGNED: u16 = 35;
     /// Shutdown message (BOLT 2).
     pub const SHUTDOWN: u16 = 38;
+    /// `tx_add_input` message (BOLT 2).
+    pub const TX_ADD_INPUT: u16 = 0x0042;
```
ekzyis:
```suggestion
    pub const TX_ADD_INPUT: u16 = 66;
```

smite/src/bolt.rs
```diff
@@ -154,6 +162,8 @@ pub enum Message {
     FundingSigned(FundingSigned),
     /// Shutdown message (type 38).
     Shutdown(Shutdown),
+    /// `tx_add_input` message (type 0x0042).
```
ekzyis:
```suggestion
    /// `tx_add_input` message (type 66).
```

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,138 @@
+//! BOLT 2 `tx_add_input` message.
+
+use super::BoltError;
+use super::types::ChannelId;
+use super::wire::WireFormat;
+use bitcoin::Transaction;
+
+/// BOLT 2 `tx_add_input` message (type 0x0042).
```
ekzyis: Why use hex here when we use decimal everywhere else?

```suggestion
/// BOLT 2 `tx_add_input` message (type 66).
```

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn decode_ignores_trailing_bytes() {
+        let original = sample_msg();
+        let mut encoded = original.encode();
+        encoded.extend_from_slice(&[0xaa, 0xbb, 0xcc]);
+        let decoded = TxAddInput::decode(&encoded).unwrap();
+        assert_eq!(decoded, original);
+    }
+
+    #[test]
+    fn truncated_decode_returns_error() {
```
ekzyis: nit: to be consistent with naming of similar tests

```suggestion
    fn decode_truncated_channel_id() {
```

smite/src/bolt/tx_add_input.rs
ekzyis, line 1: nit: could add more tests for truncated properties

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,138 @@
...
+/// input to the shared transaction.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TxAddInput {
+    /// The channel this message pertains to.
+    pub channel_id: ChannelId,
+    /// Serial ID for this input. Must be even if sent by the initiator,
+    /// odd if sent by the non-initiator (BOLT 2 parity rule).
+    pub serial_id: u64,
+    /// Previous transaction being spent (consensus-encoded on the wire).
+    pub prevtx: Transaction,
```
morehouse: The dependency on `bitcoin` pulls a lot of new crates in.

At this point in the project I'd like to avoid that dependency -- let's make `prevtx` a `Vec<u8>` instead.  Later on, when we are working on the interactive-tx flow, we can revisit whether it makes sense to take this dependency or implement our own minimal `Transaction` functionality.

ekzyis (changes requested):

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// The output index within `prevtx` being spent.
+    pub prevtx_vout: u32,
+    /// The sequence number for this input.
+    pub sequence: u32,
+}
```
morehouse, line 29: We also should support the `tlvs` field.

smite/src/bolt/tx_add_input.rs
morehouse: Ideally we should test truncation at every field (one test per field) like we do with other codecs.

smite/src/bolt/tx_add_input.rs
ekzyis, line 29: This is missing the TLV stream. You can check how `accept_channel` implemented TLVs.

ekzyis (changes requested):

morehouse (changes requested):
Thanks for tackling this.  Main feedback:

- let's avoid depending on the `bitcoin` crate for now
- let's add support for the `tlvs` field
- please structure the tests similarly to the other codecs (`open_channel` is a good example)

smite/src/bolt/tx_add_input.rs
ekzyis: This won't be needed with `prevtx` as `Vec<u8>` as mentioned in https://github.com/morehouse/smite/pull/23#discussion_r3040481451

(for some reason I can't mark my comment as resolved)

harsh04044:
Updated: dropped the bitcoin crate dependency (prevtx is now Vec<u8>), added TxAddInputTlvs following the open_channel pattern, and restructured tests to match the other codecs with one test per truncated field.

@morehouse

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,183 @@
...
+    pub prevtx_vout: u32,
+    /// The sequence number for this input.
+    pub sequence: u32,
+    /// Optional TLV extensions.
+    pub tlvs: TxAddInputTlvs,
+}
+
+/// TLV extensions for the `tx_add_input` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct TxAddInputTlvs {}
```
morehouse: Please actually implement the TLVs rather than ignoring them.

morehouse (changes requested):

smite/src/bolt.rs
```diff
@@ -495,6 +516,15 @@ mod tests {
         assert_eq!(decoded, Message::ChannelReady(cr));
     }
 
+    #[test]
+    fn message_tx_add_input_roundtrip() {
```
morehouse: Tests should be ordered by the message type number.

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+/// TLV extensions for the `tx_add_input` message.
+const TLV_SHARED_INPUT_TXID: u64 = 0;
+
+/// TLV extensions for the `tx_add_input` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct TxAddInputTlvs {
+    /// Optionally specifies a shared input transaction ID.
+    pub shared_input_txid: Option<[u8; 32]>,
```
morehouse: We should use the `Txid` type for `shared_input_txid`.

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl TxAddInput {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.serial_id.write(&mut out);
+        self.prevtx.write(&mut out);
```
morehouse, line 66: We need to assert that `prevtx` isn't more than `u16::MAX`.

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,208 @@
...
+        let mut cursor = payload;
+        let channel_id = WireFormat::read(&mut cursor)?;
+        let serial_id = WireFormat::read(&mut cursor)?;
+        let prevtx: Vec<u8> = WireFormat::read(&mut cursor)?;
+        let prevtx_vout = WireFormat::read(&mut cursor)?;
+        let sequence = WireFormat::read(&mut cursor)?;
+        let tlv_stream = TlvStream::decode_with_known(cursor, &[TLV_SHARED_INPUT_TXID])?;
+        let shared_input_txid = tlv_stream.get(TLV_SHARED_INPUT_TXID).map(|v| {
+            let mut arr = [0u8; 32];
+            arr.copy_from_slice(v);
```
morehouse: This panics if `v` has the wrong size.

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// The output index within `prevtx` being spent.
+    pub prevtx_vout: u32,
+    /// The sequence number for this input.
+    pub sequence: u32,
+    /// Optional TLV extensions.
+    pub tlvs: TxAddInputTlvs,
+}
+
+/// TLV extensions for the `tx_add_input` message.
+const TLV_SHARED_INPUT_TXID: u64 = 0;
```
morehouse:
```suggestion
/// TLV type for shared input txid.
const TLV_SHARED_INPUT_TXID: u64 = 0;
```

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,208 @@
...
+        let prevtx_field_size = 2 + sample_msg().prevtx.len();
+        let cutoff = CHANNEL_ID_SIZE + 8 + prevtx_field_size + 4 + 2;
+        assert_eq!(
+            TxAddInput::decode(&encoded[..cutoff]),
+            Err(BoltError::Truncated {
+                expected: 4,
+                actual: 2,
+            })
+        );
+    }
```
morehouse, line 222: Missing tests for `decode_empty`, unknown even TLVs, and wrong-length `shared_input_txid`.

morehouse (changes requested):

smite/src/bolt.rs
harsh04044: Done.

smite/src/bolt/tx_add_input.rs
harsh04044: Done, switched to Txid.

smite/src/bolt/tx_add_input.rs
harsh04044: Done, added the assert.

smite/src/bolt/tx_add_input.rs
harsh04044: Done.

smite/src/bolt/tx_add_input.rs
harsh04044: Added decode_empty, unknown even TLV rejection, and wrong-length shared_input_txid tests.

smite/src/bolt/tx_add_input.rs
harsh04044: Fixed, now validates length and returns BoltError::Truncated on mismatch.

smite/src/bolt/tx_add_input.rs
morehouse: Actually we no longer need this -- #49 moved the assert into `WireFormat`.

smite/src/bolt.rs
```diff
@@ -497,6 +506,18 @@ mod tests {
         }
     }
 
+    /// Valid `TxAddInput` message for testing.
+    fn sample_tx_add_input() -> TxAddInput {
```
morehouse: Let's move this helper function to directly before `message_tx_add_input_roundtrip`, for consistency.

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,267 @@
...
+                        expected: TXID_SIZE,
+                        actual: v.len(),
+                    });
+                }
+                let mut arr = [0u8; TXID_SIZE];
+                arr.copy_from_slice(v);
+                Ok(Txid::from_byte_array(arr))
+            })
+            .transpose()?;
+        let tlvs = TxAddInputTlvs { shared_input_txid };
```
morehouse: Nit: please follow the `from_stream` pattern used by other codecs:

```suggestion
        let tlvs = TxAddInputTlvs::from_stream(&tlv_stream);
```

smite/src/bolt/tx_add_input.rs
```diff
@@ -0,0 +1,267 @@
...
+            .map(|v| {
+                if v.len() != TXID_SIZE {
+                    return Err(BoltError::Truncated {
+                        expected: TXID_SIZE,
+                        actual: v.len(),
+                    });
+                }
+                let mut arr = [0u8; TXID_SIZE];
+                arr.copy_from_slice(v);
+                Ok(Txid::from_byte_array(arr))
```
morehouse: We can simplify this using `WireFormat`.


```suggestion
                let mut cursor = v;
                Txid::read(&mut cursor);
```

morehouse (commented):
Looks good once these comments are addressed.

smite/src/bolt.rs
harsh04044: Moved to directly before message_tx_add_input_roundtrip.

smite/src/bolt/tx_add_input.rs
harsh04044: Done.

smite/src/bolt/tx_add_input.rs
harsh04044: Done, using Txid::read now.

morehouse (approved):
LGTM

PR #24, opened 2026-04-03
smite: add commitment_signed message codec implementation

kartikangiras:
Add support for `commitment-signed` codec for BOLT 2 message.

### DIFF
--- a/smite/src/bolt.rs
+++ b/smite/src/bolt.rs
@@ -7,6 +7,7 @@ mod accept_channel;
 mod accept_channel2;
 mod attribution_data;
 mod channel_ready;
+mod commitment_signed;
 mod error;
 mod funding_created;
 mod funding_signed;
@@ -35,6 +36,7 @@ pub use accept_channel::{AcceptChannel, AcceptChannelTlvs};
 pub use accept_channel2::{AcceptChannel2, AcceptChannel2Tlvs};
 pub use attribution_data::{AttributionData, TruncatedHmac};
 pub use channel_ready::{ChannelReady, ChannelReadyTlvs};
+pub use commitment_signed::{CommitmentSigned, CommitmentSignedTlvs};
 pub use error::Error;
 pub use funding_created::FundingCreated;
 pub use funding_signed::FundingSigned;
@@ -143,6 +145,8 @@ pub mod msg_type {
     pub const UPDATE_FULFILL_HTLC: u16 = 130;
     /// `update_fail_htlc` message (BOLT 2).
     pub const UPDATE_FAIL_HTLC: u16 = 131;
+    /// `commitment_signed` message (BOLT 2).
+    pub const COMMITMENT_SIGNED: u16 = 132;
     /// `update_fail_malformed_htlc` message (BOLT 2).
     pub const UPDATE_FAIL_MALFORMED_HTLC: u16 = 135;
     /// Gossip timestamp filter message (BOLT 7).
@@ -195,6 +199,8 @@ pub enum Message {
     UpdateFulfillHtlc(UpdateFulfillHtlc),
     /// `update_fail_htlc` message (type 131).
     UpdateFailHtlc(UpdateFailHtlc),
+    /// `commitment_signed` message (type 132).
+    CommitmentSigned(CommitmentSigned),
     /// `update_fail_malformed_htlc` message (type 135).
     UpdateFailMalformedHtlc(UpdateFailMalformedHtlc),
     /// Gossip timestamp filter message (type 265).
@@ -237,6 +243,7 @@ impl Message {
             Self::TxAbort(_) => msg_type::TX_ABORT,
             Self::UpdateFulfillHtlc(_) => msg_type::UPDATE_FULFILL_HTLC,
             Self::UpdateFailHtlc(_) => msg_type::UPDATE_FAIL_HTLC,
+            Self::CommitmentSigned(_) => msg_type::COMMITMENT_SIGNED,
             Self::UpdateFailMalformedHtlc(_) => msg_type::UPDATE_FAIL_MALFORMED_HTLC,
             Self::GossipTimestampFilter(_) => msg_type::GOSSIP_TIMESTAMP_FILTER,
             Self::Unknown { msg_type, .. } => *msg_type,
@@ -270,6 +277,7 @@ impl Message {
             Self::TxAbort(m) => out.extend(m.encode()),
             Self::UpdateFulfillHtlc(m) => out.extend(m.encode()),
             Self::UpdateFailHtlc(m) => out.extend(m.encode()),
+            Self::CommitmentSigned(m) => out.extend(m.encode()),
             Self::UpdateFailMalformedHtlc(m) => out.extend(m.encode()),
             Self::GossipTimestampFilter(m) => out.extend(m.encode()),
             Self::Unknown { payload, .. } => out.extend(payload),
@@ -312,6 +320,9 @@ impl Message {
                 Ok(Self::UpdateFulfillHtlc(UpdateFulfillHtlc::decode(cursor)?))
             }
             msg_type::UPDATE_FAIL_HTLC => Ok(Self::UpdateFailHtlc(UpdateFailHtlc::decode(cursor)?)),
+            msg_type::COMMITMENT_SIGNED => {
+                Ok(Self::CommitmentSigned(CommitmentSigned::decode(cursor)?))
+            }
             msg_type::UPDATE_FAIL_MALFORMED_HTLC => Ok(Self::UpdateFailMalformedHtlc(
                 UpdateFailMalformedHtlc::decode(cursor)?,
             )),
@@ -726,6 +737,32 @@ mod tests {
         assert_eq!(decoded, Message::UpdateFailHtlc(msg));
     }
 
+    /// Valid `CommitmentSigned` message for testing.
+    fn sample_commitment_signed() -> CommitmentSigned {
+        let secp = Secp256k1::new();
+        let sk = SecretKey::from_byte_array([0x11; 32]).expect("valid secret");
+        let msg = secp256k1::Message::from_digest([0xaa; 32]);
+        let sig = secp.sign_ecdsa(msg, &sk);
+        let msg2 = secp256k1::Message::from_digest([0xbb; 32]);
+        let htlc_sig = secp.sign_ecdsa(msg2, &sk);
+
+        CommitmentSigned {
+            channel_id: ChannelId::new([0xab; CHANNEL_ID_SIZE]),
+            signature: sig,
+            htlc_signatures: vec![htlc_sig],
+            tlvs: CommitmentSignedTlvs::default(),
+        }
+    }
+
+    #[test]
+    fn message_commitment_signed_roundtrip() {
+        let cs = sample_commitment_signed();
+        let msg = Message::CommitmentSigned(cs.clone());
+        let encoded = msg.encode();
+        let decoded = Message::decode(&encoded).unwrap();
+        assert_eq!(decoded, Message::CommitmentSigned(cs));
+    }
+
     #[test]
     fn message_update_fail_malformed_htlc_roundtrip() {
         let msg = UpdateFailMalformedHtlc {
@@ -870,6 +907,10 @@ mod tests {
             })
             .msg_type(),
             msg_type::UPDATE_FAIL_HTLC
+        );
+        assert_eq!(
+            Message::CommitmentSigned(sample_commitment_signed()).msg_type(),
+            msg_type::COMMITMENT_SIGNED
         );
         assert_eq!(
             Message::UpdateFailMalformedHtlc(UpdateFailMalformedHtlc {
--- a/smite/src/bolt/commitment_signed.rs
+++ b/smite/src/bolt/commitment_signed.rs
@@ -1,0 +1,442 @@
+//! BOLT 2 `commitment_signed` message.
+
+use super::BoltError;
+use super::tlv::TlvStream;
+use super::types::{ChannelId, Txid};
+use super::wire::WireFormat;
+use secp256k1::ecdsa::Signature;
+use secp256k1::hashes::Hash;
+
+/// BOLT 2 `commitment_signed` message (type 132).
+///
+/// Sent to sign the counterparty's commitment transaction. Contains the signature
+/// for the commitment transaction and signatures for any pending HTLC outputs.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CommitmentSigned {
+    /// The channel ID derived from the funding outpoint
+    pub channel_id: ChannelId,
+    /// The sender's signature for the commitment transaction
+    pub signature: Signature,
+    /// The signatures for the HTLC outputs (one signature per HTLC in-flight)
+    pub htlc_signatures: Vec<Signature>,
+    /// Optional TLV extensions
+    pub tlvs: CommitmentSignedTlvs,
+}
+
+/// TLV extensions for the `commitment_signed` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct CommitmentSignedTlvs {
+    /// The funding transaction spent by this commitment transaction (TLV type 1).
+    /// Present when splicing is used.
+    pub funding_txid: Option<Txid>,
+}
+
+impl CommitmentSignedTlvs {
+    /// Encodes the TLVs to wire format.
+    fn encode(&self) -> Vec<u8> {
+        let mut stream = TlvStream::new();
+
+        if let Some(txid) = self.funding_txid {
+            stream.add(1, txid.to_byte_array().to_vec());
+        }
+
+        stream.encode()
+    }
+
+    /// Decodes the TLVs from a byte slice.
+    fn decode(data: &[u8]) -> Result<Self, BoltError> {
+        let tlv_stream = TlvStream::decode(data)?;
+
+        let funding_txid = if let Some(bytes) = tlv_stream.get(1) {
+            if bytes.len() != 32 {
+                return Err(BoltError::Truncated {
+                    expected: 32,
+                    actual: bytes.len(),
+                });
+            }
+
+            Some(Txid::from_byte_array(
+                bytes.try_into().expect("length checked above"),
+            ))
+        } else {
+            None
+        };
+
+        Ok(Self { funding_txid })
+    }
+}
+
+impl CommitmentSigned {
+    /// Encodes to wire format (without message type prefix).
+    ///
+    /// # Panics
+    ///
+    /// Panics if more than `u16::MAX` HTLC signatures are present.
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.signature.write(&mut out);
+        u16::try_from(self.htlc_signatures.len())
+            .expect("htlc_signatures length is within u16::MAX")
+            .write(&mut out);
+
+        for sig in &self.htlc_signatures {
+            sig.write(&mut out);
+        }
+
+        out.extend(self.tlvs.encode());
+        out
+    }
+
+    /// Decodes from wire format (without message type prefix).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short for any fixed field, also can return
+    /// `InvalidSignature` and TLV errors.
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        let mut cursor = payload;
+
+        let channel_id = WireFormat::read(&mut cursor)?;
+        let signature = WireFormat::read(&mut cursor)?;
+        let num_htlcs = u16::read(&mut cursor)?;
+
+        let mut htlc_signatures = Vec::with_capacity(num_htlcs as usize);
+        for _ in 0..num_htlcs {
+            let sig = WireFormat::read(&mut cursor)?;
+            htlc_signatures.push(sig);
+        }
+
+        let tlvs = CommitmentSignedTlvs::decode(cursor)?;
+
+        Ok(Self {
+            channel_id,
+            signature,
+            htlc_signatures,
+            tlvs,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::super::{CHANNEL_ID_SIZE, COMPACT_SIGNATURE_SIZE};
+    use super::*;
+    use secp256k1::hashes::Hash;
+    use secp256k1::{Message, Secp256k1, SecretKey};
+
+    /// Valid `CommitmentSigned` message with no HTLCs for testing.
+    fn sample_commitment_signed_no_htlcs() -> CommitmentSigned {
+        let secp = Secp256k1::new();
+        let sk = SecretKey::from_byte_array([0x11; 32]).expect("valid secret");
+        let msg = Message::from_digest([0xaa; 32]);
+        let sig = secp.sign_ecdsa(msg, &sk);
+
+        CommitmentSigned {
+            channel_id: ChannelId::new([0xbb; CHANNEL_ID_SIZE]),
+            signature: sig,
+            htlc_signatures: vec![],
+            tlvs: CommitmentSignedTlvs {
+                funding_txid: Some(Txid::from_byte_array([0xcc; 32])),
+            },
+        }
+    }
+
+    /// Valid `CommitmentSigned` message with 2 HTLC signatures for testing.
+    fn sample_commitment_signed_with_htlcs() -> CommitmentSigned {
+        let secp = Secp256k1::new();
+        let sk = SecretKey::from_byte_array([0x11; 32]).expect("valid secret");
+        let msg = Message::from_digest([0xaa; 32]);
+        let sig = secp.sign_ecdsa(msg, &sk);
+
+        let sk2 = SecretKey::from_byte_array([0x22; 32]).expect("valid secret");
+        let msg2 = Message::from_digest([0xbb; 32]);
+        let sig2 = secp.sign_ecdsa(msg2, &sk2);
+
+        let sk3 = SecretKey::from_byte_array([0x33; 32]).expect("valid secret");
+        let msg3 = Message::from_digest([0xcc; 32]);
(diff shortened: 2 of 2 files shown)


smite/src/bolt.rs
```diff
@@ -110,6 +112,8 @@ pub mod msg_type {
     pub const FUNDING_CREATED: u16 = 34;
     /// `funding_signed` message (BOLT 2).
     pub const FUNDING_SIGNED: u16 = 35;
+    /// `commitment_signed` message (BOLT 2).
+    pub const COMMITMENT_SIGNED: u16 = 132;
```
morehouse: In this file we have been keeping the messages sorted by type number.  Please reorder accordingly.

morehouse: Also need tests in this file for the new `CommitmentSigned` message.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,193 @@
...
+/// Sent to sign the counterparty's commitment transaction. Contains the signature
+/// for the commitment transaction and signatures for any pending HTLC outputs.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CommitmentSigned {
+    /// The channel ID
+    pub channel_id: ChannelId,
+    /// The sender's signature for the commitment transaction
+    pub signature: Signature,
+    /// The signatures for the HTLC outputs (one signature per HTLC in-flight)
+    pub htlc_signatures: Vec<Signature>,
```
morehouse, line 21: We also should support the `tlvs` field.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+impl CommitmentSigned {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.signature.write(&mut out);
+        (self.htlc_signatures.len() as u16).write(&mut out);
```
morehouse: We should panic if this cast would overflow, similar to the other codecs.  I'd suggesting adding an assert to the beginning of `encode`:

```rust
assert!(self.htlc_signatures.len() <= u16::MAX as usize, "too many HTLC signatures");
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,193 @@
...
+        out
+    }
+
+    /// Decodes from wire format (without message type prefix).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short for any fixed field,
+    /// `InvalidSignature` if any signature bytes are not valid compact ECDSA signatures,
+    /// or `InvalidSignatureCount` if the HTLC signature count doesn't match the actual data.
```
morehouse: The `InvalidSignatureCount` error does not exist.

morehouse (changes requested):
Thanks for working on this.  Please make sure CI passes before requesting another review.  You should be able to run all the CI commands locally (e.g., `cargo test`, `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings`)

smite/src/bolt/commitment_signed.rs
ekzyis: @morehouse I was wondering if making `WireFormat for Vec<u8>` generic would be useful. This passes all existing tests:

<details>
<summary>patch</summary>

```diff
diff --git a/smite/src/bolt/wire.rs b/smite/src/bolt/wire.rs
index ebfef88..513389e 100644
--- a/smite/src/bolt/wire.rs
+++ b/smite/src/bolt/wire.rs
@@ -146,7 +146,7 @@ impl WireFormat for BigSize {
     }
 }

-impl WireFormat for Vec<u8> {
+impl<T: WireFormat> WireFormat for Vec<T> {
     /// Reads a `[u16:len][len*byte]` variable-length field, advancing past both.
     ///
     /// # Errors
@@ -160,16 +160,20 @@ impl WireFormat for Vec<u8> {
                 actual: data.len(),
             });
         }
-        let bytes = data[..len].to_vec();
-        *data = &data[len..];
-        Ok(bytes)
+        let mut vec = Vec::with_capacity(len);
+        for _ in 0..len {
+            vec.push(T::read(data)?);
+        }
+        Ok(vec)
     }

     /// Writes a `[u16:len][len*byte]` variable-length field.
     #[allow(clippy::cast_possible_truncation)]
     fn write(&self, out: &mut Vec<u8>) {
         (self.len() as u16).write(out);
-        out.extend_from_slice(self);
+        for item in self {
+            item.write(out);
+        }
     }
 }

```

</details>

However, I searched for `num_` in bolt02 and only found [`witnesses` in `tx_signatures`](https://github.com/lightning/bolts/blob/a3772650d8ebc06acf457fcadf97968ebfc4dfff/02-peer-protocol.md?plain=1#L408-L416) as another case where this could be useful.

So I think the generic interface would be overkill. Writing tests for it also feels arbitrary because of the decision for which types to write tests. I wrote some for `Vec<Signature>`:

<details>
<summary>tests</summary>

```diff
diff --git a/smite/src/bolt/wire.rs b/smite/src/bolt/wire.rs
index 513389e..a3dabf9 100644
--- a/smite/src/bolt/wire.rs
+++ b/smite/src/bolt/wire.rs
@@ -542,6 +542,34 @@ mod tests {
         assert!(cursor.is_empty());
     }

+    fn sample_signatures() -> Vec<Signature> {
+        let secp = Secp256k1::new();
+        let sk = SecretKey::from_byte_array([0x12; 32]).unwrap();
+        vec![
+            secp.sign_ecdsa(secp256k1::Message::from_digest([0x00; 32]), &sk),
+            secp.sign_ecdsa(secp256k1::Message::from_digest([0xAA; 32]), &sk),
+        ]
+    }
+
+    #[test]
+    fn vec_signature_read_empty_field() {
+        let mut data: &[u8] = &[0x00, 0x00];
+        let result = Vec::<Signature>::read(&mut data).unwrap();
+        assert_eq!(result, Vec::<Signature>::new());
+        assert!(data.is_empty());
+    }
+
+    #[test]
+    fn vec_signature_write_roundtrip() {
+        let signatures = sample_signatures();
+        let mut buf = Vec::new();
+        signatures.write(&mut buf);
+        let mut data: &[u8] = &buf;
+        let result = Vec::<Signature>::read(&mut data).unwrap();
+        assert_eq!(result, signatures);
+        assert!(data.is_empty());
+    }
+
     // Test vectors from BOLT 1 Appendix A
     // https://github.com/lightning/bolts/blob/master/01-messaging.md#appendix-a-bigsize-test-vectors

```

</details>

What do you think? Not needed?

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,224 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CommitmentSigned {
+    /// The channel ID
+    pub channel_id: ChannelId,
+    /// The sender's signature for the commitment transaction
+    pub signature: Signature,
+    /// The signatures for the HTLC outputs (one signature per HTLC in-flight)
+    pub htlc_signatures: Vec<Signature>,
+
+    pub tlvs: CommitmentSignedTlvs,
```
morehouse: The field comment is missing here.


```suggestion
    /// Optional TLV extensions
    pub tlvs: CommitmentSignedTlvs,
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,224 @@
...
+/// TLV extensions for the `commitment_signed` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct CommitmentSignedTlvs {}
+
+impl CommitmentSignedTlvs {
+    /// Extracts commitment signed TLVs from a parsed TLV stream.
+    fn from_stream(_stream: &TlvStream) -> Self {
+        Self {}
+    }
+}
```
morehouse, line 67: A no-op TLV implementation isn't useful.  We should implement the `funding_txid` TLV from the [spec](https://github.com/lightning/bolts/blob/master/02-peer-protocol.md#committing-updates-so-far-commitment_signed).

smite/src/bolt/commitment_signed.rs
morehouse: I'm afraid this might make the `Vec<u8>` read/write much slower, since they would be copying one byte at a time unless `rustc` can somehow optimize for this case.

I think for now it probably makes sense to just handle these cases manually.  If we start getting lots of repetitive code we can do a refactoring later.

smite/src/bolt.rs
morehouse: You still haven't addressed the message order here or below.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,224 @@
...
+use super::wire::WireFormat;
+use secp256k1::ecdsa::Signature;
+
+/// BOLT 2 `commitment_signed` message (type 132).
+///
+/// Sent to sign the counterparty's commitment transaction. Contains the signature
+/// for the commitment transaction and signatures for any pending HTLC outputs.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CommitmentSigned {
+    /// The channel ID
```
morehouse:
```suggestion
    /// The channel ID derived from the funding outpoint
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,224 @@
...
+        out.extend(tlv_stream.encode());
+
+        out
+    }
+
+    /// Decodes from wire format (without message type prefix).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short for any fixed field.
```
morehouse: Also can return `InvalidSignature` and TLV errors.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,224 @@
...
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        assert!(
+            u16::try_from(self.htlc_signatures.len()).is_ok(),
+            "too many HTLC signatures"
+        );
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.signature.write(&mut out);
+        #[allow(clippy::cast_possible_truncation)]
```
morehouse:
```suggestion
        #[allow(clippy::cast_possible_truncation)] // HTLC signature count checked above
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,224 @@
...
+
+        assert_eq!(
+            CommitmentSigned::decode(&encoded),
+            Err(BoltError::Truncated {
+                expected: COMPACT_SIGNATURE_SIZE,
+                actual: 54
+            })
+        );
+    }
+}
```
morehouse, line 442: We should add a `decode_unknown_odd_tlv_ignored` test.

morehouse (changes requested):
Please squash into one commit.

kartikangiras:
@morehouse I have made all the necessary changes and ran all the checks locally.

morehouse:
CI still doesn't pass...

kartikangiras:
@morehouse  The command (cargo clippy --all-targets --all-features -- -D warnings) fails locally on ARM64 based systems such as MacOS. due to which I am not able to cross verify locally my code.  I have been facing this issue on all of my PR checks.

ekzyis:
> @morehouse The command (cargo clippy --all-targets --all-features -- -D warnings) fails locally on ARM64 based systems such as MacOS. due to which I am not able to cross verify locally my code. I have been facing this issue on all of my PR checks.

Oh, interesting! I think this is worth an issue, so we can fix it. Can you create one? Can you also share the logs in the issue then?

kartikangiras:
> > @morehouse The command (cargo clippy --all-targets --all-features -- -D warnings) fails locally on ARM64 based systems such as MacOS. due to which I am not able to cross verify locally my code. I have been facing this issue on all of my PR checks.
> 
> Oh, interesting! I think this is worth an issue, so we can fix it. Can you create one? Can you also share the logs in the issue then?

@ekzyis I have created an issue for this. While investigating further, I have founded the root cause of this problem and the necessary fix for it.

kartikangiras:
@morehouse I have locally tested each command on a linux based machine (it took time because i was on a macbook and it natively does not support these commands).

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,378 @@
...
+    /// Optional TLV extensions
+    pub tlvs: CommitmentSignedTlvs,
+}
+
+/// TLV extensions for the `commitment_signed` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct CommitmentSignedTlvs {
+    /// The funding transaction spent by this commitment transaction (TLV type 1).
+    /// Required per BOLT 2 spec.
+    pub funding_txid: Option<[u8; 32]>,
```
morehouse: We actually have a `Txid` type we should use.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,378 @@
...
+impl CommitmentSignedTlvs {
+    /// Encodes the TLVs to wire format.
+    fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+
+        if let Some(txid) = self.funding_txid {
+            // TLV type 1: funding_txid (32 bytes)
+            out.push(0x01); // type = 1
+            out.push(0x20); // length = 32
+            out.extend_from_slice(&txid);
```
morehouse: We should follow the existing `TlvStream` encoding conventions from the other codecs.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,378 @@
...
+    fn decode(data: &[u8]) -> Result<Self, BoltError> {
+        let tlv_stream = TlvStream::decode(data)?;
+
+        let funding_txid = if let Some(bytes) = tlv_stream.get(1) {
+            if bytes.len() == 32 {
+                let mut txid = [0u8; 32];
+                txid.copy_from_slice(bytes);
+                Some(txid)
+            } else {
+                // Malformed funding_txid; treat as absent
```
morehouse: An incorrect len should cause a decode error.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,378 @@
...
+
+    /// Optional TLV extensions
+    pub tlvs: CommitmentSignedTlvs,
+}
+
+/// TLV extensions for the `commitment_signed` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct CommitmentSignedTlvs {
+    /// The funding transaction spent by this commitment transaction (TLV type 1).
+    /// Required per BOLT 2 spec.
```
morehouse: It's not always required -- only when splicing.


```suggestion
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,378 @@
+//! BOLT 2 commitment signed message.
```
morehouse:
```suggestion
//! BOLT 2 `commitment_signed` message.
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,378 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CommitmentSigned {
+    /// The channel ID derived from the funding outpoint
+    pub channel_id: ChannelId,
+    /// The sender's signature for the commitment transaction
+    pub signature: Signature,
+    /// The signatures for the HTLC outputs (one signature per HTLC in-flight)
+    pub htlc_signatures: Vec<Signature>,
+
+    /// Optional TLV extensions
```
morehouse, line 22:
```suggestion
    pub htlc_signatures: Vec<Signature>,
    /// Optional TLV extensions
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,378 @@
...
+    pub fn encode(&self) -> Vec<u8> {
+        assert!(
+            u16::try_from(self.htlc_signatures.len()).is_ok(),
+            "too many HTLC signatures"
+        );
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.signature.write(&mut out);
+        #[allow(clippy::cast_possible_truncation)] // HTLC signature count checked above
+        (self.htlc_signatures.len() as u16).write(&mut out);
```
morehouse: We could remove the assertion above, as well as the clippy annotation here if we do:


```suggestion
        u16::try_from(self.htlc_signatures.len()).expect("htlc_signatures length is within u16::MAX").write(&mut out);
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,378 @@
...
+
+        msg1.tlvs.funding_txid = Some(txid2);
+        let encoded2 = msg1.encode();
+        let decoded2 = CommitmentSigned::decode(&encoded2).unwrap();
+        assert_eq!(decoded2.tlvs.funding_txid, Some(txid2));
+
+        // Different TLVs should produce different encodings
+        assert_ne!(encoded1, encoded2);
+    }
+
```
morehouse, line 387: We should also test:

- malformed-`funding_txid`
- invalid signature
- roundtrip with non-default TLVs

morehouse (changes requested):

smite/src/bolt/commitment_signed.rs
kartikangiras: yeah, i have updated to use the txid

smite/src/bolt/commitment_signed.rs
kartikangiras: resolved

smite/src/bolt/commitment_signed.rs
kartikangiras: indeed, i have added the decode error handling for it

smite/src/bolt/commitment_signed.rs
kartikangiras: resolved

smite/src/bolt/commitment_signed.rs
kartikangiras: i have added the tests for the mentioned scenarios

kartikangiras:
@morehouse I have made all the necessary require changes as reviewed by you.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+    pub funding_txid: Option<Txid>,
+}
+
+impl CommitmentSignedTlvs {
+    /// Encodes the TLVs to wire format.
+    fn encode(&self) -> Vec<u8> {
+        let mut stream = TlvStream::new();
+
+        if let Some(txid) = self.funding_txid {
+            stream.add(1, txid.to_byte_array().to_vec());
```
morehouse, line 40: We should create and use a `TLV_FUNDING_TXID` constant like we do for other codecs.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+        let funding_txid = if let Some(bytes) = tlv_stream.get(1) {
+            if bytes.len() != 32 {
+                return Err(BoltError::Truncated {
+                    expected: 32,
+                    actual: bytes.len(),
+                });
+            }
+
+            Some(Txid::from_byte_array(
+                bytes.try_into().expect("length checked above"),
```
morehouse, line 59: We should be using the `WireFormat` methods for `Txid` in `encode` and `decode`.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+        let signature = WireFormat::read(&mut cursor)?;
+        let num_htlcs = u16::read(&mut cursor)?;
+
+        let mut htlc_signatures = Vec::with_capacity(num_htlcs as usize);
+        for _ in 0..num_htlcs {
+            let sig = WireFormat::read(&mut cursor)?;
+            htlc_signatures.push(sig);
+        }
+
+        let tlvs = CommitmentSignedTlvs::decode(cursor)?;
```
morehouse, line 111: I'd prefer to have consistency between the codecs -- either use the `from_stream` pattern here, or update all other codecs to move that logic into TLV decode functions.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+        out.extend(self.tlvs.encode());
+        out
+    }
+
+    /// Decodes from wire format (without message type prefix).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short for any fixed field, also can return
+    /// `InvalidSignature` and TLV errors.
```
morehouse, line 97:
```suggestion
    /// Returns `Truncated` if the payload is too short, `InvalidSignature` if any signature
    /// is malformed, or a TLV error if the TLV stream is malformed.
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+            CommitmentSigned::decode(data),
+            Err(BoltError::Truncated {
+                expected: COMPACT_SIGNATURE_SIZE,
+                actual: 54
+            })
+        );
+    }
+
+    #[test]
+    fn funding_txid_roundtrip() {
```
morehouse, line 260:
```suggestion
    fn roundtrip_with_funding_txid() {
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+
+        assert_eq!(
+            decoded.tlvs.funding_txid,
+            Some(Txid::from_byte_array([0xdd; 32]))
+        );
+        assert_eq!(original, decoded);
+    }
+
+    #[test]
+    fn malformed_funding_txid_fails() {
```
morehouse, line 296:
```suggestion
    fn decode_malformed_funding_txid() {
```

morehouse, line 296:
```suggestion
    fn decode_malformed_funding_txid() {
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+            result,
+            Err(BoltError::Truncated {
+                expected: 32,
+                actual: 31,
+            })
+        );
+    }
+
+    #[test]
+    fn no_funding_txid() {
```
morehouse, line 320:
```suggestion
    fn encode_no_funding_txid() {
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+        let encoded = msg_no_txid.encode();
+        let decoded = CommitmentSigned::decode(&encoded).unwrap();
+
+        assert_eq!(decoded.tlvs.funding_txid, None);
+        // Message with no TLVs should be shorter
+        assert_eq!(encoded.len(), 98);
+    }
+
+    #[test]
+    fn invalid_signature_fails() {
```
morehouse, line 342:
```suggestion
    fn decode_invalid_signature() {
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+        encoded[sig_offset..sig_offset + COMPACT_SIGNATURE_SIZE].copy_from_slice(&bad_sig);
+
+        assert_eq!(
+            CommitmentSigned::decode(&encoded),
+            Err(BoltError::InvalidSignature(bad_sig))
+        );
+    }
+
+    #[test]
+    fn different_funding_txids() {
```
morehouse, line 357:
```suggestion
    fn encode_different_funding_txids() {
```

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+                funding_txid: Some(Txid::from_byte_array([0xdd; 32])),
+            },
+        }
+    }
+
+    #[test]
+    fn encode_no_htlcs() {
+        let msg = sample_commitment_signed_no_htlcs();
+        let encoded = msg.encode();
+        assert!(encoded.len() > 96);
```
morehouse, line 175: Why did you get rid of the exact length check here?

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    #[test]
+    fn encode_with_htlcs() {
+        let msg = sample_commitment_signed_with_htlcs();
+        let encoded = msg.encode();
+        assert!(encoded.len() > 32 + 64 + 2 + 2 * 64);
```
morehouse, line 182: Why did you get rid of the exact length check here?

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+        let msg = Message::from_digest([0xaa; 32]);
+        let sig = secp.sign_ecdsa(msg, &sk);
+
+        let sk2 = SecretKey::from_byte_array([0x22; 32]).expect("valid secret");
+        let msg2 = Message::from_digest([0xbb; 32]);
+        let sig2 = secp.sign_ecdsa(msg2, &sk2);
+
+        let sk3 = SecretKey::from_byte_array([0x33; 32]).expect("valid secret");
+        let msg3 = Message::from_digest([0xcc; 32]);
+        let sig3 = secp.sign_ecdsa(msg3, &sk3);
```
morehouse, line 159: Looks like a helper function for creating signatures could save several lines of code.

smite/src/bolt/commitment_signed.rs
```diff
@@ -0,0 +1,442 @@
...
+
+        assert_eq!(
+            decoded.tlvs.funding_txid,
+            Some(Txid::from_byte_array([0xcc; 32]))
+        );
+        assert_eq!(original, decoded);
+    }
+
+    #[test]
+    fn funding_txid_with_htlcs_roundtrip() {
```
morehouse, line 278:
```suggestion
    fn roundtrip_funding_txid_with_htlcs() {
```

morehouse (changes requested):

morehouse:
Superseded by #167.

PR #26, opened 2026-04-04
bolt: implement tx_add_output message codec

harsh04044:
Implements the `tx_add_output` codec (type 67) for BOLT 2 interactive-tx.

The script field is a variable-length scriptPubKey, length-prefixed via the WireFormat trait. Tests cover roundtrip, empty script, trailing bytes, and truncation at each field boundary.

Part of #5.

### DIFF
--- a/smite/src/bolt/tx_add_output.rs
+++ b/smite/src/bolt/tx_add_output.rs
@@ -1,0 +1,173 @@
+//! BOLT 2 `tx_add_output` message.
+
+use super::BoltError;
+use super::types::ChannelId;
+use super::wire::WireFormat;
+
+/// BOLT 2 `tx_add_output` message (type 67).
+///
+/// Sent during interactive transaction construction to propose adding an
+/// output to the shared transaction.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TxAddOutput {
+    /// The channel this message pertains to
+    pub channel_id: ChannelId,
+    /// Serial ID for this output, must be even if sent by the initiator,
+    /// odd if sent by the non-initiator (BOLT 2 parity rule)
+    pub serial_id: u64,
+    /// The value of this output in satoshis
+    pub sats: u64,
+    /// The scriptPubKey for this output
+    pub script: Vec<u8>,
+}
+
+impl TxAddOutput {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.serial_id.write(&mut out);
+        self.sats.write(&mut out);
+        self.script.write(&mut out);
+        out
+    }
+
+    /// Decodes from wire format (without message type prefix).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short.
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        let mut cursor = payload;
+        let channel_id = WireFormat::read(&mut cursor)?;
+        let serial_id = WireFormat::read(&mut cursor)?;
+        let sats = WireFormat::read(&mut cursor)?;
+        let script: Vec<u8> = WireFormat::read(&mut cursor)?;
+        Ok(Self {
+            channel_id,
+            serial_id,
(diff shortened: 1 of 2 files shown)


kushagra0902 (commented):
The PR looks good to me with everything handled properly

smite/src/bolt/tx_add_output.rs
```diff
@@ -0,0 +1,147 @@
...
+impl TxAddOutput {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.serial_id.write(&mut out);
+        self.sats.write(&mut out);
+        let script_bytes = self.script.as_bytes().to_vec();
+        script_bytes.write(&mut out);
```
Vineet1101: Also I think using entire bitcoin crate just to provide a wrapper around Vec<u8>  is unecessary. Looking at the actual encode/decode implementation, this type provides no value:


The code reads and writes raw Vec<u8> anyway — ScriptBuf is just wrapping and immediately unwrapping a byte vector with zero added value. 
```rust
pub script: Vec<u8>,
//Which reduces encode to:
self.script.write(&mut out);
//And decode to:
let script: Vec<u8> = WireFormat::read(&mut cursor)?;
```
No external dependency needed.

smite/src/bolt/tx_add_output.rs
```diff
@@ -0,0 +1,147 @@
...
+pub struct TxAddOutput {
+    /// The channel this message pertains to.
+    pub channel_id: ChannelId,
+    /// Serial ID for this output. Must be even if sent by the initiator,
+    /// odd if sent by the non-initiator (BOLT 2 parity rule).
+    pub serial_id: u64,
+    /// The value of this output in satoshis.
+    pub sats: u64,
+    /// The scriptPubKey for this output.
+    pub script: ScriptBuf,
```
morehouse: Let's use `script: Vec<u8>` for now to avoid the `bitcoin` dependency.  We can add it later if necessary.

smite/src/bolt/tx_add_output.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+impl TxAddOutput {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
```
morehouse, line 27: We should assert that `script.len() <= u16::MAX` here since `Vec<u8>::write` assumes this.

smite/src/bolt/tx_add_output.rs
```diff
@@ -0,0 +1,147 @@
...
+        // channel_id (32) + serial_id (8) + 4 bytes of sats
+        assert_eq!(
+            TxAddOutput::decode(&[0x00; CHANNEL_ID_SIZE + 8 + 4]),
+            Err(BoltError::Truncated {
+                expected: 8,
+                actual: 4
+            })
+        );
+    }
+
```
morehouse, line 121: We should also test the cases where the script length is truncated or the script itself is truncated.

morehouse (changes requested):

harsh04044:
Done, script is now Vec<u8>, added the assert on script.len() in encode(), and split out truncation tests for the script length and script data fields.
@morehouse

smite/src/bolt/tx_add_output.rs
```diff
@@ -0,0 +1,177 @@
...
+            script: vec![],
+            ..sample_msg()
+        };
+        let encoded = msg.encode();
+        let decoded = TxAddOutput::decode(&encoded).unwrap();
+        assert_eq!(decoded, msg);
+    }
+
+    #[test]
+    fn decode_truncated_script_len() {
```
morehouse, line 123: Nit: please group the truncated error tests next to each other

morehouse (commented):
CI is failing

smite/src/bolt/tx_add_output.rs
harsh04044: Grouped all truncated tests together.

morehouse (approved):
Needs a rebase to merge.

morehouse:
Superseded by #221.

PR #27, opened 2026-04-04
smite: implement update_fulfill_htlc and update_fail_htlc message codecs

devvaansh:
Add codecs for the BOLT 2 HTLC update messages:

- `update_fulfill_htlc` (type 130) — fulfills an HTLC by providing the payment preimage. Wire format: `channel_id` (32 bytes) + `id` (u64) + `payment_preimage` (32 bytes). All fixed fields, 72 bytes total.
- `update_fail_htlc` (type 131) — fails an HTLC back to the sender with an encrypted failure reason. Wire format: `channel_id` (32 bytes) + `id` (u64) + `len` (u16) + `reason` (variable bytes).

**Changes:**
- New file: `smite/src/bolt/update_fulfill_htlc.rs` — `UpdateFulfillHtlc` struct with `encode()` / `decode()`
- New file: `smite/src/bolt/update_fail_htlc.rs` — `UpdateFailHtlc` struct with `encode()` / `decode()`
- Updated `smite/src/bolt.rs` — wired both into `Message` enum

**Tests:** 6 unit tests for `UpdateFulfillHtlc` + 8 unit tests for `UpdateFailHtlc` + 4 `Message`-level tests. All 264 bolt tests pass.

Part of #5 (Milestone 3).

### DIFF
--- a/smite/src/bolt/update_fail_htlc.rs
+++ b/smite/src/bolt/update_fail_htlc.rs
@@ -1,0 +1,275 @@
+//! BOLT 2 `update_fail_htlc` message.
+
+use super::BoltError;
+use super::attribution_data::AttributionData;
+use super::tlv::TlvStream;
+use super::types::ChannelId;
+use super::wire::WireFormat;
+
+/// TLV type for attribution data.
+const TLV_ATTRIBUTION_DATA: u64 = 1;
+
+/// BOLT 2 `update_fail_htlc` message (type 131).
+///
+/// Sent to fail an HTLC back to the sender.  The `reason` field contains an
+/// encrypted failure message that is relayed back along the payment path.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UpdateFailHtlc {
+    /// The channel ID.
+    pub channel_id: ChannelId,
+    /// The HTLC ID being failed.
+    pub id: u64,
+    /// Encrypted reason for the failure, relayed back to the sender.
+    pub reason: Vec<u8>,
+    /// Optional TLV extensions.
+    pub tlvs: UpdateFailHtlcTlvs,
+}
+
+/// TLV extensions for the `update_fail_htlc` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct UpdateFailHtlcTlvs {
+    /// Attribution data for failure attribution (TLV type 1).
+    pub attribution_data: Option<AttributionData>,
+}
+
+impl UpdateFailHtlc {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.id.write(&mut out);
+        self.reason.write(&mut out);
+
+        // Encode TLVs
+        let mut tlv_stream = TlvStream::new();
+        if let Some(attr) = &self.tlvs.attribution_data {
+            tlv_stream.add(TLV_ATTRIBUTION_DATA, attr.encode());
+        }
+        out.extend(tlv_stream.encode());
+
+        out
+    }
+
+    /// Decodes from wire format (without message type prefix).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short, or TLV errors
+    /// if the TLV stream is malformed.
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        let mut cursor = payload;
+        let channel_id = WireFormat::read(&mut cursor)?;
+        let id = WireFormat::read(&mut cursor)?;
+        let reason = WireFormat::read(&mut cursor)?;
+
+        // Decode TLVs (remaining bytes)
+        // attribution_data is type 1 (odd), so no known even types
+        let tlv_stream = TlvStream::decode(cursor)?;
+        let tlvs = UpdateFailHtlcTlvs::from_stream(&tlv_stream)?;
+
+        Ok(Self {
+            channel_id,
+            id,
+            reason,
+            tlvs,
+        })
+    }
+}
+
+impl UpdateFailHtlcTlvs {
+    /// Extracts TLVs from a parsed TLV stream.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if `attribution_data` has invalid length.
+    fn from_stream(stream: &TlvStream) -> Result<Self, BoltError> {
+        let attribution_data = stream
+            .get(TLV_ATTRIBUTION_DATA)
+            .map(AttributionData::decode)
+            .transpose()?;
+        Ok(Self { attribution_data })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::super::CHANNEL_ID_SIZE;
+    use super::super::attribution_data::TruncatedHmac;
+    use super::*;
+
+    fn sample_msg() -> UpdateFailHtlc {
+        UpdateFailHtlc {
+            channel_id: ChannelId::new([0xab; CHANNEL_ID_SIZE]),
+            id: 42,
+            reason: vec![0xde, 0xad, 0xbe, 0xef],
+            tlvs: UpdateFailHtlcTlvs::default(),
+        }
+    }
+
+    #[test]
+    fn encode_field_sizes() {
+        let msg = UpdateFailHtlc {
+            channel_id: ChannelId::new([0x42; CHANNEL_ID_SIZE]),
+            id: 1,
+            reason: vec![0xaa, 0xbb],
+            tlvs: UpdateFailHtlcTlvs::default(),
+        };
+        let encoded = msg.encode();
+        // channel_id(32) + id(8) + len(2) + reason(2) = 44
+        assert_eq!(encoded.len(), 44);
+    }
+
+    #[test]
+    fn roundtrip() {
+        let original = sample_msg();
+        let encoded = original.encode();
+        let decoded = UpdateFailHtlc::decode(&encoded).unwrap();
+        assert_eq!(original, decoded);
+    }
+
+    #[test]
+    fn roundtrip_empty_reason() {
+        let original = UpdateFailHtlc {
+            channel_id: ChannelId::new([0xab; CHANNEL_ID_SIZE]),
+            id: 7,
+            reason: vec![],
+            tlvs: UpdateFailHtlcTlvs::default(),
+        };
+        let encoded = original.encode();
+        let decoded = UpdateFailHtlc::decode(&encoded).unwrap();
+        assert_eq!(original, decoded);
+    }
+
+    #[test]
+    fn roundtrip_with_attribution_data() {
+        let mut msg = sample_msg();
+        msg.tlvs.attribution_data = Some(AttributionData {
+            htlc_hold_times: [100; 20],
+            truncated_hmacs: [TruncatedHmac([0xaa; 4]); 210],
+        });
+        let encoded = msg.encode();
+        let decoded = UpdateFailHtlc::decode(&encoded).unwrap();
+        assert_eq!(msg, decoded);
+    }
+
+    #[test]
+    fn decode_truncated_attribution_data() {
+        let msg = sample_msg();
+        let mut encoded = msg.encode();
+        // Append a TLV type 1 with only 100 bytes (should be 920)
+        encoded.push(0x01); // type
+        encoded.push(0x64); // length = 100
+        encoded.extend_from_slice(&[0x00; 100]);
+        assert_eq!(
+            UpdateFailHtlc::decode(&encoded),
+            Err(BoltError::Truncated {
+                expected: AttributionData::SIZE,
+                actual: 100
+            })
+        );
+    }
+
+    #[test]
+    fn decode_unknown_odd_tlv_ignored() {
+        let mut msg = sample_msg();
+        let mut encoded = msg.encode();
+        // Append an unknown odd TLV (type 3, len 2, value 0xffff)
+        encoded.extend_from_slice(&[0x03, 0x02, 0xff, 0xff]);
+        let decoded = UpdateFailHtlc::decode(&encoded).unwrap();
+        msg.tlvs = UpdateFailHtlcTlvs::default();
+        assert_eq!(decoded.channel_id, msg.channel_id);
+        assert_eq!(decoded.id, msg.id);
+        assert_eq!(decoded.reason, msg.reason);
+    }
+
+    #[test]
+    fn decode_unknown_even_tlv_rejected() {
+        let mut encoded = sample_msg().encode();
+        // Append an unknown even TLV (type 2, len 1, value 0x00)
+        encoded.extend_from_slice(&[0x02, 0x01, 0x00]);
+        assert!(matches!(
+            UpdateFailHtlc::decode(&encoded),
+            Err(BoltError::TlvUnknownEvenType(2))
+        ));
+    }
+
+    #[test]
+    #[should_panic(expected = "exceeds maximum size")]
+    fn encode_panics_on_oversized_reason() {
+        let msg = UpdateFailHtlc {
+            channel_id: ChannelId::new([0x00; CHANNEL_ID_SIZE]),
+            id: 0,
+            reason: vec![0x00; u16::MAX as usize + 1],
+            tlvs: UpdateFailHtlcTlvs::default(),
+        };
+        let _ = msg.encode();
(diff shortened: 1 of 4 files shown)


smite/src/bolt/update_fulfill_htlc.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// Decodes from wire format (without message type prefix).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short for any fixed field.
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        let mut cursor = payload;
+        let channel_id = ChannelId::read(&mut cursor)?;
+        let id = u64::read(&mut cursor)?;
+        let payment_preimage = <[u8; PAYMENT_PREIMAGE_SIZE]>::read(&mut cursor)?;
```
morehouse: Nit: decoding syntax can be simplified slightly with type inference:


```suggestion
        let channel_id = WireFormat::read(&mut cursor)?;
        let id = WireFormat::read(&mut cursor)?;
        let payment_preimage = WireFormat::read(&mut cursor)?;
```

smite/src/bolt/update_fulfill_htlc.rs
```diff
@@ -0,0 +1,134 @@
...
+/// this message, the peer should remove the corresponding HTLC from the
+/// commitment transaction.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UpdateFulfillHtlc {
+    /// The channel ID.
+    pub channel_id: ChannelId,
+    /// The HTLC ID being fulfilled.
+    pub id: u64,
+    /// The payment preimage (32 bytes).
+    pub payment_preimage: [u8; PAYMENT_PREIMAGE_SIZE],
```
morehouse, line 27: We should also implement the `tlvs` field.

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -0,0 +1,158 @@
...
+/// Sent to fail an HTLC back to the sender.  The `reason` field contains an
+/// encrypted failure message that is relayed back along the payment path.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UpdateFailHtlc {
+    /// The channel ID.
+    pub channel_id: ChannelId,
+    /// The HTLC ID being failed.
+    pub id: u64,
+    /// Encrypted reason for the failure, relayed back to the sender.
+    pub reason: Vec<u8>,
```
morehouse, line 23: We should also implement the `tlvs` field.

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+impl UpdateFailHtlc {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
```
morehouse, line 38: We should assert that `reason.len() <= u16::MAX` since `Vec<u8>::write` assumes it.

Or perhaps we could update `Vec<u8>::write` to put the assertion there in a separate refactoring PR, if it makes sense everywhere else that call `Vec<u8>::write`.

morehouse (changes requested):

smite/src/bolt/update_fulfill_htlc.rs
devvaansh: Done, updated both files to use WireFormat::read(&mut cursor)? for consistency.

smite/src/bolt/update_fulfill_htlc.rs
devvaansh: Added with tests.

smite/src/bolt/update_fail_htlc.rs
devvaansh: Added the assertion.

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -0,0 +1,259 @@
+//! BOLT 2 `update_fail_htlc` message.
+
+use super::BoltError;
+use super::tlv::TlvStream;
+use super::types::ChannelId;
+use super::wire::WireFormat;
+
+/// TLV type for attribution data.
+const TLV_ATTRIBUTION_DATA: u64 = 1;
```
Chand-ra, line 10: DRY: Both the files define this, so maybe it's better to move it in the `tlv` module?

Chand-ra (commented):
Nit: please consider using separate commits for each message type. Reduces mental overhead while reviewing.

smite/src/bolt/update_fail_htlc.rs
morehouse: I prefer keeping the TLV types with the codecs they belong to.  It is *possible* that the TLVs for these two messages diverge in the future.

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -0,0 +1,259 @@
...
+}
+
+/// TLV extensions for the `update_fail_htlc` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct UpdateFailHtlcTlvs {
+    /// Attribution data for failure attribution (TLV type 1).
+    ///
+    /// Contains `htlc_hold_times` (20×u32) and `truncated_hmacs`
+    /// (210×4-byte truncated SHA-256 HMACs).
+    pub attribution_data: Option<Vec<u8>>,
```
morehouse: Let's implement an `AttributionData` struct so we can actually read/write these TLV values.

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -0,0 +1,259 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short, or TLV errors
+    /// if the TLV stream is malformed.
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        let mut cursor = payload;
+        let channel_id = WireFormat::read(&mut cursor)?;
+        let id = WireFormat::read(&mut cursor)?;
+        let reason = Vec::<u8>::read(&mut cursor)?;
```
morehouse:
```suggestion
        let reason = WireFormat::read(&mut cursor)?;
```

smite/src/bolt/update_fulfill_htlc.rs
```diff
@@ -0,0 +1,212 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short for any fixed field,
+    /// or TLV errors if the TLV stream is malformed.
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        let mut cursor = payload;
+        let channel_id = WireFormat::read(&mut cursor)?;
+        let id = WireFormat::read(&mut cursor)?;
+        let payment_preimage = <[u8; PAYMENT_PREIMAGE_SIZE]>::read(&mut cursor)?;
```
morehouse:
```suggestion
        let payment_preimage = WireFormat::read(&mut cursor)?;
```

morehouse (changes requested):

smite/src/bolt/update_fail_htlc.rs
devvaansh: noted :D

smite/src/bolt/update_fail_htlc.rs
devvaansh: Done. Added an AttributionData struct with hold_times: [u32; 20] and hmacs: Vec<u8> (840 bytes), matching the BOLT 4 attribution proposal's fixed 920-byte format (always padded to 20 hops). Both TLV structs now use Option<AttributionData> instead of Option<Vec<u8>>, with encode/decode and a decode_truncated_attribution_data test for each codec.

smite/src/bolt/update_fulfill_htlc.rs
morehouse: This suggestion was not addressed.

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -0,0 +1,346 @@
...
+use super::wire::WireFormat;
+
+/// TLV type for attribution data.
+const TLV_ATTRIBUTION_DATA: u64 = 1;
+
+/// Maximum number of hops for failure attribution.
+const ATTRIBUTION_MAX_HOPS: usize = 20;
+
+/// Number of truncated HMACs: `MAX_HOPS * (MAX_HOPS + 1) / 2`.
+const ATTRIBUTION_NUM_HMACS: usize = ATTRIBUTION_MAX_HOPS * (ATTRIBUTION_MAX_HOPS + 1) / 2;
```
morehouse: Where did this calculation come from?  The spec just says 210.

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -0,0 +1,346 @@
...
+///
+/// Fixed-size structure (920 bytes) containing per-hop hold times and
+/// truncated HMACs for attribution verification, always padded to the
+/// maximum of 20 hops.  Defined in the BOLT 4 attribution proposal.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AttributionData {
+    /// Hold time in milliseconds at each of the 20 possible hops.
+    pub hold_times: [u32; ATTRIBUTION_MAX_HOPS],
+    /// Truncated SHA-256 HMACs for hop-by-hop verification (840 bytes).
+    pub hmacs: Vec<u8>,
```
morehouse: Probably we want a `TruncatedHmac` newtype, and `hmacs` could just be `[TruncatedHmac; ATTRIBUTION_NUM_HMACS]`.

Then we don't have to worry about calculating the size and validating `hmacs` size on `encode` (which currently is missing).

smite/src/bolt/update_fail_htlc.rs
morehouse: This suggestion was not addressed.

smite/src/bolt/update_fail_htlc.rs
morehouse: Nit: I'd prefer to use the same field names as in the spec.

morehouse (changes requested):

smite/src/bolt/update_fail_htlc.rs
devvaansh: Apologies, fixed now.

smite/src/bolt/update_fulfill_htlc.rs
devvaansh: done

smite/src/bolt/update_fail_htlc.rs
devvaansh: formula was just my own derivation (n * (n + 1) / 2 for n = 20), not something taken from the spec.it's better to just match the spec directly, so now using the literal 210 as in the spec.

smite/src/bolt/update_fail_htlc.rs
devvaansh: I’ve added a TruncatedHmac newtype with its own WireFormat impl and switched hmacs to [TruncatedHmac; ATTRIBUTION_NUM_HMACS]. This also cleans up the size handling and avoids needing separate length validation during encode.

devvaansh: Done, renamed to htlc_hold_times and truncated_hmacs.

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -0,0 +1,369 @@
...
+pub struct AttributionData {
+    /// Per-hop hold times in milliseconds.
+    pub htlc_hold_times: [u32; ATTRIBUTION_MAX_HOPS],
+    /// Truncated HMACs for hop-by-hop verification.
+    pub truncated_hmacs: [TruncatedHmac; ATTRIBUTION_NUM_HMACS],
+}
+
+impl AttributionData {
+    /// Total wire size in bytes (920).
+    pub const SIZE: usize = ATTRIBUTION_DATA_SIZE;
```
morehouse: We could move the calculation of the size to here, instead of having it as a global above.  And potentially we don't need to calculate the size at all, perhaps we can have the compiler do that automatically:


```suggestion
    /// Total wire size in bytes (920).
    pub const SIZE: usize = std::mem::size_of<Self>();
```

devvaansh: Done. Removed ATTRIBUTION_DATA_SIZE and switched to std::mem::size_of::<Self>()

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// # Panics
+    ///
+    /// Panics if `reason` is longer than `u16::MAX` bytes.
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        assert!(
+            u16::try_from(self.reason.len()).is_ok(),
+            "reason length {} exceeds u16::MAX",
+            self.reason.len()
+        );
```
morehouse: We can remove this now that #49 moved the assert to `WireFormat`.

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -0,0 +1,365 @@
...
+pub struct AttributionData {
+    /// Per-hop hold times in milliseconds.
+    pub htlc_hold_times: [u32; ATTRIBUTION_MAX_HOPS],
+    /// Truncated HMACs for hop-by-hop verification.
+    pub truncated_hmacs: [TruncatedHmac; ATTRIBUTION_NUM_HMACS],
+}
+
+impl AttributionData {
+    /// Total wire size in bytes (920).
+    pub const SIZE: usize = std::mem::size_of::<Self>();
```
morehouse: Sorry to flip-flop on this, but on second thought using `mem::size_of` here is a bit brittle.  If the struct changes in the future and some padding is introduced, the `mem::size_of` would return the wrong value.  Perhaps we should just calculate it directly here.


```suggestion
    pub const SIZE: usize = ATTRIBUTION_MAX_HOPS * 4 + ATTRIBUTION_NUM_HMACS * TRUNCATED_HMAC_SIZE;
```

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -0,0 +1,365 @@
...
+    }
+}
+
+/// Attribution data for failure/fulfill attribution (TLV type 1).
+///
+/// Fixed-size structure (920 bytes) containing per-hop hold times and
+/// truncated HMACs for attribution verification, always padded to the
+/// maximum of 20 hops.  Defined in the BOLT 4 attribution proposal.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AttributionData {
```
morehouse: Since `AttributionData` is used by both `UpdateFailHtlc` and `UpdateFulfillHtlc`, I think it would make sense to move it to its own module at `bolt/attribution_data.rs`.

morehouse (approved):

smite/src/bolt/update_fail_htlc.rs
devvaansh: Done, We now rely on the assert that lives in WireFormat::write for Vec<u8> (from #49).

smite/src/bolt/update_fail_htlc.rs
devvaansh: Good point, Also I have added a unit test asserting AttributionData::SIZE == 920 so any accidental drift is caught.

smite/src/bolt/update_fail_htlc.rs
devvaansh: agree++, moved AttributionData, TruncatedHmac, and the related constants/tests to a new bolt/attribution_data.rs module.

smite/src/bolt/update_fail_htlc.rs
```diff
@@ -284,18 +194,6 @@ mod tests {
         ));
     }
 
-    #[test]
-    #[should_panic(expected = "reason length")]
-    fn encode_panics_on_oversized_reason() {
```
morehouse: This test could stay since we still panic in `WireFormat`, we would just need to change the panic string we're looking for.

morehouse (approved):
Thanks, looks good.  Just needs a squash and rebase before I can merge it.

smite/src/bolt/update_fail_htlc.rs
devvaansh: noted!

devvaansh:
@morehouse  branch is now squashed , rebased on the upstream

morehouse (approved):
LGTM

PR #28, opened 2026-04-04
bolt: implement tx_signatures message codec

harsh04044:
Implements the `tx_signatures` codec (type 71) for BOLT 2 interactive-tx.

The witnesses field is a nested structure -- one entry per input, each containing a witness stack of variable-length items. Encoded as num_witnesses (u16) followed by num_items (u16) and length-prefixed items for each witness.

Tests cover roundtrip, empty witnesses, empty stack items, trailing bytes, wire size, and truncation at each field boundary.

Part of #5.

### DIFF
--- a/smite/src/bolt/tx_signatures.rs
+++ b/smite/src/bolt/tx_signatures.rs
@@ -1,0 +1,229 @@
+//! BOLT 2 `tx_signatures` message.
+
+use super::BoltError;
+use super::types::{ChannelId, Txid};
+use super::wire::WireFormat;
+
+/// BOLT 2 `tx_signatures` message (type 71).
+///
+/// Sent during interactive transaction construction to provide the sender's
+/// witnesses for the negotiated transaction.  Both peers exchange
+/// `tx_signatures` before broadcasting.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TxSignatures {
+    /// The channel ID.
+    pub channel_id: ChannelId,
+    /// The transaction ID (little-endian, Bitcoin serialization).
+    pub txid: Txid,
+    /// Witnesses for the transaction inputs.
+    ///
+    /// Outer `Vec`: one entry per input.
+    /// Middle `Vec`: the witness stack for that input.
+    /// Inner `Vec<u8>`: a single witness stack item.
+    pub witnesses: Vec<Vec<Vec<u8>>>,
+}
+
+impl TxSignatures {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.txid.write(&mut out);
+        u16::try_from(self.witnesses.len()).unwrap_or(u16::MAX).write(&mut out);
+        for witness in &self.witnesses {
+            u16::try_from(witness.len()).unwrap_or(u16::MAX).write(&mut out);
+            for item in witness {
+                item.write(&mut out);
+            }
+        }
+        out
+    }
+
+    /// Decodes from wire format (without message type prefix).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short for any fixed field or
+    /// declared variable-length data.
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        let mut cursor = payload;
+        let channel_id: ChannelId = WireFormat::read(&mut cursor)?;
+        let txid: Txid = WireFormat::read(&mut cursor)?;
+        let num_witnesses: u16 = WireFormat::read(&mut cursor)?;
+        let mut witnesses = Vec::with_capacity(num_witnesses as usize);
+        for _ in 0..num_witnesses {
+            let num_items: u16 = WireFormat::read(&mut cursor)?;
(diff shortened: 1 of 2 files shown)


smite/src/bolt/tx_signatures.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// The channel ID.
+    pub channel_id: ChannelId,
+    /// The transaction ID (little-endian, Bitcoin serialization).
+    pub txid: Txid,
+    /// Witnesses for the transaction inputs.
+    ///
+    /// Outer `Vec`: one entry per input.
+    /// Middle `Vec`: the witness stack for that input.
+    /// Inner `Vec<u8>`: a single witness stack item.
+    pub witnesses: Vec<Vec<Vec<u8>>>,
```
morehouse: This looks wrong -- shouldn't there only be an outer and an inner vec (outer: `witnesses`, inner: `witness_data`)?  I think it would also help readability to define a newtype for `Witness` in this file.  Then we just have `witnesses: Vec<Witness>`.

smite/src/bolt/tx_signatures.rs
```diff
@@ -0,0 +1,229 @@
...
+}
```
morehouse, line 29: We also should implement the `tlvs` field.

smite/src/bolt/tx_signatures.rs
```diff
@@ -0,0 +1,229 @@
...
+impl TxSignatures {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.txid.write(&mut out);
+        u16::try_from(self.witnesses.len()).unwrap_or(u16::MAX).write(&mut out);
+        for witness in &self.witnesses {
+            u16::try_from(witness.len()).unwrap_or(u16::MAX).write(&mut out);
```
morehouse: Rather than silently truncating, I think we should assert that `u16::MAX` is not exceeded, since that indicates a programming error on our part.

morehouse (changes requested):

harsh04044:
Done: added the Witness newtype, TxSignaturesTlvs, and swapped the unwrap_or(u16::MAX) calls for asserts.
@morehouse

Chand-ra:
Since we're panicking in `TxSignatures::encode()` now, it might be a good idea to add tests to cover the 'panic' cases as well.

smite/src/bolt/tx_signatures.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short for any fixed field or
+    /// declared variable-length data.
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        let mut cursor = payload;
+        let channel_id: ChannelId = WireFormat::read(&mut cursor)?;
+        let txid: Txid = WireFormat::read(&mut cursor)?;
+        let num_witnesses: u16 = WireFormat::read(&mut cursor)?;
+        let mut witnesses = Vec::with_capacity(num_witnesses as usize);
```
Chand-ra, line 83: Looks like recipe for an Out-Of-Memory (OOM) crash in case the fuzzer decides on a high value for `num_witnesses` or `num_items`. This would mean the `WireFormat::read` loop doesn't even get a chance to fail on a truncated payload.

I'd say it's better to let the vector grow dynamically:
```
let mut witnesses = Vec::new();
```

Chand-ra (changes requested):

smite/src/bolt/tx_signatures.rs
```diff
@@ -0,0 +1,270 @@
...
+use super::BoltError;
+use super::tlv::TlvStream;
+use super::types::{ChannelId, Txid};
+use super::wire::WireFormat;
+
+/// A witness stack for a single transaction input.
+///
+/// Each inner `Vec<u8>` is a single witness stack item.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Witness(pub Vec<Vec<u8>>);
```
morehouse, line 12: IIUC, a witness should just be `Vec<u8>`, not `Vec<Vec<u8>>`.

smite/src/bolt/tx_signatures.rs
```diff
@@ -0,0 +1,270 @@
...
+    pub txid: Txid,
+    /// Witnesses for the transaction inputs.
+    pub witnesses: Vec<Witness>,
+    /// Optional TLV extensions.
+    pub tlvs: TxSignaturesTlvs,
+}
+
+/// TLV extensions for the `tx_signatures` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct TxSignaturesTlvs {}
```
morehouse, line 33: Please actually implement the TLVs rather than ignoring them.

smite/src/bolt/tx_signatures.rs
morehouse: The fuzzer shouldn't be choosing values that get *decoded*, only values that get *encoded*.

Either way, `num_witnesses` is capped at 64K due to `u16::MAX`.  I think it makes sense to preallocate in this case.

morehouse (changes requested):
CI is failing

PR #30, opened 2026-04-04
bolt: add tx_init_rbf and tx_ack_rbf codecs

devvaansh:
Add encoding/decoding support for two interactive transaction RBF messages from BOLT 2:

- **`tx_init_rbf`** (type 72): `channel_id` (32) + `locktime` (u32, 4) + `feerate` (u32, 4) — 40 bytes total
- **`tx_ack_rbf`** (type 73): `channel_id` (32) — 32 bytes total

Both messages are part of the interactive transaction protocol's RBF (Replace-By-Fee) flow, where `tx_init_rbf` initiates an RBF attempt and `tx_ack_rbf` acknowledges it.

## Changes

- **`smite/src/bolt/tx_init_rbf.rs`** — New codec: `TxInitRbf` struct with `encode()`/`decode()` and tests (roundtrip, field sizes, truncation for each field, empty payload)
- **`smite/src/bolt/tx_ack_rbf.rs`** — New codec: `TxAckRbf` struct with `encode()`/`decode()` and tests (roundtrip, field sizes, truncation, empty payload, trailing bytes)
- **`smite/src/bolt.rs`** — Wire both codecs into the `Message` enum with msg_type constants (`TX_INIT_RBF = 72`, `TX_ACK_RBF = 73`), `msg_type()`/`encode()`/`decode()` match arms, and Message-level roundtrip + type_values tests

## Testing

- All 261 bolt tests pass
- `cargo clippy` clean (0 warnings)
- `cargo fmt --all --check` clean

## Notes

TLV streams (`funding_output_contribution`, `require_confirmed_inputs`) are omitted for now, consistent with how existing codecs handle optional TLVs.

### DIFF
--- a/smite/src/bolt/tx_ack_rbf.rs
+++ b/smite/src/bolt/tx_ack_rbf.rs
@@ -1,0 +1,234 @@
+//! BOLT 2 `tx_ack_rbf` message.
+
+use super::BoltError;
+use super::tlv::TlvStream;
+use super::types::ChannelId;
+use super::wire::WireFormat;
+
+/// TLV type for funding output contribution.
+const TLV_FUNDING_OUTPUT_CONTRIBUTION: u64 = 0;
+
+/// TLV type for require confirmed inputs.
+const TLV_REQUIRE_CONFIRMED_INPUTS: u64 = 2;
+
+/// BOLT 2 `tx_ack_rbf` message (type 73).
+///
+/// Sent by the non-initiator to acknowledge an RBF attempt initiated by
+/// the peer via `tx_init_rbf`.  After this message, both sides begin a
+/// new interactive transaction construction for the replacement transaction.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TxAckRbf {
+    /// The channel ID.
+    pub channel_id: ChannelId,
+    /// Optional TLV extensions.
+    pub tlvs: TxAckRbfTlvs,
+}
+
+/// TLV extensions for the `tx_ack_rbf` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct TxAckRbfTlvs {
+    /// The amount the sender will contribute to the funding output (TLV type 0).
+    ///
+    /// Signed 64-bit integer; may be negative if the sender is removing funds.
+    pub funding_output_contribution: Option<i64>,
+
+    /// Whether the sender requires all inputs to be confirmed (TLV type 2).
+    ///
+    /// Presence of this TLV (even with empty value) signals the requirement.
+    pub require_confirmed_inputs: bool,
+}
+
+impl TxAckRbf {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+
+        // Encode TLVs
+        let mut tlv_stream = TlvStream::new();
+        if let Some(contribution) = self.tlvs.funding_output_contribution {
+            tlv_stream.add(
+                TLV_FUNDING_OUTPUT_CONTRIBUTION,
+                contribution.to_be_bytes().to_vec(),
+            );
+        }
+        if self.tlvs.require_confirmed_inputs {
+            tlv_stream.add(TLV_REQUIRE_CONFIRMED_INPUTS, vec![]);
+        }
+        out.extend(tlv_stream.encode());
+
+        out
+    }
+
+    /// Decodes from wire format (without message type prefix).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short, or TLV errors
+    /// if the TLV stream is malformed.
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        let mut cursor = payload;
+        let channel_id = WireFormat::read(&mut cursor)?;
+
+        // Decode TLVs (remaining bytes)
+        let tlv_stream = TlvStream::decode_with_known(
+            cursor,
+            &[
+                TLV_FUNDING_OUTPUT_CONTRIBUTION,
+                TLV_REQUIRE_CONFIRMED_INPUTS,
+            ],
+        )?;
+        let tlvs = TxAckRbfTlvs::from_stream(&tlv_stream)?;
+
+        Ok(Self { channel_id, tlvs })
+    }
+}
+
+impl TxAckRbfTlvs {
+    /// Extracts TLVs from a parsed TLV stream.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if `funding_output_contribution` has invalid length.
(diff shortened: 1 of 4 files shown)


smite/src/bolt/tx_init_rbf.rs
```diff
@@ -0,0 +1,131 @@
...
+/// locktime and feerate that will apply to the replacement transaction.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TxInitRbf {
+    /// The channel ID.
+    pub channel_id: ChannelId,
+    /// The locktime for the replacement transaction.
+    pub locktime: u32,
+    /// The feerate for the replacement transaction (satoshis per kilo-weight).
+    pub feerate: u32,
+}
```
morehouse, line 43: `tlvs` is missing.

smite/src/bolt/tx_ack_rbf.rs
```diff
@@ -0,0 +1,93 @@
...
+/// BOLT 2 `tx_ack_rbf` message (type 73).
+///
+/// Sent by the non-initiator to acknowledge an RBF attempt initiated by
+/// the peer via `tx_init_rbf`.  After this message, both sides begin a
+/// new interactive transaction construction for the replacement transaction.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TxAckRbf {
+    /// The channel ID.
+    pub channel_id: ChannelId,
+}
```
morehouse, line 39: `tlvs` is missing.

smite/src/bolt/tx_ack_rbf.rs
```diff
@@ -0,0 +1,93 @@
...
+            TxAckRbf::decode(&[]),
+            Err(BoltError::Truncated {
+                expected: CHANNEL_ID_SIZE,
+                actual: 0
+            })
+        );
+    }
+
+    #[test]
+    fn decode_ignores_trailing_bytes() {
```
morehouse: Trailing bytes should be decoded as TLVs, not ignored.

morehouse (changes requested):

smite/src/bolt/tx_ack_rbf.rs
devvaansh: trailing bytes are now decoded as a TLV stream instead of being ignored.

smite/src/bolt/tx_ack_rbf.rs
devvaansh: Added with tests.

smite/src/bolt/tx_init_rbf.rs
devvaansh: Added with tests.

smite/src/bolt/tx_ack_rbf.rs
```diff
@@ -0,0 +1,226 @@
...
+const TLV_FUNDING_OUTPUT_CONTRIBUTION: u64 = 0;
+
+/// TLV type for require confirmed inputs.
+const TLV_REQUIRE_CONFIRMED_INPUTS: u64 = 2;
+
+/// Known even TLV types for `tx_ack_rbf`.
+const KNOWN_EVEN: &[u64] = &[
+    TLV_FUNDING_OUTPUT_CONTRIBUTION,
+    TLV_REQUIRE_CONFIRMED_INPUTS,
+];
```
morehouse: Nit: since there's only two even TLVs, perhaps we should just inline this list where it's used.

(also applies to the other codec)

smite/src/bolt/tx_ack_rbf.rs
```diff
@@ -0,0 +1,226 @@
...
+        let funding_output_contribution =
+            if let Some(data) = stream.get(TLV_FUNDING_OUTPUT_CONTRIBUTION) {
+                if data.len() != 8 {
+                    return Err(BoltError::Truncated {
+                        expected: 8,
+                        actual: data.len(),
+                    });
+                }
+                let bytes: [u8; 8] = data.try_into().expect("length checked");
+                Some(i64::from_be_bytes(bytes))
```
morehouse: It may be worth adding a `WireFormat` impl for `i64` in this PR or as a follow-up.

smite/src/bolt/tx_ack_rbf.rs
```diff
@@ -0,0 +1,226 @@
...
+    #[test]
+    fn decode_truncated_channel_id() {
+        assert_eq!(
+            TxAckRbf::decode(&[0x00; 20]),
+            Err(BoltError::Truncated {
+                expected: CHANNEL_ID_SIZE,
+                actual: 20
+            })
+        );
+    }
```
morehouse, line 208: We should also test a truncated `funding_output_contribution` for both codecs.

smite/src/bolt/tx_ack_rbf.rs
devvaansh: Done, inlined the list directly at the [decode_with_known] call site in both codecs.

smite/src/bolt/tx_ack_rbf.rs
devvaansh: Added impl_wire_format_int!(i64) in wire.rs and updated both codecs to use i64::read() in from_stream, replacing the manual from_be_bytes logic.

smite/src/bolt/tx_ack_rbf.rs
devvaansh: Added decode_truncated_funding_output_contribution tests for both codecs.

smite/src/bolt/wire.rs
```diff
@@ -60,6 +60,7 @@ impl_wire_format_int!(u8);
 impl_wire_format_int!(u16);
 impl_wire_format_int!(u32);
 impl_wire_format_int!(u64);
+impl_wire_format_int!(i64);
```
morehouse, line 63: We also need to add tests for the new type, especially for negative values since that's new.

Also please make this change in its own commit that comes before the new codecs.

morehouse (commented):
Looks good, we just need tests for the new `WireFormat` impl.

smite/src/bolt/wire.rs
devvaansh: i have  moved the i64 impl and tests into a separate first commit and added tests for reading, writing, truncation, and roundtrips, including negative values like -1, -100000, and i64::MIN, plus a test verifying the wire encoding of negatives.

smite/src/bolt/wire.rs
```diff
@@ -357,6 +358,70 @@ mod tests {
         }
     }
 
+    #[test]
+    fn i64_read_valid() {
```
morehouse, line 362: Let's also test that the cursor advances, for consistency with the other `read_valid` tests.

Also would be good to test `i64` max and min values.

morehouse (approved):

smite/src/bolt/wire.rs
devvaansh: yupp!! i have Added cursor-advancement check (two back-to-back reads) and i64 max and min cases.

morehouse (approved):
LGTM

PR #32, opened 2026-04-04
smite: Clean up managed subprocess groups on shutdown

Ashish-Kumar-Dash:
Summary
Teach ManagedProcess to place spawned children in their own process groups and terminate the full group during shutdown. This makes cleanup more reliable for subprocess trees instead of only signaling the direct child.

Changes
create a dedicated process group for each managed subprocess at spawn time
send SIGTERM and, if needed, SIGKILL to the process group during shutdown
keep a direct-child signal fallback if process-group signaling fails
add a regression test covering child + grandchild cleanup
add a small CLN note documenting graceful RPC stop first, then managed cleanup fallback
add tempfile as a test-only dependency for the new regression test
Why
This improves teardown semantics for multi-process workloads and is directly relevant to future campaign management work where reliable stop/cleanup behavior matters.

Testing
cargo test -p smite ✅
cargo test ✅
Kindly review @morehouse & let me know if any changes are needed

### DIFF
--- a/smite/src/process.rs
+++ b/smite/src/process.rs
@@ -1,16 +1,20 @@
 //! Process management utilities for spawning and controlling subprocesses.
 
 use std::io;
+use std::os::unix::process::CommandExt;
 use std::process::{Child, Command, ExitStatus};
 use std::time::{Duration, Instant};
 
-use nix::sys::signal::{Signal, kill};
+use nix::sys::signal::{Signal, killpg};
 use nix::unistd::Pid;
 
 /// A managed subprocess with graceful shutdown support.
 ///
 /// Wraps a [`Child`] process and provides utilities for graceful shutdown
 /// (SIGTERM followed by SIGKILL after timeout).
+///
+/// Each managed child is placed into its own process group when spawned so
+/// that shutdown can reliably clean up grandchildren as well.
 pub struct ManagedProcess {
     child: Child,
     name: String,
@@ -23,7 +27,12 @@ impl ManagedProcess {
     ///
     /// Returns an error if the process fails to spawn.
     pub fn spawn(cmd: &mut Command, name: impl Into<String>) -> io::Result<Self> {
+        // Put the child in its own process group so shutdown can signal the
+        // whole subtree rather than just the direct child process.
+        cmd.process_group(0);
+
         let child = cmd.spawn()?;
+
         Ok(Self {
             child,
             name: name.into(),
@@ -61,7 +70,7 @@ impl ManagedProcess {
     ///
     /// Returns an error if sending signals or waiting fails.
     pub fn shutdown(&mut self, timeout: Duration) -> io::Result<ExitStatus> {
-        let pid = i32::try_from(self.child.id())
+        let process_group = i32::try_from(self.child.id())
             .map(Pid::from_raw)
             .map_err(|_| io::Error::other("pid exceeds i32::MAX"))?;
 
@@ -71,10 +80,18 @@ impl ManagedProcess {
             return Ok(status);
         }
 
-        // Send SIGTERM
-        log::debug!("{}: sending SIGTERM", self.name);
-        if let Err(e) = kill(pid, Signal::SIGTERM) {
-            log::warn!("{}: failed to send SIGTERM: {e}", self.name);
+        // Send SIGTERM to the entire process group so grandchildren are
+        // cleaned up too.
+        log::debug!(
+            "{}: sending SIGTERM to process group {}",
+            self.name,
+            process_group
+        );
+        if let Err(e) = killpg(process_group, Signal::SIGTERM) {
+            log::warn!(
+                "{}: failed to send SIGTERM to process group: {e}",
+                self.name
+            );
         }
 
         // Wait for process to exit with timeout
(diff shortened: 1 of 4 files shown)


smite/src/process.rs
```diff
@@ -23,10 +27,26 @@ impl ManagedProcess {
     ///
     /// Returns an error if the process fails to spawn.
     pub fn spawn(cmd: &mut Command, name: impl Into<String>) -> io::Result<Self> {
+        // Put the child in its own process group so shutdown can signal the
+        // whole subtree rather than just the direct child process.
+        unsafe {
```
morehouse: We should add a comment explaining why this block is safe.  For example:


```suggestion
        // Put the child in its own process group so shutdown can signal the
        // whole subtree rather than just the direct child process.
        //
        // SAFETY: This closure runs in the child process after fork, before exec.
        // We only call the async-signal-safe setpgid function and create io::Error
        // from last_os_error() which just stores an i32 errno.
        unsafe {
```

smite/src/process.rs
```diff
@@ -23,10 +27,26 @@ impl ManagedProcess {
...
+            use std::os::unix::process::CommandExt;
+
+            cmd.pre_exec(|| {
+                nix::unistd::setpgid(Pid::from_raw(0), Pid::from_raw(0))
+                    .map_err(io::Error::other)
```
morehouse: I'm afraid `map_err` is not signal safe here -- it heap allocates.  I think we should use the libc `setpgid` directly, so we can manage the error safely.


```suggestion
                if libc::setpgid(0, 0) == 0 {
                    Ok(())
                } else {
                    Err(io::Error::last_os_error())
                }
```

smite/src/process.rs
```diff
@@ -4,16 +4,20 @@ use std::io;
...
 ///
 /// Wraps a [`Child`] process and provides utilities for graceful shutdown
 /// (SIGTERM followed by SIGKILL after timeout).
+///
+/// Each managed child is placed into its own process group when spawned so
+/// that shutdown can reliably clean up grandchildren as well.
 pub struct ManagedProcess {
     child: Child,
     name: String,
+    process_group: Pid,
```
morehouse: It looks like we don't really need to store the `process_group`, since we can derive it in `shutdown` as before.

smite/src/process.rs
```diff
@@ -61,20 +81,24 @@ impl ManagedProcess {
...
+        // cleaned up too.
+        log::debug!(
+            "{}: sending SIGTERM to process group {}",
+            self.name,
+            self.process_group
+        );
+        if let Err(e) = killpg(self.process_group, Signal::SIGTERM) {
+            log::warn!("{}: failed to send SIGTERM to process group: {e}", self.name);
+            if let Err(child_err) = kill(self.process_group, Signal::SIGTERM) {
+                log::warn!("{}: failed to send SIGTERM to child: {child_err}", self.name);
```
morehouse: I can't think of any situations where `killpg` will fail but `kill` will succeed.  Perhaps we should just remove the fallback `kill`s.

morehouse (changes requested):
I think this change makes sense.  But we'll need to manually test it against each coverage workload to make sure it doesn't cause coverage files to be dropped.  In the past we've had to do several workarounds, especially for CLN, to make sure there's a "clean exit" that allows coverage profiles to be saved to disk.

Also please make sure CI passes before requesting the next review.  You should be able to run all the CI commands locally (e.g., `cargo test`, `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings`).

Ashish-Kumar-Dash:
Hey @morehouse, thanks for the review. I’ve updated the patch to address the feedback, switched the `pre_exec` path to use `libc::setpgid`, removed the stored `process_group` field, dropped the fallback direct-child kills, and added the requested safety comment. I also reran the local CI equivalent checks and fixed the issues they surfaced.

One question as I think about follow-up validation: for the manual coverage testing you mentioned, is there a target/workload you consider the most sensitive canary for shutdown-related coverage loss, or should I treat CLN as the primary one given its existing clean-exit handling?

smite/src/process.rs
```diff
@@ -23,7 +27,26 @@ impl ManagedProcess {
...
+
+            cmd.pre_exec(|| {
+                if libc::setpgid(0, 0) == 0 {
+                    Ok(())
+                } else {
+                    Err(io::Error::last_os_error())
+                }
+            });
+        }
+
```
morehouse, line 33: Just took a look at the docs, and looks like we can greatly simplify here.


```suggestion
        // Put the child in its own process group so shutdown can signal the
        // whole subtree rather than just the direct child process.
        cmd.process_group(0);
```

morehouse (changes requested):
Please squash and rebase before the next review.

We'll also need to manually test all the coverage workloads before landing this (CLN, LND, LDK, Eclair) to ensure they still work.

Ashish-Kumar-Dash:
Thanks, I’ve squashed the branch and rerun the checks successfully on the final commit. The earlier failures came from branch-state issues during the squash/rebase process rather than the intended process-group change itself.

morehouse (approved):
LGTM.  I also manually tested this against all 4 targets and didn't find any issues.

PR #36, opened 2026-04-05
added crate level documentation for smite core library

psg-19:
The smite crate's `lib.rs` had no documentation. This pr adds crate level doc comment explaining the library's purpose and descriptions of each module.

### DIFF
--- a/smite/src/lib.rs
+++ b/smite/src/lib.rs
@@ -1,3 +1,18 @@
+//! Core library for the smite coverage-guided fuzzing framework.
+//!
+//! Smite finds bugs in Lightning Network implementations by sending
+//! fuzz-derived protocol messages to target nodes (LND, LDK, CLN, Eclair)
+//! and checking whether they crash or violate invariants. This crate
+//! provides the building blocks that scenarios and targets are built on.
+//!
+//! # Modules
+//! - [`bolt`] - BOLT message encoding and decoding.
+//! - [`noise`] - BOLT 8 `Noise_XK` encrypted transport.
+//! - [`oracles`] - Post-scenario invariant checks.
+//! - [`process`] - Managed subprocess utilities.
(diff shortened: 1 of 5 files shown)


smite/src/lib.rs
```diff
@@ -1,3 +1,40 @@
...
+//!
+//! - [`process`] - Managed subprocess utilities.  Wraps [`std::process::Child`]
+//!   with graceful shutdown support and automatic cleanup on drop.
+//!
+//! - [`runners`] - Runner abstraction for delivering fuzz inputs.
+//!   `NyxRunner` communicates with the Nyx hypervisor for
+//!   fast snapshot-based fuzzing, `LocalRunner` reads input
+//!   from a file or stdin for crash reproduction.  `StdRunner`
+//!   auto-selects between them based on the `SMITE_NYX` environment
+//!   variable.
```
morehouse: Also it would be helpful to mention that `NyxRunner` requires the `nyx` cargo feature to be enabled.

morehouse (approved):
Thanks, this is definitely an improvement.

I think it could be even better if we move the detailed descriptions of each module to the module files themselves and then have short one-line summaries of each module here.  That's how rust docs are usually structured.

smite/src/lib.rs
psg-19: Sure, will do that

psg-19:
> Thanks, this is definitely an improvement.
> 
> I think it could be even better if we move the detailed descriptions of each module to the module files themselves and then have short one-line summaries of each module here. That's how rust docs are usually structured.

Okay, I'll do this way. 
Thanks for the review!!

morehouse (approved):
Please squash into one commit and rebase.

morehouse:
Please also follow existing commit naming conventions: e.g., `smite: add module-level documentation`

morehouse (approved):
LGTM

PR #39, opened 2026-04-06
ci: Check order of BOLT message type constants

ekzyis:
Just a suggestion. I noticed this gets mentioned a lot during review. I think the script is simple enough to understand to make it worth adding to the CI without adding maintenance burden.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -33,6 +33,9 @@ jobs:
       - name: Check Rust formatting
         run: cargo fmt --all --check
 
+      - name: Check order of BOLT message type constants
+        run: scripts/check-bolt-msg-type-order.sh
+
       - name: Run Clippy
         run: cargo clippy --all-targets --all-features -- -D warnings
 
--- a/scripts/check-bolt-msg-type-order.sh
+++ b/scripts/check-bolt-msg-type-order.sh
@@ -1,0 +1,125 @@
+#!/bin/bash
+#
+# This script is called in the CI pipeline. It makes sure that the BOLT message
+# type constants in smite/src/bolt.rs are listed in increasing order.
+# It also makes sure that other code blocks follow that order.
+
+set -eu
+set -o pipefail
+
+FILE=smite/src/bolt.rs
+
+order_pub_mod_msg_type() {
+    # awk '/start/,/end/ { print }' <file>
+    #   Print everything between the lines matching start and end
+    # grep x3
+    #   Print the last number of every line matching "pub const" (BOLT message type constant)
+    awk '/^pub mod msg_type {/,/^}/ { print }' $FILE | \
+        grep "pub const" | \
+        grep -oE '= [0-9]+;$' | \
+        grep -oE '[0-9]+'
+}
+
+echo "Checking pub mod msg_type { ... }"
+diff -u <(order_pub_mod_msg_type) <(order_pub_mod_msg_type | sort -n)
+echo "OK"
+
+# We now know the keys are in the correct order. We can now use a
+# canonical form of the keys to check if other blocks are in the same order.
+
+to_canonical() {
+    tr -d ':_ ' | tr '[:upper:]' '[:lower:]'
+}
+
+order_canonical_keys() {
+    # print all keys in pub mod msg_type { ... }
+    awk '/^pub mod msg_type {/,/^}/ { print }' $FILE | \
+        grep "pub const" | \
+        awk '{ print $3 }' | \
+        to_canonical
+}
+
+order_pub_enum_message() {
+    # awk:  print pub enum Message { ... }
+    # grep: only keep lines like "Warning(Warning),"
+    # sed:  remove everything after first '('
+    awk '/^pub enum Message {/,/^}/ { print }' $FILE | \
+        grep -E "[A-Za-z0-9]+\([A-Za-z0-9]+\)" | \
+        sed 's/(.*//' | \
+        to_canonical
+}
+
+impl_message_block() {
+    awk '/^impl Message {/,/^}/ { print }' $FILE
+}
+
+order_impl_message_msg_type() {
+    impl_message_block | \
+        awk '/^    pub fn msg_type/,/^    }/ { print }' | \
+        grep "Self::" | \
+        sed -e 's/Self:://' -e 's/(.*//' | \
+        grep -v "Unknown" | \
+        to_canonical
+}
+
+order_impl_message_encode() {
+    impl_message_block | \
+        awk '/^    pub fn encode/,/^    }/ { print }' | \
+        grep "Self::" | \
+        sed -e 's/Self:://' -e 's/(.*//' | \
+        grep -v "Unknown" | \
+        to_canonical
+}
+
+order_impl_message_decode() {
+    impl_message_block | \
+        awk '/^    pub fn decode/,/^    }/ { print }' | \
+        grep "msg_type::" | \
+        sed -e 's/msg_type:://' -e 's/ =>.*//' | \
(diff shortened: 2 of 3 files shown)


morehouse:
I like the idea, but it only helps with the initial const definition.  We still need to verify everything after that is in the desired order too.

ekzyis:
I agree regarding the other places. Will see what I can do about that.

ekzyis:
528c44a will now also check other blocks.

When everything is in the correct order:

```
$ bash .github/workflows/check-bolt-msg-type-order.sh
Checking pub mod msg_type { ... }
OK
Checking pub enum Message { ... }
OK
Checking impl Message::msg_type { ... }
OK
Checking impl Message::encode { ... }
OK
Checking impl Message::decode { ... }
OK
```

Switching order of `TxRemoveOutput` and `TxComplete`:

```diff
diff --git a/smite/src/bolt.rs b/smite/src/bolt.rs
index 1f3e174..3a2ba4c 100644
--- a/smite/src/bolt.rs
+++ b/smite/src/bolt.rs
@@ -207,8 +207,8 @@ impl Message {
             Self::Shutdown(_) => msg_type::SHUTDOWN,
             Self::OpenChannel2(_) => msg_type::OPEN_CHANNEL2,
             Self::TxRemoveInput(_) => msg_type::TX_REMOVE_INPUT,
-            Self::TxRemoveOutput(_) => msg_type::TX_REMOVE_OUTPUT,
             Self::TxComplete(_) => msg_type::TX_COMPLETE,
+            Self::TxRemoveOutput(_) => msg_type::TX_REMOVE_OUTPUT,
             Self::TxAbort(_) => msg_type::TX_ABORT,
             Self::GossipTimestampFilter(_) => msg_type::GOSSIP_TIMESTAMP_FILTER,
             Self::Unknown { msg_type, .. } => *msg_type,
```

```
$ bash .github/workflows/check-bolt-msg-type-order.sh
Checking pub mod msg_type { ... }
OK
Checking pub enum Message { ... }
OK
Checking impl Message::msg_type { ... }
--- /dev/fd/63  2026-04-07 02:23:43.137845606 +0200
+++ /dev/fd/62  2026-04-07 02:23:43.137845606 +0200
@@ -11,7 +11,7 @@
 shutdown
 openchannel2
 txremoveinput
-txcomplete
 txremoveoutput
+txcomplete
 txabort
 gossiptimestampfilter
```

The output unfortunately doesn't map to the original code. The big improvement is to let the computer check if it's out of order. Fixing the order is then easy for a human.

Also testing the order of the tests is a bit trickier. It comes down to filtering the test names to the list of message types, without making it fragile or having to keep it in sync with the source code. It's less predictable how the tests will change. Example attempt:

```
$ awk '/^mod tests {/,/^}/ { print }' smite/src/bolt.rs | grep "fn message_" | sed -e 's/fn message_//' -e 's/_roundtrip.*//' | grep -vi "unknown" | tr -d ' '
warning
init
error
ping
pong
open_channel
accept_channel
funding_created
funding_signed
channel_ready
gossip_timestamp_filter
shutdown
open_channel2
tx_remove_input
tx_remove_output
tx_complete
tx_abort
type_values(){
decode_truncated(){
with_type_helper(){
```

.github/workflows/rust.yml
```diff
@@ -33,6 +33,9 @@ jobs:
       - name: Check Rust formatting
         run: cargo fmt --all --check
 
+      - name: Check order of BOLT message type constants
+        run: .github/workflows/check-bolt-msg-type-order.sh
+
```
ekzyis, line 38: I put the job here because I consider it part of formatting, but it also means that everything below it won't run until this passes.

We could add a separate job just for formatting, so formatting issues don't hide build or test failures, but I'd just keep it as is for now and see if we'll ever want that.

scripts/check-bolt-msg-type-order.sh
```diff
@@ -0,0 +1,95 @@
...
+    # sed:  remove everything after first '('
+    awk '/^pub enum Message {/,/^}/ { print }' $FILE | \
+        grep -E "[A-Za-z0-9]+\([A-Za-z0-9]+\)" | \
+        sed 's/(.*//' | \
+        to_canonical
+}
+
+impl_message_block() {
+    awk '/^impl Message {/,/^}/ { print }' $FILE
+}
```
morehouse, line 54: The output from this function appears unused -- it's always piped to an `awk` command that ignores it and reads from `$FILE` instead.  Probably we should drop `$FILE` from those commands.

scripts/check-bolt-msg-type-order.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+#
+# This makes sure that the BOLT message type constants in smite/src/bolt.rs are
+# listed in increasing order.
+
+set -eu
```
morehouse, line 7: We should probably add `pipefail` too.
```suggestion
set -eu
set -o pipefail
```

morehouse (commented):
Perhaps we should move the script to `scripts/` so it's easier to find and run manually.

Also, I think we can use the same approach to verify that the roundtrip and "type values" tests are in the correct order.  I think it's OK to assume the tests continue to follow the current naming conventions -- we can always update the script if the naming changes in the future.

scripts/check-bolt-msg-type-order.sh
ekzyis: Oh, good catch! I did not realize it worked without it. Makes sense though, because there's no other `pub fn msg_type|encode|decode` with the same indentation in the file.

I decided to keep it, because it's meant to make sure we're matching exactly the block we want in `impl Message`, and not accidentally some other block that happens to have the same start and end (however unlikely that might ever be the case). Unlike the indented blocks, `impl Message` can only exist once with zero indentation.

I also think it makes the intent of the code clearer.

ekzyis:
> Perhaps we should move the script to scripts/ so it's easier to find and run manually.

:+1: 

> Also, I think we can use the same approach to verify that the roundtrip and "type values" tests are in the correct order. I think it's OK to assume the tests continue to follow the current naming conventions -- we can always update the script if the naming changes in the future.

Makes sense, looking into it! In draft until then.

ekzyis:
> we can use the same approach to verify that the roundtrip and "type values" tests are in the correct order

Done in 7f87089!

scripts/check-bolt-msg-type-order.sh
```diff
@@ -0,0 +1,125 @@
...
+        to_canonical
+}
+
+tests_block() {
+    awk '/^mod tests {/,/^}/' $FILE
+}
+
+order_tests() {
+    tests_block | \
+        grep -oE "fn message_[A-Za-z0-9_]+_roundtrip()" $FILE | \
```
morehouse: Same bug here:  `grep` ignores the output from `tests_block` because `$FILE` is passed to it.

morehouse: Also, I think we probably need to escape the parens so `grep` doesn't consider them a "capture group".
```suggestion
        grep -oE "fn message_[A-Za-z0-9_]+_roundtrip\(\)" $FILE | \
```

Or just omit them I guess since we currently aren't actually matching them anyway.

morehouse (approved):

scripts/check-bolt-msg-type-order.sh
ekzyis: Removed parentheses in `grep` and `sed` in 8121e94

ekzyis: Oh, thanks! Fixed in 8121e94

morehouse (approved):
LGTM

PR #40, opened 2026-04-06
refactor: Move ScenarioError into smite

ekzyis:
related: https://github.com/morehouse/smite/pull/13#issuecomment-4180370162

Despite the error name `ScenarioError`, `Scenario::new` was expecting `String` as errors. This also meant we had to map `ScenarioError` to `String` with `map_err` in a lot of places.

This moves `ScenarioError` to the smite package, and uses it instead of `String`.

To avoid a circular dependency between smite-scenarios and smite, `TargetError` is also moved.

### DIFF
--- a/smite-scenarios/src/scenarios/noise.rs
+++ b/smite-scenarios/src/scenarios/noise.rs
@@ -9,7 +9,7 @@ use smite::bolt::{Init, Message};
 use smite::noise::{
     ACT_TWO_SIZE, ENCRYPTED_LENGTH_SIZE, MAC_SIZE, NoiseCipher, NoiseConnection, NoiseHandshake,
 };
-use smite::scenarios::{Scenario, ScenarioResult};
+use smite::scenarios::{Scenario, ScenarioError, ScenarioResult, TargetError};
 
 use super::{EPHEMERAL_KEY, connect_to_target, ping_pong};
 use crate::targets::Target;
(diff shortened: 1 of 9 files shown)


smite/Cargo.toml
```diff
@@ -30,3 +30,4 @@ nix = { version = "0.30", default-features = false, features = [
   "process",
 ] }
 hex = "0.4"
+thiserror = "2"
```
morehouse, line 33: Now that `thiserror` is used in `smite/` there are probably other error types that could be simplified there in a follow-up PR.

smite-scenarios/src/scenarios/noise.rs
```diff
@@ -225,27 +225,27 @@ fn recv_message(stream: &mut TcpStream, cipher: &mut NoiseCipher) -> Vec<u8> {
...
+            TcpStream::connect_timeout(&target.addr(), TIMEOUT).map_err(TargetError::Io)?;
+        stream.set_nodelay(true).map_err(TargetError::Io)?;
         stream
             .set_read_timeout(Some(TIMEOUT))
-            .map_err(|e| e.to_string())?;
+            .map_err(TargetError::Io)?;
         stream
             .set_write_timeout(Some(TIMEOUT))
-            .map_err(|e| e.to_string())?;
+            .map_err(TargetError::Io)?;
```
morehouse, line 248: We could eliminate these remaining `map_err` by adding a `ScenarioError:Io` type.  Would be a good followup PR.

morehouse (approved):

smite/Cargo.toml
ekzyis: Done in #42

smite-scenarios/src/scenarios/noise.rs
ekzyis: Done in #43

PR #41, opened 2026-04-07
smite: Also check for Self::Target in is_timeout

ekzyis:
`ScenarioError::is_timeout` did not consider timeout errors derived from `TargetError`. This is currently not a bug, because `is_timeout` is only used to check `ping_pong` errors, and `ping_pong` always returns `ConnectionError` or `BoltError`, but it might become one.


morehouse (approved):
LGTM

PR #44, opened 2026-04-08
smite-ir: implement basic mutators

morehouse:
Implement two simple mutators:

- `OperationParamMutator` mutates a single operation, changing its parameter to some other type-compatible value.
- `InputSwapMutator` mutates a single instruction, changing one of its input variables to another variable of compatible type.

Ref: #5 (Milestone 1)

### DIFF
--- a/smite-ir/src/mutators/input_swap.rs
+++ b/smite-ir/src/mutators/input_swap.rs
@@ -1,0 +1,48 @@
+//! Mutator that swaps an instruction's input to a different variable of the
+//! same type.
+
+use rand::Rng;
+use rand::seq::IteratorRandom;
+
+use super::Mutator;
+use crate::Program;
+
+/// Swaps a randomly chosen input reference to point at a different variable of
+/// the same type.  This explores alternative data-flow paths while preserving
+/// type correctness.
+pub struct InputSwapMutator;
+
+impl Mutator for InputSwapMutator {
+    fn mutate(&self, program: &mut Program, rng: &mut impl Rng) -> bool {
+        // Pick a random (instruction index, input position) pair.
+        let Some((instr_idx, input_pos)) = program
+            .instructions
+            .iter()
+            .enumerate()
+            .flat_map(|(i, instr)| (0..instr.inputs.len()).map(move |j| (i, j)))
+            .choose(rng)
+        else {
+            return false;
+        };
+
+        let current_input = program.instructions[instr_idx].inputs[input_pos];
+        let expected_type = program.instructions[instr_idx].operation.input_types()[input_pos];
+
+        // Pick an alternative variable defined before this instruction with the
+        // matching type, excluding the current input.
+        let Some(new_input) = program.instructions[..instr_idx]
+            .iter()
+            .enumerate()
+            .filter_map(|(i, instr)| {
+                let out_type = instr.operation.output_type()?;
+                (out_type == expected_type && i != current_input).then_some(i)
+            })
+            .choose(rng)
+        else {
+            return false;
+        };
+
+        program.instructions[instr_idx].inputs[input_pos] = new_input;
+        true
+    }
+}
--- a/smite-ir/src/mutators/operation_param.rs
+++ b/smite-ir/src/mutators/operation_param.rs
@@ -1,0 +1,325 @@
+//! Mutator that tweaks embedded literal values in Load and Extract operations.
+
+use rand::seq::IteratorRandom;
+use rand::{Rng, RngExt};
+use smite::bolt::MAX_MESSAGE_SIZE;
+
+use super::Mutator;
+use crate::operation::AcceptChannelField;
+use crate::{Operation, Program};
+
+/// Mutates the embedded parameter of a randomly chosen `is_param_mutable`
+/// instruction. For numeric loads, applies a random arithmetic tweak. For byte
+/// loads, flips/adds/removes bytes. For extract operations, swaps to a random
+/// field with the same output type.
+pub struct OperationParamMutator;
+
+impl Mutator for OperationParamMutator {
+    fn mutate(&self, program: &mut Program, rng: &mut impl Rng) -> bool {
+        let Some(idx) = program
+            .instructions
+            .iter()
+            .enumerate()
+            .filter_map(|(i, instr)| instr.operation.is_param_mutable().then_some(i))
+            .choose(rng)
+        else {
+            return false;
+        };
+
+        mutate_operation(&mut program.instructions[idx].operation, rng)
+    }
+}
+
+/// Returns `true` if the operation was changed.
+fn mutate_operation(op: &mut Operation, rng: &mut impl Rng) -> bool {
+    match op {
+        Operation::LoadAmount(v) => {
+            *v = tweak_u64(*v, rng);
+            true
+        }
+        Operation::LoadFeeratePerKw(v) | Operation::LoadBlockHeight(v) => {
+            *v = tweak_u32(*v, rng);
+            true
+        }
+        Operation::LoadU16(v) => {
+            *v = tweak_u16(*v, rng);
+            true
+        }
+        Operation::LoadU8(v) => {
+            *v = tweak_u8(*v, rng);
+            true
+        }
+        Operation::LoadBytes(bytes) | Operation::LoadFeatures(bytes) => {
+            mutate_bytes(bytes, rng);
+            true
+        }
+        Operation::LoadPrivateKey(bytes) | Operation::LoadChannelId(bytes) => {
+            mutate_fixed_bytes(bytes, rng);
+            true
+        }
+        Operation::ExtractAcceptChannel(field) => mutate_extract_field(field, rng),
+
+        // Non-mutable variants. Reaching here means `is_param_mutable` and this
+        // match have drifted out of sync.
+        Operation::DerivePoint
+        | Operation::LoadTargetPubkeyFromContext
+        | Operation::LoadChainHashFromContext
+        | Operation::BuildOpenChannel
+        | Operation::SendMessage
+        | Operation::RecvAcceptChannel => {
+            unreachable!("is_param_mutable returned true for {op:?}")
+        }
+    }
+}
+
+// -- Numeric tweaks --
+
+fn tweak_u64(v: u64, rng: &mut impl Rng) -> u64 {
+    match rng.random_range(0..4) {
+        0 => rng.random(),
+        1 => v.wrapping_add(rng.random_range(1..=256)),
+        2 => v.wrapping_sub(rng.random_range(1..=256)),
+        _ => interesting_u64(rng),
+    }
+}
+
+fn tweak_u32(v: u32, rng: &mut impl Rng) -> u32 {
+    match rng.random_range(0..4) {
+        0 => rng.random(),
+        1 => v.wrapping_add(rng.random_range(1..=256)),
+        2 => v.wrapping_sub(rng.random_range(1..=256)),
+        _ => interesting_u32(rng),
+    }
+}
+
+fn tweak_u16(v: u16, rng: &mut impl Rng) -> u16 {
+    match rng.random_range(0..4) {
+        0 => rng.random(),
+        1 => v.wrapping_add(rng.random_range(1..=16)),
+        2 => v.wrapping_sub(rng.random_range(1..=16)),
+        _ => interesting_u16(rng),
+    }
+}
+
+fn tweak_u8(v: u8, rng: &mut impl Rng) -> u8 {
+    match rng.random_range(0..4) {
+        0 => rng.random(),
+        1 => v.wrapping_add(rng.random_range(1..=8)),
+        2 => v.wrapping_sub(rng.random_range(1..=8)),
+        _ => interesting_u8(rng),
+    }
+}
+
+// -- Byte mutations --
+
+/// Shuffle a random subrange of `bytes` using Fisher-Yates.
+fn shuffle_subrange(bytes: &mut [u8], rng: &mut impl Rng) {
+    if bytes.len() < 2 {
+        return;
+    }
+    let max_len = bytes.len().min(16);
+    let len = rng.random_range(2..=max_len);
+    let start = rng.random_range(0..=bytes.len() - len);
+    for i in (1..len).rev() {
+        let j = rng.random_range(0..=i);
+        bytes.swap(start + i, start + j);
+    }
+}
+
+fn mutate_bytes(bytes: &mut Vec<u8>, rng: &mut impl Rng) {
+    match rng.random_range(0..10) {
+        // Flip a random bit.
+        0 if !bytes.is_empty() => {
+            let idx = rng.random_range(0..bytes.len());
+            bytes[idx] ^= 1 << rng.random_range(0..8u8);
+        }
+        // Change a random byte.
+        1 if !bytes.is_empty() => {
+            let idx = rng.random_range(0..bytes.len());
+            bytes[idx] = rng.random();
+        }
+        // Shuffle a random subrange.
+        2 if bytes.len() >= 2 => {
+            shuffle_subrange(bytes, rng);
+        }
+        // Insert a random byte.
+        3 if bytes.len() < MAX_MESSAGE_SIZE => {
+            let pos = rng.random_range(0..=bytes.len());
+            bytes.insert(pos, rng.random());
+        }
+        // Insert repeated bytes. Requires room for at least 2 bytes.
+        4 if bytes.len() + 2 <= MAX_MESSAGE_SIZE => {
+            let pos = rng.random_range(0..=bytes.len());
+            let max_count = (MAX_MESSAGE_SIZE - bytes.len()).min(128);
+            let count = rng.random_range(2..=max_count);
+            let byte: u8 = rng.random();
+            bytes.splice(pos..pos, std::iter::repeat_n(byte, count));
+        }
+        // Remove a random byte.
+        5 if !bytes.is_empty() => {
+            let idx = rng.random_range(0..bytes.len());
+            bytes.remove(idx);
+        }
+        // Erase a byte range (up to half the length).
+        6 if bytes.len() >= 2 => {
+            let max_erase = bytes.len() / 2;
+            let count = rng.random_range(1..=max_erase);
+            let start = rng.random_range(0..=bytes.len() - count);
+            bytes.drain(start..start + count);
+        }
+        // Add or subtract a small delta from a random byte.
+        7 if !bytes.is_empty() => {
+            let idx = rng.random_range(0..bytes.len());
+            let delta = rng.random_range(1..=35);
+            if rng.random() {
+                bytes[idx] = bytes[idx].wrapping_add(delta);
+            } else {
+                bytes[idx] = bytes[idx].wrapping_sub(delta);
+            }
+        }
+        // Copy a chunk from one position to another.
+        8 if bytes.len() >= 2 => {
+            let max_len = (bytes.len() - 1).min(32);
+            let len = rng.random_range(1..=max_len);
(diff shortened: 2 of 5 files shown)


smite-ir/src/mutators/input_swap.rs
```diff
@@ -0,0 +1,51 @@
...
+            .iter()
+            .enumerate()
+            .flat_map(|(i, instr)| (0..instr.inputs.len()).map(move |j| (i, j)))
+            .collect();
+
+        if input_slots.is_empty() {
+            return false;
+        }
+
+        let &(instr_idx, input_pos) = &input_slots[rng.random_range(0..input_slots.len())];
```
Chand-ra: I think using vectors (which are allocated on the heap) here would hurt fuzzer throughput. Maybe [Reservoir Sampling](https://en.wikipedia.org/wiki/Reservoir_sampling) (an algorithm to randomly pick a value with uniform probability using $O(1)$ memory):

```suggestion
        let mut selected_slot = None;
        let mut count = 0;
        
        for (i, instr) in program.instructions.iter().enumerate() {
            for j in 0..instr.inputs.len() {
                count += 1;
                if rng.random_range(0..count) == 0 {
                    selected_slot = Some((i, j));
                }
            }
        }
        
        let (instr_idx, input_pos) = match selected_slot {
            Some(slot) => slot,
            None => return false,
        };
```
is better?

smite-ir/src/mutators/input_swap.rs
```diff
@@ -0,0 +1,51 @@
...
+        let current_input = program.instructions[instr_idx].inputs[input_pos];
+        let expected_type = program.instructions[instr_idx].operation.input_types()[input_pos];
+
+        // Find all variables defined before this instruction with the matching
+        // type, excluding the current input.
+        let alternatives: Vec<usize> = program.instructions[..instr_idx]
```
Chand-ra: Could use Reservoir Sampling here as well.

smite-ir/src/mutators/input_swap.rs
Chand-ra: Just, looked into it, `rand::seq::IteratorRandom` implements this algorithm, so it's better to use that than write boilerplate.

smite-ir/src/mutators/operation_param.rs
```diff
@@ -0,0 +1,261 @@
...
+
+/// Mutates the embedded parameter of a randomly chosen `is_param_mutable`
+/// instruction. For numeric loads, applies a random arithmetic tweak. For byte
+/// loads, flips/adds/removes bytes. For `ExtractAcceptChannel`, swaps to a
+/// random field with the same output type.
+pub struct OperationParamMutator;
+
+impl Mutator for OperationParamMutator {
+    fn mutate(&self, program: &mut Program, rng: &mut impl Rng) -> bool {
+        let mutable_indices: Vec<usize> = program
```
Chand-ra: Here as well.

smite-ir/src/tests.rs
```diff
@@ -430,3 +431,138 @@ fn append_type_mismatch_panics() {
...
+                expected_type,
+                "{field:?} has type {:?}, expected {expected_type:?}",
+                field.output_type(),
+            );
+        }
+    }
+}
+
+// -- InputSwapMutator tests --
+
```
Chand-ra, line 542: I think adding a test case for when `alternatives.is_empty()` is true is worthwhile as well:

```suggestion
#[test]
fn input_swap_returns_false_when_no_alternatives_exist() {
    // Instruction 0: outputs U64
    // Instruction 1: consumes U64 (Instruction 0's output)
    // There are no other U64s to swap to.
    let mut program = Program { /* ... */ };
    let mutator = InputSwapMutator;
    let mut rng = SmallRng::seed_from_u64(0);
    
    assert!(!mutator.mutate(&mut program, &mut rng));
}

```

Chand-ra (changes requested):
I was working on this as well, should've checked in to prevent duplicating effort 🙃.

As for the PR, it looks mostly good except for a few improvements that I could think of. I have a question though:

What is the plan/timeline for actually integrating these mutators with the fuzzer? Are we deferring that part until other parts of the Milestone 1 "minimum viable product" are completed?

smite-ir/src/mutators/operation_param.rs
```diff
@@ -0,0 +1,261 @@
...
+            .enumerate()
+            .filter_map(|(i, instr)| instr.operation.is_param_mutable().then_some(i))
+            .collect();
+
+        if mutable_indices.is_empty() {
+            return false;
+        }
+
+        let idx = mutable_indices[rng.random_range(0..mutable_indices.len())];
+        mutate_operation(&mut program.instructions[idx].operation, rng);
```
erickcestari: Shouldn't `mutate_operation` return bool? `mutate_extract_field` can be a no-op for fields with a unique output type (e.g., `TemporaryChannelId` is the sole `ChannelId`), but `mutate` still returns true.

smite-ir/src/mutators/operation_param.rs
```diff
@@ -0,0 +1,261 @@
...
+const INTERESTING_U32: &[u32] = &[
+    // u8 boundaries
+    0, 1, 0x7F, 0x80, 0xFF,
+    // u16 boundaries
+    0x100, 0x7FFF, 0x8000, 0xFFFF,
+    // u32 boundaries
+    0x1_0000, 0x7FFF_FFFF, 0x8000_0000, 0xFFFF_FFFF,
+    // BOLT constants
+    253, // min feerate in sat/kw
+];
```
erickcestari, line 287: Since `BlockHeight` is u32. We could add some interesting custom values for block height. It's probably fine to use the same values as `INTERESTING_U16`.

```
    // BOLT constants
    6,    // minimum confirmation depth
    144,  // blocks per day
    483,  // max accepted HTLCs (BOLT 2)
    2016, // blocks per 2 weeks
```

smite-ir/src/mutators/operation_param.rs
```diff
@@ -0,0 +1,261 @@
...
+            let pos = rng.random_range(0..=bytes.len());
+            bytes.insert(pos, rng.random());
+        }
+        // Insert repeated bytes.
+        4 => {
+            let pos = rng.random_range(0..=bytes.len());
+            let count = rng.random_range(2..=128);
+            let byte: u8 = rng.random();
+            bytes.splice(pos..pos, std::iter::repeat_n(byte, count));
+        }
```
erickcestari: `mutate_bytes` has no upper bound on length, so repeated insertions can grow it indefinitely. Should it cap at `MAX_MESSAGE_SIZE`?

smite-ir/src/mutators/operation_param.rs
```diff
@@ -0,0 +1,261 @@
...
+            bytes.remove(idx);
+        }
+        // Erase a byte range (up to half the length).
+        6 if bytes.len() >= 2 => {
+            let max_erase = bytes.len() / 2;
+            let count = rng.random_range(1..=max_erase);
+            let start = rng.random_range(0..=bytes.len() - count);
+            bytes.drain(start..start + count);
+        }
+        // Replace with random length and content.
```
erickcestari, line 189: We could probably add some more mutations to bytes and take inspiration from the ones used by Libfuzzer/AFL. For now, I think having more these two should be good.

AFL's arith8 and libFuzzer's CopyPart

```rust
        // Add or subtract a small delta from a random byte.
        7 if !bytes.is_empty() => {
            let idx = rng.random_range(0..bytes.len());
            let delta = rng.random_range(1..=35u8);
            if rng.random() {
                bytes[idx] = bytes[idx].wrapping_add(delta);
            } else {
                bytes[idx] = bytes[idx].wrapping_sub(delta);
            }
        }
        // Copy a chunk from one position to another.
        8 if bytes.len() >= 2 => {
            let max_len = bytes.len().min(32);
            let len = rng.random_range(1..=max_len);
            let src = rng.random_range(0..=bytes.len() - len);
            let chunk: Vec<u8> = bytes[src..src + len].to_vec();
            let dst = rng.random_range(0..=bytes.len() - len);
            bytes[dst..dst + len].copy_from_slice(&chunk);
        }
```

erickcestari (changes requested):
It's looking good for me. I've added some suggestions:

smite-ir/src/tests.rs
```diff
@@ -430,3 +431,138 @@ fn append_type_mismatch_panics() {
...
+#[test]
+fn param_mutator_changes_values() {
+    let original = generate_program(0);
+    let mut program = original.clone();
+    let mutator = OperationParamMutator;
+    let mut rng = SmallRng::seed_from_u64(0);
+
+    for _ in 0..100 {
+        mutator.mutate(&mut program, &mut rng);
```
ekzyis, line 433: Why mutate it 100 times?

Since I see that `for _ in 0..100` is used in several places—I considered to ask this question in #18 already—this is more a general question for the motivation of it. We're using deterministic RNG via seeds, so this isn't about making the tests less flaky.

smite-ir/src/tests.rs
ekzyis, line 1: I see there are tests for `OperationParamMutator` and `InputSwapMutator` returning `false` when they change nothing, but no tests for `true` when they change something.

smite-ir/src/mutators/operation_param.rs
```diff
@@ -0,0 +1,261 @@
...
+        Operation::LoadBytes(bytes) | Operation::LoadFeatures(bytes) => {
+            mutate_bytes(bytes, rng);
+        }
+        Operation::LoadPrivateKey(bytes) | Operation::LoadChannelId(bytes) => {
+            mutate_fixed_bytes(bytes, rng);
+        }
+        Operation::ExtractAcceptChannel(field) => {
+            mutate_extract_field(field, rng);
+        }
+        _ => unreachable!("is_param_mutable returned true for {op:?}"),
```
ekzyis: When we add new mutable operations, it could be easy to forget to add them here, too, not just in `is_param_mutable`. We also can't rely on tests to catch this yet: they only cover it implicitly and incompletely. For example, when I remove the `LoadBlockHeight` branch, the tests still pass, because `OpenChannelGenerator` never uses it.[^1] Should we consider this somehow?

If the test only mutates once, I can remove `LoadChannelId` too. So perhaps mutating 100 times is meant to hit many different operations implicitly (see my other comment)?

[^1]: mutation testing the mutators, haha

smite-ir/src/mutators/operation_param.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+}
+
+// -- Numeric tweaks --
+
+fn tweak_u64(v: u64, rng: &mut impl Rng) -> u64 {
+    match rng.random_range(0..4) {
+        0 => rng.random(),
+        1 => v.wrapping_add(rng.random_range(1..=256)),
+        2 => v.wrapping_sub(rng.random_range(1..=256)),
```
ekzyis, line 81: Is it intentional to have the same range here as for u32?

smite-ir/src/mutators/operation_param.rs
```diff
@@ -0,0 +1,261 @@
...
+        .copied()
+        .filter(|f| f.output_type() == target_type && f != field)
+        .collect();
+
+    if same_type.is_empty() {
+        return;
+    }
+
+    *field = same_type[rng.random_range(0..same_type.len())];
+}
```
ekzyis: This needs to become generic once we have more than just one `Extract` operation, right?

smite-ir/src/mutators/operation_param.rs
ekzyis: I think the same is true for `tweak_*`: While the `wrapping_add` and `wrapping_sub` branches make sure we tweak != 0 to not return the same value, the random and interesting branches don't. That a random branch might return the same value back is generally true, though (see `mutate_bytes`).

Are the implications of false positives worthy of consideration? Is the fuzzer just going to run the same fuzz input again, and it's not going to affect executions per second significantly?

ekzyis (changes requested):
Only questions, except one suggestion to add another test case.

smite-ir/src/mutators/input_swap.rs
morehouse: Much simpler, thanks!

Done in 05bfdbbe0bc6843cd8dc4c13ff3a6718d71b8046.

smite-ir/src/mutators/input_swap.rs
morehouse: Done in 05bfdbb.

smite-ir/src/mutators/operation_param.rs
morehouse: Done in 3c261f2119ee029f075dfe37826abc76913736a.

smite-ir/src/mutators/operation_param.rs
morehouse: Yes.  Done in b91b734a98b1a876c8ddf788e06d38c20b7700c.

smite-ir/src/mutators/operation_param.rs
morehouse: I've fixed the `mutate_extract_field` no-op return value in 9ae74e447325ec8347870a76b83254425aabcda.

For the other mutations, I think it's not worth addressing false positives from RNG choosing the same value that already exists.  The main effect of false positives is that a non-mutated program might get executed, wasting some fuzzing time.  But the odds of these random-equivalence false positives are fairly low, so I'd rather not complicate the code to handle them.

smite-ir/src/mutators/operation_param.rs
morehouse: Yes, definitely.  For now I prefer to keep it simple though.

smite-ir/src/mutators/operation_param.rs
morehouse: We can fully enumerate all match arms here -- that makes it so we get a compile error if we forget to update this when a new `Operation` is added.  We still won't compile-time detect if *existing* operations get unsynced with `is_param_mutable`, but it seems better overall.

Done in 341f919d469c3d2a9fd72b31138869b289f1a02.

smite-ir/src/mutators/operation_param.rs
morehouse: Yes, the idea was that values in the range 1-256 will generally only modify 1-2 bytes of the integer, so we have 3 types of mutations: random, small changes in the same neighborhood, and interesting values.

But this is just a heuristic, and I'm open to changing it if there's a good argument for it.

smite-ir/src/mutators/operation_param.rs
morehouse: Nice!  Done in b27f30b6904951ae17b2e65838d619e9d6f71a3.

smite-ir/src/mutators/operation_param.rs
morehouse: Added 6, 144, and 2016 in 7c3bc0add97ff493b38192eb0ce9737c91c097a.

smite-ir/src/tests.rs
morehouse: Usually `true` will mean the input changed, but it's best effort.  `false` should always mean there's no change.  This means `false` is easy to test, `true` less so.

I think `param_mutator_changes_values` and `input_swap_changes_references` already test the `true` branch as best as we can.

smite-ir/src/tests.rs
morehouse: It is to reduce flakiness.  Sure we could find a seed that causes `mutate` to succeed on the first try, but as soon as we change the `mutate` logic, that same test can start failing.  By doing 100 mutations, we can be very confident that this test should *always* pass regardless of what seed we choose.

smite-ir/src/tests.rs
morehouse: Yep!  Added in acf1b7f322e6d038e5773bdaeb9b7cf1769843d.

morehouse (commented):
CI is failing because I haven't rebased onto master yet after #45 landed.

Leaving as-is for now so everyone can more easily see the new changes.

morehouse:
> What is the plan/timeline for actually integrating these mutators with the fuzzer? Are we deferring that part until other parts of the Milestone 1 "minimum viable product" are completed?

These basic mutators are part of Milestone 1.  I'm working on a follow-up PR that implements AFL++'s custom mutator ABI using these mutators.  After that, we're pretty much ready to start implementing the first end-to-end IR scenario.

smite-ir/src/mutators/operation_param.rs
Chand-ra: Any reasons for leaving out 483? Seems like an interesting value to me.

morehouse: I only added values that are block counts, since those are generally `u32` (or `u16`) in the protocol.  483 is an HTLC count, which in the protocol is represented by `u16`.  And we already have 483 as an interesting `u16`.

smite-ir/src/tests.rs
ekzyis: Ahh, didn't think about flakiness that way. Makes sense!

smite-ir/src/mutators/operation_param.rs
```diff
@@ -0,0 +1,325 @@
...
+use smite::bolt::MAX_MESSAGE_SIZE;
+
+use super::Mutator;
+use crate::operation::AcceptChannelField;
+use crate::{Operation, Program};
+
+/// Mutates the embedded parameter of a randomly chosen `is_param_mutable`
+/// instruction. For numeric loads, applies a random arithmetic tweak. For byte
+/// loads, flips/adds/removes bytes. For `ExtractAcceptChannel`, swaps to a
+/// random field with the same output type.
```
ekzyis: nit: could make this comment generic so we don't forget to update it when we add more extract OPs

```suggestion
/// loads, flips/adds/removes bytes. For extract operations, swaps to a random
/// field with the same output type.
```

ekzyis (approved):
LGTM, just one nit

erickcestari (approved):
LGTM!

Chand-ra:
Question: There's this [TODO](https://github.com/morehouse/smite/blob/3eeb3c81fae718c65e81baf4675ee64168bb4c15/smite-ir/src/program.rs#L15) in `smite-ir/src/program.rs` that suggests mutators should "validate" a deserialized program before mutation. I don't think we do that in either of these mutators. Is it because they modify the program in-place and do not need to rebuild them using the panicky `ProgramBuilder`?

I think the TODO is confusing because it implies _all_ mutators need to use this validation.

morehouse:
> Question: There's this [TODO](https://github.com/morehouse/smite/blob/3eeb3c81fae718c65e81baf4675ee64168bb4c15/smite-ir/src/program.rs#L15) in `smite-ir/src/program.rs` that suggests mutators should "validate" a deserialized program before mutation. I don't think we do that in either of these mutators. Is it because they modify the program in-place and do not need to rebuild them using the panicky `ProgramBuilder`?

No, `validate` is not implemented in this PR.  I have it implemented in the follow-up PR that also implements the AFL++ custom mutator ABI.  Validation makes more sense to do after input deserialization and before calling mutators, not in the mutators themselves.

> I think the TODO is confusing because it implies _all_ mutators need to use this validation.

Sort of.  The validation needs to be done in the mutation phase *prior* to calling mutators.  It should make more sense once I post the next PR, and then we can remove this TODO.

Chand-ra:
> > I think the TODO is confusing because it implies _all_ mutators need to use this validation.
> 
> Sort of. The validation needs to be done in the mutation phase _prior_ to calling mutators. It should make more sense once I post the next PR, and then we can remove this TODO.

Sounds good. This was the only gripe I had. Other than this, it LGTM.

smite-ir/src/mutators/operation_param.rs
morehouse: Done.

PR #45, opened 2026-04-08
smite-ir: drop `Program::context` field

morehouse:
The context cannot be populated until scenario execution, so it really doesn't make sense to include it within `Program`.  Generators and mutators could write placeholder bytes for the context, but it would just be overwritten anyway by scenarios.

We can simplify and avoid unnecessary postcard (de)serialization by dropping this field.

### DIFF
--- a/smite-ir/src/builder.rs
+++ b/smite-ir/src/builder.rs
@@ -8,7 +8,7 @@ use std::collections::HashMap;
 
 use rand::{Rng, RngExt};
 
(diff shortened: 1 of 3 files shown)


erickcestari (approved):
LGTM! It makes more sense indeed. Should the #5 milestone issue be updated as well?

morehouse:
> Should the #5 milestone issue be updated as well?

Yes, updated the text and diagram.

PR #49, opened 2026-04-13
smite: assert Vec<u8> length <= MAX_MESSAGE_SIZE

ekzyis:
This asserts that `self.len()` does not exceed `u16::MAX` in `Vec<u8>::write`, as suggested in https://github.com/morehouse/smite/pull/27#discussion_r3041066644.

I looked at all existing calls to `Vec<u8>::write`. All expect a u16 length prefix to be written. None expect to write more than 65535 bytes.

The assertion that was added in #27 could stay as a more specific error message.

### DIFF
--- a/smite/src/bolt/error.rs
+++ b/smite/src/bolt/error.rs
@@ -1,7 +1,7 @@
 //! BOLT 1 error message.
 
 use super::BoltError;
-use super::types::{ChannelId, MAX_MESSAGE_SIZE};
+use super::types::ChannelId;
 use super::wire::WireFormat;
 
 /// BOLT 1 error message (type 17).
@@ -21,16 +21,8 @@ pub struct Error {
 
 impl Error {
     /// Creates an error for all channels.
-    ///
-    /// # Panics
-    ///
-    /// Panics if `msg` exceeds `MAX_MESSAGE_SIZE` bytes.
     #[must_use]
     pub fn all_channels(msg: &str) -> Self {
-        assert!(
-            msg.len() <= MAX_MESSAGE_SIZE,
-            "error message exceeds maximum size"
-        );
         Self {
             channel_id: ChannelId::ALL,
             data: msg.as_bytes().to_vec(),
@@ -38,16 +30,8 @@ impl Error {
     }
 
     /// Creates an error for a specific channel.
-    ///
-    /// # Panics
-    ///
-    /// Panics if `msg` exceeds `MAX_MESSAGE_SIZE` bytes.
     #[must_use]
     pub fn for_channel(channel_id: ChannelId, msg: &str) -> Self {
-        assert!(
-            msg.len() <= MAX_MESSAGE_SIZE,
-            "error message exceeds maximum size"
-        );
         Self {
             channel_id,
             data: msg.as_bytes().to_vec(),
@@ -85,7 +69,7 @@ impl Error {
 
 #[cfg(test)]
 mod tests {
-    use super::super::CHANNEL_ID_SIZE;
+    use super::super::{CHANNEL_ID_SIZE, MAX_MESSAGE_SIZE};
     use super::*;
 
     #[test]
@@ -185,16 +169,18 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "error message exceeds maximum size")]
+    #[should_panic(expected = "Vec<u8> length 65536 exceeds maximum size")]
     fn error_all_channels_too_long() {
         let long_msg = "x".repeat(MAX_MESSAGE_SIZE + 1);
-        let _ = Error::all_channels(&long_msg);
+        let err = Error::all_channels(&long_msg);
+        let _ = err.encode();
     }
 
     #[test]
-    #[should_panic(expected = "error message exceeds maximum size")]
+    #[should_panic(expected = "Vec<u8> length 65536 exceeds maximum size")]
     fn error_for_channel_too_long() {
         let long_msg = "x".repeat(MAX_MESSAGE_SIZE + 1);
-        let _ = Error::for_channel(ChannelId::ALL, &long_msg);
+        let err = Error::for_channel(ChannelId::ALL, &long_msg);
+        let _ = err.encode();
     }
 }
--- a/smite/src/bolt/shutdown.rs
+++ b/smite/src/bolt/shutdown.rs
@@ -1,7 +1,7 @@
 //! BOLT 2 shutdown message.
 
 use super::BoltError;
-use super::types::{ChannelId, MAX_MESSAGE_SIZE};
+use super::types::ChannelId;
 use super::wire::WireFormat;
 
 /// BOLT 2 shutdown message (type 38).
@@ -17,16 +17,8 @@ pub struct Shutdown {
 
 impl Shutdown {
     /// Creates a shutdown for a specific channel.
-    ///
-    /// # Panics
-    ///
-    /// Panics if `scriptpubkey` exceeds `MAX_MESSAGE_SIZE` bytes.
     #[must_use]
     pub fn for_channel(channel_id: ChannelId, scriptpubkey: Vec<u8>) -> Self {
-        assert!(
-            scriptpubkey.len() <= MAX_MESSAGE_SIZE,
-            "shutdown scriptpubkey exceeds maximum size"
-        );
         Self {
             channel_id,
             scriptpubkey,
--- a/smite/src/bolt/tx_abort.rs
+++ b/smite/src/bolt/tx_abort.rs
@@ -1,7 +1,7 @@
 //! BOLT 2 `tx_abort` message.
 
 use super::BoltError;
-use super::types::{ChannelId, MAX_MESSAGE_SIZE};
+use super::types::ChannelId;
 use super::wire::WireFormat;
 
 /// BOLT 2 `tx_abort` message (type 74).
@@ -23,16 +23,8 @@ pub struct TxAbort {
 
 impl TxAbort {
     /// Creates a `tx_abort` for the given channel.
-    ///
-    /// # Panics
-    ///
-    /// Panics if `msg` exceeds `MAX_MESSAGE_SIZE` bytes.
     #[must_use]
     pub fn new(channel_id: ChannelId, msg: &str) -> Self {
-        assert!(
-            msg.len() <= MAX_MESSAGE_SIZE,
-            "tx_abort message exceeds maximum size"
-        );
         Self {
             channel_id,
             data: msg.as_bytes().to_vec(),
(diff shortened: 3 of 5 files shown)


Chand-ra:
> The assertion that was added in #27 could stay as a more specific error message.

Not sure about this. If we think more specific panic messages are better, I'd say it's better to not implement this change at all and have the callers handle it on a case-by-case basis. 

Since the maximum size of any Lightning message cannot exceed 65535 bytes, I say we retain this change and roll back the assertion in #27.

ekzyis:
> Not sure about this. If we think more specific panic messages are better, I'd say it's better to not implement this change at all and have the callers handle it on a case-by-case basis.

The idea was to have this assertion here as a fallback. Where it makes sense, we could assert earlier to have more context in the error message for easier debugging. But no strong opinion on this. Might be some kind of unnecessary premature (debugging) optimization. We will see when we actually run into errors and have to debug something.

> Since the maximum size of any Lightning message cannot exceed 65535 bytes, I say we retain this change and roll back the assertion in https://github.com/morehouse/smite/pull/27.

Yes, I would definitely keep this change here.

Btw, this reminded me that I was wondering if I should use `MAX_MESSAGE_SIZE` here instead of `u16::MAX`. But I wasn't sure if `Vec<u8>` is meant to always be used in the context of a lightning _message_. But I guess everything we send over the wire is a lightning message, and there are no exceptions (?)

Chand-ra:
> Yes, I would definitely keep this change here.
> 
> Btw, this reminded me that I was wondering if I should use `MAX_MESSAGE_SIZE` here instead of `u16::MAX`. But I wasn't sure if `Vec<u8>` is meant to always be used in the context of a lightning _message_. But I guess everything we send over the wire is a lightning message, and there are no exceptions (?)

I think using `MAX_MESSAGE_SIZE` would be better and yes, everything sent over the wire (at least everything we're interested in fuzzing) is a lightning message and has to abide by the `len <= MAX_MESSAGE_SIZE` constraint.

ekzyis:
Ok, mentioning `MAX_MESSAGE_SIZE` instead of `u16::MAX` now.

Used "maximum size" in the error message because that's what other error messages use.

Chand-ra:
LGTM

NishantBansal2003 (changes requested):
I don't think we should panic (even as fallback), instead, we should return a `Result`, just like we do in `WireFormat::read`, and let the caller handle it themselves.

**Reasons:**

* If we are the ones adding the message field, it should be our responsibility to ensure the encoding rules are satisfied and that we get the desired result. Why should we panic in our own case?
* If this field is generated using a fuzzer, we might end up panicking frequently, which could lead to false positives.

NishantBansal2003:
But yes, this definitely needs to be updated because the initial version `(self.len() as u16).write(out);` silently truncates the length to `65536`. However, `out.extend_from_slice(self);` still writes the entire vec which could cause the full message to become inconsistent

morehouse:
I think it can make sense to panic.

>     * If we are the ones adding the message field, it should be our responsibility to ensure the encoding rules are satisfied and that we get the desired result. Why should we panic in our own case?

Because if we fail to follow encoding rules that's clearly a bug that we want to discover and fix ASAP.

>     * If this field is generated using a fuzzer, we might end up panicking frequently, which could lead to false positives.

Right, but I'd argue that's a symptom of a problem that we should solve -- we should avoid generating messages that we can't actually serialize, since that is wasted effort and hurts fuzzing efficiency.

The current generators and mutators are designed to never create byte vectors larger than `MAX_MESSAGE_SIZE` for this reason.  If they ever do, I'd prefer to have a panic so we immediately notice it and can fix it.

morehouse:
@ekzyis If we can all agree that it makes sense to panic, I think it would make sense to add another commit that removes any duplicate length checks from the various codecs.

Chand-ra:
> Because if we fail to follow encoding rules that's clearly a bug that we want to discover and fix ASAP.
> 
> > ```
> > * If this field is generated using a fuzzer, we might end up panicking frequently, which could lead to false positives.
> > ```
> 
> Right, but I'd argue that's a symptom of a problem that we should solve -- we should avoid generating messages that we can't actually serialize, since that is wasted effort and hurts fuzzing efficiency.
> 
> The current generators and mutators are designed to never create byte vectors larger than `MAX_MESSAGE_SIZE` for this reason. If they ever do, I'd prefer to have a panic so we immediately notice it and can fix it.

Agreed. I believe a big upside of snapshot fuzzing is the lower false positive per unique crash ratio compared to traditional fuzzing.

We're going the extra mile elsewhere in the repo (like disabling AFL++'s byte-level mutator) to ensure this, so adding this check makes sure we fail fast and correct early a message that violates this fundamental of the BOLT specs.

ekzyis:
> @ekzyis If we can all agree that it makes sense to panic, I think it would make sense to add another commit that removes any duplicate length checks from the various codecs.

Yes, I agree that panic makes sense here. I also consider it a programmer error. Removed the duplicate length checks in e0f5572.

I noticed two things:

1. I wasn't sure in 1bcac4f if I should add panic docs to `encode()` now?

<details>
<summary>patch</summary>

```diff
diff --git a/smite/src/bolt/error.rs b/smite/src/bolt/error.rs
index 8501210..851c769 100644
--- a/smite/src/bolt/error.rs
+++ b/smite/src/bolt/error.rs
@@ -55,6 +55,10 @@ impl Error {
     }
 
     /// Encodes to wire format (without message type prefix).
+    ///
+    /// # Panics
+    ///
+    /// Panics if `self.data` exceeds `MAX_MESSAGE_SIZE` bytes.
     #[must_use]
     pub fn encode(&self) -> Vec<u8> {
         let mut out = Vec::new();
diff --git a/smite/src/bolt/init.rs b/smite/src/bolt/init.rs
index 9f28138..c948d75 100644
--- a/smite/src/bolt/init.rs
+++ b/smite/src/bolt/init.rs
@@ -65,6 +65,11 @@ impl Init {
     }
 
     /// Encodes to wire format (without message type prefix).
+    ///
+    /// # Panics
+    ///
+    /// Panics if `self.globalfeatures` or `self.features` exceeds
+    /// `MAX_MESSAGE_SIZE` bytes.
     #[must_use]
     pub fn encode(&self) -> Vec<u8> {
         let mut out = Vec::new();
diff --git a/smite/src/bolt/ping.rs b/smite/src/bolt/ping.rs
index 0beaca1..93c08cb 100644
--- a/smite/src/bolt/ping.rs
+++ b/smite/src/bolt/ping.rs
@@ -34,6 +34,10 @@ impl Ping {
     }
 
     /// Encodes to wire format (without message type prefix).
+    ///
+    /// # Panics
+    ///
+    /// Panics if `self.ignored` exceeds `MAX_MESSAGE_SIZE` bytes.
     #[must_use]
     pub fn encode(&self) -> Vec<u8> {
         let mut out = Vec::new();
diff --git a/smite/src/bolt/pong.rs b/smite/src/bolt/pong.rs
index 4f58f39..d16d817 100644
--- a/smite/src/bolt/pong.rs
+++ b/smite/src/bolt/pong.rs
@@ -29,6 +29,10 @@ impl Pong {
     }
 
     /// Encodes to wire format (without message type prefix).
+    ///
+    /// # Panics
+    ///
+    /// Panics if `self.ignored` exceeds `MAX_MESSAGE_SIZE` bytes.
     #[must_use]
     pub fn encode(&self) -> Vec<u8> {
         let mut out = Vec::new();
diff --git a/smite/src/bolt/shutdown.rs b/smite/src/bolt/shutdown.rs
index 08d0c51..b96b3c6 100644
--- a/smite/src/bolt/shutdown.rs
+++ b/smite/src/bolt/shutdown.rs
@@ -34,6 +34,10 @@ impl Shutdown {
     }
 
     /// Encodes to wire format (without message type prefix).
+    ///
+    /// # Panics
+    ///
+    /// Panics if `self.scriptpubkey` exceeds `MAX_MESSAGE_SIZE` bytes.
     #[must_use]
     pub fn encode(&self) -> Vec<u8> {
         let mut out = Vec::new();
diff --git a/smite/src/bolt/tx_abort.rs b/smite/src/bolt/tx_abort.rs
index b161e69..a4c6153 100644
--- a/smite/src/bolt/tx_abort.rs
+++ b/smite/src/bolt/tx_abort.rs
@@ -40,6 +40,10 @@ impl TxAbort {
     }
 
     /// Encodes to wire format (without message type prefix).
+    ///
+    /// # Panics
+    ///
+    /// Panics if `self.data` exceeds `MAX_MESSAGE_SIZE` bytes.
     #[must_use]
     pub fn encode(&self) -> Vec<u8> {
         let mut out = Vec::new();
diff --git a/smite/src/bolt/warning.rs b/smite/src/bolt/warning.rs
index 37e9062..3eb7a62 100644
--- a/smite/src/bolt/warning.rs
+++ b/smite/src/bolt/warning.rs
@@ -55,6 +55,10 @@ impl Warning {
     }
 
     /// Encodes to wire format (without message type prefix).
+    ///
+    /// # Panics
+    ///
+    /// Panics if `self.data` exceeds `MAX_MESSAGE_SIZE` bytes.
     #[must_use]
     pub fn encode(&self) -> Vec<u8> {
         let mut out = Vec::new();
```

</details>

But looking at other code that might panic like `TlvStream::add()`, I think not. Panic docs don't "bubble up".

2. Currently, it is still possible that `Message::encode()` will return a message larger than 65535 bytes. My change only checks whether we would write more than 65535 bytes *at once*, but multiple writes together can still lead to a message larger than 65535 bytes. This would currently be caught by `NoiseConnection::send_message()`, which will throw `ConnectionError(MessageTooLarge)`. Is that expected? Or do we want to check the length of `out` before returning it in `Message::encode()` too?

morehouse (approved):
LGTM

morehouse:
>     1. I wasn't sure in [1bcac4f](https://github.com/morehouse/smite/commit/1bcac4fc16ef65888488dccbb9aba231564f305a) if I should add panic docs to `encode()` now?

If clippy doesn't complain, I'm fine leaving the panic docs off.

>     2. Currently, it is still possible that `Message::encode()` will return a message larger than 65535 bytes. My change only checks whether we would write more than 65535 bytes _at once_, but multiple writes together can still lead to a message larger than 65535 bytes. This would currently be caught by `NoiseConnection::send_message()`, which will throw `ConnectionError(MessageTooLarge)`. Is that expected? Or do we want to check the length of `out` before returning it in `Message::encode()` too?

I think it makes sense to return the error in `send_message`, since that's where the issue arises.  I.e. we can serialize messages longer than MAX_MESSAGE_SIZE no problem, the issue happens when we try to send the encrypted header over the noise connection (which uses `u16` for the message length).

NishantBansal2003 (approved):
Great 👍 

> The current generators and mutators are designed to never create byte vectors larger than `MAX_MESSAGE_SIZE` for this reason. If they ever do, I'd prefer to have a panic so we immediately notice it and can fix it.

Thinking about it that way makes sense -- I hadn't considered it from that perspective. I initially thought it was a way to avoid adding checks elsewhere

PR #51, opened 2026-04-14
smite: add BOLT 3 commitment transaction construction

NishantBansal2003:
Add CommitmentParams for building and signing commitment transactions for both legacy and anchor channel types, following the BOLT 3 spec for output construction, key derivation, and BIP 143 sighashing.

The main API provides:
- counterparty_commitment_signature: sign the counterparty's commitment
- verify_received_commitment_signature: verify a signature from the counterparty for the holder's commitment
- can_opener_afford_feerate: reject channels where the opener cannot cover push_sat + fee + anchor_cost from the funding amount

This will be used in the funding channel flow to exchange initial commitment signatures, and in normal channel operation once HTLC support is added.

For this, I am adding `bitcoin` as a dependency for constructing signed commitment transactions. This significantly reduces boilerplate compared to doing it manually. Since bitcoin re-exports secp256k1, drop the direct dependency to avoid duplicate versions and use the re-exported one throughout.

For the documentation part, I believe BOLT 3 is the best reference, as I followed all the instructions from there, particularly: https://github.com/lightning/bolts/blob/master/03-transactions.md#commitment-transaction-construction

For the tests, most parameters were taken from the BOLT 3 test vectors, and the rest were added using the LDK program below (slightly tweaked), so that we can keep the assertions consistent throughout the test cases
<details>
<summary>LDK Program (Old)</summary>

```rust
fn build_acceptor_commitment_and_print(
    feerate_per_kw: u32,
    funding_satoshis: u64,
    push_msat: u64,
    acceptor_dust_limit_satoshis: u64,
    opener_to_self_delay: u16,
    anchor: bool,
) {
    let secp = Secp256k1::new();
    let opener_funding_pubkey =
        pubkey("023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb");
    let acceptor_funding_pubkey =
        pubkey("030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c1");
    let opener_payment_basepoint =
        pubkey("034f355bdcb7cc0af728ef3cceb9615d90684bb5b2ca5f859ab0f0b704075871aa");
    let acceptor_payment_basepoint =
        pubkey("032c0b7cf95324a07d05398b240174dc0c2be444d96b159aa6c7f7b1e668680991");
    let opener_revocation_basepoint =
        pubkey("0212a140cd0c6539d07cd08dfe09984dec3251ea808b892efeac3ede9402bf2b19");
    let acceptor_delayed_payment_basepoint =
        pubkey("023c72addb4fdf09af94f0c94d7fe92a386a7e70cf8a1d85916386bb2535c7b1b1");
    let acceptor_per_commitment_point =
        pubkey("025f7117a78150fe2ef97db7cfc83bd57b2e2c0d0dd25eaf467a4a1c2a45ce1486");
    let opener_funding_privkey =
        secret("30ff4956bbdd3222d44cc5e8a1261dab1e07957bdac5ae88fe3261ef321f3749");

    // holder = opener (countersignatory), counterparty = acceptor (broadcaster)
    let holder_pubkeys = ChannelPublicKeys {
        funding_pubkey: opener_funding_pubkey,
        revocation_basepoint: RevocationBasepoint(opener_revocation_basepoint),
        payment_point: opener_payment_basepoint,
        delayed_payment_basepoint: DelayedPaymentBasepoint(opener_payment_basepoint), // NA
        htlc_basepoint: HtlcBasepoint(opener_payment_basepoint),                      // NA
    };

    let counterparty_pubkeys = ChannelPublicKeys {
        funding_pubkey: acceptor_funding_pubkey,
        revocation_basepoint: RevocationBasepoint(acceptor_payment_basepoint), // NA
        payment_point: acceptor_payment_basepoint,
        delayed_payment_basepoint: DelayedPaymentBasepoint(acceptor_delayed_payment_basepoint),
        htlc_basepoint: HtlcBasepoint(acceptor_payment_basepoint), // NA
    };

    let funding_outpoint = OutPoint {
        txid: "8984484a580b825b9972d7adb15050b3ab624ccd731946b3eeddb92f4e7ef6be"
            .parse()
            .unwrap(),
        index: 0,
    };
    let channel_type = if anchor {
        ChannelTypeFeatures::anchors_zero_htlc_fee_and_dependencies()
    } else {
        ChannelTypeFeatures::only_static_remote_key()
    };
    let channel_parameters = ChannelTransactionParameters {
        holder_pubkeys: holder_pubkeys.clone(),
        holder_selected_contest_delay: opener_to_self_delay,
        is_outbound_from_holder: true,
        counterparty_parameters: Some(CounterpartyChannelTransactionParameters {
            pubkeys: counterparty_pubkeys.clone(),
            selected_contest_delay: opener_to_self_delay,
        }),
        funding_outpoint: Some(funding_outpoint),
        splice_parent_funding_txid: None,
        channel_type_features: channel_type,
        channel_value_satoshis: funding_satoshis,
    };

    // Fee and balance computation done manually because LDK's `commit_tx_fee_sat`,
    //  `commitment_tx_base_weight`, and `total_anchors_sat` are not accessible from external crates.
    let push_sat = push_msat / 1000;
    let commitment_weight = if anchor {
        COMMITMENT_TX_BASE_ANCHOR_WEIGHT
    } else {
        COMMITMENT_TX_BASE_WEIGHT
    };
    let fee = feerate_per_kw as u64 * commitment_weight / 1000;
    let anchor_cost = if anchor {
        ANCHOR_OUTPUT_VALUE_SATOSHI * 2
    } else {
        0
    };
    let dust = |v: u64| {
        if v >= acceptor_dust_limit_satoshis {
            v
        } else {
            0
        }
    };

    let commitment_tx = CommitmentTransaction::new(
        INITIAL_COMMITMENT_NUMBER,
        &acceptor_per_commitment_point,
        dust(push_sat),
        dust(funding_satoshis.saturating_sub(push_sat + fee + anchor_cost)),
        feerate_per_kw,
        vec![],
        &channel_parameters.as_counterparty_broadcastable(),
        &secp,
    );

    let commit_tx = commitment_tx.trust().built_transaction();
    let funding_redeemscript =
        make_funding_redeemscript(&opener_funding_pubkey, &acceptor_funding_pubkey);
    let sighash = commit_tx.get_sighash_all(&funding_redeemscript, funding_satoshis);
    let sig = commit_tx.sign_counterparty_commitment(
        &opener_funding_privkey,
        &funding_redeemscript,
        funding_satoshis,
        &secp,
    );

    println!("sighash: {}", sighash);
    println!("signature: {}", hex::encode(sig.serialize_compact()));
    println!();
}
```
</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,10 +19,58 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
+name = "base58ck"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c8d66485a3a2ea485c1913c4572ce0256067a5377ac8c75c4960e1cda98605f"
+dependencies = [
+ "bitcoin-internals",
+ "bitcoin_hashes",
+]
+
+[[package]]
+name = "bech32"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "32637268377fc7b10a8c6d51de3e7fba1ce5dd371a96e342b34e6078db558e7f"
+
+[[package]]
+name = "bitcoin"
+version = "0.32.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e499f9fc0407f50fe98af744ab44fa67d409f76b6772e1689ec8485eb0c0f66"
+dependencies = [
+ "base58ck",
+ "bech32",
+ "bitcoin-internals",
+ "bitcoin-io",
+ "bitcoin-units",
+ "bitcoin_hashes",
+ "hex-conservative",
+ "hex_lit",
+ "secp256k1",
+]
+
+[[package]]
+name = "bitcoin-internals"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "30bdbe14aa07b06e6cfeffc529a1f099e5fbe249524f8125358604df99a4bed2"
+
+[[package]]
 name = "bitcoin-io"
 version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2dee39a0ee5b4095224a0cfc6bf4cc1baf0f9624b96b367e53b66d974e51d953"
+
+[[package]]
+name = "bitcoin-units"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5285c8bcaa25876d07f37e3d30c303f2609179716e11d688f51e8f1fe70063e2"
+dependencies = [
+ "bitcoin-internals",
+]
 
 [[package]]
 name = "bitcoin_hashes"
@@ -217,6 +265,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "hex_lit"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3011d1213f159867b13cfd6ac92d2cd5f1345762c63be3554e84092d85a50bbd"
+
+[[package]]
 name = "hkdf"
 version = "0.12.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -321,15 +375,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ppv-lite86"
-version = "0.2.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "85eae3c4ed2f50dcfe72643da4befc30deadb458a9b590d720cde2f2b1e97da9"
-dependencies = [
- "zerocopy",
-]
-
-[[package]]
 name = "proc-macro2"
 version = "1.0.106"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -355,40 +400,11 @@ checksum = "69cdb34c158ceb288df11e18b4bd39de994f6657d83847bdffdbd7f346754b0f"
 
 [[package]]
 name = "rand"
-version = "0.9.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6db2770f06117d490610c7488547d543617b21bfa07796d7a12f6f1bd53850d1"
-dependencies = [
- "rand_chacha",
- "rand_core 0.9.5",
-]
-
-[[package]]
-name = "rand"
 version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bc266eb313df6c5c09c1c7b1fbe2510961e5bcd3add930c1e31f7ed9da0feff8"
 dependencies = [
- "rand_core 0.10.0",
-]
-
-[[package]]
-name = "rand_chacha"
-version = "0.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d3022b5f1df60f26e1ffddd6c66e8aa15de382ae63b3a0c1bfc0e4d3e3f325cb"
-dependencies = [
- "ppv-lite86",
- "rand_core 0.9.5",
-]
-
-[[package]]
-name = "rand_core"
-version = "0.9.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "76afc826de14238e6e8c374ddcc1fa19e374fd8dd986b0d2af0d02377261d83c"
-dependencies = [
- "getrandom",
+ "rand_core",
 ]
 
 [[package]]
@@ -412,20 +428,19 @@ dependencies = [
 
 [[package]]
 name = "secp256k1"
-version = "0.31.1"
+version = "0.29.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c3c81b43dc2d8877c216a3fccf76677ee1ebccd429566d3e67447290d0c42b2"
+checksum = "9465315bc9d4566e1724f0fffcbcc446268cb522e60f9a27bcded6b19c108113"
 dependencies = [
  "bitcoin_hashes",
- "rand 0.9.2",
  "secp256k1-sys",
 ]
 
 [[package]]
 name = "secp256k1-sys"
-version = "0.11.0"
+version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dcb913707158fadaf0d8702c2db0e857de66eb003ccfdda5924b5f5ac98efb38"
+checksum = "d4387882333d3aa8cb20530a17c69a3752e97837832f34f6dccc760e715001d9"
 dependencies = [
  "cc",
 ]
@@ -504,12 +519,12 @@ dependencies = [
 name = "smite"
 version = "0.0.0"
 dependencies = [
+ "bitcoin",
  "chacha20poly1305",
  "hex",
  "hkdf",
  "log",
  "nix",
- "secp256k1",
  "sha2",
  "simple_logger",
  "smite-nyx-sys",
@@ -521,9 +536,9 @@ dependencies = [
 name = "smite-ir"
 version = "0.0.0"
 dependencies = [
+ "bitcoin",
  "postcard",
- "rand 0.10.0",
- "secp256k1",
+ "rand",
  "serde",
  "smite",
 ]
@@ -539,10 +554,10 @@ dependencies = [
 name = "smite-scenarios"
 version = "0.0.0"
 dependencies = [
+ "bitcoin",
  "hex",
  "libc",
  "log",
- "secp256k1",
  "serde",
  "serde_json",
  "simple_logger",
@@ -660,26 +675,6 @@ name = "wit-bindgen"
 version = "0.51.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d7249219f66ced02969388cf2bb044a09756a083d0fab1e566056b04d9fbcaa5"
-
-[[package]]
-name = "zerocopy"
-version = "0.8.33"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "668f5168d10b9ee831de31933dc111a459c97ec93225beb307aed970d1372dfd"
-dependencies = [
- "zerocopy-derive",
-]
-
-[[package]]
-name = "zerocopy-derive"
-version = "0.8.33"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c7962b26b0a8685668b671ee4b54d007a67d4eaf05fda79ac0ecf41e32270f1"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn",
-]
 
 [[package]]
 name = "zeroize"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,6 @@ warnings = "deny"
 log = "0.4"
 simple_logger = { version = "5", default-features = false }
 postcard = { version = "1.1", default-features = false, features = ["alloc"] }
-secp256k1 = { version = "0.31", features = ["std", "hashes"] }
+bitcoin = "0.32"
 serde = { version = "1", features = ["derive"] }
 thiserror = "2"
--- a/smite/src/bolt/commitment.rs
+++ b/smite/src/bolt/commitment.rs
@@ -1,0 +1,828 @@
+//! BOLT 3 commitment transaction construction and signing.
+
+use bitcoin::absolute::LockTime;
+use bitcoin::hashes::{Hash, HashEngine, sha256};
+use bitcoin::opcodes::all as opcodes;
+use bitcoin::script::Builder;
+use bitcoin::secp256k1::ecdsa::Signature;
+use bitcoin::secp256k1::{Message, PublicKey, Scalar, Secp256k1, SecretKey};
+use bitcoin::sighash::{EcdsaSighashType, SighashCache};
+use bitcoin::transaction::Version;
+use bitcoin::{
+    Amount, CompressedPublicKey, OutPoint, ScriptBuf, Sequence, Transaction, TxIn, TxOut, Witness,
+};
+
+/// Anchor output value in satoshis.
+const ANCHOR_OUTPUT_VALUE: u64 = 330;
+
+/// Base weight of a non-anchor commitment transaction without HTLCs.
+const COMMITMENT_WEIGHT_NON_ANCHOR: u64 = 724;
+
+/// Base weight of an anchor commitment transaction without HTLCs.
+const COMMITMENT_WEIGHT_ANCHOR: u64 = 1124;
+
+/// `option_anchors` feature bits (BOLT 9, bits 22/23).
+const OPTION_ANCHORS_EVEN_BIT: usize = 22;
+const OPTION_ANCHORS_ODD_BIT: usize = 23;
+
+/// Public keys and channel parameters for one side of a channel (opener or acceptor).
+pub struct ChannelPartyParams {
+    /// Funding pubkey used in the funding output.
+    pub funding_pubkey: PublicKey,
+    /// Payment basepoint used to derive the `to_remote` output key.
+    pub payment_basepoint: PublicKey,
+    /// Revocation basepoint used to derive keys that allow punishment of old states.
+    pub revocation_basepoint: PublicKey,
+    /// Delayed payment basepoint used to derive the time-locked `to_local` output key.
+    pub delayed_payment_basepoint: PublicKey,
+    /// Per-commitment point used to derive all commitment-specific keys.
+    pub per_commitment_point: PublicKey,
+    /// Minimum output value below which outputs are trimmed as dust.
+    pub dust_limit_satoshis: u64,
+    /// CSV delay this party imposes on the other's `to_local` output.
+    pub to_self_delay: u16,
+}
+
+/// Parameters for building a commitment transaction.
+pub struct CommitmentParams {
+    /// The commitment transaction number.
+    pub commitment_number: u64,
+    /// Whether the holder of this commitment is the channel opener.
+    pub holder_is_opener: bool,
+    /// Holder's funding private key.
+    pub holder_funding_privkey: SecretKey,
+    /// Funding transaction outpoint.
+    pub funding_outpoint: OutPoint,
+    /// Total channel funding amount in satoshis.
+    pub funding_satoshis: u64,
+    /// Amount pushed to the acceptor in millisatoshis.
+    pub push_msat: u64,
+    /// Fee rate for the commitment transaction.
+    pub feerate_per_kw: u32,
+    /// Opener's keys and parameters.
+    pub opener: ChannelPartyParams,
+    /// Acceptor's keys and parameters.
+    pub acceptor: ChannelPartyParams,
+    /// Channel type feature bits. The commitment format (anchor / legacy) is
+    /// derived from the bits set here.
+    pub channel_type: Vec<u8>,
+}
+
+impl CommitmentParams {
+    /// Builds the signature for the counterparty's commitment transaction.
+    #[must_use]
+    pub fn counterparty_commitment_signature(&self) -> Signature {
+        let sighash = self.build_commitment_sighash(false);
+        sign(&sighash, &self.holder_funding_privkey)
+    }
+
+    /// Verifies a signature received from the counterparty for the holder's
+    /// commitment transaction. Returns `true` if the signature is valid.
+    #[must_use]
+    pub fn verify_received_commitment_signature(&self, signature: &Signature) -> bool {
+        let sighash = self.build_commitment_sighash(true);
+        let secp = Secp256k1::new();
+        let msg = Message::from_digest(sighash);
+        let counterparty_funding_pubkey = if self.holder_is_opener {
+            &self.acceptor.funding_pubkey
+        } else {
+            &self.opener.funding_pubkey
+        };
+        secp.verify_ecdsa(&msg, signature, counterparty_funding_pubkey)
+            .is_ok()
+    }
+
+    /// Builds the signature for the holder's commitment transaction.
+    /// Only used to exercise BOLT 3 test vectors.
+    #[cfg(test)]
+    fn holder_commitment_signature(&self) -> Signature {
+        let sighash = self.build_commitment_sighash(true);
+        sign(&sighash, &self.holder_funding_privkey)
+    }
+
+    /// Builds the sighash for the commitment transaction. The commitment
+    /// format (legacy or anchor) is determined by the `channel_type`.
+    ///
+    /// When `for_holder` is true, builds the holder's commitment and
+    /// when false, builds the counterparty's.
+    #[allow(clippy::cast_possible_truncation)]
+    fn build_commitment_sighash(&self, for_holder: bool) -> [u8; 32] {
+        // Obscured commitment number.
+        let obscuring_factor = compute_obscuring_factor(
+            &self.opener.payment_basepoint,
+            &self.acceptor.payment_basepoint,
+        );
+        let obscured_commitment_number = self.commitment_number ^ obscuring_factor;
+
+        // Upper 8 bits of sequence are 0x80 and lower 24 bits are the upper 24 bits
+        // of the obscured commitment number.
+        let sequence = (0x80u32 << (8 * 3)) | ((obscured_commitment_number >> 24) as u32);
+
+        // Upper 8 bits of locktime are 0x20 and lower 24 bits are the lower 24 bits
+        // of the obscured commitment number.
+        let locktime =
+            (0x20u32 << (8 * 3)) | ((obscured_commitment_number & 0x00ff_ffff_u64) as u32);
+
+        // Build the commitment transaction
+        let outputs = self.build_commitment_outputs(for_holder);
+
+        // Witness is not included in the BIP 143 sighash, so we leave it empty.
+        let input = TxIn {
+            previous_output: self.funding_outpoint,
+            script_sig: ScriptBuf::new(),
+            sequence: Sequence::from_consensus(sequence),
+            witness: Witness::new(),
+        };
+
+        let tx = Transaction {
+            version: Version::TWO,
+            lock_time: LockTime::from_consensus(locktime),
+            input: vec![input],
+            output: outputs,
+        };
+
+        // Funding output redeem script.
+        let funding_redeemscript =
+            build_funding_redeemscript(&self.opener.funding_pubkey, &self.acceptor.funding_pubkey);
+
+        // Compute the BIP143 sighash
+        let mut cache = SighashCache::new(&tx);
+        let sighash = cache
+            .p2wsh_signature_hash(
+                0,
+                &funding_redeemscript,
+                Amount::from_sat(self.funding_satoshis),
+                EcdsaSighashType::All,
+            )
+            .expect("input index 0 is always in bounds for a single input transaction");
+
+        sighash.to_byte_array()
+    }
+
+    /// Builds the lexicographically sorted commitment outputs.
+    ///
+    /// When `for_holder` is true, builds outputs for the holder's commitment;
+    /// otherwise builds outputs for the counterparty's commitment.
+    fn build_commitment_outputs(&self, for_holder: bool) -> Vec<TxOut> {
+        let anchor = supports_option_anchors(&self.channel_type);
+
+        // Fee and balances.
+        let commitment_weight = if anchor {
+            COMMITMENT_WEIGHT_ANCHOR
+        } else {
+            COMMITMENT_WEIGHT_NON_ANCHOR
+        };
+        let fee = u64::from(self.feerate_per_kw) * commitment_weight / 1000;
+        let anchor_cost = if anchor { 2 * ANCHOR_OUTPUT_VALUE } else { 0 };
+
+        let push_sat = self.push_msat / 1000;
+        let acceptor_balance = push_sat;
+        let opener_balance = self
+            .funding_satoshis
+            .saturating_sub(push_sat)
+            .saturating_sub(fee)
+            .saturating_sub(anchor_cost);
+
+        // Map opener/acceptor to holder/counterparty for this commitment side.
+        let (to_local_value, to_remote_value) = if self.holder_is_opener == for_holder {
+            (opener_balance, acceptor_balance)
+        } else {
+            (acceptor_balance, opener_balance)
+        };
+
+        let (holder, counterparty) = if self.holder_is_opener == for_holder {
+            (&self.opener, &self.acceptor)
+        } else {
+            (&self.acceptor, &self.opener)
+        };
+
+        let mut outputs: Vec<TxOut> = Vec::new();
+
+        if to_local_value >= holder.dust_limit_satoshis {
+            let local_delayedpubkey = derive_pubkey(
+                &holder.delayed_payment_basepoint,
+                &holder.per_commitment_point,
+            );
+            let revocationpubkey = derive_revocation_pubkey(
+                &counterparty.revocation_basepoint,
+                &holder.per_commitment_point,
+            );
+
+            let to_local_spk = build_to_local_scriptpubkey(
+                &local_delayedpubkey,
+                &revocationpubkey,
+                counterparty.to_self_delay,
+            );
+
+            outputs.push(TxOut {
+                value: Amount::from_sat(to_local_value),
+                script_pubkey: to_local_spk,
+            });
+
+            if anchor {
+                outputs.push(TxOut {
+                    value: Amount::from_sat(ANCHOR_OUTPUT_VALUE),
+                    script_pubkey: build_anchor_scriptpubkey(&holder.funding_pubkey),
+                });
+            }
+        }
+        if to_remote_value >= holder.dust_limit_satoshis {
+            let to_remote_spk =
+                build_to_remote_scriptpubkey(&counterparty.payment_basepoint, anchor);
+
+            outputs.push(TxOut {
+                value: Amount::from_sat(to_remote_value),
+                script_pubkey: to_remote_spk,
+            });
+
+            if anchor {
+                outputs.push(TxOut {
+                    value: Amount::from_sat(ANCHOR_OUTPUT_VALUE),
+                    script_pubkey: build_anchor_scriptpubkey(&counterparty.funding_pubkey),
+                });
+            }
+        }
+
+        // BOLT 3 output ordering: sort by (value, script_pubkey).
+        outputs.sort_by(|a, b| {
+            a.value
+                .cmp(&b.value)
+                .then_with(|| a.script_pubkey.as_bytes().cmp(b.script_pubkey.as_bytes()))
+        });
+
+        outputs
+    }
+}
+
+/// Checks whether the opener can afford the commitment fee at the given
+/// feerate, after accounting for the pushed amount and anchor outputs.
+#[must_use]
+#[allow(clippy::similar_names)]
+pub fn can_opener_afford_feerate(
+    funding_satoshis: u64,
+    push_msat: u64,
+    feerate_per_kw: u32,
+    channel_type: &[u8],
+) -> bool {
+    let anchor = supports_option_anchors(channel_type);
+    let commitment_weight = if anchor {
+        COMMITMENT_WEIGHT_ANCHOR
+    } else {
+        COMMITMENT_WEIGHT_NON_ANCHOR
+    };
+    let fee = u64::from(feerate_per_kw) * commitment_weight / 1000;
+    let anchor_cost = if anchor { 2 * ANCHOR_OUTPUT_VALUE } else { 0 };
+    let push_sat = push_msat / 1000;
+
+    funding_satoshis
+        .checked_sub(push_sat)
+        .and_then(|balance| balance.checked_sub(fee))
+        .and_then(|balance| balance.checked_sub(anchor_cost))
+        .is_some()
+}
+
+/// Computes the commitment number obscuring factor per BOLT 3.
+fn compute_obscuring_factor(
+    opener_payment_basepoint: &PublicKey,
+    acceptor_payment_basepoint: &PublicKey,
+) -> u64 {
+    let mut sha = sha256::Hash::engine();
+
+    sha.input(&opener_payment_basepoint.serialize());
+    sha.input(&acceptor_payment_basepoint.serialize());
+    let hash = sha256::Hash::from_engine(sha).to_byte_array();
+
+    let mut buf = [0u8; 8];
+    buf[2..].copy_from_slice(&hash[26..32]);
+    u64::from_be_bytes(buf)
+}
+
+/// Checks whether `option_anchors` (BOLT 9, bits 22/23) is set in a
+/// big-endian `channel_type` feature bitfield.
+///
+/// Per BOLT 9, even bit (22) = required, odd bit (23) = optional.
+/// Either bit indicates anchor support.
+fn supports_option_anchors(channel_type: &[u8]) -> bool {
+    let byte_offset = OPTION_ANCHORS_EVEN_BIT / 8;
+    let len = channel_type.len();
+    if len <= byte_offset {
+        return false;
+    }
+
+    let required_mask = 1 << (OPTION_ANCHORS_EVEN_BIT % 8);
+    let optional_mask = 1 << (OPTION_ANCHORS_ODD_BIT % 8);
+
+    channel_type[len - 1 - byte_offset] & (required_mask | optional_mask) != 0
+}
+
+/// Derives a public key from a basepoint and per-commitment point per BOLT 3.
+fn derive_pubkey(basepoint: &PublicKey, per_commitment_point: &PublicKey) -> PublicKey {
+    let secp = Secp256k1::new();
+    let mut sha = sha256::Hash::engine();
+
+    sha.input(&per_commitment_point.serialize());
+    sha.input(&basepoint.serialize());
+    let tweak = sha256::Hash::from_engine(sha).to_byte_array();
+    let hashkey = PublicKey::from_secret_key(
+        &secp,
+        &SecretKey::from_slice(&tweak).expect("SHA256 output is a valid secret key"),
+    );
+
+    basepoint
+        .combine(&hashkey)
+        .expect("point addition of two valid pubkeys cannot produce infinity")
+}
+
+/// Derives the `revocationpubkey` per BOLT 3.
+fn derive_revocation_pubkey(
+    revocation_basepoint: &PublicKey,
+    per_commitment_point: &PublicKey,
+) -> PublicKey {
+    let secp = Secp256k1::new();
+
+    let rev_append_commit_hash_key = {
+        let mut sha = sha256::Hash::engine();
+        sha.input(&revocation_basepoint.serialize());
+        sha.input(&per_commitment_point.serialize());
+
+        sha256::Hash::from_engine(sha).to_byte_array()
+    };
+
+    let commit_append_rev_hash_key = {
+        let mut sha = sha256::Hash::engine();
+        sha.input(&per_commitment_point.serialize());
+        sha.input(&revocation_basepoint.serialize());
+
+        sha256::Hash::from_engine(sha).to_byte_array()
+    };
+
+    let revocation_contrib = revocation_basepoint
+        .mul_tweak(
+            &secp,
+            &Scalar::from_be_bytes(rev_append_commit_hash_key)
+                .expect("SHA256 output is a valid scalar"),
+        )
+        .expect("scalar multiplication of a valid pubkey cannot fail");
+
+    let commitment_contrib = per_commitment_point
+        .mul_tweak(
+            &secp,
+            &Scalar::from_be_bytes(commit_append_rev_hash_key)
+                .expect("SHA256 output is a valid scalar"),
+        )
+        .expect("scalar multiplication of a valid pubkey cannot fail");
+
+    revocation_contrib
+        .combine(&commitment_contrib)
+        .expect("point addition of two valid pubkeys cannot produce infinity")
+}
+
+/// Builds the `to_local` P2WSH `script_pubkey` per BOLT 3.
+fn build_to_local_scriptpubkey(
+    local_delayedpubkey: &PublicKey,
+    revocationpubkey: &PublicKey,
+    to_self_delay: u16,
+) -> ScriptBuf {
+    Builder::new()
+        .push_opcode(opcodes::OP_IF)
+        .push_slice(revocationpubkey.serialize())
+        .push_opcode(opcodes::OP_ELSE)
+        .push_int(i64::from(to_self_delay))
+        .push_opcode(opcodes::OP_CSV)
+        .push_opcode(opcodes::OP_DROP)
+        .push_slice(local_delayedpubkey.serialize())
+        .push_opcode(opcodes::OP_ENDIF)
+        .push_opcode(opcodes::OP_CHECKSIG)
+        .into_script()
+        .to_p2wsh()
+}
+
+/// Builds the `to_remote` output `script_pubkey` per BOLT 3.
+///
+/// With `option_anchors`, the output is P2WSH with a 1-block CSV lock.
+/// Without anchors, it is a simple P2WPKH to the remote payment basepoint.
+fn build_to_remote_scriptpubkey(payment_basepoint: &PublicKey, anchor: bool) -> ScriptBuf {
+    if anchor {
+        Builder::new()
+            .push_slice(payment_basepoint.serialize())
+            .push_opcode(opcodes::OP_CHECKSIGVERIFY)
+            .push_opcode(opcodes::OP_PUSHNUM_1)
+            .push_opcode(opcodes::OP_CSV)
+            .into_script()
+            .to_p2wsh()
+    } else {
+        ScriptBuf::new_p2wpkh(&CompressedPublicKey(*payment_basepoint).wpubkey_hash())
+    }
+}
+
+/// Builds the anchor output P2WSH `script_pubkey` per BOLT 3.
+fn build_anchor_scriptpubkey(funding_pubkey: &PublicKey) -> ScriptBuf {
+    Builder::new()
+        .push_slice(funding_pubkey.serialize())
+        .push_opcode(opcodes::OP_CHECKSIG)
+        .push_opcode(opcodes::OP_IFDUP)
+        .push_opcode(opcodes::OP_NOTIF)
+        .push_opcode(opcodes::OP_PUSHNUM_16)
+        .push_opcode(opcodes::OP_CSV)
+        .push_opcode(opcodes::OP_ENDIF)
+        .into_script()
+        .to_p2wsh()
+}
+
+/// Builds the funding output redeem script per BOLT 3.
+fn build_funding_redeemscript(pubkey1: &PublicKey, pubkey2: &PublicKey) -> ScriptBuf {
+    let key1_bytes = pubkey1.serialize();
+    let key2_bytes = pubkey2.serialize();
+    let (lesser, greater) = if key1_bytes < key2_bytes {
+        (&key1_bytes, &key2_bytes)
+    } else {
+        (&key2_bytes, &key1_bytes)
+    };
+    Builder::new()
+        .push_opcode(opcodes::OP_PUSHNUM_2)
+        .push_slice(lesser)
+        .push_slice(greater)
+        .push_opcode(opcodes::OP_PUSHNUM_2)
+        .push_opcode(opcodes::OP_CHECKMULTISIG)
+        .into_script()
+}
+
+/// Signs a commitment sighash with the given funding private key.
+fn sign(sighash: &[u8; 32], funding_privkey: &SecretKey) -> Signature {
+    let secp = Secp256k1::new();
+    let msg = Message::from_digest(*sighash);
+    secp.sign_ecdsa(&msg, funding_privkey)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn pubkey(hex_str: &str) -> PublicKey {
+        let bytes = hex::decode(hex_str).expect("valid hex");
+        PublicKey::from_slice(&bytes).expect("valid pubkey")
+    }
+
+    fn secret(hex_str: &str) -> SecretKey {
+        let bytes = hex::decode(hex_str).expect("valid hex");
+        SecretKey::from_slice(&bytes).expect("valid secret key")
+    }
+
+    fn der_sig(hex_str: &str) -> Signature {
+        let bytes = hex::decode(hex_str).expect("valid hex");
+        Signature::from_der(&bytes).expect("valid DER signature")
+    }
+
+    /// BOLT 3 Appendix C opener (local) funding private key.
+    const OPENER_FUNDING_PRIVKEY: &str =
+        "30ff4956bbdd3222d44cc5e8a1261dab1e07957bdac5ae88fe3261ef321f3749";
+
+    /// BOLT 3 Appendix C acceptor (remote) funding private key.
+    const ACCEPTOR_FUNDING_PRIVKEY: &str =
+        "1552dfba4f6cf29a62a0af13c8d6981d36d0ef8d61ba10fb0fe90da7634d7e13";
+
+    #[test]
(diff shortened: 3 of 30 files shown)


Cargo.lock
```diff
@@ -412,20 +428,19 @@ dependencies = [
 
 [[package]]
 name = "secp256k1"
-version = "0.31.1"
+version = "0.29.1"
```
ekzyis, line 393: Note: This downgrade means that we're using `from_slice` now which was [deprecated in 0.30.0](https://docs.rs/secp256k1/latest/secp256k1/struct.SecretKey.html#method.from_slice). I was curious about this, and found https://github.com/rust-bitcoin/rust-bitcoin/issues/3102.

smite/src/noise/tests.rs
```diff
@@ -23,7 +23,7 @@ fn hex_to_vec(s: &str) -> Vec<u8> {
 
 /// Helper to create a `SecretKey` from hex.
 fn secret_key(hex: &str) -> SecretKey {
-    SecretKey::from_byte_array(hex_to_array(hex)).expect("valid secret key")
+    SecretKey::from_slice(&hex_to_array::<32>(hex)).expect("valid secret key")
```
ekzyis: nit: could use `hex_to_vec` here like below

ekzyis (commented):
Only looked at first commit, lgtm. Just one nit and one note (nothing to do there). Will look at the second commit later today.

ekzyis:
> However, they are not used for direct validation, because BOLT 3 does not provide expected signatures specific to our use case. (it provides values for generic commitment txs (e.g with commitment number 42 etc. etc.)).

I don't understand this. Can't we add `commitment_number` to the params, and then use the test vector with zero HTLC outputs?

> ```
> name: simple commitment tx with no HTLCs
> to_local_msat: 7000000000
> to_remote_msat: 3000000000
> local_feerate_per_kw: 15000
> # base commitment transaction fee = 10860
> # actual commitment transaction fee = 10860
> # to_local amount 6989140 wscript 63210212a140cd0c6539d07cd08dfe09984dec3251ea808b892efeac3ede9402bf2b1967029000b2752103fd5960528dc152014952efdb702a88f71e3c1653b2314431701ec77e57fde83c68ac
> # to_remote amount 3000000 P2WPKH(032c0b7cf95324a07d05398b240174dc0c2be444d96b159aa6c7f7b1e668680991)
> remote_signature = 3045022100c3127b33dcc741dd6b05b1e63cbd1a9a7d816f37af9b6756fa2376b056f032370220408b96279808fe57eb7e463710804cdf4f108388bc5cf722d8c848d2c7f9f3b0
> # local_signature = 30440220616210b2cc4d3afb601013c373bbd8aac54febd9f15400379a8cb65ce7deca60022034236c010991beb7ff770510561ae8dc885b8d38d1947248c38f2ae055647142
> output commit_tx: 02000000000101bef67e4e2fb9ddeeb3461973cd4c62abb35050b1add772995b820b584a488489000000000038b02b8002c0c62d0000000000160014cc1b07838e387deacd0e5232e1e8b49f4c29e48454a56a00000000002200204adb4e2f00643db396dd120d4e7dc17625f5f2c11a40d857accc862d6b7dd80e04004730440220616210b2cc4d3afb601013c373bbd8aac54febd9f15400379a8cb65ce7deca60022034236c010991beb7ff770510561ae8dc885b8d38d1947248c38f2ae05564714201483045022100c3127b33dcc741dd6b05b1e63cbd1a9a7d816f37af9b6756fa2376b056f032370220408b96279808fe57eb7e463710804cdf4f108388bc5cf722d8c848d2c7f9f3b001475221023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb21030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c152ae3e195220
> num_htlcs: 0
> ```

NishantBansal2003:
> I don't understand this. Can't we add `commitment_number` to the params, and then use the test vector with zero HTLC outputs?

We can technically bypass this by passing it through params, but BOLT 3 provides the remote commitment signature (acceptor's) for the local commitment tx (opener). And, what we are building is the opposite, so the signature provided by BOLT will not match.

For testing purposes, you could swap the opener and acceptor here to align with how BOLT defines them, and then verify the signature (it should match with BOLT (I've tested this as well, and it works with a few small mirroring changes + `sign_ecdsa`)). However, since the definitions are inverted in that case, it becomes quite confusing and would require renaming several fields. That's why I chose not to include those changes

Cargo.lock
NishantBansal2003: Yes, fuzzamoto also only uses `secp256k1` re-exported from `bitcoin` instead of using it directly

Cargo.toml
```diff
@@ -19,6 +19,6 @@ warnings = "deny"
 log = "0.4"
 simple_logger = { version = "5", default-features = false }
 postcard = { version = "1.1", default-features = false, features = ["alloc"] }
-secp256k1 = { version = "0.31", features = ["std", "hashes"] }
+bitcoin = "0.32.8"
```
morehouse: Nit: let's just specify "0.32" for consistency with other deps.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+/// Computes the commitment number obscuring factor per BOLT 3.
+fn compute_obscuring_factor(
+    opener_payment_basepoint: &PublicKey,
+    acceptor_payment_basepoint: &PublicKey,
+) -> u64 {
+    let mut sha = sha256::Hash::engine();
+
+    sha.input(&opener_payment_basepoint.serialize());
+    sha.input(&acceptor_payment_basepoint.serialize());
+    let hash = sha256::Hash::from_engine(sha).to_byte_array();
```
morehouse: Existing code uses `sha2::Sha256` for hashing.  It would be nice to standardize on one library, potentially in a follow-up PR.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+}
+
+/// Checks whether `option_anchors` (BOLT 9, bits 22/23) is set.
+///
+/// Per BOLT 9, even bit (22) = required, odd bit (23) = optional.
+/// Either bit indicates anchor support.
+fn supports_option_anchors(channel_type: &[u8]) -> bool {
+    let byte_offset = OPTION_ANCHORS_EVEN_BIT / 8;
+    let required_mask = 1 << (OPTION_ANCHORS_EVEN_BIT - 8 * byte_offset);
+    let optional_mask = 1 << (OPTION_ANCHORS_ODD_BIT - 8 * byte_offset);
```
morehouse:
```suggestion
    let required_mask = 1 << (OPTION_ANCHORS_EVEN_BIT % 8);
    let optional_mask = 1 << (OPTION_ANCHORS_ODD_BIT % 8);
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+    channel_type
+        .get(byte_offset)
+        .is_some_and(|&b| b & (required_mask | optional_mask) != 0)
```
morehouse: I believe `channel_type` is actually in big-endian order, which would mean we're choosing the wrong offset here.


```suggestion
    channel_type
        .get(channel_type.len() - byte_offset - 1)
        .is_some_and(|&b| b & (required_mask | optional_mask) != 0)
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+};
+
+/// Anchor output value in satoshis.
+const ANCHOR_OUTPUT_VALUE: u64 = 330;
+
+/// Base weight of a non-anchor commitment transaction without HTLCs.
+const COMMITMENT_WEIGHT_NON_ANCHOR: u64 = 724;
+
+/// Base weight of an anchor commitment transaction without HTLCs.
+const COMMITMENT_WEIGHT_ANCHOR: u64 = 1124;
```
morehouse: Note that there are other commitment types we'll want to support in the future (e.g., taproot, 0FC).  I think it's good to keep things basic for now though.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+
+    // Acceptor's commitment: push_msat is the acceptor's balance.
+    let push_sat = params.push_msat / 1000;
+    let to_local_value = push_sat;
+    let anchor_cost = if anchor { 2 * ANCHOR_OUTPUT_VALUE } else { 0 };
+    let to_remote_value = params
+        .funding_satoshis
+        .saturating_sub(push_sat)
+        .saturating_sub(fee)
+        .saturating_sub(anchor_cost);
```
morehouse: `saturating_sub` will silently hide invalid parameters (`push_sat + fee + anchor_cost > funding_satoshis`).  *Maybe* it would make sense to surface an error in that case instead, so if the target actually accepts the invalid parameters we can detect it and report a failure?

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+fn derive_pubkey(basepoint: &PublicKey, per_commitment_point: &PublicKey) -> PublicKey {
+    let secp = Secp256k1::new();
+    let mut sha = sha256::Hash::engine();
+
+    sha.input(&per_commitment_point.serialize());
+    sha.input(&basepoint.serialize());
+    let tweak = sha256::Hash::from_engine(sha).to_byte_array();
+    let hashkey = PublicKey::from_secret_key(&secp, &SecretKey::from_slice(&tweak).unwrap());
+
+    basepoint.combine(&hashkey).unwrap()
```
morehouse: Nit: replace `unwrap` throughout this file with `expect` to better document the reason this is safe

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+        let local_delayedpubkey = derive_pubkey(
+            &params.acceptor_delayed_payment_basepoint,
+            &params.acceptor_per_commitment_point,
+        );
+        let revocationpubkey = derive_revocation_pubkey(
+            &params.opener_revocation_basepoint,
+            &params.acceptor_per_commitment_point,
+        );
+
+        let to_local_spk = build_to_local_redeemscript(
```
morehouse: It's a `scriptPubkey`, not a redeem script.


```suggestion
        let to_local_spk = build_to_local_scriptpubkey(
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+
+        outputs.push(TxOut {
+            value: Amount::from_sat(to_local_value),
+            script_pubkey: to_local_spk,
+        });
+
+        if anchor {
+            outputs.push(TxOut {
+                value: Amount::from_sat(ANCHOR_OUTPUT_VALUE),
+                script_pubkey: build_anchor_redeemscript(&params.acceptor_funding_pubkey),
```
morehouse:
```suggestion
                script_pubkey: build_anchor_scriptpubkey(&params.acceptor_funding_pubkey),
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+            });
+        }
+    }
+    if to_remote_value >= params.acceptor_dust_limit_satoshis {
+        let to_remote_spk = if anchor {
+            build_to_remote_anchor_redeemscript(&params.opener_payment_basepoint)
```
morehouse: This name is quite confusing:  it sounds like a remote anchor output script.  Perhaps we should refactor to `build_to_remote_scriptpubkey(basepoint, anchor)` and move the two cases into that function.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+    let outputs = build_commitment_outputs(params);
+
+    let input = TxIn {
+        previous_output: OutPoint {
+            txid: params.funding_txid,
+            vout: u32::from(params.funding_output_index),
+        },
+        script_sig: ScriptBuf::new(),
+        sequence: Sequence::from_consensus(sequence),
+        witness: Witness::new(),
```
morehouse: I think a comment would be helpful explaining that the `witness` is not used in the sighash calculation, which is why we can leave it empty here.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+        .push_opcode(opcodes::OP_PUSHNUM_16)
+        .push_opcode(opcodes::OP_CSV)
+        .push_opcode(opcodes::OP_ENDIF)
+        .into_script()
+        .to_p2wsh()
+}
+
+/// Builds the funding output redeem script per BOLT 3.
+#[must_use]
+fn make_funding_redeemscript(pubkey1: &PublicKey, pubkey2: &PublicKey) -> ScriptBuf {
```
morehouse: Nit: why is this function called `make_*` instead of `build_*` like the other ones?

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+        .push_opcode(opcodes::OP_CHECKMULTISIG)
+        .into_script()
+}
+
+/// Signs a commitment sighash with the opener's funding private key.
+#[must_use]
+pub fn sign_commitment(sighash: &[u8; 32], opener_funding_privkey: &SecretKey) -> Signature {
+    let secp = Secp256k1::new();
+    let msg = Message::from_digest(*sighash);
+    secp.sign_ecdsa_low_r(&msg, opener_funding_privkey)
```
morehouse: Do we need low-R signatures?  I believe the BOLT 3 test vectors don't enforce low-R, so if we want to use those, we'll need `sign_ecdsa` anyway.  Also there's a slight performance improvement for `sign_ecdsa`.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+        let factor =
+            compute_obscuring_factor(&opener_payment_basepoint, &acceptor_payment_basepoint);
+        assert_eq!(factor, 0x2bb0_3852_1914);
+    }
+
+    #[test]
+    fn supports_option_anchors_detection() {
+        // Required (bit 22), optional (bit 23).
+        assert!(supports_option_anchors(&[0x00, 0x00, 0x40]));
+        assert!(supports_option_anchors(&[0x00, 0x00, 0x80]));
```
morehouse: I believe these test cases are actually reversed from how `channel_type` is decoded from the wire.  We need to fix that, and I think it would make sense to also test that the incorrect byte order does not support anchors.


```suggestion
        // Required (bit 22), optional (bit 23).
        assert!(supports_option_anchors(&[0x40, 0x00, 0x00]));
        assert!(supports_option_anchors(&[0x80, 0x00, 0x00]));
        assert!(!supports_option_anchors(&[0x00, 0x00, 0x40]));
        assert!(!supports_option_anchors(&[0x00, 0x00, 0x80]));
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,583 @@
...
+/// Builds the funding output redeem script per BOLT 3.
+#[must_use]
+fn make_funding_redeemscript(pubkey1: &PublicKey, pubkey2: &PublicKey) -> ScriptBuf {
+    let key1_bytes = pubkey1.serialize();
+    let key2_bytes = pubkey2.serialize();
+    let (lesser, greater) = if key1_bytes < key2_bytes {
+        (&key1_bytes, &key2_bytes)
+    } else {
+        (&key2_bytes, &key1_bytes)
+    };
```
morehouse, line 550: We should add a test for this ordering.

morehouse (changes requested):
Concept ACK.

I think it would be better if we can generalize this code more, so we can handle any commitment number and either the local *or* remote commitments, with a `local_is_opener` bool to specify which party is the opener/funder.  At first we will only be using commitment 0 for the remote with `local_is_opener = true`, but eventually we'll need to sign other commitments as well.  A huge benefit of generalizing is that we can directly use all the BOLT 3 test vectors without needing to adapt them.

We will also eventually want HTLC support, though we can probably wait and do that later.

I also think it may be helpful to rework the API a bit and make it more object-oriented.  Perhaps something like this:

```rust
impl CommitmentParams {
  pub fn local_commitment_signature(&self, commitment_number: u64) -> Signature;
  pub fn remote_commitment_signature(&self, commitment_number: u64) -> Signature;
}
```

Also a nit: the commit messages could have more description about the reason for the changes.  e.g., the `bitcoin` dep is needed for constructing signed commitments.

smite/src/bolt/commitment.rs
NishantBansal2003: Hmm, in BOLT 2 there was no mention of endianness, but I later found a general mention in BOLT 1 (https://github.com/lightning/bolts/blob/master/01-messaging.md#overview) after confirming it from LND and CLN (that channel_type is actually in big-endian order)

smite/src/bolt/commitment.rs
NishantBansal2003: I think it would be better to have a separate public API like `can_opener_afford_feerate` that returns a bool based on `push_sat + fee + anchor_cost > funding_satoshis`, so we can check this before building the commitment

smite/src/bolt/commitment.rs
NishantBansal2003: No particular reason, I just picked the name from LDK, But we can switch to `build_*` to keep things consistent

NishantBansal2003:
Addressed the comments in the latest rebase and made things a bit more general for both the opener and acceptor sides

For the BOLT 3 test vectors, I used all those provided (and relevant currently for us) by BOLT and added the rest using the above LDK program (slightly tweaked, let me know if you need that program), so we can keep the assertions consistent throughout the test cases

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,828 @@
...
+    /// Per-commitment point used to derive all commitment-specific keys.
+    pub per_commitment_point: PublicKey,
+    /// Minimum output value below which outputs are trimmed as dust.
+    pub dust_limit_satoshis: u64,
+    /// CSV delay this party imposes on the other's `to_local` output.
+    pub to_self_delay: u16,
+}
+
+/// Parameters for building a commitment transaction.
+pub struct CommitmentParams {
```
morehouse: I think we'll want to split the "channel config" parameters from the "per commitment" state.  The channel config will be the *static* parameters that don't change, while the per-commitment state will change for every commitment generated.  I think this will create a cleaner API for our future fuzzing scenarios.

Something like this:

```rust
pub struct ChannelConfig {
  pub funding_outpoint: OutPoint,
  pub funding_satoshis: u64,
  pub channel_type: Vec<u8>,

  // ChannelPartyConfig is the current ChannelPartyParams without per_commitment_point
  pub opener: ChannelPartyConfig,
  pub acceptor: ChannelPartyConfig,
}

pub struct CommitmentState {
  pub commitment_number: u64,
  pub feerate_per_kw: u32,
  pub opener_per_commitment_point: PublicKey,
  pub acceptor_per_commitment_point: PublicKey,
  pub to_opener_msat: u64,
  pub to_acceptor_msat: u64,

  // In the future, HTLCs go here.
}

pub struct LocalIdentity {
  pub side: Side, // Opener or Acceptor
  pub funding_privkey: SecretKey,
}

impl ChannelConfig {
  pub fn sign_counterparty_commitment(&self, state: &CommitmentState, ident: &LocalIdentity) -> Signature;
  pub fn verify_counterparty_signature(&self, state: &CommitmentState, ident: &LocalIdentity, sig: &Signature) -> bool;
}
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,828 @@
...
+    /// Whether the holder of this commitment is the channel opener.
+    pub holder_is_opener: bool,
+    /// Holder's funding private key.
+    pub holder_funding_privkey: SecretKey,
+    /// Funding transaction outpoint.
+    pub funding_outpoint: OutPoint,
+    /// Total channel funding amount in satoshis.
+    pub funding_satoshis: u64,
+    /// Amount pushed to the acceptor in millisatoshis.
+    pub push_msat: u64,
```
morehouse: To make this API more general, we should probably remove `push_msat` since that's only applicable to the initial funding.  We can get the same functionality with `to_local_msat` and `to_remote_msat`.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,828 @@
...
+    /// Opener's keys and parameters.
+    pub opener: ChannelPartyParams,
+    /// Acceptor's keys and parameters.
+    pub acceptor: ChannelPartyParams,
+    /// Channel type feature bits. The commitment format (anchor / legacy) is
+    /// derived from the bits set here.
+    pub channel_type: Vec<u8>,
+}
+
+impl CommitmentParams {
```
morehouse: Thinking roughly about the funding flow, we'll probably want to expose more functions in the API.

For example, the `funding_created` message will require us to get the funding txid.  And we'll actually need the fully-constructed transaction so we can mine it.  It would be good to start prototyping a full funding-flow scenario to help nail down the initial API.

In the future, we'll need to extend the API further:
- We'll need a way to get the holder's commitment signature for testing force closes.
- We'll need ways to construct HTLC transactions and get signatures.
- etc.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,828 @@
...
+        // Obscured commitment number.
+        let obscuring_factor = compute_obscuring_factor(
+            &self.opener.payment_basepoint,
+            &self.acceptor.payment_basepoint,
+        );
+        let obscured_commitment_number = self.commitment_number ^ obscuring_factor;
+
+        // Upper 8 bits of sequence are 0x80 and lower 24 bits are the upper 24 bits
+        // of the obscured commitment number.
+        let sequence = (0x80u32 << (8 * 3)) | ((obscured_commitment_number >> 24) as u32);
```
morehouse: Nit:  I know it *shouldn't* happen, but if the `commitment_number` is ever more than 48 bits we get silent truncation in this function.  Perhaps we should use `u32::try_from().expect()` throughout this function to make sure we panic if it ever happens.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,828 @@
...
+        // Fee and balances.
+        let commitment_weight = if anchor {
+            COMMITMENT_WEIGHT_ANCHOR
+        } else {
+            COMMITMENT_WEIGHT_NON_ANCHOR
+        };
+        let fee = u64::from(self.feerate_per_kw) * commitment_weight / 1000;
+        let anchor_cost = if anchor { 2 * ANCHOR_OUTPUT_VALUE } else { 0 };
+
+        let push_sat = self.push_msat / 1000;
```
morehouse: I think this premature conversion will lead to rounding issues and incorrect commitment transactions.

Suppose `funding_satoshis = 1` and `push_msat = 999`.  The expected initial balances would be `opener_balance_msat = 1` and `accepter_balance_msat = 999`.  When constructing the commitment transaction, both balances get rounded down to the nearest satoshi, so we should get `to_opener = 0` and `to_acceptor = 0` with the extra 1 sat allocated to fees.

But due to the premature conversion to sats, what actually happens here is `to_opener = 1` and `to_acceptor = 0` with no extra allocated to fees.

We should also add a test that detects this.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,828 @@
...
+            holder_is_opener: false,
+            holder_funding_privkey: secret(ACCEPTOR_FUNDING_PRIVKEY),
+            ..bolt3_commitment_params(15_000, 0, vec![])
+        };
+        assert!(acceptor_params.verify_received_commitment_signature(&acceptor_commit_sig));
+    }
+
+    #[test]
+    fn simple_commitment_tx_with_no_htlcs_and_single_remote_output_legacy() {
+        let opener_params = bolt3_commitment_params(9_667_817, 3_000_000_000, vec![]);
```
morehouse: The parameters in this test differ from the Appendix C test vectors.  In general it would be good to add comments to all the tests indicating which test vectors they correspond to, and if we are modifying a test vector we should explain why.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,828 @@
...
+            holder_funding_privkey: secret(ACCEPTOR_FUNDING_PRIVKEY),
+            ..bolt3_commitment_params(15_000, 0, vec![0x40, 0x00, 0x00])
+        };
+        assert!(acceptor_params.verify_received_commitment_signature(&acceptor_commit_sig));
+    }
+
+    #[test]
+    fn simple_commitment_tx_with_no_htlcs_and_single_remote_output_anchor() {
+        let opener_params =
+            bolt3_commitment_params(9_667_817, 3_000_000_000, vec![0x40, 0x00, 0x00]);
```
morehouse: The parameters in this test differ from the Appendix F test vectors

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,828 @@
...
+        let acceptor_params = CommitmentParams {
+            holder_is_opener: false,
+            holder_funding_privkey: secret(ACCEPTOR_FUNDING_PRIVKEY),
+            ..bolt3_commitment_params(15_000, 3_000_000_000, vec![])
+        };
+        assert!(acceptor_params.verify_received_commitment_signature(&acceptor_commit_sig));
+    }
+
+    #[test]
+    fn simple_commitment_tx_with_no_htlcs_and_single_local_output_legacy() {
```
morehouse: What is being used as the oracle for this test?

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,828 @@
...
+
+        // Opener signs the acceptor's commitment, then the acceptor verifies it.
+        let acceptor_commit_sig = opener_params.counterparty_commitment_signature();
+        let acceptor_params = CommitmentParams {
+            holder_is_opener: false,
+            holder_funding_privkey: secret(ACCEPTOR_FUNDING_PRIVKEY),
+            ..bolt3_commitment_params(9_667_817, 3_000_000_000, vec![0x40, 0x00, 0x00])
+        };
+        assert!(acceptor_params.verify_received_commitment_signature(&acceptor_commit_sig));
+    }
```
morehouse, line 1237: I think we can also support the Appendix F "commitment tx with two outputs untrimmed (minimum dust limit)" test with the current code.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,828 @@
...
+        ));
+
+        // Fails: push + fee fit but anchor cost does not
+        assert!(!can_opener_afford_feerate(
+            17_500,
+            0,
+            feerate_per_kw,
+            anchors
+        ));
+    }
```
morehouse, line 1348: Some other tests that would be valuable:

- Using non-identical basepoints (detects if we mix up basepoints in our construction)
- `push_msat % 1000 != 0` bug mentioned above
- equal output amounts (should be sorted by scriptPubkey)

If we need oracles for any new tests, we could generate them using LDK (just need to document it).

morehouse (changes requested):

smite/src/bolt/commitment.rs
NishantBansal2003: Isn't `to_local_msat` and `to_remote_msat` in the BOLT 3 test vectors wrong? It mentions (https://github.com/lightning/bolts/blob/master/03-transactions.md#appendix-f-commitment-and-htlc-transaction-test-vectors-anchors) that both are before subtracting any fees, anchor outputs, and pending HTLCs, but still both offered and received HTLCs seem to have been subtracted from `to_local_msat`

smite/src/bolt/commitment.rs
NishantBansal2003: Yes, since it belongs to a different part of the funding flow, I thought of separating it from the commitment tx part, So, I will add it in a follow up PR

smite/src/bolt/commitment.rs
NishantBansal2003: hmm, so this will always cause an off by one issue when `push_msat % 1000 != 0`. Also, since now `funding_satoshis` can be any value, this won't be ruled out by the minimum channel capacity checks in LN. I also checked CLN and LDK, and both handle this conversion and subtraction in msat, so I'll update this

smite/src/bolt/commitment.rs
NishantBansal2003: Yes, because of the initial `push_msat` and `funding_satoshis` approach, I wasn't able to match the balance with BOLT (without being hacky), and the signature didn't match. But I think now, with the `to_local_msat` and `to_remote_msat` approach, we can include all those test vectors as well

smite/src/bolt/commitment.rs
NishantBansal2003: Yeah, I calculated this manually so we could add a case where `local_value < dust limit`. But now, with the `to_local_msat` and `to_remote_msat` approach, we can fully use the BOLT 3 test vectors, which will provide the same coverage. Thanks for this approach

smite/src/bolt/commitment.rs
NishantBansal2003: LDK program shared in the PR description (with a minor mirroring tweak since the case is inversed)

<details>
<summary>diff from LDK program in PR desc (for holder_commitment_signature case)</summary>

```rust
    let commitment_tx = CommitmentTransaction::new(
        INITIAL_COMMITMENT_NUMBER - 42,
        &per_commitment_point,
        dust((opener_msat / 1000).saturating_sub(fee + anchor_cost)),
        dust(push_sat),
        feerate_per_kw,
        vec![],
        &channel_parameters.as_holder_broadcastable(),
        &secp,
    );

    let commit_tx = commitment_tx.trust().built_transaction();
    let funding_redeemscript =
        make_funding_redeemscript(&opener_funding_pubkey, &acceptor_funding_pubkey);
    let sighash = commit_tx.get_sighash_all(&funding_redeemscript, 10_000_000);
    let sig = secp.sign_ecdsa(&sighash, &opener_funding_privkey);

    println!("signature: {}", hex::encode(sig.serialize_der()));
    println!();
```
</details>

And, the same program (with some tweaks) exists for remote signing as well. Let me know if you need the complete both program to validate against

smite/src/bolt/commitment.rs
NishantBansal2003: FYI, and for the opener signing its own commitment case, the program I used is https://github.com/morehouse/smite/pull/51#discussion_r3183586971
<details>
<summary>LDK program used for validation (acceptor signs opener’s commitment case)</summary>

```rust
    let counterparty_pubkeys = ChannelPublicKeys {
        funding_pubkey: opener_funding_pubkey,
        revocation_basepoint: RevocationBasepoint(opener_revocation_basepoint),
        payment_point: opener_payment_basepoint,
        delayed_payment_basepoint: DelayedPaymentBasepoint(opener_delayed_payment_basepoint),
        htlc_basepoint: HtlcBasepoint(opener_delayed_payment_basepoint), // NA
    };

    let holder_pubkeys = ChannelPublicKeys {
        funding_pubkey: acceptor_funding_pubkey,
        revocation_basepoint: RevocationBasepoint(acceptor_revocation_basepoint),
        payment_point: acceptor_payment_basepoint,
        delayed_payment_basepoint: DelayedPaymentBasepoint(acceptor_delayed_payment_basepoint),
        htlc_basepoint: HtlcBasepoint(acceptor_revocation_basepoint), // NA
    };

    let funding_outpoint = OutPoint {
        txid: "8984484a580b825b9972d7adb15050b3ab624ccd731946b3eeddb92f4e7ef6be"
            .parse()
            .unwrap(),
        index: 0,
    };
    let channel_type = if anchor {
        ChannelTypeFeatures::anchors_zero_htlc_fee_and_dependencies()
    } else {
        ChannelTypeFeatures::only_static_remote_key()
    };
    let channel_parameters = ChannelTransactionParameters {
        holder_pubkeys: holder_pubkeys.clone(),
        holder_selected_contest_delay: opener_to_self_delay,
        is_outbound_from_holder: false,
        counterparty_parameters: Some(CounterpartyChannelTransactionParameters {
            pubkeys: counterparty_pubkeys.clone(),
            selected_contest_delay: opener_to_self_delay,
        }),
        funding_outpoint: Some(funding_outpoint),
        splice_parent_funding_txid: None,
        channel_type_features: channel_type,
        channel_value_satoshis: 10_000_000,
    };

    // Fee and balance computation done manually because LDK's `commit_tx_fee_sat`,
    //  `commitment_tx_base_weight`, and `total_anchors_sat` are not accessible from external crates.
    let push_sat = acceptor_msat / 1000;
    let commitment_weight = if anchor {
        COMMITMENT_TX_BASE_ANCHOR_WEIGHT
    } else {
        COMMITMENT_TX_BASE_WEIGHT
    };
    let fee = feerate_per_kw as u64 * commitment_weight / 1000;
    let anchor_cost = if anchor {
        ANCHOR_OUTPUT_VALUE_SATOSHI * 2
    } else {
        0
    };
    let dust = |v: u64| {
        if v >= acceptor_dust_limit_satoshis {
            v
        } else {
            0
        }
    };

    let commitment_tx = CommitmentTransaction::new(
        INITIAL_COMMITMENT_NUMBER - 42,
        &per_commitment_point,
        dust((opener_msat / 1000).saturating_sub(fee + anchor_cost)),
        dust(push_sat),
        feerate_per_kw,
        vec![],
        &channel_parameters.as_counterparty_broadcastable(),
        &secp,
    );

    let commit_tx = commitment_tx.trust().built_transaction();
    let funding_redeemscript =
        make_funding_redeemscript(&opener_funding_pubkey, &acceptor_funding_pubkey);
    let sig = commit_tx.sign_counterparty_commitment(
        &acceptor_funding_privkey,
        &funding_redeemscript,
        10000000,
        &secp,
    );

    println!("signature: {}", hex::encode(sig.serialize_der()));
    println!();
}
```

</details>

NishantBansal2003:
Addressed the comments in https://github.com/morehouse/smite/pull/51/commits/e77a38e1e26f5a4b83d70736f167a56d6b8f4818
With the new approach, I was able to include all non-htlc BOLT 3 test vectors with the same coverage. I also added some TODOs to extend the current flow for HTLC support. I haven't fully thought through the HTLC design yet, but I’ve outlined where things could fit when we add support. Let me know if that makes sense.

ekzyis:
If @morehouse agrees, I think it would help with review to put the first commit into its own PR and merge it soon. 28/30 files were only touched because of it.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1342 @@
...
+use bitcoin::transaction::Version;
+use bitcoin::{
+    Amount, CompressedPublicKey, OutPoint, ScriptBuf, Sequence, Transaction, TxIn, TxOut, Witness,
+};
+
+/// Anchor output value in satoshis.
+const ANCHOR_OUTPUT_VALUE: u64 = 330;
+
+/// Base weight of a non-anchor commitment transaction without HTLCs.
+const COMMITMENT_WEIGHT_NON_ANCHOR: u64 = 724;
```
ekzyis: nit: I found "base weight" confusing. bolt03 doesn't mention "base weight", only "base fee", which [doesn't seem to be related to "without HTLCs" but "trimmed HTLCs"](https://github.com/lightning/bolts/blob/444805d12ab98c30006173bb190cd9d6fce9e405/03-transactions.md?plain=1#L517-L524). I think with "base" you mean "without HTLCs"? If so, I suggest this change:

```diff
-/// Base weight of a non-anchor commitment transaction without HTLCs.
-const COMMITMENT_WEIGHT_NON_ANCHOR: u64 = 724;
+/// Weight of a non-anchor commitment transaction without HTLCs.
+const BASE_COMMITMENT_WEIGHT_NON_ANCHOR: u64 = 724;

-/// Base weight of an anchor commitment transaction without HTLCs.
-const COMMITMENT_WEIGHT_ANCHOR: u64 = 1124;
+/// Weight of an anchor commitment transaction without HTLCs.
+const BASE_COMMITMENT_WEIGHT_ANCHOR: u64 = 1124;
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1342 @@
...
+
+/// Base weight of an anchor commitment transaction without HTLCs.
+const COMMITMENT_WEIGHT_ANCHOR: u64 = 1124;
+
+/// `option_anchors` feature bits (BOLT 9, bits 22/23).
+const OPTION_ANCHORS_EVEN_BIT: usize = 22;
+const OPTION_ANCHORS_ODD_BIT: usize = 23;
```
ekzyis: nit: could use `OPTION_ANCHORS_BITS` like in #56

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1342 @@
...
+
+/// Channel configuration including funding details and both parties configuration.
+pub struct ChannelConfig {
+    /// Funding transaction outpoint.
+    pub funding_outpoint: OutPoint,
+    /// Total channel funding amount in satoshis.
+    pub funding_satoshis: u64,
+    /// Channel type feature bits. The commitment format (anchor / legacy) is
+    /// derived from the bits set here.
+    pub channel_type: Vec<u8>,
```
ekzyis, line 78: bolt03 mentions [this](https://github.com/lightning/bolts/blob/444805d12ab98c30006173bb190cd9d6fce9e405/02-peer-protocol.md?plain=1#L1041-L1044):

> We generate the commitment transaction at this point, using the `channel_type`
> that was communicated in the `open_channel` and `accept_channel` messages.
> This `channel_type` determines the channel commitment format for the total
> lifetime of the channel.

We use `Option<Vec<u8>>` there. Does that mean we should also use `Option<Vec<u8>>` here?

It also mentions [this](https://github.com/lightning/bolts/blob/444805d12ab98c30006173bb190cd9d6fce9e405/02-peer-protocol.md?plain=1#L839-L840) though:

> The receiving node MUST:
>  - if the message doesn't include a `channel_type`:
>    - fail the channel.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1342 @@
...
+        });
+
+        outputs
+    }
+}
+
+impl CommitmentState {
+    /// Constructs the initial commitment state after channel funding.
+    ///
+    /// # Errors
```
ekzyis:
```suggestion
    /// # Errors
    ///
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1342 @@
...
+
+/// Get the fee cost of a commitment tx in satoshis.
+fn commit_tx_fee_sat(feerate_per_kw: u32, channel_type: &[u8]) -> u64 {
+    let commitment_weight = if supports_option_anchors(channel_type) {
+        COMMITMENT_WEIGHT_ANCHOR
+    } else {
+        COMMITMENT_WEIGHT_NON_ANCHOR
+    };
+
+    u64::from(feerate_per_kw) * commitment_weight / 1000
```
ekzyis, line 392: nit: This return value is always multiplied by 1000 before it's used. Maybe rename to `commit_tx_fee_msat` and drop `/ 1000` here?

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1342 @@
...
+}
+
+/// Get the anchor cost of a commitment tx in satoshis.
+fn total_anchors_sat(channel_type: &[u8]) -> u64 {
```
ekzyis, line 396: same nit here: returns satoshis, but we always need millisatoshis

morehouse:
> If @morehouse agrees, I think it would help with review to put the first commit into its own PR and merge it soon. 28/30 files were only touched because of it.

This is a very good idea and would speed up review for the remainder of the PR.

ekzyis (commented):
withholding approval because I didn't review everything yet

smite/src/bolt/commitment.rs
NishantBansal2003: IIUC, BOLT specifies that the `channel_type` field should be present, so technically there should be an assert after receiving `accept_channel` that the message contains a `channel_type`. If not, we should fail the test case, since that would mean the target is not fully compliant with the BOLT spec. Else, we can always use`channel_type: Vec<u8>` when constructing the commitment tx

smite/src/bolt/commitment.rs
NishantBansal2003: Division by 1000 is actually part of the spec so that the commitment fee is rounded down (See: https://github.com/lightning/bolts/blob/444805d12ab98c30006173bb190cd9d6fce9e405/03-transactions.md#requirements-3), so the / 1000 is required. To avoid confusion around dividing by 1000 and then later multiplying by 1000, I kept the function as `commit_tx_fee_sat` and later converted it to msat

Though this function is more or less directly copied from LDK (See: https://github.com/lightningdevkit/rust-lightning/blob/38f5651ce0c423d27d7dc4d7abb1b2adc5cc4eb0/lightning/src/ln/chan_utils.rs#L280-L294)

smite/src/bolt/commitment.rs
NishantBansal2003: This too is taken from LDK (See: https://github.com/lightningdevkit/rust-lightning/blob/38f5651ce0c423d27d7dc4d7abb1b2adc5cc4eb0/lightning/src/sign/tx_builder.rs#L101-L107), and I intentionally kept it in sat so that things remain consistent with `commit_tx_fee_sat`

smite/src/bolt/commitment.rs
morehouse: You might be right there.

In Appendix C it does look like HTLC amounts are subtracted.  But even then I don't think they're correctly distributed -- the offered HTLCs should be subtracted from `to_local_msat` and the received HTLCs should be subtracted from `to_remote_msat`.

Perhaps we should create an issue in the BOLTs repo.

smite/src/bolt/commitment.rs
morehouse: It looks like both callers of these functions could do the resulting calculation in sat instead of msat, which would help eliminate the slight confusion.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1342 @@
...
+    /// Returns an error if:
+    /// - `funding_satoshis` is too large to convert to millisatoshis
+    /// - `push_msat` exceeds the total channel funding in millisatoshis
+    pub fn new_initial_from_funding(
+        funding_satoshis: u64,
+        push_msat: u64,
+        feerate_per_kw: u32,
+        opener_per_commitment_point: PublicKey,
+        acceptor_per_commitment_point: PublicKey,
+    ) -> Result<Self, String> {
```
morehouse: We should use a public `thiserror` enum for errors, or panic if we expect the caller to only past sane values.

NishantBansal2003: spec says: `The receiving node MUST fail the channel if: .... push_msat is greater than funding_satoshis * 1000`, so let's panic here so that we can detect if some target is not following the spec

spec also says: `The receiving node MUST fail the channel if: .... the funder's amount for the initial commitment transaction is not sufficient for full fee payment`. For that, I think we can later add an assert using `can_opener_afford_feerate` to verify whether the target is following the spec or not

smite/src/bolt/commitment.rs
NishantBansal2003: I used `COMMITMENT_TX_BASE_WEIGHT_NON_ANCHOR` and `COMMITMENT_TX_BASE_WEIGHT_ANCHOR`, which includes your suggestion as well as the LDK naming style. Hope this fixes the confusion

smite/src/bolt/commitment.rs
NishantBansal2003: ref: https://github.com/lightning/bolts/issues/1335

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1312 @@
...
+        });
+
+        outputs
+    }
+}
+
+impl CommitmentState {
+    /// Constructs the initial commitment state after channel funding.
+    #[must_use]
+    #[allow(clippy::missing_panics_doc)]
```
morehouse: Nit: I think it's worth documenting the panics so it's clear the user should sanitize the inputs

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1312 @@
...
+            channel_type,
+            opener: ChannelPartyConfig {
+                funding_pubkey: pubkey(
+                    "023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb",
+                ),
+                payment_basepoint: pubkey(
+                    "034f355bdcb7cc0af728ef3cceb9615d90684bb5b2ca5f859ab0f0b704075871aa",
+                ),
+                revocation_basepoint: pubkey(
+                    "0212a140cd0c6539d07cd08dfe09984dec3251ea808b892efeac3ede9402bf2b19",
```
morehouse: Nit:  In the test vectors this value is actually supposed to be the `local_revocation_pubkey`, not the basepoint.  IIUC it doesn't really matter because this value is unused in the current tests.  But perhaps we should set it to some clearly arbitrary value for clarity.


```suggestion
                revocation_basepoint: PublicKey::from_secret_key(
                    &Secp256k1::new(),
                    &secret("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"),
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1312 @@
...
+                per_commitment_point: acceptor_per_commitment_point,
+                balance_msat: to_acceptor_balance_msat,
+            },
+        }
+    }
+
+    /// Checks whether the opener can afford the commitment fee at the given
+    /// feerate, after accounting for the anchor outputs.
+    #[must_use]
+    pub fn can_opener_afford_feerate(&self, channel_type: &[u8]) -> bool {
```
morehouse: Requiring the user to extract the `channel_type` is slightly awkward.  Maybe this should be a method on `ChannelConfig` instead, with `CommitmentState` as the param.


```suggestion
    pub fn can_opener_afford_feerate(&self, state: &CommitmentState) -> bool {
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1312 @@
...
+    /// Constructs the initial commitment state after channel funding.
+    #[must_use]
+    #[allow(clippy::missing_panics_doc)]
+    pub fn new_initial_from_funding(
+        funding_satoshis: u64,
+        push_msat: u64,
+        feerate_per_kw: u32,
+        opener_per_commitment_point: PublicKey,
+        acceptor_per_commitment_point: PublicKey,
+    ) -> Self {
```
morehouse: `funding_satoshis` is also needed in the `ChannelConfig`.  Maybe this function should be a method on `ChannelConfig` instead?  WDYT?


```suggestion
    pub fn new_initial_commitment(
        &self,
        push_msat: u64,
        feerate_per_kw: u32,
        opener_per_commitment_point: PublicKey,
        acceptor_per_commitment_point: PublicKey,
    ) -> CommitmentState {
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1312 @@
...
+            (&self.opener, &self.acceptor)
+        } else {
+            (&self.acceptor, &self.opener)
+        };
+
+        let local_per_commitment_point = if holder.is_opener == for_holder {
+            state.opener.per_commitment_point
+        } else {
+            state.acceptor.per_commitment_point
+        };
```
morehouse: The repeated `holder.is_opener == for_holder` check is a bit confusing, and IIUC we don't actually need both -- we just need to know which commitment to build.

I think the code could be more readable if we introduce a `Side` enum like this:

```rust
enum Side {
  Opener,
  Acceptor,
}

impl Side {
  fn other(self) -> Self {
    match self {
      Self::Opener => Self::Acceptor,
      Self::Acceptor => Self::Opener,
    }
  }
}

impl ChannelConfig {
  fn party(&self, side: Side) -> &ChannelPartyConfig {
    match side {
      Side::Opener => self.opener,
      Side::Acceptor => self.acceptor,
    }
  }
}

impl CommitmentState {
  fn party(&self, side: Size) -> &CommitmentPartyState { ... }
}
```

For example, then `build_commitment_outputs` just takes a `local_side: Side` param and the code here becomes:

```suggestion
        // Map opener/acceptor to local/remote for this commitment side.
        let (to_local_value, to_remote_value) =  match local_side {
            Side::Opener => (opener_balance, acceptor_balance),
            Side::Acceptor => (acceptor_balance, opener_balance),
        };
        let local = self.party(local_side);
        let remote = self.party(local_side.other());
        let local_per_commitment_point = state.party(local_side).per_commitment_point;
```

I'd also expect simplifications elsewhere.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1312 @@
...
+            &self.opener.funding_pubkey
+        };
+        secp.verify_ecdsa(&msg, signature, counterparty_funding_pubkey)
+            .is_ok()
+    }
+
+    /// Builds the signature for the holder's commitment transaction.
+    /// Only used to exercise BOLT 3 test vectors.
+    #[cfg(test)]
+    fn holder_commitment_signature(
```
morehouse: Nit: naming consistency

```suggestion
    fn sign_holder_commitment(
```

morehouse (approved):
Functionality all looks good to me. Only style and readability nits.

smite/src/bolt/commitment.rs
NishantBansal2003: This doc suppression was intentional since, the caller does not need to worry about sanitizing the input, since according to the spec, if `push_msat is greater than funding_satoshis * 1000`, then `accept_channel` should not be sent. So if we receive an `accept_channel` and try to construct the commitment tx here, it should panic so that we can detect if any target is not following the spec

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1334 @@
...
+    ) -> Signature {
+        let sighash = self.build_commitment_sighash(state, &holder.side);
+        sign(&sighash, &holder.funding_privkey)
+    }
+
+    /// Builds the sighash for the commitment transaction. The commitment
+    /// format (legacy or anchor) is determined by the `channel_type`.
+    ///
+    /// When `for_holder` is true, builds the holder's commitment and
+    /// when false, builds the counterparty's.
```
erickcestari: nit: This rustdoc needs to be updated.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1334 @@
...
+            .expect("input index 0 is always in bounds for a single input transaction");
+
+        sighash.to_byte_array()
+    }
+
+    /// Builds the lexicographically sorted commitment outputs.
+    ///
+    /// When `for_holder` is true, builds outputs for the holder's commitment;
+    /// otherwise builds outputs for the counterparty's commitment.
+    fn build_commitment_outputs(&self, state: &CommitmentState, local_side: &Side) -> Vec<TxOut> {
```
erickcestari: nit: This rustdoc needs to be updated.

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1334 @@
...
+        // BOLT 3 output ordering: sort by (value, script_pubkey).
+        outputs.sort_by(|a, b| {
+            a.value
+                .cmp(&b.value)
+                .then_with(|| a.script_pubkey.as_bytes().cmp(b.script_pubkey.as_bytes()))
+        });
+
+        outputs
+    }
+}
```
erickcestari, line 369: nit: We could simplify `build_commitment_outputs` by creating two functions: one to create local output and one to create remote output.

```suggestion
/// Builds the lexicographically sorted commitment outputs, following the
    /// algorithm in BOLT 3 "Commitment Transaction Construction".
    fn build_commitment_outputs(&self, state: &CommitmentState, local_side: &Side) -> Vec<TxOut> {
        let anchor = supports_option_anchors(&self.channel_type);
        let local = self.party(local_side);
        let remote = self.party(local_side.other());
        let local_per_commitment_point = state.party(local_side).per_commitment_point;

        let (to_local_value, to_remote_value) = self.local_balances(state, local_side);

        let mut outputs: Vec<TxOut> = Vec::new();

        if to_local_value >= local.dust_limit_satoshis {
            outputs.push(build_to_local_output(
                to_local_value,
                local,
                remote,
                &local_per_commitment_point,
            ));
            if anchor {
                outputs.push(build_anchor_output(&local.funding_pubkey));
            }
        }
        if to_remote_value >= local.dust_limit_satoshis {
            outputs.push(build_to_remote_output(
                to_remote_value,
                &remote.payment_basepoint,
                anchor,
            ));
            if anchor {
                outputs.push(build_anchor_output(&remote.funding_pubkey));
            }
        }

        // BOLT 3 output ordering: sort by (value, script_pubkey).
        outputs.sort_by(|a, b| {
            a.value
                .cmp(&b.value)
                .then_with(|| a.script_pubkey.as_bytes().cmp(b.script_pubkey.as_bytes()))
        });

        outputs
    }

    /// Returns `(to_local, to_remote)` values in satoshis from the perspective
    /// of `local_side`, after subtracting the base commitment fee and anchor
    /// costs from the funder.
    fn local_balances(&self, state: &CommitmentState, local_side: &Side) -> (u64, u64) {
        let fee = commit_tx_fee_sat(state.feerate_per_kw, &self.channel_type);
        let anchor_cost = total_anchors_sat(&self.channel_type);

        let acceptor_balance = state.acceptor.balance_msat / 1000;
        let opener_balance = (state.opener.balance_msat / 1000)
            .saturating_sub(fee)
            .saturating_sub(anchor_cost);

        match local_side {
            Side::Opener => (opener_balance, acceptor_balance),
            Side::Acceptor => (acceptor_balance, opener_balance),
        }
    }
}

/// Builds the `to_local` commitment output.
fn build_to_local_output(
    value: u64,
    local: &ChannelPartyConfig,
    remote: &ChannelPartyConfig,
    per_commitment_point: &PublicKey,
) -> TxOut {
    let local_delayedpubkey = derive_pubkey(&local.delayed_payment_basepoint, per_commitment_point);
    let revocationpubkey =
        derive_revocation_pubkey(&remote.revocation_basepoint, per_commitment_point);
    TxOut {
        value: Amount::from_sat(value),
        script_pubkey: build_to_local_scriptpubkey(
            &local_delayedpubkey,
            &revocationpubkey,
            remote.to_self_delay,
        ),
    }
}

/// Builds the `to_remote` commitment output.
fn build_to_remote_output(value: u64, payment_basepoint: &PublicKey, anchor: bool) -> TxOut {
    TxOut {
        value: Amount::from_sat(value),
        script_pubkey: build_to_remote_scriptpubkey(payment_basepoint, anchor),
    }
}

/// Builds an anchor commitment output for the given funding pubkey.
fn build_anchor_output(funding_pubkey: &PublicKey) -> TxOut {
    TxOut {
        value: Amount::from_sat(ANCHOR_OUTPUT_VALUE),
        script_pubkey: build_anchor_scriptpubkey(funding_pubkey),
    }
}
```

smite/src/bolt/commitment.rs
```diff
@@ -0,0 +1,1334 @@
...
+        Sha256::from_engine(sha).to_byte_array()
+    };
+
+    let commit_append_rev_hash_key = {
+        let mut sha = Sha256::engine();
+        sha.input(&per_commitment_point.serialize());
+        sha.input(&revocation_basepoint.serialize());
+
+        Sha256::from_engine(sha).to_byte_array()
+    };
```
erickcestari: nit: We could have a helper function to calculate the tweak.

```rust
/// SHA256 of `a || b` where each is the compressed serialization of a pubkey.
fn hash_pubkeys(a: &PublicKey, b: &PublicKey) -> [u8; 32] {
    let mut sha = Sha256::engine();
    sha.input(&a.serialize());
    sha.input(&b.serialize());
    Sha256::from_engine(sha).to_byte_array()
}
```

erickcestari (approved):
LGTM! Left a few nits. Feel free to address if you think they're worth it.

smite/src/bolt/commitment.rs
NishantBansal2003: I will defer this one for now, since it feels more like segregation than simplification, and personally I feel it affects the reading flow a bit in some cases. But I’m totally fine with going with it if you feel strongly about it, so just let me know

smite/src/bolt/commitment.rs
NishantBansal2003: Used an arbitrary `pubkey("02c6047f9441ed7d6d3045406e95c07cd85a0f5f0f3b9b3f3d5f9b1e5e4a7c4f09")` to keep it consistent with the other parts of the code

smite/src/bolt/commitment.rs
NishantBansal2003: I looked into this again, and I think we still need to sanitize the input. Even if the target rejects invalid input, we might still end up with an invalid push_msat here while constructing the commitment tx coz of the mutators. So yes, we should sanitize this beforehand. I’ll update this now

morehouse (approved):
LGTM

PR #53, opened 2026-04-17
smite-ir-mutator: add AFL++ custom mutator library

morehouse:
Implement the AFL++ custom mutator C ABI as a new crate. The resulting cdylib should be used with `AFL_CUSTOM_MUTATOR_LIBRARY`, along with `AFL_CUSTOM_MUTATOR_ONLY=1` and `AFL_DISABLE_TRIM=1`.

The library exports five functions for use by AFL++:
- `afl_custom_init` and `afl_custom_deinit` create and destroy a boxed `MutatorState` for use by the other functions.
- `afl_custom_fuzz` decodes and validates a provided input as a `Program`,  then mutates the `Program` or generates a new one from scratch.
- `afl_custom_splice_optout` tells AFL++ not to provide inputs for a splice mutator.
- `afl_custom_describe` returns the sequence of mutators that produced the latest mutated input.

The AFL++ custom mutator [documentation](https://aflplus.plus/docs/custom_mutators/) and [examples](https://github.com/AFLplusplus/AFLplusplus/tree/stable/custom_mutators/examples) will be helpful for review.

Ref: #5 (Milestone 1)

### DIFF
--- a/smite-ir-mutator/src/lib.rs
+++ b/smite-ir-mutator/src/lib.rs
@@ -1,0 +1,400 @@
+//! AFL++ custom mutator C ABI shim.
+//!
+//! This crate exports the `afl_custom_*` symbols that AFL++ looks up via
+//! `dlsym` when a shared library is supplied through
+//! `AFL_CUSTOM_MUTATOR_LIBRARY`. The resulting `.so` (built via
+//! `crate-type = ["cdylib"]`) is the *only* way AFL++ produces inputs for the
+//! IR fuzzer.
+//!
+//! # AFL++ environment contract
+//!
+//! Set the following when launching `afl-fuzz`:
+//! - `AFL_CUSTOM_MUTATOR_LIBRARY=/path/to/libsmite_ir_mutator.so`
+//! - `AFL_CUSTOM_MUTATOR_ONLY=1` -- disable AFL++'s byte mutators. This also
+//!   disables the havoc stage entirely, so we deliberately do not implement
+//!   `afl_custom_havoc_mutation`.
+//! - `AFL_DISABLE_TRIM=1` -- this library does not implement custom trim and
+//!   AFL++'s default byte-level trim would corrupt our structured programs.
+//!
+//! # Buffer ownership
+//!
+//! `MutatorState` owns a `Vec<u8>` (`out_buf`) that holds the serialized
+//! mutated program. When AFL++ requests a new input from us, we hand it a
+//! pointer to that `Vec<u8>` along with its length. AFL++ never saves the
+//! pointer across calls, so we're safe to reuse the buffer for the next input.
+
+use std::os::raw::{c_char, c_uint, c_void};
+use std::slice;
+
+use rand::rngs::SmallRng;
+use rand::{RngExt, SeedableRng};
+
+use smite_ir::generators::OpenChannelGenerator;
+use smite_ir::mutators::{InputSwapMutator, OperationParamMutator};
+use smite_ir::{Generator, Mutator, Program, ProgramBuilder};
+
+/// Mutator state owned by AFL++ across calls. Allocated by [`afl_custom_init`],
+/// destroyed by [`afl_custom_deinit`].
+struct MutatorState {
+    rng: SmallRng,
+    /// Reusable buffer for serialized mutated programs. The pointer to its
+    /// contents is what we hand back to AFL++ when it requests an input.
+    out_buf: Vec<u8>,
+    /// Reusable null-terminated description string returned by
+    /// [`afl_custom_describe`].
+    description: Vec<u8>,
+    /// Sequence of actions taken in the last [`afl_custom_fuzz`] call, used by
+    /// [`afl_custom_describe`] to name queue entries.
+    last_sequence: Vec<&'static str>,
+}
+
+impl MutatorState {
+    fn new(seed: u32) -> Self {
+        Self {
+            rng: SmallRng::seed_from_u64(u64::from(seed)),
+            out_buf: Vec::new(),
+            description: Vec::new(),
+            last_sequence: vec!["init"],
+        }
+    }
+
+    /// Generates a fresh program from scratch using `OpenChannelGenerator`.
+    fn generate_fresh(&mut self) -> Program {
+        let mut builder = ProgramBuilder::new();
+        OpenChannelGenerator.generate(&mut builder, &mut self.rng);
+        self.last_sequence.clear();
+        self.last_sequence.push("fresh");
+        builder.build()
+    }
+
+    /// Applies a random number of stacked mutations to `program`. The stack
+    /// count is a power of two in `[1, 16]` to give a mix of small tweaks and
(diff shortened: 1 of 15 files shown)


smite-ir-mutator/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+        panic!("no stacked mutations of existing input after 10 attempts");
+    }
+
+    #[test]
+    fn describe_respects_max_description_len() {
+        let state = State::new(7);
+        let _ = fuzz_via_ffi(&state, Vec::new(), 1 << 16);
+        // Cap the description at 5 bytes.
+        let ptr = unsafe { afl_custom_describe(state.0, 5) };
+        let bytes = unsafe { CStr::from_ptr(ptr) }.to_bytes();
```
Chand-ra: [`CStr::to_bytes()`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.to_bytes) explicitly strips the null terminator before returning the slice. We should use the `with_nul` variant instead.

```suggestion
        let bytes = unsafe { CStr::from_ptr(ptr) }.to_bytes_with_nul();
```

smite-ir-mutator/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+        let mut buf = std::mem::take(&mut self.out_buf);
+        buf.clear();
+        match postcard::to_extend(program, buf) {
+            Ok(buf) => {
+                self.out_buf = buf;
+                self.out_buf.len() <= max_size
+            }
+            Err(_) => false,
+        }
+    }
```
Chand-ra: When `postcard::to_extend` fails, it returns a `postcard::Error` and drops the `Vec`, destroying the underlying heap allocation. The next time the fuzzer calls `serialize`, it will be forced to perform a fresh heap allocation, defeating the purpose of hoisting `out_buf` into `MutatorState`.

Since `&mut Vec<T>` implements the `Extend<T>` trait, We can instead pass a mutable reference to `to_extend`. If it fails, the reference's lifetime ends, but the `Vec` and its memory capacity stay right where they are.

```suggestion
    fn serialize(&mut self, program: &Program, max_size: usize) -> bool {
        self.out_buf.clear();
        match postcard::to_extend(program, &mut self.out.buf) {
            Ok(buf) => self.out_buf.len() <= max_size
            Err(_) => false,
        }
    }
```

Makes the code cleaner as well.

smite-ir-mutator/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+    state.description.clear();
+    state.description.extend_from_slice(b"smite-ir:");
+    for (i, name) in state.last_sequence.iter().enumerate() {
+        if i > 0 {
+            state.description.push(b',');
+        }
+        state.description.extend_from_slice(name.as_bytes());
+    }
+    // Leave room for the trailing null terminator.
+    if state.description.len() > max_description_len {
```
devvaansh, line 231: Shouldn't this truncate to max_description_len.saturating_sub(1) becuase The AFL++ docs say max_description_len is the total budget. After truncation to max_description_len bytes, pushing the \0 makes it max_description_len + 1. The test describe_respects_max_description_len doesn't catch this because CStr::to_bytes() excludes the null, the assertion checks content length, not total buffer length.

Chand-ra: The description above `afl_custom_describe()` [clearly notes](https://github.com/morehouse/smite/pull/53/changes#diff-e4c2054a2abea3bb148c17bd66d85187783bf518ad07dc6e9fca263b6b3cee0bR208) that `max_description_len` doesn't count the NULL terminator byte.

devvaansh: You're right, thanks for the correction

smite-ir-mutator/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+        assert!(!state.0.is_null());
+    }
+
+    #[test]
+    fn fuzz_with_empty_input_generates_fresh() {
+        let state = State::new(1);
+        let (out, len) = fuzz_via_ffi(&state, Vec::new(), 1 << 16);
+        assert!(len > 0);
+        decode_and_validate(out, len);
+    }
```
erickcestari, line 316: Let's verify that it's actually creating a "fresh" `Program` by examining the description.

```suggestion
    #[test]
    fn fuzz_with_empty_input_generates_fresh() {
        let state = State::new(1);
        let (out, len) = fuzz_via_ffi(&state, Vec::new(), 1 << 16);
        assert!(len > 0);
        decode_and_validate(out, len);
        let ptr = unsafe { afl_custom_describe(state.0, 64) };
        let s = unsafe { CStr::from_ptr(ptr) }
            .to_str()
            .expect("valid utf-8");
        assert_eq!(s, "smite-ir:fresh");
    }
```

smite-ir-mutator/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+        }
+    }
+
+    #[test]
+    fn fuzz_with_garbage_input_generates_fresh() {
+        let state = State::new(3);
+        let (out, len) = fuzz_via_ffi(&state, vec![0xFFu8; 16], 1 << 16);
+        assert!(len > 0);
+        decode_and_validate(out, len);
+    }
```
erickcestari, line 340: Same here:

```suggestion
    #[test]
    fn fuzz_with_garbage_input_generates_fresh() {
        let state = State::new(3);
        let (out, len) = fuzz_via_ffi(&state, vec![0xFFu8; 16], 1 << 16);
        assert!(len > 0);
        decode_and_validate(out, len);
        let ptr = unsafe { afl_custom_describe(state.0, 64) };
        let s = unsafe { CStr::from_ptr(ptr) }
            .to_str()
            .expect("valid utf-8");
        assert_eq!(s, "smite-ir:fresh");
    }
```

erickcestari (approved):
LGTM! I've pointed out some nits. Also:

Afl++ docs says that don't having a custom trimmer is bad. This is probably designed to be done in a follow up PR, also that would probably require having different minimizers technics at `smite-ir`, right?

> Setting AFL_DISABLE_TRIM tells afl-fuzz not to trim test cases. This is usually a bad idea!

Chand-ra:
Took a quick look, LGTM.

smite-ir-mutator/src/lib.rs
morehouse: Done.

smite-ir-mutator/src/lib.rs
morehouse: Done.

morehouse:
> Afl++ docs says that don't having a custom trimmer is bad. This is probably designed to be done in a follow up PR, also that would probably require having different minimizers technics at `smite-ir`, right?

Current mutators and generators have a fixed upper bound on how large they can make programs, so trim mutators aren't super important yet.  I definitely want them in the future, but for now I'm focused on the MVP.

PR #55, opened 2026-04-24
smite-ir: Add unit tests for `OperationParamMutator`

Chand-ra:
This PR is a follow-up to #44, adding targeted, deterministic unit tests for the `OperationParamMutator`'s helper functions. 

Current tests for this mutator rely primarily on top-level integration loops. Directly testing the helper functions with seeded RNGs guarantees that every individual mutation branch is structurally sound and prevents future regressions (like off-by-one errors).

### DIFF
--- a/smite-ir/src/mutators/operation_param.rs
+++ b/smite-ir/src/mutators/operation_param.rs
@@ -392,3 +392,54 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
 fn interesting_u64(rng: &mut impl Rng) -> u64 {
     INTERESTING_U64[rng.random_range(0..INTERESTING_U64.len())]
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rand::{SeedableRng, rngs::SmallRng};
+
+    #[test]
+    fn shuffle_subrange_preserves_elements() {
+        let mut rng = SmallRng::seed_from_u64(0);
+        let sorted = vec![1, 2, 3, 4, 5, 6, 7, 8];
+        let mut shuffled = sorted.clone();
+
+        for _ in 0..100 {
+            shuffle_subrange(&mut shuffled, &mut rng);
+            shuffled.sort_unstable();
+            assert_eq!(
+                shuffled, sorted,
+                "shuffle_subrange must preserve all original elements"
+            );
+        }
+    }
+
+    #[test]
+    fn shuffle_subrange_mutates_input() {
+        let mut rng = SmallRng::seed_from_u64(0);
+        let data = vec![1, 2, 3, 4, 5, 6, 7, 8];
+        let mut mutated = false;
+
+        for _ in 0..50 {
+            let mut cloned = data.clone();
+            shuffle_subrange(&mut cloned, &mut rng);
+            if data != cloned {
+                mutated = true;
+            }
+        }
+        assert!(mutated, "shuffle_subrange doesn't mutate input");
+    }
+
+    #[test]
+    fn shuffle_subrange_empty_and_single() {
+        let mut rng = SmallRng::seed_from_u64(0);
+
+        let mut empty: Vec<u8> = vec![];
+        shuffle_subrange(&mut empty, &mut rng);
+        assert!(empty.is_empty());
+
+        let mut single = vec![0xFF];
+        shuffle_subrange(&mut single, &mut rng);
+        assert_eq!(single, vec![0xFF]);
+    }
+}


smite-ir/Cargo.toml
```diff
@@ -9,7 +9,7 @@ workspace = true
 
 [dependencies]
 smite = { path = "../smite" }
-rand = { version = "0.10", default-features = false }
+rand = { version = "0.10", default-features = false, features = ["std_rng"] }
```
Chand-ra: `default-features` includes `std_rng` but it's explicitly excluded here, so unsure if this is okay to do. I could add a simple PRNG in the tests module to seed the tests instead, but that'd be writing boilerplate.

smite-ir/src/mutators/operation_param.rs
```diff
@@ -323,3 +323,131 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
 fn interesting_u64(rng: &mut impl Rng) -> u64 {
     INTERESTING_U64[rng.random_range(0..INTERESTING_U64.len())]
 }
+
+#[cfg(test)]
+mod internal_operation_param_mutator_tests {
```
ekzyis: Any reason why not use `mod tests` here?

smite-ir/Cargo.toml
ekzyis: Can we use `SmallRng` instead of `StdRng` like in smite-ir/src/tests.rs?

smite-ir/src/mutators/operation_param.rs
```diff
@@ -323,3 +323,131 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
...
+    use super::*;
+    use rand::{SeedableRng, rngs::StdRng};
+
+    #[test]
+    fn test_shuffle_subrange_preserves_elements() {
+        let mut rng = StdRng::seed_from_u64(42);
```
ekzyis: nit: the tests in smite-ir/src/tests.rs always use 0 as the seed

smite-ir/src/mutators/operation_param.rs
```diff
@@ -323,3 +323,131 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
...
+        let mut rng = StdRng::seed_from_u64(555);
+
+        // Find a field that has NO other fields with the same output type
+        let unique_field = AcceptChannelField::ALL.iter().find(|&&f| {
+            AcceptChannelField::ALL
+                .iter()
+                .filter(|&&other| other.output_type() == f.output_type())
+                .count()
+                == 1
+        });
```
ekzyis: nit: I don't expect changes to `AcceptChannelField`, so I think we could hard-code the field here to improve readability.

smite-ir/src/mutators/operation_param.rs
```diff
@@ -323,3 +323,131 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
...
     INTERESTING_U64[rng.random_range(0..INTERESTING_U64.len())]
 }
+
+#[cfg(test)]
+mod internal_operation_param_mutator_tests {
+    use super::*;
+    use rand::{SeedableRng, rngs::StdRng};
+
+    #[test]
+    fn test_shuffle_subrange_preserves_elements() {
```
ekzyis: nit: I think the following tests the same while being slightly more readable + it tests that `shuffle_subrange` does something when there's no edge case, which seems to be missing as a test:

```diff
diff --git a/smite-ir/src/mutators/operation_param.rs b/smite-ir/src/mutators/operation_param.rs
index 5bcc2cb..0343998 100644
--- a/smite-ir/src/mutators/operation_param.rs
+++ b/smite-ir/src/mutators/operation_param.rs
@@ -332,24 +332,15 @@ mod internal_operation_param_mutator_tests {
     #[test]
     fn test_shuffle_subrange_preserves_elements() {
         let mut rng = StdRng::seed_from_u64(42);
+        let sorted = vec![1, 2, 3, 4, 5, 6, 7, 8];
+        let mut shuffled = sorted.clone();

-        let mut data = vec![1, 2, 3, 4, 5, 6, 7, 8];
-        let original_len = data.len();
+        shuffle_subrange(&mut shuffled, &mut rng);

-        shuffle_subrange(&mut data, &mut rng);
-
-        assert_eq!(
-            data.len(),
-            original_len,
-            "Shuffle must not change vector length"
-        );
-
-        let mut sorted_data = data.clone();
-        sorted_data.sort_unstable();
-
-        let expected = vec![1, 2, 3, 4, 5, 6, 7, 8];
+        assert_ne!(shuffled, sorted, "Shuffle must shuffle elements");
+        shuffled.sort_unstable();
         assert_eq!(
-            sorted_data, expected,
+            shuffled, sorted,
             "Shuffle must preserve all original elements"
         );
     }
```

smite-ir/Cargo.toml
ekzyis, line 1: When I run `cargo test -p smite-ir`, it updates Cargo.lock. Could it be that you forgot to push these changes? But I think the changes to Cargo.lock aren't needed if we don't enable `std_rng`, see other comment.

smite-ir/Cargo.toml
Chand-ra: A lot better, thanks.

smite-ir/src/mutators/operation_param.rs
Chand-ra: Wanted to make the name more specific, probably better to go with plain old 'test' though.

smite-ir/Cargo.toml
Chand-ra: Done.

smite-ir/src/mutators/operation_param.rs
Chand-ra: Does read better, thanks!

ekzyis (commented):
GitHub is having issues ([again](https://news.ycombinator.com/item?id=48034587)). Couldn't create review comments for individual lines, sorry.

L340:

```diff
-        assert_ne!(shuffled, sorted, "Shuffle must shuffle elements");
+        assert_ne!(shuffled, sorted, "shuffle_subrange must shuffle elements");
```

or

```diff
-        assert_ne!(shuffled, sorted, "Shuffle must shuffle elements");
+        assert_ne!(shuffled, sorted, "must shuffle elements");
```

---

L344: same suggestion

---

L424: I think "field that has NO other fields" is confusing

---

L425:

`u32` is slightly inaccurate, block heights are their own type:

```diff
-        // MinimumDepth is the only field with a `u32` output type.
+        // MinimumDepth is the only field with a `BlockHeight` output type.
```

smite-ir/src/mutators/operation_param.rs
```diff
@@ -380,3 +380,115 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
...
     INTERESTING_U64[rng.random_range(0..INTERESTING_U64.len())]
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rand::{SeedableRng, rngs::SmallRng};
+
+    #[test]
+    fn test_shuffle_subrange_preserves_elements() {
```
morehouse: Nit: follow codebase naming convention


```suggestion
    fn shuffle_subrange_preserves_elements() {
```

smite-ir/src/mutators/operation_param.rs
```diff
@@ -380,3 +380,115 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
...
+        let mut shuffled = sorted.clone();
+
+        shuffle_subrange(&mut shuffled, &mut rng);
+
+        assert_ne!(shuffled, sorted, "shuffle_subrange must shuffle elements");
+        shuffled.sort_unstable();
+        assert_eq!(
+            shuffled, sorted,
+            "shuffle_subrange must preserve all original elements"
+        );
```
morehouse: This test would be stronger if we looped for many iterations.

smite-ir/src/mutators/operation_param.rs
```diff
@@ -380,3 +380,115 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
...
+    }
+
+    #[test]
+    fn test_shuffle_subrange_empty_and_single() {
```
morehouse: Nit: follow codebase naming convention


```suggestion
    fn shuffle_subrange_empty_and_single() {
```

smite-ir/src/mutators/operation_param.rs
```diff
@@ -380,3 +380,115 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
...
+        shuffle_subrange(&mut empty, &mut rng);
+        assert!(empty.is_empty());
+
+        let mut single = vec![0xFF];
+        shuffle_subrange(&mut single, &mut rng);
+        assert_eq!(single, vec![0xFF]);
+    }
+
+    #[test]
+    fn test_mutate_bytes_respects_max_size() {
```
morehouse: This is already covered by `param_mutator_caps_byte_length`.

smite-ir/src/mutators/operation_param.rs
```diff
@@ -380,3 +380,115 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
...
+            assert!(
+                data.len() <= MAX_MESSAGE_SIZE,
+                "mutate_bytes exceeded MAX_MESSAGE_SIZE. Current len: {}",
+                data.len()
+            );
+        }
+    }
+
+    #[test]
+    fn test_mutate_bytes_handles_empty_input() {
```
morehouse: This test is almost useless.  It runs one time while `mutate_bytes` has 9 randomly selected arms that could execute.  It's also likely already tested in `param_mutator_caps_byte_length`.

smite-ir/src/mutators/operation_param.rs
```diff
@@ -380,3 +380,115 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
...
+
+        for _ in 0..500 {
+            mutate_fixed_bytes(&mut bytes, &mut rng);
+        }
+        // Ensure the array actually mutated from all zeros
+        assert_ne!(bytes, [0x00; 32], "Array should have been mutated");
+    }
+
+    #[test]
+    fn test_mutate_extract_field_preserves_type() {
```
morehouse: This is already covered by `param_mutator_preserves_extract_field_type`.

smite-ir/src/mutators/operation_param.rs
```diff
@@ -380,3 +380,115 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
...
+            assert_eq!(
+                field.output_type(),
+                original_type,
+                "Output type must be preserved"
+            );
+        }
+    }
+
+    #[test]
+    fn test_mutate_extract_field_handles_unique_types() {
```
morehouse: This is already covered by `param_mutator_returns_false_for_singleton_extract_field`.

morehouse (changes requested):
Most of these tests are already covered in the public API tests.  I think we should drop all the tests except the `shuffle_subrange` ones that actually provide some benefit.

smite-ir/src/mutators/operation_param.rs
```diff
@@ -390,3 +380,38 @@ fn interesting_u32(rng: &mut impl Rng) -> u32 {
...
+        let mut shuffled = sorted.clone();
+
+        for _ in 0..100 {
+            shuffle_subrange(&mut shuffled, &mut rng);
+            shuffled.sort_unstable();
+            assert_eq!(
+                shuffled, sorted,
+                "shuffle_subrange must preserve all original elements"
+            );
+        }
```
morehouse, line 414: I think we also should test that shuffle generally does something.


```suggestion
        let mut shuffled = sorted.clone();
        let mut diff_count = 0;

        for _ in 0..100 {
            shuffle_subrange(&mut shuffled, &mut rng);
            if shuffled != sorted {
                diff_count += 1;
            }
            shuffled.sort_unstable();
            assert_eq!(
                shuffled, sorted,
                "shuffle_subrange must preserve all original elements"
            );
        }
        assert(diff_count >= 70, "shuffle_subrange frequently does nothing");
```

Chand-ra: Doesn't look quite right to me.

Since our RNG is deterministic, `diff_count` will evaluate to the exact same number every time for a given seed, at which point we're more so testing the statistical distribution of the `rand` crate rather than the code's logic.

I believe it's better to test that `shuffle_subrange` mutates the input _at least once_. This way we can be sure the function isn't broken and just returning the identical array all the time.

Chand-ra:
Rebased on top of the latest master to resolve a merge conflict.

smite-ir/src/mutators/operation_param.rs
morehouse: The idea was to test that there isn't some unexpected bias towards doing nothing (either from our mutator implementation or the RNG), while also reducing the test flakiness in case later code changes cause slightly different mutations to occur.

But either way this is a strict improvement in testing coverage.

morehouse (approved):

PR #57, opened 2026-04-24
smite-ir: generate valid channel types

morehouse:
Helps the fuzzer choose valid channel types so that deeper paths can be reached in the `open_channel->accept_channel` flow.

Ref: https://github.com/morehouse/smite/issues/5 (Milestone 1)

### DIFF
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -44,6 +44,11 @@ pub enum Operation {
     /// Produces a [`VariableType::Bytes`] value whose contents match one of the
     /// standard script templates required by BOLT 2.
     LoadShutdownScript(ShutdownScriptVariant),
+    /// Load a valid `channel_type` feature vector.
+    ///
+    /// Produces a [`VariableType::Features`] value encoding one of a set of
+    /// channel types known to be accepted by at least one target.
+    LoadChannelType(ChannelTypeVariant),
     /// Load the target node's public key from the program context.
     LoadTargetPubkeyFromContext,
     /// Load the chain hash from the program context.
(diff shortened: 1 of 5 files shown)


smite-ir/src/operation.rs
```diff
@@ -88,6 +93,163 @@ pub enum Operation {
...
+/// A specific BOLT 2 `channel_type` feature-bit combination.
+///
+/// Each variant corresponds to a channel type accepted by at least one target
+/// implementation:
+///
+/// - `option_static_remotekey` (bit 12)
+/// - `option_anchors` (bits 22 and 12)
+/// - `zero_fee_commitments` (bit 40)
+/// - `option_simple_taproot` (bit 80)
+/// - `option_simple_taproot` staging (bit 180)
```
NishantBansal2003:
```suggestion
/// - `option_simple_taproot_staging` (bit 180)
```

smite-ir/src/operation.rs
```diff
@@ -88,6 +93,163 @@ pub enum Operation {
...
+    /// The feature bits (even/required) contained in this channel type.
+    #[must_use]
+    pub fn bits(self) -> &'static [usize] {
+        // BOLT 9 feature bits:
+        //   12 = option_static_remotekey
+        //   22 = option_anchors_zero_fee_htlc_tx
+        //   40 = option_anchor_zero_fee_commitments
+        //   46 = option_scid_alias
+        //   50 = option_zeroconf
+        //   80 = option_simple_taproot_final
```
NishantBansal2003: nit: would be great if we stay consistent with what the spec says:
```suggestion
        //   22 = option_anchors
        //   40 = zero_fee_commitments
        //   46 = option_scid_alias
        //   50 = option_zeroconf
        //   80 = option_simple_taproot
```

smite-ir/src/tests.rs
```diff
@@ -269,6 +269,82 @@ fn accept_channel_field_all_is_complete() {
...
+            | ChannelTypeVariant::ScriptEnforcedLeaseZeroConf
+            | ChannelTypeVariant::ScriptEnforcedLeaseScidAliasZeroConf => 24,
+        }
+    };
+    assert_eq!(
+        ChannelTypeVariant::ALL.len(),
+        variant_count(ChannelTypeVariant::ALL[0]),
+    );
+}
+#[test]
```
NishantBansal2003, line 472:
```suggestion
}

#[test]
```

ekzyis (approved):
LGTM

NishantBansal2003 (commented):
Looks fine, I only have a few nits

Also, if you go ahead with those nits, I think the enum variant names should be updated accordingly based on the comments

NishantBansal2003:
One more thing -- how about adding cases where mutually exclusive channel types are set, so we can verify that implementations reject it? for eg `option_anchors` with `option_simple_taproot`

smite-ir/src/operation.rs
morehouse: Done.

smite-ir/src/operation.rs
morehouse: Thanks, good catch.  Done, and updated enum variants.

smite-ir/src/tests.rs
morehouse: Done.

morehouse:
> One more thing -- how about adding cases where mutually exclusive channel types are set, so we can verify that implementations reject it? for eg `option_anchors` with `option_simple_taproot`

I'm not sure about this.  The idea was to help the fuzzer generate correct channel types, and adding lots of invalid combinations works against that.

In order to verify rejection of incorrect types, we also need a special fuzz target that verifies the rejection occurs.  If we're going to do that, we could probably hard-code all the other open_channel fields to be valid and then just use fuzz input against `channel_type`.  That would be a much more effective targeted test.

erickcestari (approved):
LGTM

NishantBansal2003:
> I'm not sure about this. The idea was to help the fuzzer generate correct channel types, and adding lots of invalid combinations works against that.

I thought we could do that using some probabilistic way (90% valid, 10% anything) in `OpenChannelGenerator`, it would very much resonate with `pick_variable`, but very specific for channel type. With this, we have a chance that the invalid channel type case is also covered.

Now, for asserting that the target always rejects it, we can check the channel type when building `open_channel` in the `executor`, and when waiting to receive `accept_channel`, we will have data that we sent a bad channel type (the one with a variant other than what the valid types we have), so we should only get an unexpected message and not `accept_channel` for sure. Also, we can assert that the channel type in received `accept_channel` matches what we originally sent in `open_channel`

Though this is somewhat hacky I think, so yeah, we can defer this for MVP. But later on, I think we can give some thought to integrating this check in the normal funding flow to avoid having specific tests where all fields are fixed and only the channel type is invalid.

PR #58, opened 2026-04-24
smite-ir: generate compliant shutdown scripts

morehouse:
Helps the fuzzer choose BOLT 2 compliant shutdown scripts so that deeper paths can be reached in the `open_channel->accept_channel` flow.

This change in combination with #57 are required to get past `open_channel` message validation in LDK.

Ref: #5 (Milestone 1)

### DIFF
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -11,6 +11,7 @@
 use std::fmt;
 use std::fmt::Write;
 
+use rand::{Rng, RngExt};
 use serde::{Deserialize, Serialize};
 
 use super::VariableType;
@@ -38,6 +39,11 @@ pub enum Operation {
     LoadPrivateKey([u8; 32]),
     /// Load a 32-byte channel identifier.
     LoadChannelId([u8; 32]),
+    /// Load a BOLT 2 compliant `upfront_shutdown_script`.
+    ///
+    /// Produces a [`VariableType::Bytes`] value whose contents match one of the
+    /// standard script templates required by BOLT 2.
+    LoadShutdownScript(ShutdownScriptVariant),
     /// Load the target node's public key from the program context.
     LoadTargetPubkeyFromContext,
     /// Load the chain hash from the program context.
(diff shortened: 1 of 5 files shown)


smite-ir/src/operation.rs
```diff
@@ -88,6 +94,184 @@ pub enum Operation {
...
+/// BOLT 2 for the upfront shutdown TLV. `Empty` opts out of upfront shutdown
+/// entirely and is accepted regardless of feature negotiation.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum ShutdownScriptVariant {
+    /// Zero-length script. Opts out of upfront shutdown.
+    Empty,
+    /// `OP_DUP OP_HASH160 <20-byte hash> OP_EQUALVERIFY OP_CHECKSIG`.
+    P2pkh([u8; 20]),
+    /// `OP_HASH160 <20-byte hash> OP_EQUAL`.
+    P2sh([u8; 20]),
```
NishantBansal2003, line 109: I don't see legacy P2PKH or P2SH mentioned in BOLT 2 (https://github.com/lightning/bolts/blob/master/02-peer-protocol.md#requirements-31)

However, from the code, I can see that CLN, LDK, and Eclair accept them, but LND does not (is it fine to have it like this?)

NishantBansal2003 (approved):
LGTM 👍

smite-ir/src/operation.rs
ekzyis: bolt02 mentions this:

> The `scriptpubkey` forms include only standard segwit forms accepted by the Bitcoin network, which ensures the resulting transaction will propagate to miners. However old nodes may send non-segwit scripts, which may be accepted for backwards-compatibility (with a caveat to force-close if this output doesn't meet dust relay requirements).

So I think it's for backward compatibility and LND started to drop it

smite-scenarios/src/executor.rs
```diff
@@ -327,7 +328,10 @@ fn build_open_channel(
         first_per_commitment_point: resolve_pubkey(variables, inputs[16])?,
         channel_flags: resolve_u8(variables, inputs[17])?,
         tlvs: OpenChannelTlvs {
-            upfront_shutdown_script: nonempty_or_none(resolve_bytes(variables, inputs[18])?),
+            // Always send the TLV: a zero-length value is the BOLT 2 opt-out
+            // signal when option_upfront_shutdown_script is negotiated.
+            // Omitting it is a protocol violation in that case.
+            upfront_shutdown_script: Some(resolve_bytes(variables, inputs[18])?.to_vec()),
```
ekzyis, line 335: nit: I verified that it's also okay to send `upfront_shutdown_script` when we did not signal `option_upfront_shutdown_script` during `init` exchange.

I see that this is documented for `ShutdownScriptVariant`:

> accepted regardless of feature negotiation.

Could make sense to also add a small note here?

```diff
diff --git a/smite-scenarios/src/executor.rs b/smite-scenarios/src/executor.rs
index f05481a..b0e541e 100644
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -330,7 +330,8 @@ fn build_open_channel(
         tlvs: OpenChannelTlvs {
             // Always send the TLV: a zero-length value is the BOLT 2 opt-out
             // signal when option_upfront_shutdown_script is negotiated.
-            // Omitting it is a protocol violation in that case.
+            // Omitting it is a protocol violation in that case. Including if
+            // not negotiated is not.
             upfront_shutdown_script: Some(resolve_bytes(variables, inputs[18])?.to_vec()),
             channel_type: nonempty_or_none(resolve_features(variables, inputs[19])?),
         },
```

smite-ir/src/operation.rs
```diff
@@ -88,6 +94,184 @@ pub enum Operation {
...
+
+    /// Encodes the variant as raw scriptpubkey bytes.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `AnySegwit.version` is outside
+    /// `ANYSEGWIT_MIN_VERSION..=ANYSEGWIT_MAX_VERSION`, if the program is
+    /// outside `ANYSEGWIT_MIN_PROGRAM_LEN..=ANYSEGWIT_MAX_PROGRAM_LEN` bytes,
+    /// or if `OpReturn` data is outside
+    /// `OP_RETURN_MAX_DATA_LEN..=OP_RETURN_MAX_DATA_LEN` bytes.
```
ekzyis:
```suggestion
    /// `OP_RETURN_MIN_DATA_LEN..=OP_RETURN_MAX_DATA_LEN` bytes.
```

smite-ir/src/operation.rs
```diff
@@ -88,6 +94,184 @@ pub enum Operation {
...
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        match self {
+            Self::Empty => Vec::new(),
+            Self::P2pkh(h) => {
+                let mut out = Vec::with_capacity(25);
+                out.extend_from_slice(&[0x76, 0xa9, 0x14]);
```
ekzyis, line 192: note: #51 adds the `bitcoin` crate. We could then use `bitcoin::opcodes` here in a follow-up PR. But it would require casting `OpCode` to `u8` or updating our types to use `OpCode` instead of `u8`.

ekzyis (changes requested):
Just one typo, LGTM otherwise

smite-ir/src/operation.rs
morehouse: Yes, those types have been removed from the spec.  But since the legacy code exists in multiple implementations, I figured we should exercise that code.

smite-ir/src/operation.rs
morehouse: Good catch!  Done.

smite-scenarios/src/executor.rs
morehouse: Done.

erickcestari (approved):
LGTM

PR #60, opened 2026-04-25
smite-ir: Add `InstructionDeleteMutator`

Chand-ra:
Add an `InstructionDelete` mutator for Smite IR. Mutates a given program by deleting a randomly selected instruction. Deletion of the said instruction happens by removing it from the instructions list and reindexing the subsequent instructions.

### DIFF
--- a/smite-ir-mutator/src/lib.rs
+++ b/smite-ir-mutator/src/lib.rs
@@ -30,7 +30,7 @@ use rand::rngs::SmallRng;
 use rand::{RngExt, SeedableRng};
 
 use smite_ir::generators::OpenChannelGenerator;
-use smite_ir::mutators::{InputSwapMutator, OperationParamMutator};
+use smite_ir::mutators::{InputSwapMutator, InstructionDeleteMutator, OperationParamMutator};
 use smite_ir::{Generator, Mutator, Program, ProgramBuilder};
 
 /// Mutator state owned by AFL++ across calls. Allocated by [`afl_custom_init`],
@@ -77,12 +77,20 @@ impl MutatorState {
         let stack = 1u32 << self.rng.random_range(0..=4);
         for _ in 0..stack {
             // Uniform pick between the available mutators.
-            let name = if self.rng.random() {
-                OperationParamMutator.mutate(program, &mut self.rng);
-                "op-param"
-            } else {
-                InputSwapMutator.mutate(program, &mut self.rng);
-                "input-swap"
+            let name = match self.rng.random_range(0..3) {
+                0 => {
+                    OperationParamMutator.mutate(program, &mut self.rng);
+                    "op-param"
+                }
+                1 => {
+                    InputSwapMutator.mutate(program, &mut self.rng);
+                    "input-swap"
+                }
+                2 => {
+                    InstructionDeleteMutator.mutate(program, &mut self.rng);
+                    "instr-delete"
+                }
+                _ => unreachable!(),
             };
             self.last_sequence.push(name);
         }
@@ -369,7 +377,7 @@ mod tests {
             }
             for name in suffix.split(',') {
                 assert!(
-                    name == "op-param" || name == "input-swap",
+                    name == "op-param" || name == "input-swap" || name == "instr-delete",
                     "unexpected mutator name in description: {name:?} (full: {s:?})",
                 );
             }
--- a/smite-ir-mutator/src/lib.rs
+++ b/smite-ir-mutator/src/lib.rs
@@ -44,7 +44,7 @@ use smite_ir::generators::{
     OpenChannelGenerator,
 };
 use smite_ir::minimizers::{CommonSubexpressionEliminator, DeadCodeEliminator, Minimizer};
-use smite_ir::mutators::{InputSwapMutator, OperationParamMutator};
+use smite_ir::mutators::{InputSwapMutator, InstructionDeleteMutator, OperationParamMutator};
 use smite_ir::{Generator, Mutator, Program, ProgramBuilder};
 
 /// Mutator state owned by AFL++ across calls. Allocated by [`afl_custom_init`],
@@ -98,12 +98,20 @@ impl MutatorState {
         let stack = 1u32 << self.rng.random_range(0..=4);
         for _ in 0..stack {
             // Uniform pick between the available mutators.
-            let name = if self.rng.random() {
-                OperationParamMutator.mutate(program, &mut self.rng);
-                "op-param"
-            } else {
-                InputSwapMutator.mutate(program, &mut self.rng);
-                "input-swap"
+            let name = match self.rng.random_range(0..3) {
+                0 => {
+                    OperationParamMutator.mutate(program, &mut self.rng);
+                    "op-param"
+                }
+                1 => {
+                    InputSwapMutator.mutate(program, &mut self.rng);
+                    "input-swap"
+                }
+                2 => {
+                    InstructionDeleteMutator.mutate(program, &mut self.rng);
+                    "instr-delete"
+                }
+                _ => unreachable!("random_range() bound out of sync with match arms"),
             };
             self.last_sequence.push(name);
         }
@@ -540,7 +548,7 @@ mod tests {
             }
             for name in suffix.split(',') {
                 assert!(
-                    name == "op-param" || name == "input-swap",
+                    name == "op-param" || name == "input-swap" || name == "instr-delete",
                     "unexpected mutator name in description: {name:?} (full: {s:?})",
                 );
             }
--- a/smite-ir/src/mutators/instruction_delete.rs
+++ b/smite-ir/src/mutators/instruction_delete.rs
@@ -1,0 +1,76 @@
+//! Mutator that removes an instruction by replacing it with a Nop or
+//! index renumbering.
+
+use rand::{Rng, RngExt, seq::IteratorRandom};
+
+use super::Mutator;
+use crate::{Instruction, Operation, Program};
+
+/// Deletes a randomly selected instruction by:
+/// (i) Replacing it with a Nop (50% chance).
+/// (ii) Removing it from the instuctions list and reindexing the
+/// subsequent instructions (50% chance).
+pub struct InstructionDeleteMutator;
+
+impl Mutator for InstructionDeleteMutator {
+    fn mutate(&self, program: &mut Program, rng: &mut impl Rng) -> bool {
+        if program.instructions.is_empty() {
+            return false;
+        }
+        // Pick a random instruction to delete.
+        let deleted_idx = rng.random_range(0..program.instructions.len());
+        let deleted_type = program.instructions[deleted_idx].operation.output_type();
+
+        // Heal downstream dependencies by swapping references to a prior,
+        // type-matching variable.
+        if program.instructions[(deleted_idx + 1)..]
+            .iter()
+            .any(|instr| instr.inputs.contains(&deleted_idx))
+        {
+            // Abort if no valid replacement variable exists in the preceding scope.
+            let Some(replacement_idx) = program.instructions[..deleted_idx]
+                .iter()
+                .enumerate()
+                .filter_map(|(i, instr)| {
+                    let out_type = instr.operation.output_type();
+                    (out_type.is_some() && out_type == deleted_type).then_some(i)
+                })
+                .choose(rng)
+            else {
+                return false;
+            };
+            // Update dependent inputs.
+            for instr in &mut program.instructions[(deleted_idx + 1)..] {
+                for input in &mut instr.inputs {
+                    if *input == deleted_idx {
+                        *input = replacement_idx;
+                    }
+                }
+            }
+        }
+
+        // Choose either with a uniform 50/50 split.
+        if rng.random() {
+            // Replace with Nop.
+            program.instructions[deleted_idx] = Instruction {
+                operation: Operation::Nop,
+                inputs: vec![],
+            };
+        } else {
+            // Delete from the program.
+            program.instructions.remove(deleted_idx);
+
+            // Decrement subsequent references pointing past the deleted index.
+            program.instructions[deleted_idx..]
+                .iter_mut()
+                .for_each(|instr| {
+                    for input in &mut instr.inputs {
+                        if *input > deleted_idx {
+                            *input -= 1;
+                        }
+                    }
+                });
+        }
+        true
+    }
+}
--- a/smite-ir/src/mutators/instruction_delete.rs
+++ b/smite-ir/src/mutators/instruction_delete.rs
@@ -1,0 +1,88 @@
+//! Mutator that removes an instruction.
+
+use rand::{Rng, RngExt, seq::IteratorRandom};
+
+use super::Mutator;
+use crate::Program;
+
+/// Deletes a randomly selected instruction by removing it from
+/// the instructions list and reindexing the subsequent instructions.
+pub struct InstructionDeleteMutator;
+
+impl Mutator for InstructionDeleteMutator {
+    fn mutate(&self, program: &mut Program, rng: &mut impl Rng) -> bool {
+        if program.instructions.is_empty() {
+            return false;
+        }
+        // Pick a random instruction to delete.
+        let deleted_idx = rng.random_range(0..program.instructions.len());
+        let deleted_type = program.instructions[deleted_idx].operation.output_type();
+
+        // Find the first instruction that uses the deleted instruction.
+        let first_use_idx = program
+            .instructions
+            .iter()
+            .position(|instr| instr.inputs.contains(&deleted_idx));
+
+        // If any instruction downstream depends on the deleted one, pick a prior
+        // type-matching variable to redirect those inputs to.
+        let replacement_idx = if let Some(first_use_idx) = first_use_idx {
+            let mut is_consumed = vec![false; program.instructions.len()];
+            if deleted_type
+                .expect("`None` shouldn't be consumed")
+                .is_affine()
+            {
+                for (i, instr) in program.instructions.iter().enumerate() {
+                    if i == deleted_idx {
+                        continue;
+                    }
+                    for &input in &instr.inputs {
+                        if program.instructions[input].operation.output_type() == deleted_type {
+                            is_consumed[input] = true;
+                        }
+                    }
+                }
+            }
+            match program.instructions[..first_use_idx]
+                .iter()
+                .enumerate()
+                .filter_map(|(i, instr)| {
+                    (instr.operation.output_type() == deleted_type
+                        && i != deleted_idx
+                        && !is_consumed[i])
+                        .then_some(i)
+                })
+                .choose(rng)
+            {
+                Some(idx) => {
+                    if idx > deleted_idx {
+                        // Deleting an instruction shifts every element past it by -1.
+                        Some(idx - 1)
+                    } else {
+                        Some(idx)
+                    }
+                }
+                // Abort if no valid replacement variable exists in the preceding scope.
+                None => return false,
+            }
+        } else {
+            None
+        };
+
+        // Delete from the program.
+        program.instructions.remove(deleted_idx);
+
+        // Heal downstream inputs: redirect references to the deleted index, and
+        // decrement references past it.
+        for instr in &mut program.instructions[deleted_idx..] {
+            for input in &mut instr.inputs {
+                if *input == deleted_idx {
+                    *input = replacement_idx.expect("dependent input implies replacement");
+                } else if *input > deleted_idx {
+                    *input -= 1;
+                }
+            }
+        }
+        true
+    }
+}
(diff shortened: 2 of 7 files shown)


erickcestari:
> Deletion of an instruction can happen in one of two ways: (i) Replacing it with a `Operation::Nop` (50% chance). (ii) Removing it from the instructions list and reindexing the subsequent instructions (50% chance).

Are there any advantages to having two ways to delete an operation? If not, I would prefer to use only `Operation::Nop` for simplicity and performance. Also, noping a random instruction could invalidate program correctness (e.g., if any operation references a NOP instruction, it will cause a panic). So we can only Nop instructions with no consumers.

Chand-ra:
> Are there any advantages to having two ways to delete an operation? If not, I would prefer to use only `Operation::Nop` for simplicity and performance.

If we only keep the `Nop` mutation path and the fuzzer triggers this mutator multiple times, we run into the risk of replacing most of the program with `Nop`, which will severely affect subsequent mutations. Index renumbering allows us to actually shrink the program and occasionally get rid of these inserted `Nop`s as well.

If we want to only keep one mutation scheme, I say we should go with index renumbering.

> Also, noping a random instruction could invalidate program correctness (e.g., if any operation references a NOP instruction, it will cause a panic). So we can only Nop instructions with no consumers.

Yes, we handle that in [this part](https://github.com/morehouse/smite/pull/60/changes#diff-768dac23cecd4ad40efc9a5295c5cfe2df3f617e26ee568b0f5c4c05d6d9f87eR26) of the code.

We first check if any operation uses our deleted instruction as input. If yes, we search for a previously defined instruction with the same type as the deleted one (returns false if no such instruction exists) and then replace instances of our deleted instruction with the found replacement.

morehouse:
Perhaps there's a middle path -- we could have our mutators insert `Nop`, and have our nop-minimizer (see #54) rewrite the minimized program without nops?  That would allow mutators to do the simple, quick thing, while minimizers periodically clean up the nops during the afl-trim phase.  WDYT?

erickcestari:
> Perhaps there's a middle path -- we could have our mutators insert `Nop`, and have our nop-minimizer (see #54) rewrite the minimized program without nops? That would allow mutators to do the simple, quick thing, while minimizers periodically clean up the nops during the afl-trim phase. WDYT?

Sounds good to me.

Chand-ra:
> Perhaps there's a middle path -- we could have our mutators insert `Nop`, and have our nop-minimizer (see #54) rewrite the minimized program without nops? That would allow mutators to do the simple, quick thing, while minimizers periodically clean up the nops during the afl-trim phase. WDYT?

If I'm not wrong, the [`nop-minimizer`](https://github.com/morehouse/smite/pull/54/changes#diff-4238813de9864c079ea5cd150cb6d61c8337796a6f96a0eb49b99c114fffbebb) replaces unused instructions with `NOP`. So to achieve your suggestion, we would need to modify the minimizer to remove the added `NOP`s, or write a separate minimizer that cleans up these `NOP`s (although I guess `nop-minimizer` should do this anyway).

Looks to me like we're deferring complexity in the mutator to other parts of the codebase. Why not make the mutator never introduce `NOP`s in the first place?

Unless of course, if it's better for `nop-minimizer` to always get rid of dead instructions instead of nopping them, in which case I agree that this would be a good idea.

erickcestari:
After reconsidering, it likely makes sense to remove `NOP` entirely. Its only real justification would be performance, but the fuzzer mutator is not the bottleneck. The executor, performing cryptographic operations and I/O over a TCP, handshake-bound connection, is orders of magnitude more expensive than any overhead introduced here. I also think this PR could drop the 50% chance of replacing with NOP and keep only the remove-and-reindex path.

I'll rewrite my trimmer/minimizer PR to also do remove-and-reindex, but more aggressively, for all instructions that have no consumers, while maintaining the same coverage. Subsequent mutations should then be more efficient, since they won't waste exec budget mutating dead IR.

What do you think @morehouse @Chand-ra ?

Chand-ra:
> After reconsidering, it likely makes sense to remove `NOP` entirely. Its only real justification would be performance, but the fuzzer mutator is not the bottleneck. The executor, performing cryptographic operations and I/O over a TCP, handshake-bound connection, is orders of magnitude more expensive than any overhead introduced here. I also think this PR could drop the 50% chance of replacing with NOP and keep only the remove-and-reindex path.
> 
> I'll rewrite my trimmer/minimizer PR to also do remove-and-reindex, but more aggressively, for all instructions that have no consumers, while maintaining the same coverage. Subsequent mutations should then be more efficient, since they won't waste exec budget mutating dead IR.
> 
> What do you think @morehouse @Chand-ra ?

Sounds good to me. But as Matt mentioned [earlier](https://github.com/morehouse/smite/pull/60#issuecomment-4330677455), maybe inserting `NOP`s here and deferring cleanup to the minimizer is a better idea? That way we can get rid of some complexity in this mutator AND avoid duplicating reindexing logic.

erickcestari:
> Sounds good to me. But as Matt mentioned [earlier](https://github.com/morehouse/smite/pull/60#issuecomment-4330677455), maybe inserting `NOP`s here and deferring cleanup to the minimizer is a better idea? That way we can get rid of some complexity in this mutator AND avoid duplicating reindexing logic.

Either way is fine to me. If you both prefer going with `NOP`s I'll update my PR for it.

Chand-ra:
> > Sounds good to me. But as Matt mentioned [earlier](https://github.com/morehouse/smite/pull/60#issuecomment-4330677455), maybe inserting `NOP`s here and deferring cleanup to the minimizer is a better idea? That way we can get rid of some complexity in this mutator AND avoid duplicating reindexing logic.
> 
> Either way is fine to me. If you both prefer going with `NOP`s I'll update my PR for it.

I'll let Matt take the call on this, seeing how `afl_custom_trim` isn't always called after a custom mutate, so we could still run into the risk of nopping the entire program. Either way, we need to keep the following in mind when choosing one of the mutation schemes:

- Only nopping runs into the risk of replacing most of the program with `NOP`, which may severely affect subsequent mutations, especially long runs. It is however lower on complexity (in terms of both execution and implementation) and the too-much-nopping problem may be mitigated by the newly introduced [`dead_code`](https://github.com/morehouse/smite/pull/54/changes#diff-c410667f78f547c2725e44b3ff6a60e654242c41d2ae6a31eca7c39fea3d0dc5) custom trimmer.

- Reindexing has higher complexity (again, in terms of both execution and implementation) but it never introduces dead code into the program, which is a more efficient use of fuzzing cycles. However, we duplicate the re-indexing logic implemented in the `dead_code` trimmer. 

I'm personally leaning towards reindexing because:
1. I'm unsure about the frequency of custom trimming after each custom mutation, which is the only way nopping makes sense.
2. As Erick noted, the mutator's time complexity is nothing compared to other phases of fuzzing.
3. The implementation complexity of either of these paths is minuscule compared to the mutator's healing phase, which we always need to perform.
4. Duplicating a bit of healing logic doesn't sound like the end of world to me.

morehouse:
Let's ditch Nop entirely and rewrite instead.

Chand-ra:
Rewrote the mutator and accompanying tests without the `NOP` path and rebased on top of the latest master to resolve a merge conflict.

smite-ir/src/mutators/instruction_delete.rs
```diff
@@ -0,0 +1,64 @@
...
+                return false;
+            };
+            // Update dependent inputs.
+            for instr in &mut program.instructions[(deleted_idx + 1)..] {
+                for input in &mut instr.inputs {
+                    if *input == deleted_idx {
+                        *input = replacement_idx;
+                    }
+                }
+            }
```
morehouse: This pass can be combined with the decrement pass below.

smite-ir/src/mutators/instruction_delete.rs
```diff
@@ -0,0 +1,64 @@
...
+            .iter()
+            .any(|instr| instr.inputs.contains(&deleted_idx))
+        {
+            // Abort if no valid replacement variable exists in the preceding scope.
+            let Some(replacement_idx) = program.instructions[..deleted_idx]
+                .iter()
+                .enumerate()
+                .filter_map(|(i, instr)| {
+                    let out_type = instr.operation.output_type();
+                    (out_type.is_some() && out_type == deleted_type).then_some(i)
```
morehouse: Nit: the `is_some` check is unnecessary since `deleted_type` should never be `None` if we get here.


```suggestion
                    (instr.operation.output_type() == deleted_type).then_some(i)
```

smite-ir/src/mutators/instruction_delete.rs
```diff
@@ -0,0 +1,64 @@
+//! Mutator that removes an instruction.
+
+use rand::{Rng, RngExt, seq::IteratorRandom};
+
+use super::Mutator;
+use crate::Program;
+
+/// Deletes a randomly selected instruction by removing it from
+/// the instuctions list and reindexing the subsequent instructions.
```
morehouse:
```suggestion
/// the instructions list and reindexing the subsequent instructions.
```

smite-ir/src/mutators/instruction_delete.rs
```diff
@@ -0,0 +1,64 @@
...
+        // Decrement subsequent references pointing past the deleted index.
+        program.instructions[deleted_idx..]
+            .iter_mut()
+            .for_each(|instr| {
+                for input in &mut instr.inputs {
+                    if *input > deleted_idx {
+                        *input -= 1;
+                    }
+                }
+            });
```
morehouse: Let's match the regular for-loop style used elsewhere in this file and the rest of the codebase.


```suggestion
        for instr in &mut program.instructions[deleted_idx..] {
            for input in &mut instr.inputs {
                if *input > deleted_idx {
                    *input -= 1;
                }
            }
        }
```

smite-ir/src/tests.rs
```diff
@@ -1058,3 +1058,137 @@ fn input_swap_preserves_types() {
...
+
+    for _ in 0..100 {
+        mutator.mutate(&mut program, &mut rng);
+    }
+    assert_ne!(
+        program, original,
+        "InstructionDeleteMutator never changed the program"
+    );
+}
+
```
morehouse, line 1708: We should test that programs are unchanged when `mutate` returns false.

smite-ir/src/tests.rs
```diff
@@ -1058,3 +1058,137 @@ fn input_swap_preserves_types() {
...
+    let mutator = InstructionDeleteMutator;
+    let mut rng = SmallRng::seed_from_u64(0);
+    let mut verified_shift = false;
+
+    for _ in 0..100 {
+        let mut program = original.clone();
+        if mutator.mutate(&mut program, &mut rng) &&
+            program.instructions.len() == 2 &&
+            // Check if it deleted index 1, meaning DerivePoint is now at index 1
+            program.instructions[1].operation == Operation::DerivePoint
```
morehouse, line 1784: The comment isn't quite accurate -- `DerivePoint` could also be at index 1 if index 0 was deleted.

smite-ir/src/tests.rs
```diff
@@ -1058,3 +1058,137 @@ fn input_swap_preserves_types() {
...
+            verified_shift = true;
+            break;
+        }
+    }
+    assert!(
+        verified_shift,
+        "mutator never took the expected target shift path"
+    );
+}
+
```
morehouse, line 1797: It would be good to also test deletion of a void-output operation (i.e. `SendMessage`).

smite-ir/src/tests.rs
```diff
@@ -1058,3 +1058,137 @@ fn input_swap_preserves_types() {
...
+                            panic!("instruction {i} input {j}: references void at {input_idx}")
+                        });
+                    assert_eq!(
+                        actual_type, expected_types[j],
+                        "instruction {i} input {j}: expected {:?}, got {actual_type:?}",
+                        expected_types[j],
+                    );
+                }
+            }
+        }
```
morehouse, line 1914: We can simplify and strengthen the test using `Program::validate`.


```suggestion
        if mutator.mutate(&mut program, &mut rng) {
            program.validate().expect("InstructionDeleteMutator produces valid programs");
        }
```

morehouse (commented):
The same concern described in https://github.com/morehouse/smite/pull/61#pullrequestreview-4239823431 applies here:  if we delete a `SendMessage` that precedes a `Recv*` operation, we may force a 5s timeout when executing the input.

I think we need a good solution to this issue before we can merge this mutator.

smite-ir/src/mutators/instruction_delete.rs
Chand-ra: That would require elevating `replacement_idx` to a higher scope, before we even check whether we need a replacement or not. I don't think it's a good idea to pollute the higher scope like that. Also, then we would need to add extra logic to the decrement pass for when `replacement_idx` is `None`.

I think this is better as is, where we divide the healing phase into two logical chunks: i) finding and integrating a replacement if needed ii) healing subsequent operations.

smite-ir/src/tests.rs
Chand-ra: I cannot say I see the merit in it, the only "interesting" thing that happens (from the mutator's perspective) on deleting a void-output operation is index-shifting of downstream instructions, which is tested here anyway. Maybe I'm missing something?

Chand-ra:
Haven't squashed the changes yet because we won't merge this without solving the 'implicit dependence' issue anyway, and it should ease the reviewing.

smite-ir/src/mutators/instruction_delete.rs
morehouse: It's a style choice, so feel free to ignore it, but I do think my suggestion leads to slightly simpler code overall.

> That would require elevating `replacement_idx` to a higher scope, before we even check whether we need a replacement or not. I don't think it's a good idea to pollute the higher scope like that.

It's a 50-line function; I don't think a variable at that scope is *pollution*.  We're just elevating `replacement_idx` to the same scope as `deleted_idx` and `deleted_type`.

> Also, then we would need to add extra logic to the decrement pass for when `replacement_idx` is `None`.

Not really -- just an `expect()`.

Here's what my suggestion would look like.  It's ~10 lines shorter and eliminates one pass.  You can decide if you like the style better or worse.

```rust
  let deleted_idx = rng.random_range(0..program.instructions.len());                                                                                                                                                
  let deleted_type = program.instructions[deleted_idx].operation.output_type();                                                                                                                                     
                                                                                                                                                                                                                    
  // If anything downstream depends on the deleted instruction, pick a prior                                                                                                                                        
  // type-matching variable to redirect those inputs to.                                                                                                                                                            
  let replacement_idx = if program.instructions[(deleted_idx + 1)..]                                                                                                                                                
      .iter()                                                                                                                                                                                                       
      .any(|instr| instr.inputs.contains(&deleted_idx))                                                                                                                                                             
  {                                                                                                                                                                                                                 
      match program.instructions[..deleted_idx]                                                                                                                                                           
          .iter()                                                                                                                                                                                                   
          .enumerate()                                                                                                                                                                                              
          .filter_map(|(i, instr)| {                                                                                                                                                                                
              (instr.operation.output_type() == deleted_type).then_some(i)                                                                                                                                          
          })                                                                                                                                                                                                        
          .choose(rng)
      {                                                                                                                                                                                              
          Some(idx) => Some(idx),
          None => return false,                                                                                                                                                                                                                                                                                                                                                                
      }                                                                                                                                                                                                                                                                                                                                                                                                        
  } else {                                                                                                                                                                                                          
      None                                                                                                                                                                                                          
  };                                                                                                                                                                                                                
                                                                                                                                                                                                                    
  program.instructions.remove(deleted_idx);                                                                                                                                                                         
                                                                                                                                                                                                                    
  // Heal downstream inputs: redirect references to the deleted index, and                                                                                                                                          
  // decrement references past it.                                                                                                                                                                                  
  for instr in &mut program.instructions[deleted_idx..] {                                                                                                                                                           
      for input in &mut instr.inputs {                                                                                                                                                                              
          if *input == deleted_idx {                                                                                                                                                                                
              *input = replacement_idx.expect("dependent input implies replacement");                                                                                                                                     
          } else if *input > deleted_idx {                                                                                                                                                                          
              *input -= 1;                                                                                                                                                                                          
          }                                                                                                                                                                                                         
      }                                                                                                                                                                                                             
  }                                                                                                       
  true
```

Chand-ra: Hmm, I think both the approaches have similar levels of complexity, but this one manages to perform the healing in one go which gives it an upper hand. Will rewrite, thanks!

Chand-ra:
We need to make this mutator respect affine rules by construction now, but I am confused on how to go about this. If we select an affine-producing operation to delete, we can either:
1. If the affine producer has consumers, return false (abort).
2. Try to heal the downstream consumer with another candidate. If it's an affine type, strictly require an unspent candidate. If none exists, abort (return false).
3. If the healing pass cannot find an unspent replacement for the consumer, delete the consumer too. So basically, cascading deletions.

I don't think we want to do 1, but I'm unsure which one is better among 2 and 3. WDYT @morehouse @NishantBansal2003?

Edit: After some thought, I think 3 wouldn't be right either. We'd have to implement an entirely different scheme for affine types and it would delete entire protocol flows which is not what we're trying to do with this mutator, but rather delete a single instruction.

NishantBansal2003:
IMO, deletion of both `Send` and `Recv` (option 3) doesn’t really buy us anything. Once that is removed, the target never sends or receives any messages, so the mutation has no observable effect from its perspective.

I’d go with option 2: try to heal the `Recv` by redirecting it to another unspent `Sent`. If that isn’t possible, then just abort the deletion and leave the original `Sent` in place as well

EDIT: I didn’t see your edit ;). Looks like we came to the same conclusion

smite-ir/src/mutators/instruction_delete.rs
```diff
@@ -0,0 +1,77 @@
...
+        // when deleted_type is affine.
+        let mut is_consumed = vec![false; program.instructions.len()];
+        for idx in 0..program.instructions.len() {
+            if idx == deleted_idx {
+                continue;
+            }
+            for input in &program.instructions[idx].inputs {
+                is_consumed[*input] = true;
+            }
+        }
```
morehouse: This pass is only needed if we delete an instruction that outputs an affine variable that is consumed by a later operation.  So we can move it into the block below, guarded by a `deleted_type.is_affine()` check.

We also only need set `is_consumed[i] = true` when the output type matches `deleted_type`, which then simplifies the logic in the `filter_map` below.

smite-ir-mutator/src/lib.rs
```diff
@@ -96,12 +96,20 @@ impl MutatorState {
...
+                }
+                1 => {
+                    InputSwapMutator.mutate(program, &mut self.rng);
+                    "input-swap"
+                }
+                2 => {
+                    InstructionDeleteMutator.mutate(program, &mut self.rng);
+                    "instr-delete"
+                }
+                _ => unreachable!(),
```
morehouse: Nit:


```suggestion
                _ => unreachable!("random_range() bound out of sync with match arms"),
```

smite-ir/src/mutators/instruction_delete.rs
```diff
@@ -0,0 +1,75 @@
...
+                        continue;
+                    }
+                    for &input in &instr.inputs {
+                        if program.instructions[input].operation.output_type() == deleted_type {
+                            is_consumed[input] = true;
+                        }
+                    }
+                }
+            }
+            match program.instructions[..deleted_idx]
```
morehouse: We should actually be able to choose a replacement from the wider range `[..first_use_idx]`, which would increase the effectiveness of this mutator.

smite-ir/src/tests.rs
```diff
@@ -1522,6 +1438,172 @@ fn input_swap_preserves_affine() {
...
+    }
+    assert!(
+        verified_shift,
+        "InstructionDeleteMutator never took the expected target shift path"
+    );
+}
+
+#[test]
+fn instr_delete_respects_affine_rules() {
```
morehouse: It would also be worth testing that an affine consumer can have its input redirected to another unconsumed affine variable.

smite-ir/src/tests.rs
```diff
@@ -1522,6 +1438,172 @@ fn input_swap_preserves_affine() {
...
+        if mutator.mutate(&mut program, &mut rng) {
+            assert!(program.instructions.len() == original.instructions.len() - 1);
+            assert!(
+                // The first `RecvAcceptChannel` was deleted.
+                program.instructions[2] == original.instructions[3] ||
+                // The second `RecvAcceptChannel` was deleted.
+                program.instructions == original.instructions[0..original.instructions.len() - 1],
+                "InstructionDeleteMutator deleted an unhealable instruction"
+            );
+        }
```
morehouse, line 1841: We should probably also assert that at least one mutation succeeds.

morehouse (approved):
LGTM

morehouse:
Smoke tested a 10-minute IR fuzz run against LDK.  No issues.

PR #61, opened 2026-04-25
smite-ir: Add `InstructionReorderMutator`

Chand-ra:
Add an `InstructionReorder` mutator for Smite IR. Mutates a given program by swapping two '"Act" instructions that have no data dependencies between them. This explores alternative execution orderings while preserving SSA invariants.

### DIFF
--- a/smite-ir-mutator/src/lib.rs
+++ b/smite-ir-mutator/src/lib.rs
@@ -30,7 +30,7 @@ use rand::rngs::SmallRng;
 use rand::{RngExt, SeedableRng};
 
 use smite_ir::generators::OpenChannelGenerator;
-use smite_ir::mutators::{InputSwapMutator, OperationParamMutator};
+use smite_ir::mutators::{InputSwapMutator, InstructionReorderMutator, OperationParamMutator};
 use smite_ir::{Generator, Mutator, Program, ProgramBuilder};
 
 /// Mutator state owned by AFL++ across calls. Allocated by [`afl_custom_init`],
@@ -77,12 +77,20 @@ impl MutatorState {
         let stack = 1u32 << self.rng.random_range(0..=4);
         for _ in 0..stack {
             // Uniform pick between the available mutators.
-            let name = if self.rng.random() {
-                OperationParamMutator.mutate(program, &mut self.rng);
-                "op-param"
-            } else {
-                InputSwapMutator.mutate(program, &mut self.rng);
-                "input-swap"
+            let name = match self.rng.random_range(0..3) {
+                0 => {
+                    OperationParamMutator.mutate(program, &mut self.rng);
+                    "op-param"
+                }
+                1 => {
+                    InputSwapMutator.mutate(program, &mut self.rng);
+                    "input-swap"
+                }
+                2 => {
+                    InstructionReorderMutator.mutate(program, &mut self.rng);
+                    "instr-reorder"
+                }
+                _ => unreachable!(),
             };
             self.last_sequence.push(name);
         }
@@ -369,7 +377,7 @@ mod tests {
             }
             for name in suffix.split(',') {
                 assert!(
-                    name == "op-param" || name == "input-swap",
+                    name == "op-param" || name == "input-swap" || name == "instr-reorder",
                     "unexpected mutator name in description: {name:?} (full: {s:?})",
                 );
             }
--- a/smite-ir-mutator/src/lib.rs
+++ b/smite-ir-mutator/src/lib.rs
@@ -44,7 +44,9 @@ use smite_ir::generators::{
     OpenChannelGenerator,
 };
 use smite_ir::minimizers::{CommonSubexpressionEliminator, DeadCodeEliminator, Minimizer};
-use smite_ir::mutators::{InputSwapMutator, InstructionDeleteMutator, OperationParamMutator};
+use smite_ir::mutators::{
+    InputSwapMutator, InstructionDeleteMutator, InstructionReorderMutator, OperationParamMutator,
+};
 use smite_ir::{Generator, Mutator, Program, ProgramBuilder};
 
 /// Mutator state owned by AFL++ across calls. Allocated by [`afl_custom_init`],
@@ -98,7 +100,7 @@ impl MutatorState {
         let stack = 1u32 << self.rng.random_range(0..=4);
         for _ in 0..stack {
             // Uniform pick between the available mutators.
-            let name = match self.rng.random_range(0..3) {
+            let name = match self.rng.random_range(0..4) {
                 0 => {
                     OperationParamMutator.mutate(program, &mut self.rng);
                     "op-param"
@@ -110,6 +112,10 @@ impl MutatorState {
                 2 => {
                     InstructionDeleteMutator.mutate(program, &mut self.rng);
                     "instr-delete"
+                }
+                3 => {
+                    InstructionReorderMutator.mutate(program, &mut self.rng);
+                    "instr-reorder"
                 }
                 _ => unreachable!("random_range() bound out of sync with match arms"),
             };
@@ -548,7 +554,10 @@ mod tests {
             }
             for name in suffix.split(',') {
                 assert!(
-                    name == "op-param" || name == "input-swap" || name == "instr-delete",
+                    name == "op-param"
+                        || name == "input-swap"
+                        || name == "instr-delete"
+                        || name == "instr-reorder",
                     "unexpected mutator name in description: {name:?} (full: {s:?})",
                 );
             }
--- a/smite-ir/src/mutators/instruction_reorder.rs
+++ b/smite-ir/src/mutators/instruction_reorder.rs
@@ -1,0 +1,81 @@
+//! Mutator that swaps Act instructions.
+
+use rand::{Rng, RngExt, seq::IteratorRandom};
+
+use super::Mutator;
+use crate::Program;
+
+/// Swaps two `Act` instructions that have no data dependencies between them.
+/// This explores alternative execution orderings while preserving SSA invariants.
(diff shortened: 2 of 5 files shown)


Chand-ra:
Drafting right now because tests for the mutator are yet to be implemented.

smite-ir/src/mutators/instruction_reorder.rs
```diff
@@ -0,0 +1,81 @@
...
+            candidates_count += 1;
+            if rng.random_range(0..candidates_count) == 0 {
+                act2_idx = i;
+            }
+        }
+
+        // Abort if no valid independent Act instructions exist in the range.
+        if act2_idx == act1_idx {
+            return false;
+        }
```
morehouse: We can simplify using the built-in reservoir sampling.  Something like this:


```suggestion
        let Some(act2_idx) = ((act1_idx + 1)..=usage_boundary)
            .filter(|&i| {
                program.instructions[i].operation.is_act() &&
                    !program.instructions[i].inputs.iter().any(|&input| input >= act1_idx)
            })
            .choose(rng)
        else {
            return false;
        }
```

smite-ir/src/mutators/instruction_reorder.rs
```diff
@@ -0,0 +1,81 @@
...
+            .instructions
+            .iter()
+            .enumerate()
+            .filter_map(|(i, instr)| {
+                if instr.operation.is_act() {
+                    Some(i)
+                } else {
+                    None
+                }
+            })
```
morehouse:
```suggestion
            .filter_map(|(i, instr)| instr.operation.is_act().then_some(i))
```

smite-ir/src/mutators/instruction_reorder.rs
```diff
@@ -0,0 +1,81 @@
...
+        if act2_idx == act1_idx {
+            return false;
+        }
+
+        // Swap Act_1 and Act_2.
+        program.instructions.swap(act1_idx, act2_idx);
+
+        // Healing: Update downstream references to Act_1 and Act_2.
+        for instr in &mut program.instructions[(act2_idx + 1)..] {
+            instr.inputs.iter_mut().for_each(|input| {
```
morehouse:
```suggestion
            for input in &mut instr.inputs {
```

morehouse (commented):
This all looks correctly implemented to me.

Unfortunately I don't think it's useful until we implement more Act operations and start generating more interesting programs.  Currently we just have `SendMessage` and `RecvAcceptChannel`, which we actually don't want to swap (doing so would lead to 5s timeout delays as we wait to receive a message that never comes, which would register as a hang).

I think it would be best to defer further work on this mutator until we have some more interesting things to swap.  Then we can figure out how to avoid swapping a `Recv*` operation to a location before a `SendMessage`.  We will also then be able to collect better data about how often swapping is actually possible with the current implementation, which may lead us to investigate further changes to this mutator (e.g., moving entire dependency trees).

smite-ir/src/mutators/instruction_reorder.rs
Chand-ra: Takes O(N) space now but I guess it's worth it given the massive simplification.

morehouse: It looks O(1) to me.  What part allocates O(N)?

Chand-ra: Ugh, my bad, I mistook `filter()` as an allocating operation.

smite-ir/src/operation.rs
```diff
@@ -924,4 +924,42 @@ impl Operation {
             | Self::BroadcastTransaction => false,
         }
     }
+
+    /// Returns true for Act instructions.
+    #[must_use]
+    pub fn is_act(&self) -> bool {
```
morehouse: `has_side_effects` is almost identical to this function.  We should add a comment explaining why we need a different function.

smite-ir-mutator/src/lib.rs
```diff
@@ -98,12 +98,20 @@ impl MutatorState {
...
+                }
+                1 => {
+                    InputSwapMutator.mutate(program, &mut self.rng);
+                    "input-swap"
+                }
+                2 => {
+                    InstructionReorderMutator.mutate(program, &mut self.rng);
+                    "instr-reorder"
+                }
+                _ => unreachable!(),
```
morehouse: Nit:


```suggestion
                _ => unreachable!("random_range() bound out of sync with match arms"),
```

smite-ir/src/tests.rs
```diff
@@ -1688,6 +1690,133 @@ fn input_swap_preserves_affine() {
...
+        mutated,
+        "InstructionReorderMutator never mutated the program"
+    );
+
+    // RecvAcceptChannel and SendMessage should've been swapped.
+    assert!(program.instructions[2].operation == Operation::SendMessage);
+    assert!(program.instructions[4].operation == Operation::RecvAcceptChannel);
+    // The inteleaved instruction shouldn't have been changed.
+    assert!(program.instructions[3].operation == Operation::LoadChainHashFromContext);
+    // Downstream instructions should be healed.
```
morehouse, line 2052: Nit: prefer using `assert_eq!`

smite-ir/src/tests.rs
```diff
@@ -1688,6 +1690,133 @@ fn input_swap_preserves_affine() {
...
+    assert_eq!(program.instructions[5].inputs[0], 4);
+}
+
+#[test]
+fn instr_reorder_preserves_well_formedness() {
+    let original = generate_open_channel_program(0);
+    assert_mutator_preserves_well_formedness(&InstructionReorderMutator, &original);
```
morehouse, line 2076: I'm pretty sure this test is vacuous.  `generate_open_channel_program` is never able to be mutated by `InstructionReorderMutator`.

smite-ir/src/operation.rs
```diff
@@ -924,4 +924,42 @@ impl Operation {
...
+        match self {
+            Self::SendMessage
+            | Self::SendOpenChannel
+            | Self::RecvAcceptChannel
+            | Self::MineBlocks(_)
+            | Self::BroadcastTransaction
+            | Self::SendFundingCreated => true,
```
morehouse: Please keep match arms in declaration order.

morehouse (approved):
I think this is a good first implementation of a reordering mutator.  Currently it isn't useful, since we never generate programs with independent act operations, but once we have splice or generator-insertion mutators it should be more useful.

I also think it would be worth evaluating this mutator against an alternative implementation that moves and *inserts* an act operation elsewhere instead of swapping.  That seems like a more general mutation that *may* be more effective.

Chand-ra:
> I also think it would be worth evaluating this mutator against an alternative implementation that moves and _inserts_ an act operation elsewhere instead of swapping. That seems like a more general mutation that _may_ be more effective.

Hmm, not very sure about this. The target does not, after all, care about whether we `LoadBytes` or `DerivePoint` before sending it a message or mining blocks. I do agree that it is a more general version of this mutator. Maybe there's some edge case that I cannot think of.

smite-ir/src/operation.rs
Chand-ra: This confused me as well. _Do_ we need a different function, instead of making do with `has_side_effects()`? Here's how Act instructions are defined:
```
/ -- Act: side effects against the target --
```
The only difference between the two is that `is_act()` returns false for `CreateFundingTransaction` and `has_side_effects()` returns true. 

This is because `CreateFundingTransaction` isn't mentioned in the list of Act instructions, but I think it should because it interacts with the Bitcoin CLI similar to `MineBlocks` and `MineBlocks` is listed as an Act instruction.

morehouse: Either way is fine with me.

Really the only benefit of having a separate function is to remove any `Create*` operations that only interact with the Bitcoin CLI and not the target.  In practice I'd expect such functions to have minimal effect on the target itself, and I'm not sure rearranging them will make a big difference.  For corpus minimization we might care more about the side effects because we're trying to preserve the *exact* same coverage on the target (or else AFL++ rejects the minimization).

morehouse:
> > I also think it would be worth evaluating this mutator against an alternative implementation that moves and _inserts_ an act operation elsewhere instead of swapping. That seems like a more general mutation that _may_ be more effective.
> 
> Hmm, not very sure about this. The target does not, after all, care about whether we `LoadBytes` or `DerivePoint` before sending it a message or mining blocks. I do agree that it is a more general version of this mutator. Maybe there's some edge case that I cannot think of.

Here's an example where the insertion mutator could mutate the program in ways a swap mutator could not:

```
v1 = BuildOpenChannel(...)
v2 = SendOpenChannel(v1)
v3 = RecvAcceptChannel(v2)
v4 = ExtractAcceptChannelChannelType(v3)
MineBlocks(6)
```

The swap mutator cannot modify the program at all.  The insertion mutator can move `MineBlocks` to any point between or before the other Act operations.

smite-ir/src/operation.rs
Chand-ra: I'll go ahead and drop `is_act()`.
1. I don't see a point in artificially separating operations that perform Bitcoin CLI interactions.
2.  Maintaining two methods that are 99% identical sounds like a maintenance trap.
3. It may be that swapping such functions has minimal effect on the target itself, but we're not decreasing the mutator's surface area. Since the change is a net-positive, or neutral at worst, I think it's worth making.

Chand-ra:
Will rebase after review.

morehouse (approved):
Ready to rebase

Chand-ra:
Rebased on top of the latest master.

morehouse (approved):
LGTM.

Smoke tested against LDK IR scenario for 10 minutes.

PR #64, opened 2026-04-29
Remove cargo afl installation

tokatoka:
If I understand the Dockerfile correctly, smite is not really instrumenting any Rust code. (and neither is it using `cargo afl` subcommands)

so I think we can delete it from the `Dockerfile`

fwiw; it was [removed from fuzzamoto](https://github.com/dergoegge/fuzzamoto/commit/98dc4d223e46f9470360cf4b69ef8b51d0aa0f1e) too

### DIFF
--- a/workloads/ldk/Dockerfile
+++ b/workloads/ldk/Dockerfile
@@ -51,7 +51,6 @@ WORKDIR /ldk-wrapper
 COPY workloads/ldk/Cargo.toml workloads/ldk/Cargo.lock ./
 COPY workloads/ldk/src/ src/
 ENV AFL_NO_CFG_FUZZING=1
-RUN cargo afl build --release
 
 # Copy smite workspace files and build the Rust ldk-scenario binary
 WORKDIR /smite


ekzyis (approved):
makes sense to me

erickcestari:
> If I understand the Dockerfile correctly, smite is not really instrumenting any Rust code. (and neither is it using `cargo afl` subcommands)

It's instrumenting the rust-lighting (LDK) code, that's why it's only used in the LDK Dockerfile. It won't instrument any Smite code. If we remove it, the fuzzer will not get any coverage feedback from LDK workloads.

Here, it compiles rust-lightning with instrumentation.
```rust
cargo afl build --release
```
 
> fwiw; it was [removed from fuzzamoto](https://github.com/dergoegge/fuzzamoto/commit/98dc4d223e46f9470360cf4b69ef8b51d0aa0f1e) too

I believe they removed it because they only need to instrument the C++/Go code, and they don't need to compile Fuzzamoto with instrumentation.

Here's a bit more context: #2

tokatoka:
I see. 
I thought this cargo-afl came from the dockerfile of fuzzamoto, but didn't know it was really actually used.
thanks for clarifying this!

PR #56, opened 2026-04-24
smite-scenarios: implement `IrScenario`

morehouse:
Enables end-to-end IR fuzzing of all 4 targets.

`IrScenario` is parameterized by a `SnapshotSetup` that determines the snapshot restore point.  Currently we only implement a `PostInitSetup` tailored towards the `open_channel` funding flow, but we can add more snapshot setups as needed.

Ref: #5 (Milestone 1)

PR #67, opened 2026-05-07
ci: Build every target image with all scenarios

ekzyis:
This should catch if a PR breaks building a target image.

#64 did, but I didn't run `docker build` before approving.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,32 @@
+name: Docker
+
+on:
+  push:
+    branches: ["master"]
+  pull_request:
+    branches: ["master"]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target: [lnd, ldk, cln, eclair]
+      # don't skip other targets if one fails
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: docker/setup-buildx-action@v4
+
+      - uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: workloads/${{ matrix.target }}/Dockerfile
+          build-args: SCENARIO=ir
+          cache-from: type=gha,scope=${{ matrix.target }}
+          cache-to: type=gha,mode=max,scope=${{ matrix.target }}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -19,7 +19,7 @@ jobs:
       # don't skip other targets if one fails
       fail-fast: false
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - uses: docker/setup-buildx-action@v4
 
--- a/workloads/cln/Dockerfile
+++ b/workloads/cln/Dockerfile
@@ -4,7 +4,6 @@ FROM debian:bookworm AS builder
 ARG LLVM_V=19
 ARG CLN_VERSION=v25.12.1
 ARG BITCOIN_VERSION=29.2
-ARG SCENARIO
 
 ENV LLVM_V=${LLVM_V}
 
@@ -81,7 +80,7 @@ RUN CC=afl-clang-lto ./configure --prefix=/usr/local --disable-rust \
 COPY workloads/cln/build-cln-lto.sh /tmp/
 RUN chmod +x /tmp/build-cln-lto.sh && /tmp/build-cln-lto.sh
 
-# Copy smite workspace files and build the Rust cln-scenario binary
+# Copy smite workspace files and build all scenario binaries
 WORKDIR /smite
 COPY Cargo.toml Cargo.lock ./
 COPY smite/ smite/
@@ -89,8 +88,10 @@ COPY smite-ir/ smite-ir/
 COPY smite-ir-mutator/ smite-ir-mutator/
 COPY smite-nyx-sys/ smite-nyx-sys/
 COPY smite-scenarios/ smite-scenarios/
-RUN TARGET_MAP_SIZE=$(cat /tmp/total-map-size) \
-    cargo build -p smite-scenarios --bin cln_${SCENARIO} --release --features nyx
+RUN set -eu; for f in smite-scenarios/src/bin/cln_*.rs; do \
+        TARGET_MAP_SIZE=$(cat /tmp/total-map-size) \
+            cargo build -p smite-scenarios --bin "$(basename $f .rs)" --release --features nyx; \
+    done
 
 # Build crash handler shared libraries, LD_PRELOADed into CLN to bypass its
 # blocking crashdump signal handler. Two variants:
--- a/workloads/lnd/Dockerfile
+++ b/workloads/lnd/Dockerfile
@@ -5,7 +5,6 @@ ARG LLVM_V=19
 ARG GO_VERSION=1.24.11
 ARG LND_VERSION=v0.20.0-beta
 ARG BITCOIN_VERSION=29.2
-ARG SCENARIO
 
 ENV LLVM_V=${LLVM_V}
 
@@ -62,7 +61,7 @@ RUN cd cmd/lncli && go build
 COPY ./workloads/lnd/sancov.go /lnd/sancov.go
 RUN cd cmd/lnd && CGO_ENABLED=1 go build -v -tags=libfuzzer -gcflags=all=-d=libfuzzer
 
-# Copy smite workspace files and build the Rust lnd-scenario binary
+# Copy smite workspace files and build all scenario binaries
 WORKDIR /smite
 COPY Cargo.toml Cargo.lock ./
 COPY smite/ smite/
@@ -70,7 +69,10 @@ COPY smite-ir/ smite-ir/
 COPY smite-ir-mutator/ smite-ir-mutator/
 COPY smite-nyx-sys/ smite-nyx-sys/
 COPY smite-scenarios/ smite-scenarios/
-RUN TARGET_PATH=/lnd/cmd/lnd/lnd cargo build -p smite-scenarios --bin lnd_${SCENARIO} --release --features nyx
+RUN set -eu; for f in smite-scenarios/src/bin/lnd_*.rs; do \
+        TARGET_PATH=/lnd/cmd/lnd/lnd \
+            cargo build -p smite-scenarios --bin "$(basename $f .rs)" --release --features nyx; \
+    done
 
 # Runtime image
 FROM debian:bookworm-slim
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
--- a/workloads/eclair/Dockerfile
+++ b/workloads/eclair/Dockerfile
@@ -4,7 +4,6 @@ FROM eclipse-temurin:21 AS builder
 # Build arguments
 ARG ECLAIR_VERSION=v0.13.1
 ARG BITCOIN_VERSION=29.2
-ARG SCENARIO
 
 # Install build dependencies.
 RUN apt-get update && apt-get install -y --no-install-recommends \
@@ -61,7 +60,7 @@ RUN java -cp target/eclair-sancov-0.0.0.jar EclairEdgeCounter \
     > /tmp/eclair-edge-count.txt && \
     echo "Eclair edge count: $(cat /tmp/eclair-edge-count.txt)"
 
-# Copy smite workspace and build the eclair scenario binary.
+# Copy smite workspace files and build all scenario binaries
 WORKDIR /smite
 COPY Cargo.toml Cargo.lock ./
 COPY smite/ smite/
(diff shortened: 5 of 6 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,36 @@
...
+  push:
+    branches: ["master"]
+  pull_request:
+    branches: ["master"]
+    paths:
+      - '.github/workflows/docker.yml'
+      - 'workloads/**'
+      - 'smite-scenarios/src/bin/*'
+      - 'smite-scenarios/src/scenarios/*'
+      - 'smite-scenarios/src/targets/*'
```
ekzyis: GitHub provides 2,000 free CI/CD minutes, see https://github.com/pricing and https://github.com/settings/billing.

Since I don't see any minutes consumed on my account, I think all actions run in this repository are from @morehouse's budget:

Image

But even if it's per contributor, not per owner or repository, the jobs introduced in this PR consume a lot: 16 jobs with ~10 minutes each, so one PR would consume 160 minutes. 13 PRs per month consume the whole budget.

So I decided to only run these when related files have been changed. 

It's also possible to trigger the workflow (with all jobs) manually, see [here](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow).

morehouse: I think we'll be okay on CI minutes if we can cache the Docker stages.  I'd prefer to do that and then get rid of the paths config.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Build smite-${{ matrix.target }}-${{ matrix.scenario }}
+        run: |
+          TARGET="${{ matrix.target }}"
+          SCENARIO="${{ matrix.scenario }}"
+          docker build \
+            -t "smite-$TARGET-$SCENARIO" \
+            -f "workloads/$TARGET/Dockerfile" \
+            --build-arg SCENARIO=$SCENARIO \
+            .
```
morehouse: Here's where we can do the caching.

```suggestion
      - uses: actions/checkout@v4
       
      - uses: docker/setup-buildx-action@v4

      - name: Build smite-${{ matrix.target }}-${{ matrix.scenario }}
        uses: docker/build-push-action@v7
        with:
          context: .
          file: workloads/${{ matrix.target }}/Dockerfile
          tags: smite-${{ matrix.target }}-${{ matrix.scenario }}
          build-args: SCENARIO=${{ matrix.scenario }}
          cache-from: type=gha,scope=${{ matrix.target }}-${{ matrix.scenario }}
          cache-to: type=gha,mode=max,scope=${{ matrix.target }}-${{ matrix.scenario }}
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,36 @@
...
+    branches: ["master"]
+  pull_request:
+    branches: ["master"]
+    paths:
+      - '.github/workflows/docker.yml'
+      - 'workloads/**'
+      - 'smite-scenarios/src/bin/*'
+      - 'smite-scenarios/src/scenarios/*'
+      - 'smite-scenarios/src/targets/*'
+
```
morehouse, line 12: We could also add a concurrency group to cancel any previous runs on double-push.


```suggestion
concurrency:
  group: docker-${{ github.ref }}
  cancel-in-progress: true
```

morehouse (commented):
Concept ACK, thanks!

.github/workflows/docker.yml
ekzyis: Nice! Done in 8e5c056. Used `${{ github.workflow }}` to match the [official documentation](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-using-concurrency-and-the-default-behavior).

.github/workflows/docker.yml
ekzyis: Also nice! Done in 8e5c056. I didn't consider how smart docker already is when it comes to caching, and that we can use it here.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
...
+      - uses: docker/setup-buildx-action@v4
+
+      - name: Build smite-${{ matrix.target }}-${{ matrix.scenario }}
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: workloads/${{ matrix.target }}/Dockerfile
+          build-args: SCENARIO=${{ matrix.scenario }}
+          cache-from: type=gha,scope=${{ matrix.target }}-${{ matrix.scenario }}
+          cache-to: type=gha,mode=max,scope=${{ matrix.target }}-${{ matrix.scenario }}
```
morehouse: Let's try widening the cache scope to share all builds within each target.


```suggestion
          cache-from: type=gha,scope=${{ matrix.target }}
          cache-to: type=gha,mode=max,scope=${{ matrix.target }}
```

morehouse (changes requested):
I reran CI and it looks like 5 of the builds were properly cached, while everything else got rebuilt from scratch.  The lnd-noise build also failed due to some kind of cache issue.

It seems like we're running into some kind of cache limit.  Let's try reducing how much stuff we cache and see if it helps.

ekzyis:
Mhh, interesting. There's a [10GB limit per repository](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#default-limits):

> There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited. By default, the limit is 10 GB per repository, [...]

I'm currently building all images to see if we might be hitting that.

---

Unrelated to this: I noticed that actions/checkout@v4 is deprecated. Will switch to v5 or v6 in a follow-up PR.

.github/workflows/docker.yml
ekzyis: Done in e8ea9ca.

We could also look into using `mode=min`:

> In `min` cache mode (the default), only layers that are exported into the resulting image are cached, while in `max` cache mode, all layers are cached, even those of intermediate steps.
>
> While `min` cache is typically smaller (which speeds up import/export times, and reduces storage costs), `max` cache is more likely to get more cache hits. Depending on the complexity and location of your build, you should experiment with both parameters to find the results that work best for you.

https://docs.docker.com/build/cache/backends/#cache-mode

morehouse:
I reran CI again -- looks like caching didn't do anything with the scopes shared within each target.  I'll need to look into this more.

morehouse:
On a second rerun, 2 of the builds got cached: cln-noise and lnd-init.  I also checked the Actions cache and saw that 22 GB of data was stored during the rerun, which probably explains why we're having issues.  As soon as the next run tries to load from the cache, more than half of the cache gets evicted to stay at the 10 GB limit.

I think `mode=min` is worth trying.  If that's not enough, we could try executing scenario builds in sequence (so only one build per target runs at a time), and maybe that would reduce the concurrent cache usage so that every build after the first scenario benefits from the cache.

ekzyis:
51ec8e1 uses `mode=min` now

> I also checked the Actions cache and saw that 22 GB of data was stored during the rerun, which probably explains why we're having issues. As soon as the next run tries to load from the cache, more than half of the cache gets evicted to stay at the 10 GB limit.

Ahh, nice, I didn't see the [Actions cache](https://github.com/morehouse/smite/actions/caches) before. Then I can stop trying to build all the images locally. Ironically, I'm running out of disk space right now myself, haha.

> If that's not enough, we could try executing scenario builds in sequence (so only one build per target runs at a time), and maybe that would reduce the concurrent cache usage so that every build after the first scenario benefits from the cache.

If that's also not enough, we could consider to [self-host runners](https://docs.github.com/en/actions/concepts/runners/self-hosted-runners). I would be willing to host them, and I intend to be around for a while. I think for anything interesting that involves running snapshot fuzzing in CI, we would need to self-host runners anyway (i.e. making sure we get `accept_channel` like in #56).

There is [this warning](https://docs.github.com/en/actions/how-tos/manage-runners/self-hosted-runners/add-runners) though:

> We recommend that you only use self-hosted runners with private repositories. This is because forks of your public repository can potentially run dangerous code on your self-hosted runner machine by creating a pull request that executes the code in a workflow.

ekzyis:
[Looks like `mode=min` is not enough](https://github.com/morehouse/smite/actions/runs/25582044333).[^1] Only `lnd-noise`, `ldk-init`, `cln-noise` and `eclair-encrypted_bytes` hit the cache in the second attempt. In the third attempt, `lnd-encrypted_bytes`, `ldk-ir`, `cln-init` and `eclair-init` hit the cache.

Trying this now:

> we could try executing scenario builds in sequence (so only one build per target runs at a time), and maybe that would reduce the concurrent cache usage so that every build after the first scenario benefits from the cache.

Will push the changes, but I think I can't trigger CI again.

[^1]: Not sure it even did something. Where can I see how much was cached during a run? I only see the current size of the cache [here](https://github.com/morehouse/smite/actions/runs/25559661500).

morehouse:
I noticed with `mode=min` and fully parallel builds that we got one build cached per target and everything else rebuilt from scratch.  IIUC, `mode=min` means no intermediate steps get cached, only the final image.  So possibly only one final image per target was able to fit into the 10 GB cache -- the first rebuild for each target hits the cache and everything else got evicted.

With the serialized builds I think we should try `mode=max` again.  My hope is that we can fill the cache with the intermediate steps for each target, so that builds for subsequent scenarios don't have to start from scratch.

If that doesn't work, we could try restructuring the Dockerfiles into static base images and small per-scenario derived images.  Then we could cache just the 4 base images and build all the derived images in parallel.

NishantBansal2003:
> If that doesn't work, we could try restructuring the Dockerfiles into static base images and small per-scenario derived images. Then we could cache just the 4 base images and build all the derived images in parallel.

Can we do that in the normal workflow as well? I felt it was quite time-consuming when building the docker image for other scenarios of the same target, even though most of the base components are pretty much the same

workloads/lnd/Dockerfile
```diff
@@ -62,15 +61,17 @@ RUN cd cmd/lncli && go build
...
 COPY Cargo.toml Cargo.lock ./
 COPY smite/ smite/
 COPY smite-ir/ smite-ir/
 COPY smite-ir-mutator/ smite-ir-mutator/
 COPY smite-nyx-sys/ smite-nyx-sys/
 COPY smite-scenarios/ smite-scenarios/
-RUN TARGET_PATH=/lnd/cmd/lnd/lnd cargo build -p smite-scenarios --bin lnd_${SCENARIO} --release --features nyx
+RUN for s in encrypted_bytes noise init ir; do \
+        TARGET_PATH=/lnd/cmd/lnd/lnd cargo build -p smite-scenarios --bin lnd_$s --release --features nyx; \
+    done
```
morehouse: To reduce maintenance burden, we should use a glob instead of enumerating fully.  Also `set -eu` ensures we detect if one of the builds fails:


```suggestion
RUN set -eu; \
    for f in smite-scenarios/src/bin/lnd_*.rs; do \
        TARGET_PATH=/lnd/cmd/lnd/lnd cargo build -p smite-scenarios --bin "$(basename "$f" .rs)" --release --features nyx; \
    done
```

.github/workflows/docker-target.yml
```diff
@@ -0,0 +1,29 @@
...
+      target:
+        required: true
+        type: string
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        scenario: [encrypted_bytes, noise, init, ir]
```
morehouse: Now that all scenarios use the same per-target base image, I'm not sure it's worth testing every scenario individually anymore -- they only differ in which binary gets copied to the runner image.

I'd probably hardcode `SCENARIO=ir` since that is likely to be the most used scenario.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,22 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        target: [lnd, ldk, cln, eclair]
```
morehouse, line 18: Assuming we remove the per-scenario builds, we just have 4 builds in CI now.  I think it's worth trying to cache them, to see if we get any cache hits across CI runs.  But if we can't get it to work, we're probably still OK.

workloads/cln/Dockerfile
```diff
@@ -81,16 +80,18 @@ RUN CC=afl-clang-lto ./configure --prefix=/usr/local --disable-rust \
...
 WORKDIR /smite
 COPY Cargo.toml Cargo.lock ./
 COPY smite/ smite/
 COPY smite-ir/ smite-ir/
 COPY smite-ir-mutator/ smite-ir-mutator/
 COPY smite-nyx-sys/ smite-nyx-sys/
 COPY smite-scenarios/ smite-scenarios/
-RUN TARGET_MAP_SIZE=$(cat /tmp/total-map-size) \
-    cargo build -p smite-scenarios --bin cln_${SCENARIO} --release --features nyx
+RUN export TARGET_MAP_SIZE=$(cat /tmp/total-map-size) && \
```
morehouse: Nit: I think it's cleaner to inline the variable in the `cargo build` command instead of exporting it

morehouse (changes requested):
Nice fix for the dockerfiles -- I think this is the right direction to go

workloads/lnd/Dockerfile
ekzyis: done in 036bd37

.github/workflows/docker-target.yml
ekzyis: done in 036bd37

.github/workflows/docker.yml
ekzyis: caching looks good now in 036bd37, see [latest CI run](https://github.com/morehouse/smite/actions/runs/26096153566)

ekzyis:
Addressed all feedback in 036bd37.

Caching looks good now without `ARG SCENARIO` in the builder stage. [Latest CI run](https://github.com/morehouse/smite/actions/runs/26096153566) finished in 25s.

Updated to actions/checkout@v6 in 35f75c4 because Node.js 20 actions are deprecated, see [announcement](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/).

morehouse (approved):
LGTM, thank you!

PR #68, opened 2026-05-07
multi: drop direct secp256k1 and use re-exported from bitcoin

NishantBansal2003:
Add bitcoin as a dependency for constructing signed commitment transactions. This significantly reduces boilerplate compared to doing it manually.

Since bitcoin re-exports secp256k1, drop the direct dependency to avoid duplicate versions and use the re-exported one throughout.


morehouse (approved):
LGTM

morehouse:
Existing code uses `sha2::Sha256`.  If it makes sense, maybe we can use the `bitcoin` library for this instead now and remove the `sha2` dependency.

PR #73, opened 2026-05-11
multi: add BitcoinCli support and MineBlock IR operation

NishantBansal2003:
Added a `BitcoinCli` struct that contains the connection information required to invoke bitcoin-cli against the regtest bitcoind started by a target, and added it to the program context. This will be useful for future IR funding flow support, where the IR executor may need a bitcoin-cli reference to mine blocks, broadcast tx etc. The executor can access this `BitcoinCli` reference through `ProgramContext`.

Added a `BitcoinCli` variable type and `LoadBitcoinCliFromContext` so IR programs can access the bitcoin-cli connection handle from the program context and make Bitcoin RPC calls.

Added a `MineBlock` IR operation that takes a `BitcoinCli` input and invokes `bitcoin-cli -generate 1` to mine a block on regtest. This will be required once we add the funding flow scenario, where the funding tx is broadcast and then mined into a block to confirm it on-chain, since peers can exchange the `channel_ready` message only after the funding tx is confirmed.

I also verified the working of `MineBlock` using the diff below:
<details>
<summary>smite diff</summary>

```diff
diff --git a/smite-ir/src/generators/open_channel.rs b/smite-ir/src/generators/open_channel.rs
index f58f1bf..bf94823 100644
--- a/smite-ir/src/generators/open_channel.rs
+++ b/smite-ir/src/generators/open_channel.rs
@@ -25,6 +25,7 @@ impl Generator for OpenChannelGenerator {
         let delayed_payment_basepoint = builder.generate_fresh(VariableType::Point, rng);
         let htlc_basepoint = builder.generate_fresh(VariableType::Point, rng);
         let first_per_commitment_point = builder.generate_fresh(VariableType::Point, rng);
+        let bitcoin_cli = builder.generate_fresh(VariableType::BitcoinCli, rng);
 
         // Channel parameters.
         let chain_hash = builder.pick_variable(VariableType::ChainHash, rng);
@@ -77,5 +78,7 @@ impl Generator for OpenChannelGenerator {
 
         // Receive accept_channel.
         builder.append(Operation::RecvAcceptChannel, &[]);
+
+        builder.append(Operation::MineBlock,&[bitcoin_cli]);
     }
 }
 ```

</details>

<details>
<summary>smite and bitcoind logs</summary>

```bash
INFO  [smite_scenarios::targets::bitcoind] Preserving data directory: /data
INFO  [smite_scenarios::targets::bitcoind] Starting bitcoind...
[...]

PR #74, opened 2026-05-11
smite: add funding transaction construction

NishantBansal2003:
Depends-on: https://github.com/morehouse/smite/pull/51 and https://github.com/morehouse/smite/pull/73

Added support for creating a funding tx. This requires a few prerequisite RPC calls: `get_utxos` to fetch spendable inputs for the tx, `get_new_address_script_pubkey` to generate a change output address script pubkey, and `sign_and_broadcast_tx` to sign and broadcast the tx. Signing must be handled by `bitcoind` since it controls the private keys. As part of this, I also implemented a very basic coin selection algorithm with a time complexity of O(N log N).

Since the available spendable inputs may be less than the required funding amount, I added a new `InsufficientFunds` error type to signal to the caller that there are not enough funds available.

Note that for our use case, the `bitcoin-cli -generate`, `bitcoin-cli listunspent`, `bitcoin-cli getnewaddress`, `bitcoin-cli signrawtransactionwithwallet`, and `bitcoin-cli sendrawtransaction` RPC commands should not fail. Because of this, I added panics and asserts for the error conditions to ensure that we catch and fix any edge cases we might have missed if any of these commands ever return an error.

These functions will later be wired into IR operations and the executor to support a v1 funding flow scenario.

For testing, I used this script. Please note that I added a bunch of logging to get the oracle for the unit tests. I removed those (since they were a bit messy), but let me know if they are needed for validation:

<details>
<summary>I/O</summary>

```bash
# Case of output ordering
brcli getnewaddress
## bcrt1qv04x6qc8fm29zmh0g6tgd6nf7ll5xy8sdmtwrg
brcli sendtoaddress bcrt1qv04x6qc8fm29zmh0g6tgd6nf7ll5xy8sdmtwrg 0.10010653
## 9f7ab35ebf8cbc57c1099b2ec8c2badb0e6fe636a0bbb1706cabf8d82cba4f80
cargo run
## UTXO: amount=0.10010653 BTC | txid=9f7ab35ebf8cbc57c1099b2ec8c2badb0e6fe636a0bbb1706cabf8d82cba4f80 | vout=1 | script_pubkey=001463ea6d03074ed4516eef469686ea69f7ff4310f0 | privkey=9865bded76c28e6fe38ed09c642a372bfe7869179b975e89bdd06eb3387f6d43
## change_spk=001433ddddc311bb4ac2b23f0880f2f4876443c52c33 | Fee 0.00007290 BTC
## serialize_hex=02000000000101804fba2cd8f8ab6c70b1bba036e66f0edbbac2c82e9b09c157bc8cbf5eb37a9f0100000000ffffffff02df0500000000000016001433ddddc311bb4ac2b23f0880f2f4876443c52c338096980000000000220020c015c4a6be010e21657068fc2e6a9d02b27ebe4d490a25846f7237f104d1a3cd0247304402203c826aa74c3e971cc6bea32210c30b02b8fd288fa4c480d74d6e6f48316074b702207fdc14adedc53f526c10fa4789b1e54a04c79ad8ccacd76c1de7284b6896c6f7012103596e8168073fc881f2022fac98ed600decb43fd141a42d863f21a7c511b3def900000000
## txid=393a79d1b8ac352464e151eeb903260f68e3331d00969c3d1b2d304fece32c73

# Case of No Change
brcli getnewaddress
## bcrt1qmqjyey7wv3csvqxuz0vkvumuff5yw3dn4sevh2
brcli sendtoaddress bcrt1qmqjyey7wv3csvqxuz0vkvumuff5yw3dn4sevh2 0.10008942
## 1261e35df559b1cfb28225f356ef6897496cdeb859aabba54d3e7dc257834937
cargo run
## UTXO: amount=0.10008942 BTC | txid=1261e35df559b1cfb28225f356ef6897496cdeb859aabba54d3e7dc257834937 | vout=0 | script_pubkey=0014d8244c93ce64710600dc13d966737c4a684745b3 | privkey=775b7c755fbdab9572385dac3e2a0a4fb5afda48689df10f4f5cf43e0197f4ce
## change_spk=00142e532c12351a5c81e23c8a76d19345ca7b6de57a | Fee 0.00007290 BTC
## serialize_hex=0200000000010137498357c27d3e4da5bbaa59b8de6c499768ef56f32582b2cfb159f55de361120000000000ffffffff018096980000000000220020c015c4a6be010e21657068fc2e6a9d02b27ebe4d490a25846f7237f104d1a3cd0247304402204b7b97df9655c5bb514f2b2f9e20d1bc90dfbeaaf35b0340832aa54350247cac0220589457afc5e148d2be2a984f8d3a14c58bb7b558c46d6b1f08ab6cff1ca14ca6012102b1b0bf21ad828f6f7ff3b899cd7500a59ef566e4a4c8e893748c37c85b59b7c900000000
## txid=a38288eacddcd549cbad9f000968aad38b24c5f56732e9c9778cf73b824ab9a9
```

</details>

<details>
<summary>Test script</summary>

```rust
use std::path::PathBuf;
use std::process::Command;
use std::str::FromStr;

use bitcoin::absolute::LockTime;
use bitcoin::consensus::encode::serialize_hex;
use bitcoin::opcodes::all as opcodes;
use bitcoin::script::Builder;
use bitcoin::secp256k1::PublicKey;
use bitcoin::transaction::predict_weight;
use bitcoin::transaction::InputWeightPrediction;
use bitcoin::transaction::Version;
use bitcoin::{
    Address, Amount, Network, OutPoint, ScriptBuf, Sequence, Transaction, TxIn, TxOut, Txid,
    Witness,
};

use serde::Deserialize;

/// Connection info for invoking `bitcoin-cli` against the regtest `bitcoind`
/// started by a target.
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct BitcoinCli {
    /// RPC port exposed by the regtest `bitcoind` instance.
    pub rpc_port: u16,
    /// Path passed to `bitcoin-cli -datadir`.
    pub bitcoind_dir: PathBuf,
}

impl BitcoinCli {
    /// Creates a `bitcoin-cli` command preconfigured with the connection
    /// arguments for this regtest node.
    #[must_use]
    pub fn run(&self) -> Command {
        let mut cmd = Command::new("bitcoin-cli");
        cmd.arg("-regtest")
            .arg(format!("-datadir={}", self.bitcoind_dir.display()))
            .arg(format!("-rpcport={}", self.rpc_port))
            .arg("-rpcuser=rpcuser")
            .arg("-rpcpassword=rpcpass");
        cmd
    }

    /// Mines a single block.
    #[allow(clippy::missing_panics_doc)]
    pub fn mine_block(&self) {
        let mine_out = self
            .run()
            .arg("-generate")
            .arg("1")
            .output()
            .expect("bitcoin-cli -generate should not fail");
        assert!(
            mine_out.status.success(),
            "bitcoin-cli -generate failed: {}",
            String::from_utf8_lossy(&mine_out.stderr)
        );
    }

    /// Returns all sorted spendable UTXOs in the wallet.
    #[must_use]
    #[allow(clippy::missing_panics_doc)]
    pub fn get_utxos(&self) -> Vec<(Amount, OutPoint, ScriptBuf)> {
        #[derive(Deserialize)]
        struct UnspentOutput {
            txid: String,
            vout: u32,
            amount: f64,
            #[serde(rename = "scriptPubKey")]
            script_pubkey: String,
            spendable: bool,
        }

        let utxo_out = self
            .run()
            .arg("listunspent")
            .output()
            .expect("bitcoin-cli listunspent should not fail");
        assert!(
            utxo_out.status.success(),
            "bitcoin-cli listunspent failed: {}",
            String::from_utf8_lossy(&utxo_out.stderr)
        );

        let utxos: Vec<UnspentOutput> =
            serde_json::from_slice(&utxo_out.stdout).expect("listunspent should return valid JSON");

        let mut spendable: Vec<(Amount, OutPoint, ScriptBuf)> = utxos
            .into_iter()
            .filter(|u| u.spendable)
            .map(|u| {
                (
                    Amount::from_btc(u.amount).expect("listunspent amount should be valid BTC"),
                    OutPoint::new(
                        Txid::from_str(&u.txid).expect("listunspent should return valid txid"),
                        u.vout,
                    ),
                    ScriptBuf::from(
                        hex::decode(&u.script_pubkey)
                            .expect("listunspent should return valid hex scriptPubKey"),
                    ),
                )
            })
            .collect();
        spendable.sort();

        spendable
    }

    /// Returns the scriptPubKey for a newly generated wallet address.
    #[must_use]
    #[allow(clippy::missing_panics_doc)]
    pub fn get_new_address_script_pubkey(&self) -> ScriptBuf {
        let addr_out = self
            .run()
            .arg("getnewaddress")
            .output()
            .expect("bitcoin-cli getnewaddress should not fail");
        assert!(
            addr_out.status.success(),
            "bitcoin-cli getnewaddress failed: {}",
            String::from_utf8_lossy(&addr_out.stderr)
        );

        let addr_str = String::from_utf8_lossy(&addr_out.stdout);
        Address::from_str(addr_str.trim())
            .and_then(|a| a.require_network(Network::Regtest))
            .expect("getnewaddress should return a valid address")
            .script_pubkey()
    }

    /// Signs and broadcasts a transaction.
    #[allow(clippy::missing_panics_doc)]
    pub fn sign_and_broadcast_tx(&self, tx: &Transaction) {
        #[derive(Deserialize)]
        struct SignRawTransactionResponse {
            hex: String,
            complete: bool,
        }

        let tx_hex = serialize_hex(tx);

        let signed_out = self
            .run()
            .arg("signrawtransactionwithwallet")
            .arg(&tx_hex)
            .output()
            .expect("bitcoin-cli signrawtransactionwithwallet should not fail");
        assert!(
            signed_out.status.success(),
            "bitcoin-cli signrawtransactionwithwallet failed: {}",
            String::from_utf8_lossy(&signed_out.stderr)
        );

        let signed_tx: SignRawTransactionResponse = serde_json::from_slice(&signed_out.stdout)
            .expect("signrawtransactionwithwallet should return valid JSON");
        assert!(
            signed_tx.complete,
            "signrawtransactionwithwallet returned complete=false"
        );

        let broadcast_out = self
            .run()
            .arg("sendrawtransaction")
            .arg(&signed_tx.hex)
            .output()
            .expect("bitcoin-cli sendrawtransaction should not fail");
        assert!(
            broadcast_out.status.success(),
            "bitcoin-cli sendrawtransaction failed: {}",
            String::from_utf8_lossy(&broadcast_out.stderr)
        );

        let broadcast_txid = String::from_utf8_lossy(&broadcast_out.stdout)
            .trim()
            .to_string();
        println!("serialize_hex={}", signed_tx.hex);
        println!("txid={}", broadcast_txid);
    }
}

/// Error returned when available UTXOs cannot cover the funding amount plus estimated miner fee.
#[derive(Debug, thiserror::Error)]
#[error("insufficient funds to cover funding amount and fee")]
pub struct InsufficientFunds;

/// Builds a funding transaction with a 2-of-2 P2WSH output between the opener and acceptor.
///
/// # Errors
///
/// Returns [`InsufficientFunds`] if the provided inputs do not contain enough
/// value to cover `funding_satoshis` and the required transaction fees.
#[allow(clippy::missing_panics_doc)]
pub fn build_funding_transaction(
    opener_funding_pubkey: &PublicKey,
    acceptor_funding_pubkey: &PublicKey,
    funding_satoshis: u64,
    feerate_per_kw: u32,
    utxos: Vec<(Amount, OutPoint, ScriptBuf)>,
    change_spk: ScriptBuf,
) -> Result<(Transaction, u32), InsufficientFunds> {
    let funding_amt = Amount::from_sat(funding_satoshis);
    let funding_script =
        build_funding_redeemscript(opener_funding_pubkey, acceptor_funding_pubkey).to_p2wsh();

    let mut inputs = Vec::new();
    let mut inputs_weight = Vec::new();
    let mut outputs = vec![TxOut {
        value: funding_amt,
        script_pubkey: funding_script.clone(),
    }];
    let mut total = Amount::ZERO;

    for (amount, outpoint, script) in utxos {
        total += amount;

        inputs.push(TxIn {
            previous_output: outpoint,
            script_sig: ScriptBuf::new(),
            sequence: Sequence::MAX,
            witness: Witness::new(),
        });

        // By default, the address we generate coins to is always a P2WPKH address,
        // but to support the BOLT 3 test vectors, we also include P2PKH.
        inputs_weight.push(if script.is_p2pkh() {
            InputWeightPrediction::P2PKH_COMPRESSED_MAX
        } else {
            InputWeightPrediction::P2WPKH_MAX
        });

        // Check whether the selected inputs can cover the funding amount and fees.
        let expected_fee =
            funding_tx_fee_sat(feerate_per_kw, &inputs_weight, &[funding_script.len()]);
        if total >= funding_amt + expected_fee {
            break;
        }
    }

    // Verify that the selected inputs can cover the funding amount and fees.
    let expected_fee = funding_tx_fee_sat(feerate_per_kw, &inputs_weight, &[funding_script.len()]);
    if total < funding_amt + expected_fee {
        return Err(InsufficientFunds);
    }

    // Add remaining funds after accounting for fees as a change output,
    // unless the resulting change would be dust.
    let expected_fee_with_change = funding_tx_fee_sat(
        feerate_per_kw,
        &inputs_weight,
        &[funding_script.len(), change_spk.len()],
    );
    let dust = change_spk.minimal_non_dust();
    if let Some(change) = total
        .checked_sub(funding_amt + expected_fee_with_change)
        .filter(|c| *c >= dust)
    {
        outputs.push(TxOut {
            value: change,
            script_pubkey: change_spk,
        });
    }

    // BIP 69 output ordering: sort by (value, script_pubkey).
    outputs.sort_by(|a, b| {
        a.value
            .cmp(&b.value)
            .then_with(|| a.script_pubkey.as_bytes().cmp(b.script_pubkey.as_bytes()))
    });

    let funding_vout = outputs
        .iter()
        .position(|o| o.script_pubkey == funding_script)
        .and_then(|vout| u32::try_from(vout).ok())
        .expect("funding output always present");

    Ok((
        Transaction {
            version: Version::TWO,
            lock_time: LockTime::ZERO,
            input: inputs,
            output: outputs,
        },
        funding_vout,
    ))
}

/// Get the predicted fee cost of a funding tx in satoshis.
fn funding_tx_fee_sat(
    feerate_per_kw: u32,
    inputs_weight: &[InputWeightPrediction],
    scripts: &[usize],
) -> Amount {
    let weight = predict_weight(inputs_weight.iter().copied(), scripts.iter().copied());
    Amount::from_sat((u64::from(feerate_per_kw) * weight.to_wu()) / 1000)
}

/// Builds the funding output redeem script per BOLT 3.
pub fn build_funding_redeemscript(pubkey1: &PublicKey, pubkey2: &PublicKey) -> ScriptBuf {
    let key1_bytes = pubkey1.serialize();
    let key2_bytes = pubkey2.serialize();
    let (lesser, greater) = if key1_bytes < key2_bytes {
        (&key1_bytes, &key2_bytes)
    } else {
        (&key2_bytes, &key1_bytes)
    };
    Builder::new()
        .push_opcode(opcodes::OP_PUSHNUM_2)
        .push_slice(lesser)
        .push_slice(greater)
        .push_opcode(opcodes::OP_PUSHNUM_2)
        .push_opcode(opcodes::OP_CHECKMULTISIG)
        .into_script()
}

fn pubkey(hex_str: &str) -> PublicKey {
    let bytes = hex::decode(hex_str).expect("valid hex");
    PublicKey::from_slice(&bytes).expect("valid pubkey")
}

fn main() {
    let cli = BitcoinCli {
        rpc_port: 18443,
        bitcoind_dir: PathBuf::from("/home/smite-test/.bitcoin"),
    };

    let utxos = cli.get_utxos();
    let change_spk = cli.get_new_address_script_pubkey();

    let (tx, _vout) = build_funding_transaction(
        &pubkey("023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb"),
        &pubkey("030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c1"),
        10_000_000,
        15_000,
        utxos,
        change_spk,
    )
    .expect("inputs should cover funding amount and fees");

    cli.sign_and_broadcast_tx(&tx);
    for _i in 1..=6 {
        cli.mine_block();
    }
}
```

</details>

### DIFF
--- a/smite/src/bitcoin.rs
+++ b/smite/src/bitcoin.rs
@@ -1,8 +1,43 @@
 //! This module implements utilities for interacting with regtest
 //! `bitcoind` instances via `bitcoin-cli`.
 
+use std::cmp::Ordering;
 use std::path::PathBuf;
 use std::process::Command;
+use std::str::FromStr;
+
+use bitcoin::consensus::encode::serialize_hex;
+use bitcoin::{Address, Amount, Network, OutPoint, ScriptBuf, Transaction, Txid};
+use serde::Deserialize;
+
+/// A spendable UTXO used as a transaction input.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Utxo {
+    /// The value of the UTXO.
+    pub amount: Amount,
+    /// The transaction outpoint identifying the UTXO.
+    pub outpoint: OutPoint,
+    /// The script pubkey of the UTXO being spent.
+    pub script_pubkey: ScriptBuf,
+}
+
+impl Ord for Utxo {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // Sort in decreasing order of amount to support largest-first coin
+        // selection (as used in `bdk_wallet`) and ensure deterministic ordering.
+        other
+            .amount
+            .cmp(&self.amount)
+            .then_with(|| other.script_pubkey.cmp(&self.script_pubkey))
+            .then_with(|| other.outpoint.cmp(&self.outpoint))
+    }
+}
+
+impl PartialOrd for Utxo {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
 
 /// Connection info for invoking `bitcoin-cli` against the regtest `bitcoind`
 /// started by a target.
@@ -46,6 +81,149 @@ impl BitcoinCli {
             "bitcoin-cli -generate {} failed: {}",
             num_blocks,
             String::from_utf8_lossy(&mine_out.stderr)
+        );
+    }
+
+    /// Returns the wallet's spendable UTXOs, sorted deterministically.
+    ///
+    /// # Panics
+    ///
+    /// - If `bitcoin-cli listunspent` fails to execute or exits non-zero.
+    /// - If the output is not valid JSON, or any entry has an invalid amount,
+    ///   txid, or hex scriptPubKey.
+    #[must_use]
+    pub fn get_utxos(&self) -> Vec<Utxo> {
+        #[derive(Deserialize)]
+        struct UnspentOutput {
+            txid: String,
+            vout: u32,
+            amount: f64,
+            #[serde(rename = "scriptPubKey")]
+            script_pubkey: String,
+            spendable: bool,
+        }
+
+        let utxo_out = self
+            .run()
+            .arg("listunspent")
+            .output()
+            .expect("bitcoin-cli listunspent should not fail");
+        assert!(
+            utxo_out.status.success(),
+            "bitcoin-cli listunspent failed: {}",
+            String::from_utf8_lossy(&utxo_out.stderr)
+        );
+
+        let utxos: Vec<UnspentOutput> =
+            serde_json::from_slice(&utxo_out.stdout).expect("listunspent should return valid JSON");
+
+        let mut spendable: Vec<Utxo> = utxos
+            .into_iter()
+            .filter(|u| u.spendable)
+            .map(|u| Utxo {
+                amount: Amount::from_btc(u.amount).expect("listunspent amount should be valid BTC"),
+                outpoint: OutPoint::new(
+                    Txid::from_str(&u.txid).expect("listunspent should return valid txid"),
+                    u.vout,
+                ),
+                script_pubkey: ScriptBuf::from(
+                    hex::decode(&u.script_pubkey)
+                        .expect("listunspent should return valid hex scriptPubKey"),
+                ),
+            })
+            .collect();
+        // Sorted for determinism and to support largest-first coin selection
+        // during transaction construction.
+        spendable.sort();
+
+        spendable
+    }
+
+    /// Returns the scriptPubKey for a newly generated wallet address.
+    ///
+    /// # Panics
+    ///
+    /// - If `bitcoin-cli getnewaddress` fails to execute or exits non-zero.
+    /// - If the output is not valid UTF-8 or not a valid regtest address.
+    #[must_use]
+    pub fn get_new_address_script_pubkey(&self) -> ScriptBuf {
+        let addr_out = self
+            .run()
+            .arg("getnewaddress")
+            .output()
+            .expect("bitcoin-cli getnewaddress should not fail");
+        assert!(
+            addr_out.status.success(),
+            "bitcoin-cli getnewaddress failed: {}",
+            String::from_utf8_lossy(&addr_out.stderr)
+        );
+
+        let addr_str = String::from_utf8(addr_out.stdout).expect("bitcoin address is valid UTF-8");
+        Address::from_str(addr_str.trim())
+            .and_then(|a| a.require_network(Network::Regtest))
+            .expect("getnewaddress should return a valid address")
+            .script_pubkey()
+    }
+
+    /// Signs and broadcasts a transaction.
+    ///
+    /// # Panics
+    ///
+    /// - If `bitcoin-cli signrawtransactionwithwallet` or `sendrawtransaction`
+    ///   fails to execute or exits non-zero.
+    /// - If the sign output is not valid JSON.
+    /// - If signing returns `complete=false`.
+    /// - If `sendrawtransaction` does not return a valid UTF-8 txid.
+    /// - If the broadcasted txid does not match the given transaction's txid.
+    pub fn sign_and_broadcast_tx(&self, tx: &Transaction) {
+        #[derive(Deserialize)]
+        struct SignRawTransactionResponse {
+            hex: String,
+            complete: bool,
+        }
+
+        let tx_hex = serialize_hex(tx);
+
+        let signed_out = self
+            .run()
+            .arg("signrawtransactionwithwallet")
+            .arg(&tx_hex)
+            .output()
+            .expect("bitcoin-cli signrawtransactionwithwallet should not fail");
+        assert!(
+            signed_out.status.success(),
+            "bitcoin-cli signrawtransactionwithwallet failed: {}",
+            String::from_utf8_lossy(&signed_out.stderr)
+        );
+
+        let signed_tx: SignRawTransactionResponse = serde_json::from_slice(&signed_out.stdout)
+            .expect("signrawtransactionwithwallet should return valid JSON");
+        assert!(
+            signed_tx.complete,
+            "signrawtransactionwithwallet returned complete=false"
+        );
+
+        let broadcast_out = self
+            .run()
+            .arg("sendrawtransaction")
+            .arg(&signed_tx.hex)
+            .output()
+            .expect("bitcoin-cli sendrawtransaction should not fail");
+        assert!(
+            broadcast_out.status.success(),
+            "bitcoin-cli sendrawtransaction failed: {}",
+            String::from_utf8_lossy(&broadcast_out.stderr)
+        );
+
+        // Safe because bitcoind descriptor wallets currently default to native
+        // SegWit, so signing does not alter the txid computed from the unsigned
+        // Transaction.
+        let broadcast_txid = String::from_utf8(broadcast_out.stdout)
+            .expect("sendrawtransaction should return a valid UTF-8 txid");
+        assert_eq!(
+            broadcast_txid.trim(),
+            tx.compute_txid().to_string(),
+            "sendrawtransaction returned unexpected txid"
         );
     }
 }
--- a/smite/src/bolt/funding.rs
+++ b/smite/src/bolt/funding.rs
@@ -1,0 +1,626 @@
+//! BOLT 3 funding transaction construction.
+
+use bitcoin::absolute::LockTime;
+use bitcoin::opcodes::all as opcodes;
+use bitcoin::script::Builder;
+use bitcoin::secp256k1::PublicKey;
+use bitcoin::transaction::{InputWeightPrediction, Version, predict_weight};
+use bitcoin::{Amount, ScriptBuf, Sequence, Transaction, TxIn, TxOut, Witness};
+
+use crate::bitcoin::Utxo;
+
+/// Error returned when available UTXOs cannot cover the funding amount plus
+/// estimated miner fee.
+#[derive(Debug, thiserror::Error)]
+#[error(
+    "insufficient funds to cover funding amount and fee: required {required}, available {available}"
+)]
+pub struct InsufficientFunds {
+    /// Total amount required, including fees.
+    pub required: Amount,
+    /// Total spendable amount available from the selected UTXOs.
+    pub available: Amount,
+}
+
+/// A constructed funding transaction along with the index of the 2-of-2
+/// funding output within it.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FundingTransaction {
+    /// The Bitcoin transaction containing the funding output.
+    pub tx: Transaction,
+    /// Index of the funding output.
+    pub vout: u32,
+}
+
+/// Builds a funding transaction with a 2-of-2 P2WSH output between the opener
+/// and acceptor.
+///
+/// Coins are selected for spending in the order the `utxos` are provided.
+///
+/// # Errors
+///
+/// Returns [`InsufficientFunds`] if the provided inputs do not contain enough
+/// value to cover `funding_satoshis` and the required transaction fees.
+///
+/// # Panics
+///
+/// Panics if an input has an unsupported script pubkey type, since fee
+/// estimation currently only supports P2PKH and P2WPKH inputs.
+pub fn build_funding_transaction(
+    opener_funding_pubkey: &PublicKey,
+    acceptor_funding_pubkey: &PublicKey,
+    funding_satoshis: u64,
+    feerate_per_kw: u32,
+    utxos: Vec<Utxo>,
+    change_spk: ScriptBuf,
+) -> Result<FundingTransaction, InsufficientFunds> {
+    let funding_amt = Amount::from_sat(funding_satoshis);
+    let funding_spk =
+        build_funding_witness_script(opener_funding_pubkey, acceptor_funding_pubkey).to_p2wsh();
+
+    let mut inputs = Vec::new();
+    let mut input_weights = Vec::new();
+    let mut outputs = vec![TxOut {
+        value: funding_amt,
+        script_pubkey: funding_spk.clone(),
+    }];
+    let mut total = Amount::ZERO;
+    let mut expected_fee_no_change =
+        predict_tx_fee(feerate_per_kw, &input_weights, &[funding_spk.len()]);
+
+    for utxo in utxos {
+        total += utxo.amount;
+
+        inputs.push(TxIn {
+            previous_output: utxo.outpoint,
+            script_sig: ScriptBuf::new(),
+            sequence: Sequence::ENABLE_RBF_NO_LOCKTIME,
+            witness: Witness::new(),
+        });
+
+        // By default, the address bitcoind generates coins to is always a P2WPKH
+        // address, but to support the BOLT 3 test vectors, we also include P2PKH.
+        let input_weight = if utxo.script_pubkey.is_p2pkh() {
+            InputWeightPrediction::P2PKH_COMPRESSED_MAX
+        } else {
+            // Assert this so fee estimation breaks loudly if the default ever changes.
+            assert!(
+                utxo.script_pubkey.is_p2wpkh(),
+                "unsupported input script pubkey; fee estimation only handles P2PKH and P2WPKH"
+            );
+            InputWeightPrediction::P2WPKH_MAX
+        };
+        input_weights.push(input_weight);
+
+        // Check whether the selected inputs can cover the funding amount and fees.
+        expected_fee_no_change =
+            predict_tx_fee(feerate_per_kw, &input_weights, &[funding_spk.len()]);
+        if total >= funding_amt + expected_fee_no_change {
+            break;
+        }
+    }
+
+    // Verify that the selected inputs can cover the funding amount and fees.
+    if total < funding_amt + expected_fee_no_change {
+        return Err(InsufficientFunds {
+            required: funding_amt + expected_fee_no_change,
+            available: total,
+        });
+    }
+
+    // Add remaining funds after accounting for fees as a change output,
+    // unless the resulting change would be dust.
+    let expected_fee_with_change = predict_tx_fee(
+        feerate_per_kw,
+        &input_weights,
+        &[funding_spk.len(), change_spk.len()],
+    );
+    let dust = change_spk.minimal_non_dust();
+    if let Some(change) = total
+        .checked_sub(funding_amt + expected_fee_with_change)
+        .filter(|c| *c >= dust)
+    {
+        outputs.push(TxOut {
+            value: change,
+            script_pubkey: change_spk,
+        });
+    }
+
+    Ok(FundingTransaction {
+        tx: Transaction {
+            version: Version::TWO,
+            lock_time: LockTime::ZERO,
+            input: inputs,
+            output: outputs,
+        },
+        vout: 0,
+    })
+}
+
+/// Returns the predicted fee cost of a transaction.
+fn predict_tx_fee(
+    feerate_per_kw: u32,
+    input_weights: &[InputWeightPrediction],
+    output_scriptlens: &[usize],
+) -> Amount {
+    let weight = predict_weight(
+        input_weights.iter().copied(),
+        output_scriptlens.iter().copied(),
+    );
+    Amount::from_sat((u64::from(feerate_per_kw) * weight.to_wu()) / 1000)
+}
+
+/// Builds the funding output witness script per BOLT 3.
+pub fn build_funding_witness_script(pubkey1: &PublicKey, pubkey2: &PublicKey) -> ScriptBuf {
+    let key1_bytes = pubkey1.serialize();
+    let key2_bytes = pubkey2.serialize();
+    let (lesser, greater) = if key1_bytes < key2_bytes {
+        (&key1_bytes, &key2_bytes)
+    } else {
+        (&key2_bytes, &key1_bytes)
+    };
+    Builder::new()
+        .push_opcode(opcodes::OP_PUSHNUM_2)
+        .push_slice(lesser)
+        .push_slice(greater)
+        .push_opcode(opcodes::OP_PUSHNUM_2)
+        .push_opcode(opcodes::OP_CHECKMULTISIG)
+        .into_script()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use bitcoin::OutPoint;
+    use bitcoin::consensus::encode::serialize_hex;
+    use bitcoin::ecdsa::Signature;
+    use bitcoin::secp256k1::{Secp256k1, SecretKey};
+    use bitcoin::sighash::{EcdsaSighashType, SighashCache};
+
+    fn pubkey(hex_str: &str) -> PublicKey {
+        let bytes = hex::decode(hex_str).expect("valid hex");
+        PublicKey::from_slice(&bytes).expect("valid pubkey")
+    }
+
+    fn secret(hex_str: &str) -> SecretKey {
+        let bytes = hex::decode(hex_str).expect("valid hex");
+        SecretKey::from_slice(&bytes).expect("valid secret key")
+    }
+
+    // Signs a P2PKH input (input 0) of the funding transaction, producing the
+    // script_sig with signature and public key.
+    fn sign_p2pkh_input(funding: &mut FundingTransaction, utxo: &Utxo, input_privkey: &SecretKey) {
+        let input_pubkey = PublicKey::from_secret_key(&Secp256k1::signing_only(), input_privkey);
+        let sighash = SighashCache::new(&funding.tx)
+            .legacy_signature_hash(0, &utxo.script_pubkey, EcdsaSighashType::All.to_u32())
+            .expect("valid sighash");
+        let sig = Signature::sighash_all(
+            Secp256k1::signing_only().sign_ecdsa(&sighash.into(), input_privkey),
+        );
+
+        funding.tx.input[0].script_sig = Builder::new()
+            .push_slice(sig.serialize())
+            .push_slice(input_pubkey.serialize())
+            .into_script();
+    }
+
+    // Signs P2WPKH inputs of the funding transaction, producing the witness
+    // with signature and public key.
+    fn sign_p2wpkh_input(
+        funding: &mut FundingTransaction,
+        utxos: &[Utxo],
+        input_privkeys: &[SecretKey],
+    ) {
+        assert_eq!(utxos.len(), input_privkeys.len());
+
+        let secp = Secp256k1::signing_only();
+
+        for (i, (utxo, input_privkey)) in utxos.iter().zip(input_privkeys).enumerate() {
+            let amount = utxo.amount;
+            let input_spk = &utxo.script_pubkey;
+
+            let input_pubkey = PublicKey::from_secret_key(&secp, input_privkey);
+            let sighash = SighashCache::new(&funding.tx)
+                .p2wpkh_signature_hash(i, input_spk, amount, EcdsaSighashType::All)
+                .expect("valid sighash");
+            let sig = Signature::sighash_all(secp.sign_ecdsa_low_r(&sighash.into(), input_privkey));
+
+            funding.tx.input[i].witness =
+                Witness::from_slice(&[sig.serialize().as_ref(), &input_pubkey.serialize()[..]]);
+        }
+    }
+
+    // BOLT 3 Appendix B: Funding Transaction Test Vectors
+    //    https://github.com/lightning/bolts/blob/master/03-transactions.md#appendix-b-funding-transaction-test-vectors
+
+    #[test]
+    fn valid_funding_tx_with_p2pkh_input_and_p2wpkh_change() {
+        let utxos = vec![Utxo {
+            amount: Amount::from_sat(5_000_000_000),
+            outpoint: OutPoint {
+                txid: "fd2105607605d2302994ffea703b09f66b6351816ee737a93e42a841ea20bbad"
+                    .parse()
+                    .expect("valid txid"),
+                vout: 0,
+            },
+            // P2PKH scriptpubkey of the block 1 coinbase output being spent.
+            script_pubkey: ScriptBuf::from(
+                hex::decode("76a9143ca33c2e4446f4a305f23c80df8ad1afdcf652f988ac")
+                    .expect("valid P2PKH scriptpubkey hex"),
+            ),
+        }];
+
+        // P2WPKH scriptpubkey for the change destination.
+        let change_spk = ScriptBuf::from(
+            hex::decode("00143ca33c2e4446f4a305f23c80df8ad1afdcf652f9")
+                .expect("valid P2WPKH scriptpubkey hex"),
+        );
+        let mut funding = build_funding_transaction(
+            &pubkey("023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb"),
+            &pubkey("030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c1"),
+            10_000_000,
+            15_000,
+            utxos.clone(),
+            change_spk,
+        )
+        .expect("inputs should cover funding amount and fees");
+
+        assert_eq!(funding.vout, 0);
+        assert_eq!(funding.tx.output.len(), 2);
+        assert_eq!(
+            funding.tx.output[funding.vout as usize].value,
+            Amount::from_sat(10_000_000)
+        );
+
+        // The BOLT 3 test vectors assume that the sequence used in the inputs
+        // disables absolute locktime and replace-by-fee. However, our funding
+        // transaction construction enables replace-by-fee, so we override the
+        // sequence here to validate against the BOLT 3 test vectors.
+        funding.tx.input[0].sequence = Sequence::MAX;
+
+        // Sign the P2PKH input with the block 1 coinbase privkey to verify the
+        // BOLT 3 txid.
+        sign_p2pkh_input(
+            &mut funding,
+            &utxos[0],
+            &secret("6bd078650fcee8444e4e09825227b801a1ca928debb750eb36e6d56124bb20e8"),
+        );
+
+        assert_eq!(
+            serialize_hex(&funding.tx),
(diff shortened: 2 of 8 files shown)


smite/src/bitcoin.rs
```diff
@@ -43,4 +49,119 @@ impl BitcoinCli {
             String::from_utf8_lossy(&mine_out.stderr)
         );
     }
+
+    /// Returns all sorted spendable UTXOs in the wallet.
+    #[must_use]
+    #[allow(clippy::missing_panics_doc)]
+    pub fn get_utxos(&self) -> Vec<(Amount, OutPoint, ScriptBuf)> {
```
morehouse: I think a `Utxo` struct would make the API cleaner.

smite/src/bitcoin.rs
```diff
@@ -0,0 +1,167 @@
...
+            .arg("getnewaddress")
+            .output()
+            .expect("bitcoin-cli getnewaddress should not fail");
+        assert!(
+            addr_out.status.success(),
+            "bitcoin-cli getnewaddress failed: {}",
+            String::from_utf8_lossy(&addr_out.stderr)
+        );
+
+        let addr_str = String::from_utf8_lossy(&addr_out.stdout);
```
morehouse:
```suggestion
        let addr_str = String::from_utf8(&addr_out.stdout).expect("bitcoin address is valid UTF-8");
```

smite/src/bitcoin.rs
```diff
@@ -0,0 +1,167 @@
...
+                        u.vout,
+                    ),
+                    ScriptBuf::from(
+                        hex::decode(&u.script_pubkey)
+                            .expect("listunspent should return valid hex scriptPubKey"),
+                    ),
+                )
+            })
+            .collect();
+        spendable.sort();
```
morehouse, line 137:
```suggestion
        spendable.sort();  // Sorted for determinism
```

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+use bitcoin::opcodes::all as opcodes;
+use bitcoin::script::Builder;
+use bitcoin::secp256k1::PublicKey;
+use bitcoin::transaction::{InputWeightPrediction, Version, predict_weight};
+use bitcoin::{Amount, OutPoint, ScriptBuf, Sequence, Transaction, TxIn, TxOut, Witness};
+
+/// Error returned when available UTXOs cannot cover the funding amount plus estimated miner fee.
+#[derive(Debug, thiserror::Error)]
+#[error("insufficient funds to cover funding amount and fee")]
+pub struct InsufficientFunds;
```
morehouse: Including the needed and actual funding amounts would make the error more helpful.

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+    feerate_per_kw: u32,
+    inputs_weight: &[InputWeightPrediction],
+    scripts: &[usize],
+) -> Amount {
+    let weight = predict_weight(inputs_weight.iter().copied(), scripts.iter().copied());
+    Amount::from_sat((u64::from(feerate_per_kw) * weight.to_wu()) / 1000)
+}
+
+/// Builds the funding output redeem script per BOLT 3.
+pub fn build_funding_redeemscript(pubkey1: &PublicKey, pubkey2: &PublicKey) -> ScriptBuf {
```
morehouse: I think this is technically the *witness* script since it's for P2WSH, not P2SH.  I think we also got this naming mixed up in `commitment.rs`.


```suggestion
pub fn build_funding_witness_script(pubkey1: &PublicKey, pubkey2: &PublicKey) -> ScriptBuf {
```

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+    opener_funding_pubkey: &PublicKey,
+    acceptor_funding_pubkey: &PublicKey,
+    funding_satoshis: u64,
+    feerate_per_kw: u32,
+    utxos: Vec<(Amount, OutPoint, ScriptBuf)>,
+    change_spk: ScriptBuf,
+) -> Result<(Transaction, u32), InsufficientFunds> {
+    let funding_amt = Amount::from_sat(funding_satoshis);
+    let funding_script =
+        build_funding_redeemscript(opener_funding_pubkey, acceptor_funding_pubkey).to_p2wsh();
```
morehouse: I think `to_p2wsh` actually returns the script pubkey.


```suggestion
    let funding_spk =
        build_funding_redeemscript(opener_funding_pubkey, acceptor_funding_pubkey).to_p2wsh();
```

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+        total += amount;
+
+        inputs.push(TxIn {
+            previous_output: outpoint,
+            script_sig: ScriptBuf::new(),
+            sequence: Sequence::MAX,
+            witness: Witness::new(),
+        });
+
+        // By default, the address we generate coins to is always a P2WPKH address,
```
morehouse:
```suggestion
        // By default, the address bitcoind generates coins to is always a P2WPKH address,
```

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+
+/// Get the predicted fee cost of a funding tx in satoshis.
+fn funding_tx_fee_sat(
+    feerate_per_kw: u32,
+    inputs_weight: &[InputWeightPrediction],
+    scripts: &[usize],
+) -> Amount {
+    let weight = predict_weight(inputs_weight.iter().copied(), scripts.iter().copied());
+    Amount::from_sat((u64::from(feerate_per_kw) * weight.to_wu()) / 1000)
+}
```
morehouse: This function returns an `Amount`, not sats.  Also I don't think it's specific to the funding tx.


```suggestion
/// Get the predicted fee cost of a transaction.
fn predict_tx_fee(
    feerate_per_kw: u32,
    input_weights: &[InputWeightPrediction],
    output_scriptlens: &[usize],
) -> Amount {
    let weight = predict_weight(input_weights.iter().copied(), output_scriptlens.iter().copied());
    Amount::from_sat((u64::from(feerate_per_kw) * weight.to_wu()) / 1000)
}
```

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+    feerate_per_kw: u32,
+    utxos: Vec<(Amount, OutPoint, ScriptBuf)>,
+    change_spk: ScriptBuf,
+) -> Result<(Transaction, u32), InsufficientFunds> {
+    let funding_amt = Amount::from_sat(funding_satoshis);
+    let funding_script =
+        build_funding_redeemscript(opener_funding_pubkey, acceptor_funding_pubkey).to_p2wsh();
+
+    let mut inputs = Vec::new();
+    let mut inputs_weight = Vec::new();
```
morehouse:
```suggestion
    let mut input_weights = Vec::new();
```

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+            sequence: Sequence::MAX,
+            witness: Witness::new(),
+        });
+
+        // By default, the address we generate coins to is always a P2WPKH address,
+        // but to support the BOLT 3 test vectors, we also include P2PKH.
+        inputs_weight.push(if script.is_p2pkh() {
+            InputWeightPrediction::P2PKH_COMPRESSED_MAX
+        } else {
+            InputWeightPrediction::P2WPKH_MAX
```
morehouse, line 91: If the bitcoind default ever changes to P2TR, we'll start estimating fees incorrectly.  Perhaps we should add an assertion to detect that.

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+            script_pubkey: change_spk,
+        });
+    }
+
+    // BIP 69 output ordering: sort by (value, script_pubkey).
+    outputs.sort_by(|a, b| {
+        a.value
+            .cmp(&b.value)
+            .then_with(|| a.script_pubkey.as_bytes().cmp(b.script_pubkey.as_bytes()))
+    });
```
morehouse: I'm fairly sure BIP 69 ordering is not required for funding transactions.

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+            hex::decode("00142e532c12351a5c81e23c8a76d19345ca7b6de57a")
+                .expect("valid P2WPKH scriptpubkey hex"),
+        );
+        let (mut tx, funding_vout) = build_funding_transaction(
+            &pubkey("023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb"),
+            &pubkey("030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c1"),
+            10_000_000,
+            15_000,
+            vec![(
+                Amount::from_btc(0.100_089_42).expect("valid amount"),
```
morehouse: Nit: the mental conversion from BTC->sat is annoying (here and in other tests), and the separators add to the confusion


```suggestion
                Amount::from_sat(10_008_942).expect("valid amount"),
```

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+                Amount::from_btc(0.100_106_53).expect("valid amount"),
+                EcdsaSighashType::All,
+            )
+            .expect("valid sighash");
+        let sig = Signature::sighash_all(
+            Secp256k1::signing_only().sign_ecdsa(&sighash.into(), &input_privkey),
+        );
+
+        tx.input[0].witness =
+            Witness::from_slice(&[sig.serialize().as_ref(), &input_pubkey.serialize()[..]]);
```
morehouse: The logic for signing is fairly ugly.  We could hide the complexity in a couple helper functions `sign_p2wpkh_input` and `sign_p2pkh_input`.

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+
+        assert_eq!(
+            serialize_hex(&tx),
+            "0200000000010137498357c27d3e4da5bbaa59b8de6c499768ef56f32582b2cfb159f55de361120000000000ffffffff018096980000000000220020c015c4a6be010e21657068fc2e6a9d02b27ebe4d490a25846f7237f104d1a3cd0247304402204b7b97df9655c5bb514f2b2f9e20d1bc90dfbeaaf35b0340832aa54350247cac0220589457afc5e148d2be2a984f8d3a14c58bb7b558c46d6b1f08ab6cff1ca14ca6012102b1b0bf21ad828f6f7ff3b899cd7500a59ef566e4a4c8e893748c37c85b59b7c900000000"
+        );
+        assert_eq!(
+            tx.compute_txid().to_string(),
+            "a38288eacddcd549cbad9f000968aad38b24c5f56732e9c9778cf73b824ab9a9"
+        );
+    }
```
morehouse, line 357: IIUC, you generated these expected values using your own code (i.e. basically the same as the code implemented here).  This doesn't seem like a great way to test.

It would be better to get the expected values from another implementation.  Or, since we're really just testing that the change output is omitted, we could drop this additional verification.

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+
+        assert_eq!(
+            serialize_hex(&tx),
+            "02000000000101804fba2cd8f8ab6c70b1bba036e66f0edbbac2c82e9b09c157bc8cbf5eb37a9f0100000000ffffffff02df0500000000000016001433ddddc311bb4ac2b23f0880f2f4876443c52c338096980000000000220020c015c4a6be010e21657068fc2e6a9d02b27ebe4d490a25846f7237f104d1a3cd0247304402203c826aa74c3e971cc6bea32210c30b02b8fd288fa4c480d74d6e6f48316074b702207fdc14adedc53f526c10fa4789b1e54a04c79ad8ccacd76c1de7284b6896c6f7012103596e8168073fc881f2022fac98ed600decb43fd141a42d863f21a7c511b3def900000000"
+        );
+        assert_eq!(
+            tx.compute_txid().to_string(),
+            "393a79d1b8ac352464e151eeb903260f68e3331d00969c3d1b2d304fece32c73"
+        );
+    }
```
morehouse, line 509: Again here, I don't think it's great to test against our own implementation.

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,415 @@
...
+            build_funding_redeemscript(&remote_funding_pubkey, &local_funding_pubkey);
+
+        // Argument order must not matter as keys are sorted lexicographically.
+        assert_eq!(funding_redeemscript_1, funding_redeemscript_2);
+
+        assert_eq!(
+            hex::encode(funding_redeemscript_1.as_bytes()),
+            "5221023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb21030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c152ae"
+        );
+    }
```
morehouse, line 625: It would also be good to test the case where multiple input UTXOs are used to pay for the funding tx.

smite/src/bolt/funding.rs
NishantBansal2003: Not explicitly mentioned in the BOLTs (but I think many wallets are moving toward BIP 69 for both inputs and outputs), In any case, this is done because the test vectors assume BIP 69 ordering of outputs: https://github.com/lightning/bolts/blob/master/03-transactions.md#appendix-b-funding-transaction-test-vectors

smite/src/bolt/funding.rs
NishantBansal2003: Yeah, that makes sense. Even the BOLT test vectors use the witness script for this, not the redeem script

smite/src/bolt/funding.rs
NishantBansal2003: Amount is actually a wrapper around u64 sats

morehouse: That's arguably an implementation detail for `Amount` that users shouldn't have to know, and which could theoretically be changed by rust-bitcoin in the future.  Amount provides `to_sat` and `to_btc` to get the value in u64 sats or in f64 BTC.

IMO if we want to return sats specifically, it should be a u64.  If we want to return an `Amount`, we should just call it `fee`, not `fee_sat`.

smite/src/bolt/funding.rs
morehouse: FWIW, if we want to reuse some of this code for dual funding in the future, BIP 69 sorting will have to be removed.  The interactive-tx protocol sorts by `serial_id` instead.

NishantBansal2003: > FWIW, if we want to reuse some of this code for dual funding in the future

In that case, we also need to change the `TxIn` sequence since interactive-tx requires the `sequence to be less than or equal to 4294967293`

smite/src/bolt/funding.rs
NishantBansal2003: > IIUC, you generated these expected values using your own code (i.e. basically the same as the code implemented here). This doesn't seem like a great way to test.

Yeah, I get your point. The main thing I was testing was that bitcoind accepts (signs and broadcasts) the transaction we are building and interacting with. But yes, the assertions for the txid and hex are coming directly from my program, so that is not ideal

> It would be better to get the expected values from another implementation. Or, since we're really just testing that the change output is omitted, we could drop this additional verification.

Since we’ve now added assertions for the txid and hex, let’s include them as well. It will also make things symmetric with the BOLT test vectors.
I checked LDK, and it uses BDK wallet for this stuff. Although LDK is not directly asserting the txid and hex, I created a simple program using `bdk_wallet` to get the tx hex and txid.

Please note that I changed some of the semantics. I am now using `LargestFirstCoinSelection`, which was the simplest and most deterministic strategy used in BDK wallet. For the tests, I am also using `sign_ecdsa_low_r` since BDK uses that by default. Also, I used the `ENABLE_RBF_NO_LOCKTIME` sequence, as BDK also uses that by default.

<details>
<summary>Deterministic program:</summary>

```rust
use std::str::FromStr;

use bdk_chain::BlockId;
use bdk_wallet::coin_selection::LargestFirstCoinSelection;
use bdk_wallet::keys::DescriptorSecretKey;
use bdk_wallet::test_utils::{
    get_test_wpkh_and_change_desc, insert_checkpoint, receive_output_in_latest_block,
};
use bdk_wallet::tx_builder::TxOrdering;
use bdk_wallet::{KeychainKind, Wallet};
use bitcoin::consensus::encode::serialize_hex;
use bitcoin::hashes::Hash;
use bitcoin::hex::DisplayHex;
use bitcoin::opcodes::all as opcodes;
use bitcoin::script::Builder;
use bitcoin::secp256k1::{PublicKey, Secp256k1};
use bitcoin::{Amount, BlockHash, FeeRate, Network, ScriptBuf, absolute};

fn build_funding_witness_script(pk1: &PublicKey, pk2: &PublicKey) -> ScriptBuf {
    let a = pk1.serialize();
    let b = pk2.serialize();
    let (lesser, greater) = if a < b { (&a, &b) } else { (&b, &a) };
    Builder::new()
        .push_opcode(opcodes::OP_PUSHNUM_2)
        .push_slice(lesser)
        .push_slice(greater)
        .push_opcode(opcodes::OP_PUSHNUM_2)
        .push_opcode(opcodes::OP_CHECKMULTISIG)
        .into_script()
}

fn main() {
    let (desc, change_desc) = get_test_wpkh_and_change_desc();
    let mut wallet = Wallet::create(desc, change_desc)
        .network(Network::Regtest)
        .create_wallet_no_persist()
        .unwrap();
    insert_checkpoint(
        &mut wallet,
        BlockId {
            height: 5,
            hash: BlockHash::all_zeros(),
        },
    );

    let amounts = [
        Amount::from_sat(10_000_000),
        Amount::from_sat(10_010_000),
        Amount::from_sat(10_011_000),
    ];
    for amount in amounts {
        receive_output_in_latest_block(&mut wallet, amount);
    }
    for (i, utxo) in wallet.list_unspent().enumerate() {
        println!(
            "[{i}] UTXO outpoint = {}  amount = {}  spk = {:x}",
            utxo.outpoint,
            utxo.txout.value.to_sat(),
            utxo.txout.script_pubkey,
        );
    }

    let funding_spk = build_funding_witness_script(
        &PublicKey::from_str("023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb")
            .unwrap(),
        &PublicKey::from_str("030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c1")
            .unwrap(),
    )
    .to_p2wsh();
    let funding_amt = Amount::from_sat(15_000_000);

    let mut tx_builder = wallet.build_tx().coin_selection(LargestFirstCoinSelection);
    tx_builder
        .ordering(crate::TxOrdering::Untouched)
        .nlocktime(absolute::LockTime::ZERO)
        .add_recipient(funding_spk, funding_amt)
        .fee_rate(FeeRate::from_sat_per_kwu(15_000));
    let mut psbt = tx_builder.finish().unwrap();

    let mut change_idx = 0;
    for output in psbt.unsigned_tx.output.iter() {
        if wallet.is_mine(output.script_pubkey.clone()) {
            println!("[{change_idx}] change_spk = {:x}", output.script_pubkey,);
            change_idx += 1;
        }
    }

    let input_derivations: Vec<_> = psbt
        .inputs
        .iter()
        .map(|i| i.bip32_derivation.clone())
        .collect();
    wallet.sign(&mut psbt, Default::default()).unwrap();

    let secp = Secp256k1::new();
    let xpriv = wallet
        .get_signers(KeychainKind::External)
        .as_key_map(&secp)
        .into_values()
        .find_map(|sk| match sk {
            DescriptorSecretKey::XPrv(xkey) => Some(xkey.xkey),
            _ => None,
        })
        .unwrap();

    let mut key_idx = 0;
    for (derivations, _) in input_derivations.iter().zip(&psbt.inputs) {
        for (_pk, (_fp, path)) in derivations {
            let derived = xpriv.derive_priv(&secp, path).unwrap();
            println!(
                "[{key_idx}] input signing key (hex) = {}",
                derived.private_key.secret_bytes().to_lower_hex_string(),
            );
            key_idx += 1;
        }
    }

    let tx = psbt.extract_tx().unwrap();
    println!("Final txid = {}", tx.compute_txid());
    println!("Final tx (hex) = {}", serialize_hex(&tx));
}

```

</details>

smite/src/bitcoin.rs
```diff
@@ -1,8 +1,47 @@
...
+}
+
+impl Ord for Utxo {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // Sort in decreasing order of amount, and then in increasing order of
+        // `script_pubkey` and `outpoint`.
+        //
+        // The rationale is that the coin selection algorithm uses largest-first
+        // coin selection (as used in `bdk_wallet`). We then sort in increasing
+        // order of `script_pubkey` to help optimize fees.
```
morehouse: Does sorting by script pubkey actually optimize fees?  I thought it was just for determinism.

smite/src/bitcoin.rs
```diff
@@ -48,4 +87,147 @@ impl BitcoinCli {
...
+        // Safe because bitcoind descriptor wallets currently default to native
+        // SegWit, so signing does not alter the txid computed from the unsigned
+        // Transaction.
+        let broadcast_txid = String::from_utf8(broadcast_out.stdout)
+            .expect("sendrawtransaction should return a valid UTF-8 txid");
+        assert_eq!(
+            broadcast_txid.trim(),
+            tx.compute_txid().to_string(),
+            "sendrawtransaction returned unexpected txid"
+        );
```
morehouse, line 227: If we were to ever provide a transaction to this function that did not have only segwit inputs, this assertion would fail, right?

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,540 @@
+//! BOLT 3 funding transaction construction.
```
morehouse, line 1: I think neither `funding.rs` nor `commitment.rs` really belong in the `bolt` module.  I think we should move them to a new `tx` module in a follow-up PR.

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,540 @@
...
+    /// Index of the funding output.
+    pub vout: u32,
+}
+
+/// Builds a funding transaction with a 2-of-2 P2WSH output between the opener
+/// and acceptor.
+///
+/// The coin selection algorithm uses largest-first coin selection, so `utxos`
+/// are expected to already be sorted in decreasing order of amount, and then
+/// in increasing order of script pubkey and outpoint.
```
morehouse: Nit: the coin selection algorithm in this function is "try in the order of `utxos`", which means it could be largest-first, smallest-first, or something else depending on how `utxos` was presorted.


```suggestion
/// Coins are selected for spending in the order of `utxos`.
```

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,540 @@
...
+) -> Amount {
+    let weight = predict_weight(
+        input_weights.iter().copied(),
+        output_scriptlens.iter().copied(),
+    );
+    Amount::from_sat((u64::from(feerate_per_kw) * weight.to_wu()) / 1000)
+}
+
+/// Builds the funding output witness script per BOLT 3.
+pub(super) fn build_funding_witness_script(pubkey1: &PublicKey, pubkey2: &PublicKey) -> ScriptBuf {
```
morehouse: Nit:  we haven't used `pub(super)` before  -- elsewhere we mark internal functions `pub` and then never re-export them.  I slightly prefer to use that pattern here for consistency, unless we want to switch to the stronger `pub(super)` way everywhere in a follow-up PR

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,540 @@
...
+            .into_script();
+    }
+
+    // Signs P2WPKH inputs of the funding transaction, producing the witness
+    // with signature and public key.
+    fn sign_p2wpkh_input(
+        funding: &mut FundingTransaction,
+        utxos: &[Utxo],
+        input_privkeys: &[SecretKey],
+    ) {
```
morehouse, line 213:
```suggestion
    ) {
        assert_eq!(utxos.len(), input_privkeys.len());
```

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,540 @@
...
+        assert_eq!(
+            funding.tx.output[funding.vout as usize].value,
+            Amount::from_sat(10_000_000)
+        );
+
+        // The BOLT 3 test vectors assume that the sequence used in the inputs
+        // disables absolute locktime and replace-by-fee. However, our funding
+        // transaction construction enables both absolute locktime and
+        // replace-by-fee, so we override the sequence here to validate against
+        // the BOLT 3 test vectors.
```
morehouse:
```suggestion
        // The BOLT 3 test vectors assume that the sequence used in the inputs
        // disables absolute locktime and replace-by-fee. However, our funding
        // transaction construction enables replace-by-fee, so we override the
        // sequence here to validate against the BOLT 3 test vectors.
```

smite/src/bolt/funding.rs
morehouse: Let's at least document for each test that the expected values were generated by `bdk_wallet`.

smite/src/bolt/funding.rs
```diff
@@ -0,0 +1,540 @@
...
+            build_funding_witness_script(&remote_funding_pubkey, &local_funding_pubkey);
+
+        // Argument order must not matter as keys are sorted lexicographically.
+        assert_eq!(funding_witness_script_1, funding_witness_script_2);
+
+        assert_eq!(
+            hex::encode(funding_witness_script_1.as_bytes()),
+            "5221023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb21030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c152ae"
+        );
+    }
```
morehouse, line 625: A couple more edge cases that might be worth testing:

- empty UTXO set
- change == dust exactly

morehouse (approved):

smite/src/bitcoin.rs
NishantBansal2003: Hmm, I think I completely confused this with the `input_weights` and `output_script_lens` in `predict_tx_fee`. I’ll fix the comment. Also, sorting the `script_pubkeys` in any order does not make much difference, since the UTXOs we have will always contain coins sent to the default P2WPKH (or a single script type) address

smite/src/bitcoin.rs
NishantBansal2003: Yes, so if the tx has non-segwit or non-taproot inputs i.e legacy inputs, this will fail. I think this is safe, since the current and subsequent versions of Bitcoin Core do not generate legacy addresses by default

smite/src/bolt/funding.rs
NishantBansal2003: Yeah, initially I was thinking of sorting the UTXOs inside the `build_funding_transaction` function, but then I thought the caller may not expect the inputs to be reordered, so I added the comment instead. But yes, I think leaving the coin selection to the caller makes sense

smite/src/bolt/funding.rs
NishantBansal2003: This is not correct (yet), since not all UTXOs will be used as inputs to construct the tx. But let me fix this by only sending those UTXOs that will actually be used as inputs, to clear up the confusion

morehouse (approved):
LGTM

PR #76, opened 2026-05-14
feat(smitebot): scaffold CLI crate and command surface

Ashish-Kumar-Dash:
## Summary
This PR introduces the foundation for `smitebot` as a new workspace crate.

### Included
- Add `smitebot` crate to workspace members.
- Add shared `clap` workspace dependency.
- Add `smitebot` CLI skeleton with command dispatch and typed errors.
- Add command tree:
  - `doctor`
  - `build`
  - `start`
  - `stop`
  - `status`
  - `corpus merge`
  - `corpus minimize`
  - `crashes`
  - `reproduce`
  - `coverage`
- Add CLI options (skeleton):
  - `start --startup-timeout-secs --auto-build`
  - `reproduce --input --retries`
  - `corpus minimize --input-dir... --out-dir --sharedir`

### Not included
- No runtime implementation yet (all handlers intentionally return `NotImplemented`).
- No behavior changes to existing crates.

## Validation
- `cargo fmt --all --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo build -p smitebot --verbose`
- `cargo run -p smitebot -- --help`
- `cargo run -p smitebot -- reproduce --help`
- `cargo run -p smitebot -- corpus minimize --help`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,6 +13,56 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is_terminal_polyfill",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle"
+version = "1.0.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
+
+[[package]]
+name = "anstyle-parse"
+version = "0.2.7"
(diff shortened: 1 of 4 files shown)


Ashish-Kumar-Dash:
ping @morehouse  @NishantBansal2003

NishantBansal2003 (changes requested):
Can we avoid stubs and only add working commands? I don’t think we can really know how each command will evolve in the future, so having dead structs does not make much sense to me.

Maybe we can start from how we currently run smite, where we need some prerequisite checks before running anything. So I assume `smitebot doctor` is what you are planning to proceed with first.

In that case, I think it would be better to focus on each command as a single feature. For `smitebot doctor`, I assume the code will not be too large to review, but even then, if you feel the PR is becoming large, you can split it into smaller commits or PRs for each part. That would help keep the work focused on individual commands and also make reviewing easier.

I also think having a concise, to the point README for each command would be good for explaining usage.

Ashish-Kumar-Dash:
I understand, one PR for one command and no skeleton, only add actual cmds when actual implementation is done, works for me
I will introduce a Smitebot guide, README basically, to explain each command's behaviour and gradually edit it as I implement actual cmds
I'm closing this PR, next up- smitebot doctor, smitebot build & config shape and validation (TOML schema + parsing tests + sample config).  Thanks for the review

Issue #71, opened 2026-05-09
# Milestone 6: Gossip Message Fuzzing (BOLT 7)


devvaansh:
> Implements structure-aware fuzzing of BOLT 7 gossip messages — `channel_announcement`,
> `node_announcement`, `channel_update`, and `announcement_signatures` — with valid
> secp256k1 signatures, so the fuzzer can penetrate gossip validation logic that
> raw-bytes fuzzing cannot reach.
>
> Scope: implements Milestone 6 from the [IR Design and Implementation Plan (#5)](https://github.com/morehouse/smite/issues/5).
> Non-goals: routing-table semantics, onion routing, BOLT 11 invoices.

## 1. Motivation

Every Lightning node accepts gossip from untrusted peers. Bugs in gossip parsing, signature verification, or storage have led to memory exhaustion, CPU DoS, and corrupted routing tables across implementations. The current
`EncryptedBytesScenario` only stresses early parsing, because raw bytes never satisfy:

- four chained `secp256k1` signatures over a double-SHA256 of the message tail
  (`channel_announcement`),
- the lexicographic `node_id_1 < node_id_2` ordering rule,
- `chain_hash` and `short_channel_id` consistency between
  `channel_announcement` → `channel_update` → `announcement_signatures`,
- the `must_be_one` flag and `htlc_minimum_msat ≤ htlc_maximum_msat ≤ capacity`
  invariants in `channel_update`.

On Extending the IR with gossip operations lets the fuzzer construct *structurally valid* messages whose mutations exercise the deep code paths behind these checks.

## 2. Current State (baseline)

| Area | Status |
|---|---|
| BOLT 7 codecs | only [`gossip_timestamp_filter`](https://github.com/morehouse/smite/blob/master/smite/src/bolt/gossip_timestamp_filter.rs) exists |
| IR `Operation`s | [`BuildOpenChannel`](https://github.com/morehouse/smite/blob/master/smite-ir/src/operation.rs) + load/compute primitives only |
| IR `VariableType`s | no `Signature`, `ShortChannelId`, or `Timestamp` types |
| Generators / executor | not yet upstream (Milestones 1–5) |
| `msg_type` constants | gossip types 256–259 missing from [`bolt.rs`](https://github.com/morehouse/smite/blob/master/smite/src/bolt.rs) |

This milestone assumes Milestone 1 (executor + `ProgramBuilder` + at least one `Generator`) has landed. It does **not** depend on Milestones 2–5.

[...]

PR #78, opened 2026-05-15
smite: add `node_announcement` codec

morehouse:
The first commit implements the standard codec, and the second commit adds the `sign` and `verify` methods we'll want for fuzzing.

Ref: #71

### DIFF
--- a/smite/src/bolt/node_announcement.rs
+++ b/smite/src/bolt/node_announcement.rs
@@ -1,0 +1,273 @@
+//! BOLT 7 node announcement message.
+
+use super::BoltError;
+use super::wire::WireFormat;
+use bitcoin::secp256k1::PublicKey;
+use bitcoin::secp256k1::ecdsa::Signature;
+
+/// BOLT 7 `node_announcement` message (type 257).
+///
+/// Allows a node to advertise extra data associated with its public key
+/// (features, alias, color, network addresses).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct NodeAnnouncement {
+    /// Signature by `node_id` over the rest of the message body.
+    pub signature: Signature,
+    /// Feature bits, BOLT 9.
+    pub features: Vec<u8>,
+    /// Monotonic per-node timestamp; receivers reject older announcements.
+    pub timestamp: u32,
+    /// Compressed secp256k1 public key identifying the announcing node.
+    pub node_id: PublicKey,
+    /// RGB color triple for UI display.
+    pub rgb_color: [u8; 3],
+    /// UTF-8 node alias, zero-padded to 32 bytes.
+    pub alias: [u8; 32],
+    /// Network address descriptors.
+    pub addresses: Vec<u8>,
+}
+
+impl NodeAnnouncement {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.signature.write(&mut out);
+        self.features.write(&mut out);
+        self.timestamp.write(&mut out);
+        self.node_id.write(&mut out);
+        self.rgb_color.write(&mut out);
+        self.alias.write(&mut out);
+        self.addresses.write(&mut out);
+        out
+    }
+
+    /// Decodes from wire format (without message type prefix).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short for any field,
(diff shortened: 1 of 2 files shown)


smite/src/bolt/node_announcement.rs
```diff
@@ -0,0 +1,328 @@
...
+        self.signature = secp.sign_ecdsa(&digest, sk);
+    }
+
+    /// Verifies `self.signature` against the embedded `node_id` per BOLT 7.
+    ///
+    /// # Errors
+    ///
+    /// Returns the underlying `secp256k1` error if the signature does not
+    /// verify against the double-SHA256 digest of the post-signature body.
+    pub fn verify(&self) -> Result<(), secp256k1::Error> {
```
NishantBansal2003: Here we are using a bool for signature verification. I think it would be better to standardize on one approach, either use bool on both sides or use `secp256k1::Error`. For our case, I think the error will always be `IncorrectSignature`, so we can probably just return a bool?
https://github.com/morehouse/smite/blob/dc2ce06a3c3a7340e9615f988446b9bfbb1fc884/smite/src/bolt/commitment.rs#L204-L217

smite/src/bolt/node_announcement.rs
```diff
@@ -0,0 +1,328 @@
...
+        na
+    }
+
+    #[test]
+    fn roundtrip() {
+        let original = sample_node_announcement();
+        let encoded = original.encode();
+        let decoded = NodeAnnouncement::decode(&encoded).unwrap();
+        assert_eq!(original, decoded);
+    }
```
NishantBansal2003, line 151: nit: just for the e2e test of the actual usage of the message
```suggestion
        assert_eq!(original, decoded);
        assert!(decoded.verify());
    }
```

morehouse: Currently `roundtrip` tests that `original == decoded`.  The `verify_succeeds_after_sign` test below ensures that `original.verify()` is true.  So if both pass, we already know `decoded.verify()` is true.

I mildly prefer to have the tests separated like they currently are.

smite/src/bolt/node_announcement.rs
morehouse: Switched to bool in da400742db929dfcd61bde9f89515372d7c08016.

morehouse:
Added another commit **25acf4ecb8a45a18cfa0683fdbd06ebfee4ee6e4** that includes trailing message bytes in the signature, as required by the spec:

https://github.com/lightning/bolts/blob/aff897f8f6ac0a7ba071b3b61fede7dc813be0b5/07-routing-gossip.md?plain=1#L362-L364

This is important in case an implementation eventually adds an extension to the `node_announcement` message.

ekzyis (approved):
LGTM, just one question: What about requirements for `addresses` like:

> [The origin node] MUST place address descriptors in ascending order.

> [The receiving node] SHOULD ignore the first `address descriptor` that does NOT match the types defined above.

Are they not relevant for fuzzing (yet)?

morehouse:
> LGTM, just one question: What about requirements for `addresses` like:
> 
> > [The origin node] MUST place address descriptors in ascending order.
> 
> > [The receiving node] SHOULD ignore the first `address descriptor` that does NOT match the types defined above.
> 
> Are they not relevant for fuzzing (yet)?

On the receiving side, we currently do not parse the address descriptors at all, we just save the bytes.  For the forseeable future, I don't think we can gain much from extracting those addresses anyway.

On the sending side, we currently just let the fuzzer pick arbitrary bytes.  In the future we could add a new IR type if the fuzzer needs some help to exercise the address parsing logic, or if our announcements are all getting rejected due to invalid descriptors.

NishantBansal2003 (approved):
LGTM 👍

> Added another commit **[25acf4e](https://github.com/morehouse/smite/commit/25acf4ecb8a45a18cfa0683fdbd06ebfee4ee6e4)** that includes trailing message bytes in the signature, as required by the spec:
> 
> https://github.com/lightning/bolts/blob/aff897f8f6ac0a7ba071b3b61fede7dc813be0b5/07-routing-gossip.md?plain=1#L362-L364
> 
> This is important in case an implementation eventually adds an extension to the `node_announcement` message.

In the future, any added fields will likely be optional for forward compatibility, so outbound signature verification should still pass, though inbound messages could cause issues. I was thinking of adding an assert that this extra field is empty, so future additions are caught explicitly, but I don’t think it adds much value. The current approach achieves the desired behavior well enough for now

PR #79, opened 2026-05-15
smite: use `bitcoin::Txid` instead of local newtype

NishantBansal2003:
Drop the local `Txid` `hash_newtype` and import `bitcoin::Txid` directly at the call sites.


morehouse (approved):
LGTM

PR #80, opened 2026-05-15
smite-ir: add `LoadTimestamp` operation

morehouse:
The timestamp type will be used when generating `node_announcement` and `channel_update` messages.

Tests are omitted for now since `LoadTimestamp` will be adequately tested by `BuildNodeAnnouncement` tests in the next PR.

Ref: #71

### DIFF
--- a/smite-ir/src/builder.rs
+++ b/smite-ir/src/builder.rs
@@ -140,6 +140,7 @@ impl ProgramBuilder {
                 self.append(Operation::LoadFeeratePerKw(rng.random()), &[])
             }
             VariableType::BlockHeight => self.append(Operation::LoadBlockHeight(rng.random()), &[]),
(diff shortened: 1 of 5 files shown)


NishantBansal2003 (approved):
LGTM 👍, Needs rebase

Q: Though we are at a very initial stage, but later on, if we insert a new Operation in the middle of `pub enum Operation` (instead of appending it), that would invalidate the collected corpus for some scenarios, right?

morehouse:
> Q: Though we are at a very initial stage, but later on, if we insert a new Operation in the middle of `pub enum Operation` (instead of appending it), that would invalidate the collected corpus for some scenarios, right?

Yes.  And actually this is also the case even if we append operations, since postcard isn't designed to be backward compatible.

If we want backward-compatibility, we should probably use protobuf.  Of course that's a much heavier option, and I don't think we would benefit from it much at this point in the project.  We can re-evaluate later if corpus invalidation becomes a bigger issue.

PR #81, opened 2026-05-15
smite: add channel_update and short_channel_id codecs

devvaansh:
Adds the wire codec, `Message` enum wiring, and ECDSA sign/verify helpers for BOLT 7 `channel_update` (msg type 258) and the supporting `short_channel_id` packed type.

This is the first vertical slice of the BOLT 7 gossip-fuzzing milestone (#71): the codec layer that everything else (IR generators, gossip scenario, harnesses) will sit on top of.

### What's in the PR

**Commit 1 — `smite: implement short_channel_id and channel_update codecs`**
- `bolt::short_channel_id::ShortChannelId` — packed 8-byte type with `new` / `from_u64` / `as_u64` / `block` / `tx_index` / `output_index` accessors, `Display` in BOLT 7 `<block>x<tx>x<out>` form, and `WireFormat` impl.
- `bolt::channel_update::ChannelUpdate` — full BOLT 7 wire layout (signature, chain_hash, scid, timestamp, flags, cltv_expiry_delta, htlc min/max msat, fees), plus `signing_region()` exposing the post-signature body bytes.
- `Message::ChannelUpdate` variant + msg_type/encode/decode arms + roundtrip test, so the dispatcher can actually receive/emit `channel_update` from the wire (not just dead-code the codec).

**Commit 2 — `smite: add sign and verify methods to ChannelUpdate`**
- `ChannelUpdate::sign(&mut self, sk)` — ECDSA over double-SHA256 of the signing region.
- `ChannelUpdate::verify(&self, pk)` — takes the pubkey explicitly, since unlike `node_announcement`, `channel_update` does not embed `node_id` on the wire (the receiver looks it up from the corresponding `channel_announcement` for the scid).
- 4 verify tests: succeeds-after-sign, fails-on-tampered-body, fails-on-wrong-pubkey, roundtrips-through-encode-decode.

This commit shape mirrors #78.

### Notes for review

- The decoder is intentionally lenient about `message_flags` / `channel_flags`: it preserves all bits and leaves policy enforcement (e.g. `must_be_one`) to the caller. This is what we want when fuzzing implementations that disagree on which bits are reserved.
- `signing_region()` is kept as a `pub` method (rather than only inlining inside `sign` / `verify`) so the gossip generator can reuse it when producing deliberately-malformed signatures.
- All 12 new tests pass; `cargo clippy -p smite` is clean on the touched files.

### Out of scope

- `channel_announcement` codec — will be a follow-up; `channel_update` was chosen first because it's by far the more frequent gossip message and is the one most useful for fuzzing forwarding-policy logic.
- IR generator (`ChannelUpdateGenerator`) and the gossip scenario — will land in subsequent PRs once Matt's `node_announcement` work (#78) and this codec are both in.

Refs #71.

### DIFF
--- a/smite/src/bolt.rs
+++ b/smite/src/bolt.rs
@@ -7,6 +7,7 @@ mod accept_channel;
 mod accept_channel2;
 mod attribution_data;
 mod channel_ready;
+mod channel_update;
 mod error;
 mod funding_created;
 mod funding_signed;
@@ -37,6 +38,7 @@ pub use accept_channel::{AcceptChannel, AcceptChannelTlvs};
 pub use accept_channel2::{AcceptChannel2, AcceptChannel2Tlvs};
 pub use attribution_data::{AttributionData, TruncatedHmac};
 pub use channel_ready::{ChannelReady, ChannelReadyTlvs};
+pub use channel_update::ChannelUpdate;
 pub use error::Error;
 pub use funding_created::FundingCreated;
 pub use funding_signed::FundingSigned;
@@ -58,7 +60,7 @@ pub use tx_remove_input::TxRemoveInput;
 pub use tx_remove_output::TxRemoveOutput;
 pub use types::{
     BigSize, CHANNEL_ID_SIZE, COMPACT_SIGNATURE_SIZE, ChannelId, MAX_MESSAGE_SIZE, PUBLIC_KEY_SIZE,
-    SHA256_HASH_SIZE, TXID_SIZE,
+    SHA256_HASH_SIZE, SHORT_CHANNEL_ID_SIZE, ShortChannelId, TXID_SIZE,
 };
 pub use update_fail_htlc::{UpdateFailHtlc, UpdateFailHtlcTlvs};
 pub use update_fail_malformed_htlc::UpdateFailMalformedHtlc;
@@ -153,6 +155,8 @@ pub mod msg_type {
     pub const UPDATE_FAIL_MALFORMED_HTLC: u16 = 135;
     /// `node_announcement` message (BOLT 7).
     pub const NODE_ANNOUNCEMENT: u16 = 257;
+    /// `channel_update` message (BOLT 7).
+    pub const CHANNEL_UPDATE: u16 = 258;
     /// Gossip timestamp filter message (BOLT 7).
     pub const GOSSIP_TIMESTAMP_FILTER: u16 = 265;
 }
@@ -209,6 +213,8 @@ pub enum Message {
     UpdateFailMalformedHtlc(UpdateFailMalformedHtlc),
     /// `node_announcement` message (type 257).
     NodeAnnouncement(NodeAnnouncement),
+    /// `channel_update` message (type 258).
+    ChannelUpdate(ChannelUpdate),
     /// Gossip timestamp filter message (type 265).
     GossipTimestampFilter(GossipTimestampFilter),
     /// Unknown message type.
@@ -252,6 +258,7 @@ impl Message {
             Self::UpdateFailHtlc(_) => msg_type::UPDATE_FAIL_HTLC,
             Self::UpdateFailMalformedHtlc(_) => msg_type::UPDATE_FAIL_MALFORMED_HTLC,
             Self::NodeAnnouncement(_) => msg_type::NODE_ANNOUNCEMENT,
+            Self::ChannelUpdate(_) => msg_type::CHANNEL_UPDATE,
             Self::GossipTimestampFilter(_) => msg_type::GOSSIP_TIMESTAMP_FILTER,
             Self::Unknown { msg_type, .. } => *msg_type,
         }
@@ -287,6 +294,7 @@ impl Message {
             Self::UpdateFailHtlc(m) => out.extend(m.encode()),
             Self::UpdateFailMalformedHtlc(m) => out.extend(m.encode()),
             Self::NodeAnnouncement(m) => out.extend(m.encode()),
+            Self::ChannelUpdate(m) => out.extend(m.encode()),
             Self::GossipTimestampFilter(m) => out.extend(m.encode()),
             Self::Unknown { payload, .. } => out.extend(payload),
         }
@@ -335,6 +343,7 @@ impl Message {
             msg_type::NODE_ANNOUNCEMENT => {
                 Ok(Self::NodeAnnouncement(NodeAnnouncement::decode(cursor)?))
             }
+            msg_type::CHANNEL_UPDATE => Ok(Self::ChannelUpdate(ChannelUpdate::decode(cursor)?)),
             msg_type::GOSSIP_TIMESTAMP_FILTER => Ok(Self::GossipTimestampFilter(
                 GossipTimestampFilter::decode(cursor)?,
             )),
@@ -809,6 +818,38 @@ mod tests {
         assert_eq!(decoded, Message::NodeAnnouncement(na));
     }
 
+    fn sample_channel_update() -> ChannelUpdate {
+        let sk = SecretKey::from_slice(&[0x22; 32]).expect("valid secret");
+
+        let mut cu = ChannelUpdate {
+            // Placeholder; overwritten by `sign` below.
+            signature: Signature::from_compact(&[0; COMPACT_SIGNATURE_SIZE])
+                .expect("zero signature parses"),
+            chain_hash: [0x6f; CHAIN_HASH_SIZE],
+            short_channel_id: ShortChannelId::from_u64(0x0000_0824_0000_0035),
+            timestamp: 1_715_000_000,
+            message_flags: 1, // must_be_one
+            channel_flags: 0,
+            cltv_expiry_delta: 144,
+            htlc_minimum_msat: 1_000,
+            fee_base_msat: 1_000,
+            fee_proportional_millionths: 100,
+            htlc_maximum_msat: 99_000_000,
+            extra: Vec::new(),
+        };
+        cu.sign(&sk);
+        cu
+    }
+
+    #[test]
+    fn message_channel_update_roundtrip() {
+        let cu = sample_channel_update();
+        let msg = Message::ChannelUpdate(cu.clone());
+        let encoded = msg.encode();
+        let decoded = Message::decode(&encoded).unwrap();
+        assert_eq!(decoded, Message::ChannelUpdate(cu));
+    }
+
     #[test]
     fn message_gossip_timestamp_filter_roundtrip() {
         let chain_hash = [0x6f; 32];
@@ -958,6 +999,10 @@ mod tests {
         assert_eq!(
             Message::NodeAnnouncement(sample_node_announcement()).msg_type(),
             msg_type::NODE_ANNOUNCEMENT
+        );
+        assert_eq!(
+            Message::ChannelUpdate(sample_channel_update()).msg_type(),
+            msg_type::CHANNEL_UPDATE
         );
         assert_eq!(
             Message::GossipTimestampFilter(GossipTimestampFilter::no_gossip([0u8; 32])).msg_type(),
--- a/smite/src/bolt/channel_update.rs
+++ b/smite/src/bolt/channel_update.rs
@@ -1,0 +1,318 @@
+//! BOLT 7 `channel_update` message.
+
+use bitcoin::hashes::{Hash, sha256d};
+use bitcoin::secp256k1::ecdsa::Signature;
+use bitcoin::secp256k1::{self, PublicKey, Secp256k1, SecretKey};
+
+use super::BoltError;
+use super::types::{CHAIN_HASH_SIZE, ShortChannelId};
+use super::wire::WireFormat;
+
+/// BOLT 7 `channel_update` message (type 258).
+///
+/// Each side of a channel independently announces its forwarding parameters
+/// using `channel_update`. The message is signed by the originating node.
+///
+/// Wire layout (per [BOLT 7]):
+///
+/// ```text
+/// [signature:64]
+/// [chain_hash:32]
+/// [short_channel_id:8]
+/// [u32:timestamp]
+/// [byte:message_flags]
+/// [byte:channel_flags]
+/// [u16:cltv_expiry_delta]
+/// [u64:htlc_minimum_msat]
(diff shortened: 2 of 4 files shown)


smite/src/bolt.rs
```diff
@@ -37,6 +39,10 @@ pub use accept_channel::{AcceptChannel, AcceptChannelTlvs};
 pub use accept_channel2::{AcceptChannel2, AcceptChannel2Tlvs};
 pub use attribution_data::{AttributionData, TruncatedHmac};
 pub use channel_ready::{ChannelReady, ChannelReadyTlvs};
+pub use channel_update::{
+    CHANNEL_FLAGS_DIRECTION, CHANNEL_FLAGS_DISABLE, CHANNEL_UPDATE_SIZE, ChannelUpdate,
+    MESSAGE_FLAGS_DONT_FORWARD, MESSAGE_FLAGS_MUST_BE_ONE,
+};
```
morehouse: I don't think we need to make `CHANNEL_UPDATE_SIZE` public.

smite/src/bolt/short_channel_id.rs
```diff
@@ -0,0 +1,189 @@
...
+impl ShortChannelId {
+    /// Constructs a `short_channel_id` from its components.
+    ///
+    /// Returns `None` if `block` or `tx_index` exceed their 24-bit field
+    /// (output index is constrained at the type level).
+    #[must_use]
+    pub const fn new(block: u32, tx_index: u32, output_index: u16) -> Option<Self> {
+        if block > MAX_BLOCK || tx_index > MAX_TX_INDEX {
+            return None;
+        }
```
morehouse: I think we could just panic if `block` or `tx_index` are out-of-range for simplicity.  I think it would indicate a programmer error in every case.


```suggestion
    pub const fn new(block: u32, tx_index: u32, output_index: u16) -> Self {
        assert!(block <= MAX_BLOCK, "block is out-of-range");
        assert!(tx_index <= MAX_TX_INDEX, "tx_index is out-of-range");
```

smite/src/bolt/short_channel_id.rs
```diff
@@ -0,0 +1,189 @@
...
+        let packed = ((block as u64) << 40) | ((tx_index as u64) << 16) | (output_index as u64);
+        Some(Self(packed))
+    }
+
+    /// Constructs a `short_channel_id` from its raw packed form.
+    ///
+    /// The packed value follows the BOLT 7 layout. Any 64-bit value is
+    /// representable, so this constructor is infallible -- it is the inverse
+    /// of [`Self::as_u64`].
```
morehouse: Nit: I think this comment is self-evident from the code


```suggestion
```

smite/src/bolt/short_channel_id.rs
```diff
@@ -0,0 +1,189 @@
...
+
+impl WireFormat for ShortChannelId {
+    /// Reads a packed 8-byte big-endian `short_channel_id`.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if fewer than 8 bytes remain.
+    fn read(data: &mut &[u8]) -> Result<Self, BoltError> {
+        let bytes: [u8; SHORT_CHANNEL_ID_SIZE] = WireFormat::read(data)?;
+        Ok(Self(u64::from_be_bytes(bytes)))
```
morehouse: We could read into a u64 for a slight simplification.

```suggestion
        let scid = u64::read(data)?;
        Ok(Self(scid))
```

smite/src/bolt/short_channel_id.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    fn write(&self, out: &mut Vec<u8>) {
+        self.0.to_be_bytes().write(out);
```
morehouse: Likewise since `self.0` is already a u64, I think all we need is:

```suggestion
        self.0.write(out);
```

smite/src/bolt/short_channel_id.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    #[test]
+    fn ord_matches_packed_u64() {
+        // Required so receivers can sort/dedupe `encoded_short_ids`.
+        let a = ShortChannelId::new(100, 0, 0).unwrap();
+        let b = ShortChannelId::new(100, 0, 1).unwrap();
+        let c = ShortChannelId::new(100, 1, 0).unwrap();
+        let d = ShortChannelId::new(101, 0, 0).unwrap();
+        assert!(a < b && b < c && c < d);
```
morehouse: Nit: separating these assertions will make it easier to tell what happened on test failure.

smite/src/bolt/channel_update.rs
```diff
@@ -0,0 +1,365 @@
...
+const SIGNED_BODY_SIZE: usize = 32 + 8 + 4 + 1 + 1 + 2 + 8 + 4 + 4 + 8;
+
+/// Total encoded size of a `channel_update` message body (without msg type).
+pub const CHANNEL_UPDATE_SIZE: usize = COMPACT_SIGNATURE_SIZE + SIGNED_BODY_SIZE;
+
+impl ChannelUpdate {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::with_capacity(CHANNEL_UPDATE_SIZE);
```
morehouse: We've generally avoided preallocating vectors for encoding, to keep the codecs simpler.  I'd prefer to continue this pattern unless we have data showing that vector allocations are measurably hurting fuzzing performance.

smite/src/bolt/channel_update.rs
```diff
@@ -0,0 +1,365 @@
...
+
+    /// Returns the bytes that are double-SHA256-hashed for signing -- i.e.
+    /// the message body excluding the leading `signature` field.
+    ///
+    /// Per BOLT 7, the signature covers everything after `signature`, including
+    /// any future fields appended to the end. Exposed for testing and for
+    /// callers that need the pre-image; prefer [`Self::sign`] / [`Self::verify`]
+    /// for the common case.
+    #[must_use]
+    pub fn signing_region(&self) -> Vec<u8> {
```
morehouse: Is this method useful?  I don't think I needed it for `node_announcement`.

smite/src/bolt/channel_update.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// Number of blocks to subtract from an incoming HTLC's `cltv_expiry`.
+    pub cltv_expiry_delta: u16,
+    /// Minimum HTLC value (millisatoshi) the channel peer will accept.
+    pub htlc_minimum_msat: u64,
+    /// Base fee charged per HTLC (millisatoshi).
+    pub fee_base_msat: u32,
+    /// Proportional fee per transferred satoshi (millionths).
+    pub fee_proportional_millionths: u32,
+    /// Maximum HTLC value (millisatoshi) the channel peer will route.
+    pub htlc_maximum_msat: u64,
```
morehouse, line 57: After studying the spec some more, I think we need to store any extra (unknown) bytes from the end of the message, so we can include them in the double-hash.

The spec says the hash is over "the entire message following the `signature` field (including unknown fields following `fee_proportional_millionths`)".  (I also need to fix this for #78).

smite/src/bolt/channel_update.rs
```diff
@@ -0,0 +1,365 @@
...
+    #[must_use]
+    pub const fn is_node_2(&self) -> bool {
+        self.channel_flags & CHANNEL_FLAGS_DIRECTION != 0
+    }
+
+    /// Returns true if the `disable` bit is set.
+    #[must_use]
+    pub const fn is_disabled(&self) -> bool {
+        self.channel_flags & CHANNEL_FLAGS_DISABLE != 0
+    }
```
morehouse: Are these methods useful for fuzzing?  If not, let's remove them.

smite/src/bolt/channel_update.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// Unlike `node_announcement`, `channel_update` does not embed the signing
+    /// node's public key on the wire: the receiver looks it up from the
+    /// previously-seen `channel_announcement` for `short_channel_id`. Callers
+    /// must therefore pass the expected key explicitly.
+    ///
+    /// # Errors
+    ///
+    /// Returns the underlying `secp256k1` error if the signature does not
+    /// verify against the double-SHA256 digest of the post-signature body.
+    pub fn verify(&self, pk: &PublicKey) -> Result<(), secp256k1::Error> {
```
morehouse: We should standardize on one API.  Following @NishantBansal2003's feedback, I changed `verify` to return `bool` in #78.  Let's do the same here unless we have a good reason to do otherwise.

smite/src/bolt/channel_update.rs
```diff
@@ -0,0 +1,365 @@
...
+        let secp = Secp256k1::new();
+        let mut body = Vec::with_capacity(SIGNED_BODY_SIZE);
+        self.write_signed_body(&mut body);
+        let digest = secp256k1::Message::from_digest(sha256d::Hash::hash(&body).to_byte_array());
+        secp.verify_ecdsa(&digest, &self.signature, pk)
+    }
+
+    fn write_signed_body(&self, out: &mut Vec<u8>) {
```
morehouse: Nit:  I slightly prefer the name `write_body` since it's shorter.

```suggestion
    fn write_body(&self, out: &mut Vec<u8>) {
```

smite/src/bolt/channel_update.rs
```diff
@@ -0,0 +1,365 @@
...
+    const BITCOIN_MAINNET: [u8; CHAIN_HASH_SIZE] = [
+        0x6f, 0xe2, 0x8c, 0x0a, 0xb6, 0xf1, 0xb3, 0x72, 0xc1, 0xa6, 0xa2, 0x46, 0xae, 0x63, 0xf7,
+        0x4f, 0x93, 0x1e, 0x83, 0x65, 0xe1, 0x5a, 0x08, 0x9c, 0x68, 0xd6, 0x19, 0x00, 0x00, 0x00,
+        0x00, 0x00,
+    ];
+
+    fn sample() -> ChannelUpdate {
+        let secp = Secp256k1::new();
+        let sk = SecretKey::from_slice(&[0x11; 32]).expect("valid secret");
+        let sig = secp.sign_ecdsa(&Message::from_digest([0xaa; 32]), &sk);
```
morehouse: For the `node_announcement` codec I actually signed the message in this function.  I think it's slightly cleaner that way.

morehouse (changes requested):

smite/src/bolt.rs
devvaansh: Removed `CHANNEL_UPDATE_SIZE` from the public re-exports. Also dropped the related flag re-exports since they became unused after removing the helper methods.

smite/src/bolt/short_channel_id.rs
devvaansh: Changed new() to return Self and use assert! for out-of-range block / tx_index, since that is a programmer error here. Updated the tests accordingly.

smite/src/bolt/short_channel_id.rs
devvaansh: i agree

morehouse (approved):
LGTM

smite/src/bolt/channel_update.rs
```diff
@@ -0,0 +1,319 @@
...
+        self.fee_base_msat.write(out);
+        self.fee_proportional_millionths.write(out);
+        self.htlc_maximum_msat.write(out);
+        out.extend_from_slice(&self.extra);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::super::types::COMPACT_SIGNATURE_SIZE;
```
NishantBansal2003: nit
```suggestion
    use super::super::COMPACT_SIGNATURE_SIZE;
```

smite/src/bolt/short_channel_id.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// Returns `Truncated` if fewer than 8 bytes remain.
+    fn read(data: &mut &[u8]) -> Result<Self, BoltError> {
+        let scid = u64::read(data)?;
+        Ok(Self(scid))
+    }
+
+    fn write(&self, out: &mut Vec<u8>) {
+        self.0.write(out);
+    }
+}
```
NishantBansal2003: Better to move this to `wire.rs` along with the related tests there, to keep things consistent

smite/src/bolt/short_channel_id.rs
```diff
@@ -0,0 +1,200 @@
+//! BOLT 7 `short_channel_id`: a packed reference to a channel funding output.
```
NishantBansal2003: Feels like this should go in `types.rs`, where we have `ChannelId` and other general types. WDYT?

devvaansh: makes sense to me `ShortChannelId` is a plain data type (just a packed u64 with accessors), it'd sit alongside `ChannelId` and `BigSize`. @morehouse what's your call on this since it would be an architectural change, I'm ready to collapse it into types.rs if that's the preferred direction.

morehouse: I do like having everything related to `ShortChannelId` in one place.  But for consistency it probably makes sense to move it to `types.rs` and `wire.rs`.

smite/src/bolt/short_channel_id.rs
devvaansh: done

smite/src/bolt.rs
```diff
@@ -809,6 +818,40 @@ mod tests {
...
     }
 
+    fn sample_channel_update() -> ChannelUpdate {
+        let sk = SecretKey::from_slice(&[0x22; 32]).expect("valid secret");
+
+        let mut cu = ChannelUpdate {
+            // Placeholder; overwritten by `sign` below.
+            signature: bitcoin::secp256k1::ecdsa::Signature::from_compact(
+                &[0; COMPACT_SIGNATURE_SIZE],
+            )
```
morehouse: Nit


```suggestion
            Signature::from_compact(&[0; COMPACT_SIGNATURE_SIZE])
```

morehouse (approved):

PR #83, opened 2026-05-18
scripts: drop `--user` flag for rootless docker

morehouse:
After switching to [rootless Docker](https://docs.docker.com/engine/security/rootless/), the coverage report script stopped working for me.

This fixes the script for rootless Docker while preserving the previous behavior when Docker runs as root.


ekzyis (approved):
didn't test rootless mode but LGTM

PR #69, opened 2026-05-08
smite: replace hkdf/sha2 with bitcoin hash-based derivation

NishantBansal2003:
ref: https://github.com/morehouse/smite/pull/68#issuecomment-4400056232

Removed the `hkdf` and `sha2` dependencies and implemented the same behavior using hash-based derivation from the `bitcoin` crate. Going forward, we will standardize on the `bitcoin` library wherever possible

I also verified the derivation by comparing it against LDK: https://github.com/lightningdevkit/rust-lightning/blob/a1d4e2f26eea696a0a1ebc2fe56abd884905bd5f/lightning/src/crypto/utils.rs#L8-L24

PR #86, opened 2026-05-19
noise: small refactor to make code match comment

ekzyis:
the suggestion from https://github.com/morehouse/smite/pull/69#discussion_r3208696106

### DIFF
--- a/smite/src/noise/handshake.rs
+++ b/smite/src/noise/handshake.rs
@@ -137,18 +137,19 @@ impl NoiseHandshake {
     fn initialize_state(responder_static: &PublicKey) -> ([u8; 32], [u8; 32]) {
         // h = SHA256(protocolName)
         // ck = h
-        let ck = Sha256::hash(PROTOCOL_NAME).to_byte_array();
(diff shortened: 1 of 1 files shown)


smite/src/noise/handshake.rs
```diff
@@ -137,18 +137,19 @@ impl NoiseHandshake {
...
 
         // h = SHA256(h || prologue)
-        let h = {
+        h = {
             let mut sha = Sha256::engine();
-            sha.input(&ck);
+            sha.input(&h);
             sha.input(PROLOGUE);
-            Sha256::from_engine(sha).to_byte_array()
+            Sha256::from_engine(sha.clone()).to_byte_array()
```
morehouse: Seems like an unrelated and unnecessary change.

```suggestion
            Sha256::from_engine(sha).to_byte_array()
```

morehouse (changes requested):

smite/src/noise/handshake.rs
ekzyis: Oh, that got accidentally pushed while experimenting, fixed in b6b6767

morehouse (approved):
LGTM

PR #87, opened 2026-05-19
smite-ir: Use bitcoin::opcodes to encode shutdown scripts

ekzyis:
Originally suggested in https://github.com/morehouse/smite/pull/58#discussion_r3162141946.

Raw opcodes are still used in tests, but I think it's fine to handle raw bytes there.

### DIFF
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -11,6 +11,7 @@
 use std::fmt;
 use std::fmt::Write;
 
+use bitcoin::{opcodes::all as opcodes, script::Builder, script::PushBytes};
 use rand::{Rng, RngExt};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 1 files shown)


smite-ir/src/operation.rs
```diff
@@ -196,32 +197,26 @@ impl ShutdownScriptVariant {
...
-                out.extend_from_slice(h);
-                out
-            }
-            Self::P2wsh(h) => {
-                let mut out = Vec::with_capacity(34);
-                out.extend_from_slice(&[0x00, 0x20]);
-                out.extend_from_slice(h);
-                out
-            }
+            Self::P2pkh(h) => Builder::from(Vec::with_capacity(25))
```
morehouse: Nit: I think this works to preallocate within the Builder, but it seems a bit hacky.  Maybe we should just drop the preallocation for clarity (perf is likely not an issue here).


```suggestion
            Self::P2pkh(h) => Builder::new()
```

morehouse (approved):

smite-ir/src/operation.rs
ekzyis: 1bae497 uses `Builder::new()` everywhere now

morehouse (approved):
LGTM

PR #88, opened 2026-05-20
Ignore startup exits in JVM crash handler

morehouse:
When Eclair starts up it runs several commands as subprocesses, some of which call `exit` and trigger our crash handler.

Previously we resorted to a couple hacks to filter out these startup exits:
- wrap the Java binary to ensure the crash handler is only installed on the JVM
- pass `-no-version-check` to the Eclair startup script to avoid `java --version` triggering our crash handler

This worked for a while, but when I recently updated my machine the crash handler started reporting a startup exit again.

This PR removes the ad-hoc filtering methods in favor of a more robust mechanism:  ignore all exits until startup is complete.  This is accomplished by using a marker file to signal to the crash handler whether startup has completed or not.

### DIFF
--- a/smite/src/runners.rs
+++ b/smite/src/runners.rs
@@ -10,6 +10,15 @@
 #[cfg(feature = "nyx")]
 use smite_nyx_sys::{nyx_fail, nyx_get_fuzz_input, nyx_init, nyx_release, nyx_skip};
 
+/// Marker file created right before the first fuzz input is delivered, so
+/// crash handlers can filter out expected subprocess exits that occur during
+/// node startup.
+const STARTUP_COMPLETE_MARKER: &str = "/tmp/smite-startup-complete";
+
+fn create_startup_complete_marker() {
+    std::fs::File::create(STARTUP_COMPLETE_MARKER).expect("startup complete file created");
+}
+
 /// `Runner` provides an abstraction for a smite test case runner (e.g. run under nyx,
 /// local system for reproduction, etc.)
 pub trait Runner {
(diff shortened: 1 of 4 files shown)


smite/src/runners.rs
```diff
@@ -10,6 +10,15 @@
 #[cfg(feature = "nyx")]
 use smite_nyx_sys::{nyx_fail, nyx_get_fuzz_input, nyx_init, nyx_release, nyx_skip};
 
+/// Marker file created right before the first fuzz input is delivered, so
+/// crash handlers can filter out expected subprocess exits that occur during
+/// node startup.
+const STARTUP_COMPLETE_MARKER: &str = "/tmp/smite-startup-complete";
+
+fn create_startup_complete_marker() {
+    let _ = std::fs::File::create(STARTUP_COMPLETE_MARKER);
```
ekzyis: Should we panic if creating the file fails? I think if it fails, the fuzzer can return false negatives (there was a crash, but no crash was reported).

morehouse: Yes, done.  FWIW I also tested this change by injecting a crash during the noise handshake and verifying that we could detect it with the `noise` scenario.

smite/src/runners.rs
```diff
@@ -10,6 +10,15 @@
 #[cfg(feature = "nyx")]
 use smite_nyx_sys::{nyx_fail, nyx_get_fuzz_input, nyx_init, nyx_release, nyx_skip};
 
+/// Marker file created right before the first fuzz input is delivered, so
+/// crash handlers can filter out expected subprocess exits that occur during
+/// node startup.
+const STARTUP_COMPLETE_MARKER: &str = "/tmp/smite-startup-complete";
+
+fn create_startup_complete_marker() {
+    let _ = std::fs::File::create(STARTUP_COMPLETE_MARKER).expect("startup complete file created");
```
ekzyis: nit: `let _ =` is no longer needed

```suggestion
    std::fs::File::create(STARTUP_COMPLETE_MARKER).expect("startup complete file created");
```

ekzyis (approved):
LGTM. Tested by fuzzing eclair with the encrypted_bytes scenario.

smite/src/runners.rs
morehouse: Done.

PR #89, opened 2026-05-20
smite-ir: implement `BuildNodeAnnouncement` operation

morehouse:
Enables programs that build signed `node_announcement` messages and send them to the target.

The `rgb_color` and `alias` fields are fixed-length literal params instead of `Bytes` variables so that they are always the correct size.  This means we lose the ability to cross-pollinate these two fields with other messages, but since these fields serve a purely cosmetic purpose the tradeoff is good.

Ref: #71

### DIFF
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -92,6 +92,22 @@ pub enum Operation {
     ///  18: `upfront_shutdown_script` (`Bytes`, empty = omit TLV)
     ///  19: `channel_type` (`Features`, empty = omit TLV)
     BuildOpenChannel,
+    /// Build a `node_announcement` message (BOLT 7, type 257).
+    ///
+    /// `rgb_color` and `alias` are op-level params (not variable inputs) so the
+    /// mutator can flip bits inside them but cannot change their lengths.
+    ///
+    /// Inputs (4):
+    ///   0: `node_sk` (`PrivateKey`) -- derives `node_id` and signs the body
+    ///   1: `features` (`Features`)
+    ///   2: `timestamp` (`Timestamp`)
+    ///   3: `addresses` (`Bytes`, raw network address descriptor bytes)
+    BuildNodeAnnouncement {
+        /// 3-byte RGB color for UI display.
+        rgb_color: [u8; 3],
+        /// 32-byte node alias, zero-padded.
+        alias: [u8; 32],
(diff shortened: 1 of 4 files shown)


smite-ir/src/tests.rs
```diff
@@ -1043,6 +1097,40 @@ fn param_mutator_caps_byte_length() {
...
+    }
+
+    let Operation::BuildNodeAnnouncement { rgb_color, alias } =
+        &program.instructions[build_idx].operation
+    else {
+        panic!("operation type changed");
+    };
+    assert_ne!(*rgb_color, original_rgb, "rgb_color never mutated");
+    assert_ne!(*alias, original_alias, "alias never mutated");
+}
```
NishantBansal2003, line 1132:
```suggestion
    assert_eq!(rgb_color.len(), 3, "rgb_color length changed");
    assert_eq!(alias.len(), 32, "alias length changed");
}
```

smite-ir/src/operation.rs
```diff
@@ -92,6 +92,22 @@ pub enum Operation {
...
+    /// Build a `node_announcement` message (BOLT 7, type 257).
+    ///
+    /// `rgb_color` and `alias` are op-level params (not variable inputs) so the
+    /// mutator can flip bits inside them but cannot change their lengths.
+    ///
+    /// Inputs (4):
+    ///   0: `node_sk` (`PrivateKey`) -- derives `node_id` and signs the body
+    ///   1: `features` (`Features`)
+    ///   2: `timestamp` (`Timestamp`)
+    ///   3: `addresses` (`Bytes`, raw u16-length-prefixed address descriptors)
```
NishantBansal2003: nit: this comment makes it seem like the input is u16 length-prefixed, but it is actually the output that is length-prefixed
```suggestion
    ///   3: `addresses` (`Bytes`, raw network address descriptor bytes)
```

NishantBansal2003 (approved):

smite-ir/src/tests.rs
morehouse: `rgb_color` and `alias` are fixed-size arrays, so it's impossible for them to ever have lengths other than 3 and 32.

smite-ir/src/operation.rs
morehouse: Done.

PR #91, opened 2026-05-21
smite: move funding/commitment to channel_tx

NishantBansal2003:
ref: https://github.com/morehouse/smite/pull/74#discussion_r3275176867


morehouse (approved):
Do you plan to implement a transaction builder object as the name suggests?

NishantBansal2003:
> Do you plan to implement a transaction builder object as the name suggests?

Do we need that? I mean, I don't really see a need for a separate object that returns the constructed tx. I named it `tx_builder` because, in these modules, we are kind of building the tx, even though we are not explicitly returning it

If you think the name is confusing because of that, then I can change it to just `transaction` maybe?

morehouse:
I'm not sure what the best name is, but it should probably have `tx` or `transaction` in it somewhere.  "Builder" might imply some kind of builder pattern/object is contained within, which we currently don't have, and it also clashes a bit with the naming of future `Build*Tx` operations we'll implement.  But having just `transaction` also collides with `bitcoin::Transaction`, and having just `tx` collides with `bolt::tx_*`.

So really it's all just bike-shedding.  Whatever you think fits best works for me.

morehouse (approved):
LGTM

PR #93, opened 2026-05-22
smite-ir: implement `NodeAnnouncementGenerator`

morehouse:
Generates programs that send correctly-signed `node_announcement` messages to the target.

Verified that all 4 targets successfully decode the `node_announcement`s we send and pass initial validation, including signature checks (except LND, which only checks signatures if we send a matching `channel_announcement` first).  Once we can send `channel_announcement`s that match the `node_announcement`s, we will be able to exercise even deeper code paths.

[ldk-coverage.tar.gz](https://github.com/user-attachments/files/28158045/ldk-coverage.tar.gz)
[lnd-coverage.tar.gz](https://github.com/user-attachments/files/28158049/lnd-coverage.tar.gz)
[cln-coverage.tar.gz](https://github.com/user-attachments/files/28158059/cln-coverage.tar.gz)
[eclair-coverage.tar.gz](https://github.com/user-attachments/files/28158062/eclair-coverage.tar.gz)

Ref: #71

### DIFF
--- a/smite-ir-mutator/src/lib.rs
+++ b/smite-ir-mutator/src/lib.rs
@@ -32,7 +32,7 @@ use std::slice;
 use rand::rngs::SmallRng;
 use rand::{RngExt, SeedableRng};
 
-use smite_ir::generators::OpenChannelGenerator;
+use smite_ir::generators::{NodeAnnouncementGenerator, OpenChannelGenerator};
 use smite_ir::mutators::{InputSwapMutator, OperationParamMutator};
 use smite_ir::{Generator, Mutator, Program, ProgramBuilder};
 
@@ -61,10 +61,14 @@ impl MutatorState {
         }
     }
 
-    /// Generates a fresh program from scratch using `OpenChannelGenerator`.
+    /// Generates a fresh program from scratch using our custom generators.
     fn generate_fresh(&mut self) -> Program {
         let mut builder = ProgramBuilder::new();
-        OpenChannelGenerator.generate(&mut builder, &mut self.rng);
+        if self.rng.random() {
+            OpenChannelGenerator.generate(&mut builder, &mut self.rng);
+        } else {
+            NodeAnnouncementGenerator.generate(&mut builder, &mut self.rng);
+        }
         self.last_sequence.clear();
         self.last_sequence.push("fresh");
         builder.build()
--- a/smite-ir/src/generators/node_announcement.rs
+++ b/smite-ir/src/generators/node_announcement.rs
@@ -1,0 +1,35 @@
+//! Generator for `node_announcement` gossip message.
+
+use rand::{Rng, RngExt};
+
+use super::Generator;
+use crate::builder::ProgramBuilder;
+use crate::{Operation, VariableType};
+
+/// Generates an unsolicited `node_announcement` send.
+pub struct NodeAnnouncementGenerator;
+
+impl Generator for NodeAnnouncementGenerator {
+    fn generate(&self, builder: &mut ProgramBuilder, rng: &mut impl Rng) {
+        let node_sk = builder.pick_variable(VariableType::PrivateKey, rng);
+        let features = builder.pick_variable(VariableType::Features, rng);
+        let timestamp = builder.pick_variable(VariableType::Timestamp, rng);
+        let addresses = builder.pick_variable(VariableType::Bytes, rng);
+
+        // rgb_color and alias are op-level params (not variable inputs) so the
+        // mutator can flip bits inside them without changing their lengths.
+        let rgb_color: [u8; 3] = rng.random();
+        let mut alias = [0u8; 32];
+        for b in &mut alias {
+            // BOLT 7 requires alias to be valid UTF-8, which Eclair actually
+            // enforces on decode.
+            *b = rng.random_range(0x00..=0x7F);
+        }
+
+        let msg = builder.append(
+            Operation::BuildNodeAnnouncement { rgb_color, alias },
+            &[node_sk, features, timestamp, addresses],
+        );
+        builder.append(Operation::SendMessage, &[msg]);
+    }
+}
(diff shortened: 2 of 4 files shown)


smite-ir-mutator/src/lib.rs
```diff
@@ -64,7 +64,11 @@ impl MutatorState {
     /// Generates a fresh program from scratch using `OpenChannelGenerator`.
```
NishantBansal2003: rustdoc needs to be updated

smite-ir/src/generators/node_announcement.rs
```diff
@@ -0,0 +1,35 @@
...
+
+        // rgb_color and alias are op-level params (not variable inputs) so the
+        // mutator can flip bits inside them without changing their lengths.
+        let rgb_color: [u8; 3] = rng.random();
+        let mut alias = [0u8; 32];
+        for b in &mut alias {
+            // BOLT 7 requires alias to be valid UTF-8, which Eclair actually
+            // enforces on decode.
+            *b = rng.random_range(0x00..=0x7F);
+        }
```
NishantBansal2003, line 27: I am assuming this is done just to give the fuzzer both valid UTF-8 and invalid UTF-8 paths, since after mutation the alias can certainly end up containing invalid UTF-8

smite-ir/src/tests.rs
```diff
@@ -669,9 +669,57 @@ fn generated_program_structure() {
     );
 }
 
+fn generate_node_announcement_program(seed: u64) -> Program {
```
NishantBansal2003, line 672: `generate_open_channel_program` has 2 more tests that we could also add for `generate_node_announcement_program` if you want parity with open channel generator (since I see a lot of the same cases): `validate_accepts_generated_program` and `generated_program_postcard_roundtrip`.

Though they do not really test very different invariants, so it's up to you whether you want to add them

NishantBansal2003 (approved):

smite-ir/src/generators/node_announcement.rs
ekzyis: I think this is done so the initial `alias` is always valid UTF-8

smite-ir/src/generators/node_announcement.rs
```diff
@@ -0,0 +1,35 @@
...
+
+impl Generator for NodeAnnouncementGenerator {
+    fn generate(&self, builder: &mut ProgramBuilder, rng: &mut impl Rng) {
+        let node_sk = builder.pick_variable(VariableType::PrivateKey, rng);
+        let features = builder.pick_variable(VariableType::Features, rng);
+        let timestamp = builder.pick_variable(VariableType::Timestamp, rng);
+        let addresses = builder.pick_variable(VariableType::Bytes, rng);
+
+        // rgb_color and alias are op-level params (not variable inputs) so the
+        // mutator can flip bits inside them without changing their lengths.
```
ekzyis, line 20: nit: Do we expect there to be more cases? Wondering if this means we're missing a `VariableType`, like `FixedBytes` or `UTF8`.

ekzyis (approved):
LGTM!

> [ldk-coverage.tar.gz](https://github.com/user-attachments/files/28158045/ldk-coverage.tar.gz)
> [lnd-coverage.tar.gz](https://github.com/user-attachments/files/28158049/lnd-coverage.tar.gz)
> [cln-coverage.tar.gz](https://github.com/user-attachments/files/28158059/cln-coverage.tar.gz)
> [eclair-coverage.tar.gz](https://github.com/user-attachments/files/28158062/eclair-coverage.tar.gz)

Mhh, maybe this could be added to CI.

Btw, why is branch coverage always `- (0/0)` for LDK? [This](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html#interpreting-reports) doesn't mention it needs to be enabled, and bdk_wallet-2.3.0/src/wallet/changeset.rs contains branches in `ChangeSet::merge()`, and there is branch coverage for CLN.

smite-ir-mutator/src/lib.rs
morehouse: Done.

smite-ir/src/generators/node_announcement.rs
morehouse: I don't think there's currently any other 3-byte fields or 32-byte UTF-8 that would benefit from a dedicated variable type.  `rgb_color` and `alias` are mostly cosmetic, and I would be surprised if we found a benefit to cross-pollinating them between different messages.  Of course we can always do that later if we do find some benefit.

smite-ir/src/generators/node_announcement.rs
morehouse: It turns out 32 completely random bytes are extremely unlikely to be valid UTF-8.  LND, LDK, and CLN don't care about valid UTF-8 aliases for initial validation, but Eclair rejects such messages immediately.

This line just limits to ASCII characters to ensure valid UTF-8 initially, while mutators are still able to create invalid aliases.

BTW I discovered a signature-verification issue in Eclair while exploring this: https://github.com/ACINQ/eclair/issues/3314.

smite-ir/src/tests.rs
morehouse: Added both tests.

morehouse:
> Mhh, maybe this could be added to CI.

My goal is for Smitebot (#70) to help with this eventually.  The goal is to have a server constantly fuzzing with the smite master branch and generating updated coverage reports each day.

> Btw, why is branch coverage always `- (0/0)` for LDK? [This](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html#interpreting-reports) doesn't mention it needs to be enabled, and bdk_wallet-2.3.0/src/wallet/changeset.rs contains branches in `ChangeSet::merge()`, and there is branch coverage for CLN.

I looked into this and discovered the unstable `-Z coverage-options=branch` [option](https://doc.rust-lang.org/rustc/instrument-coverage.html) that should give us branch instrumentation.  But when I attempted to use it on LDK I got segfaults.  It looks like this LLVM bug: https://github.com/llvm/llvm-project/issues/189169.

So at least for now we're stuck with just line coverage.

PR #95, opened 2026-05-23
smite-ir/tests: Assert mutators don't modify program on false return

Chand-ra:
Assert that mutators (`InputSwapMutator` and `OperationParamMutator`) do not modify the `Program` if mutation is aborted (that is, if `false` is returned).

### DIFF
--- a/smite-ir/src/tests.rs
+++ b/smite-ir/src/tests.rs
@@ -1094,6 +1094,34 @@ fn append_rejects_affine_overuse() {
 
 // -- OperationParamMutator tests --
 
+fn assert_false_is_noop<M: Mutator>(mutator: &M, original: &Program) {
+    let mut rng = SmallRng::seed_from_u64(0);
+    for _ in 0..100 {
+        let mut program = original.clone();
+        if !mutator.mutate(&mut program, &mut rng) {
+            assert_eq!(&program, original, "program modified on false return");
+        }
+    }
+}
+
+#[test]
+fn param_mutator_false_is_noop() {
+    let original = Program {
+        // A type-valid program containing ONLY immutable operations.
+        instructions: vec![
+            Instruction {
+                operation: Operation::LoadChainHashFromContext,
+                inputs: vec![],
+            },
+            Instruction {
+                operation: Operation::LoadTargetPubkeyFromContext,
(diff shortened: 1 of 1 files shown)


smite-ir/src/tests.rs
```diff
@@ -1229,26 +1163,9 @@ fn input_swap_preserves_types() {
...
-                    assert_eq!(
-                        actual_type, expected_types[j],
-                        "instruction {i} input {j}: expected {:?}, got {actual_type:?}",
-                        expected_types[j],
-                    );
-                }
-            }
+            program
+                .validate()
+                .expect("InputSwapMutator produces valid programs");
```
morehouse: Nit: the first commit adds the `validate` call and the second commit reformats it.  Move the reformat to the first commit.

smite-ir/src/tests.rs
```diff
@@ -991,6 +937,22 @@ fn validate_rejects_oversized_features() {
...
+            assert_eq!(&program, original, "program modified on false return");
+        }
+    }
+}
+
+#[test]
+fn param_mutator_false_is_noop() {
+    let original = generate_program(0);
+    assert_false_is_noop(&OperationParamMutator, &original);
+}
```
morehouse, line 1123: `OperationParamMutator.mutate` never returns false in this test, so it basically tests nothing.

Chand-ra: Maybe using a minimal program as the mutator's input is better?

smite-ir/src/tests.rs
```diff
@@ -1381,6 +1381,33 @@ fn validate_rejects_oversized_features() {
...
+        instructions: vec![
+            Instruction {
+                operation: Operation::LoadChainHashFromContext,
+                inputs: vec![],
+            },
+            Instruction {
+                operation: Operation::SendMessage,
+                inputs: vec![0],
+            },
+        ],
```
morehouse, line 1120: Nit: this program is type-invalid.  We can just as easily create a valid program instead.

morehouse (changes requested):
Needs a rebase and the first commit can be dropped now that #94 is merged.

morehouse (approved):
LGTM

PR #97, opened 2026-05-25
smite-ir: Introduce affine state variables for strict state enforcement

Chand-ra:
This PR introduces affine (single-use) types to the IR to strictly enforce the Lightning Network state machine during fuzzing, in order to prevent timeout hangs caused by semantically invalid mutation sequences.

Solves issue #75, see it for more details.

### DIFF
--- a/smite-ir/src/builder.rs
+++ b/smite-ir/src/builder.rs
@@ -71,6 +71,27 @@ impl ProgramBuilder {
                 actual_type, expected_type,
                 "{operation:?} input {i}: expected {expected_type:?}, got {actual_type:?}",
             );
+            if actual_type.is_affine() {
+                let cands = self.candidates.get_mut(&actual_type).unwrap_or_else(|| {
+                    panic!("{operation:?} input {i}: no candidates for {actual_type:?}")
+                });
+
+                // Find the position, or panic if it doesn't exist.
+                let pos = cands
+                    .iter()
+                    .position(|c| match c {
+                        Candidate::Direct(idx) => *idx == input_idx,
+                        Candidate::Extract { .. } => {
+                            panic!("extractable affine variable in candidate list")
+                        }
+                    })
+                    .unwrap_or_else(|| {
+                        panic!("{operation:?} input {i}: affine {actual_type:?} already consumed")
+                    });
+
+                // Consume it.
+                cands.remove(pos);
+            }
         }
 
         let idx = self.instructions.len();
@@ -78,6 +99,11 @@ impl ProgramBuilder {
         if let Some(out_type) = operation.output_type() {
             // Register extractable fields for compound types.
             for (extract_op, field_type) in operation.extractable_fields() {
+                assert!(
+                    !field_type.is_affine(),
+                    "affine variables cannot be extracted"
+                );
+                assert!(!out_type.is_affine(), "compound variables cannot be affine");
                 self.candidates
                     .entry(field_type)
                     .or_default()
@@ -103,11 +129,30 @@ impl ProgramBuilder {
 
     /// Selects or creates a variable of the given type using probabilistic
     /// variable selection (75% most recent, 15% any existing, 10% fresh).
-    #[allow(clippy::missing_panics_doc)] // candidates is always non-empty
+    /// For affine variables the most recent candidate is always selected.
+    ///
+    /// # Panics
+    ///
+    /// Panics in the following scenarios:
+    /// * An affine type is requested but its candidate pool is empty (all
+    ///   instances have been consumed).
+    /// * Attempts to generate a fresh variable for a type that cannot be
+    ///   instantiated out-of-thin-air (e.g., `Message`, `AcceptChannel`,
+    ///   or affine types).
+    /// * Resolving the chosen candidate triggers a panic in the underlying
+    ///   `append` operation.
     pub fn pick_variable(&mut self, var_type: VariableType, rng: &mut impl Rng) -> usize {
         let Some(candidates) = self.candidates.get(&var_type) else {
             return self.generate_fresh(var_type, rng);
         };
+
+        if var_type.is_affine() {
+            let chosen_candidate = candidates
+                .last()
+                .unwrap_or_else(|| panic!("no candidates for {var_type:?}"))
+                .clone();
+            return self.resolve_candidate(chosen_candidate);
+        }
 
         let roll = rng.random_range(0..20);
         match roll {
@@ -176,6 +221,9 @@ impl ProgramBuilder {
             }
             VariableType::FundingTransaction => {
                 panic!("cannot generate fresh FundingTransaction: requires composed inputs")
+            }
+            VariableType::SentOpenChannel => {
+                panic!("cannot generate fresh SentOpenChannel: affine type")
             }
         }
     }
--- a/smite-ir/src/builder.rs
+++ b/smite-ir/src/builder.rs
@@ -216,6 +216,9 @@ impl ProgramBuilder {
             VariableType::Message => {
                 panic!("cannot generate fresh Message: requires composed inputs")
             }
+            VariableType::OpenChannelMessage => {
+                panic!("cannot generate fresh OpenChannelMessage: requires composed inputs")
+            }
             VariableType::AcceptChannel => {
                 panic!("cannot generate fresh AcceptChannel: requires protocol interaction")
             }
--- a/smite-ir/src/mutators/input_swap.rs
+++ b/smite-ir/src/mutators/input_swap.rs
@@ -19,7 +19,16 @@ impl Mutator for InputSwapMutator {
             .instructions
             .iter()
             .enumerate()
-            .flat_map(|(i, instr)| (0..instr.inputs.len()).map(move |j| (i, j)))
+            .flat_map(|(i, instr)| {
+                instr
+                    .operation
+                    .input_types()
+                    .into_iter()
+                    .enumerate()
+                    // Affine variables carry no data, so swapping one affine variable with
+                    // another has no practical effect.
+                    .filter_map(move |(j, ty)| (!ty.is_affine()).then_some((i, j)))
+            })
             .choose(rng)
         else {
             return false;
--- a/smite-ir/src/mutators/operation_param.rs
+++ b/smite-ir/src/mutators/operation_param.rs
@@ -103,6 +103,7 @@ fn mutate_operation(op: &mut Operation, rng: &mut impl Rng) -> bool {
         | Operation::BuildOpenChannel
         | Operation::BuildChannelAnnouncement
         | Operation::SendMessage
+        | Operation::SendOpenChannel
         | Operation::RecvAcceptChannel
         | Operation::BroadcastTransaction => {
             unreachable!("is_param_mutable returned true for {op:?}")
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -141,6 +141,10 @@ pub enum Operation {
     /// Send an encoded message over the connection.
     /// Input: `Message`.
     SendMessage,
+    /// Send an `open_channel` message over the connection.
+    /// Produces a `SentOpenChannel` variable.
+    /// Input: `OpenChannelMessage`.
+    SendOpenChannel,
     /// Receive and parse an `accept_channel` response.
     /// Produces an `AcceptChannel` compound variable.
     RecvAcceptChannel,
@@ -594,6 +598,7 @@ impl fmt::Display for Operation {
                 format_hex(alias),
             ),
             Self::SendMessage => write!(f, "SendMessage"),
+            Self::SendOpenChannel => write!(f, "SendOpenChannel"),
         }
     }
 }
@@ -625,6 +630,7 @@ impl Operation {
                 Some(VariableType::Message)
             }
             Self::SendMessage | Self::MineBlocks(_) | Self::BroadcastTransaction => None,
+            Self::SendOpenChannel => Some(VariableType::SentOpenChannel),
             Self::RecvAcceptChannel => Some(VariableType::AcceptChannel),
         }
     }
@@ -649,8 +655,7 @@ impl Operation {
             | Self::LoadChannelType(_)
             | Self::LoadTargetPubkeyFromContext
             | Self::LoadChainHashFromContext
-            | Self::MineBlocks(_)
-            | Self::RecvAcceptChannel => vec![],
+            | Self::MineBlocks(_) => vec![],
 
             Self::DerivePoint => vec![VariableType::PrivateKey],
             Self::ExtractAcceptChannel(_) => vec![VariableType::AcceptChannel],
@@ -661,6 +666,8 @@ impl Operation {
                 VariableType::FeeratePerKw, // feerate_per_kw
             ],
             Self::SendMessage => vec![VariableType::Message],
+            Self::SendOpenChannel => vec![VariableType::OpenChannelMessage],
+            Self::RecvAcceptChannel => vec![VariableType::SentOpenChannel],
             Self::BroadcastTransaction => vec![VariableType::FundingTransaction],
 
             Self::BuildOpenChannel => vec![
@@ -736,6 +743,7 @@ impl Operation {
             | Self::BuildChannelAnnouncement
             | Self::BuildNodeAnnouncement { .. }
             | Self::SendMessage
+            | Self::SendOpenChannel
             | Self::MineBlocks(_)
             | Self::BroadcastTransaction => vec![],
 
@@ -776,6 +784,7 @@ impl Operation {
             | Self::BuildOpenChannel
             | Self::BuildChannelAnnouncement
             | Self::SendMessage
+            | Self::SendOpenChannel
             | Self::RecvAcceptChannel
             | Self::BroadcastTransaction => false,
         }
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -620,9 +620,10 @@ impl Operation {
             Self::LoadChainHashFromContext => Some(VariableType::ChainHash),
             Self::ExtractAcceptChannel(field) => Some(field.output_type()),
             Self::CreateFundingTransaction => Some(VariableType::FundingTransaction),
-            Self::BuildOpenChannel
-            | Self::BuildChannelAnnouncement
-            | Self::BuildNodeAnnouncement { .. } => Some(VariableType::Message),
+            Self::BuildOpenChannel => Some(VariableType::OpenChannelMessage),
+            Self::BuildChannelAnnouncement | Self::BuildNodeAnnouncement { .. } => {
+                Some(VariableType::Message)
+            }
             Self::SendMessage | Self::MineBlocks(_) | Self::BroadcastTransaction => None,
             Self::RecvAcceptChannel => Some(VariableType::AcceptChannel),
         }
--- a/smite-ir/src/program.rs
+++ b/smite-ir/src/program.rs
@@ -85,6 +85,14 @@ pub enum ValidateError {
         /// Actual byte length.
         len: usize,
     },
+    /// An affine variable is used more than once.
+    #[error("Variable {index}: affine {var_type:?} consumed twice (max 1)")]
+    AffineOverUse {
+        /// The overused affine variable index.
+        index: usize,
+        /// Type of the affine variable.
+        var_type: VariableType,
+    },
 }
 
 impl Program {
@@ -101,6 +109,8 @@ impl Program {
     ///
     /// Returns the first violation encountered.
     pub fn validate(&self) -> Result<(), ValidateError> {
+        let mut affine_consumed = vec![false; self.instructions.len()];
+
         for (instr_idx, instr) in self.instructions.iter().enumerate() {
             let expected = instr.operation.input_types();
             if instr.inputs.len() != expected.len() {
@@ -134,6 +144,15 @@ impl Program {
                         expected: expected_type,
                         got: actual_type,
                     });
+                }
+                if expected_type.is_affine() {
+                    if affine_consumed[input_idx] {
+                        return Err(ValidateError::AffineOverUse {
+                            index: input_idx,
+                            var_type: expected_type,
+                        });
+                    }
+                    affine_consumed[input_idx] = true;
                 }
             }
             if let Operation::LoadBytes(b) | Operation::LoadFeatures(b) = &instr.operation
--- a/smite-ir/src/tests.rs
+++ b/smite-ir/src/tests.rs
@@ -1096,6 +1096,39 @@ fn pick_variable_reuses_existing() {
 }
 
 #[test]
+fn pick_variable_uses_unspent_affine() {
+    let mut rng = SmallRng::seed_from_u64(0);
+    let mut builder = ProgramBuilder::new();
+    OpenChannelGenerator.generate(&mut builder, &mut rng);
+
+    let msg_idx = builder.pick_variable(VariableType::OpenChannelMessage, &mut rng);
+    let unspent_sent_open_channel = builder.append(Operation::SendOpenChannel, &[msg_idx]);
+
+    // pick_variable should prefer an unspent affine variable.
+    let idx = builder.pick_variable(VariableType::SentOpenChannel, &mut rng);
+    assert_eq!(idx, unspent_sent_open_channel);
+}
+
+#[test]
+#[should_panic(expected = "cannot generate fresh SentOpenChannel: affine type")]
+fn pick_variable_panics_on_empty_affine() {
+    let mut rng = SmallRng::seed_from_u64(0);
+    let mut builder = ProgramBuilder::new();
+    builder.pick_variable(VariableType::SentOpenChannel, &mut rng);
+}
+
+#[test]
+#[should_panic(expected = "no candidates for SentOpenChannel")]
+fn pick_variable_panics_on_all_affine_consumed() {
+    let mut rng = SmallRng::seed_from_u64(0);
+    let mut builder = ProgramBuilder::new();
+    OpenChannelGenerator.generate(&mut builder, &mut rng);
+
+    // All `SentOpenChannel` have been already consumed.
+    builder.pick_variable(VariableType::SentOpenChannel, &mut rng);
+}
+
+#[test]
 #[should_panic(expected = "cannot generate fresh Message")]
 fn generate_fresh_message_panics() {
     let mut rng = SmallRng::seed_from_u64(0);
@@ -1117,6 +1150,14 @@ fn generate_fresh_funding_transaction_panics() {
     let mut rng = SmallRng::seed_from_u64(0);
     let mut builder = ProgramBuilder::new();
     builder.generate_fresh(VariableType::FundingTransaction, &mut rng);
+}
+
+#[test]
+#[should_panic(expected = "cannot generate fresh SentOpenChannel: affine type")]
+fn generate_fresh_sent_open_channel_panics() {
+    let mut rng = SmallRng::seed_from_u64(0);
+    let mut builder = ProgramBuilder::new();
+    builder.generate_fresh(VariableType::SentOpenChannel, &mut rng);
 }
 
 #[test]
@@ -1163,6 +1204,20 @@ fn append_type_mismatch_panics() {
     let mut builder = ProgramBuilder::new();
     let amount = builder.generate_fresh(VariableType::Amount, &mut rng);
     builder.append(Operation::DerivePoint, &[amount]);
+}
+
+#[test]
+#[should_panic(expected = "RecvAcceptChannel input 0: affine SentOpenChannel already consumed")]
+fn append_rejects_affine_overuse() {
+    let mut rng = SmallRng::seed_from_u64(0);
+    let mut builder = ProgramBuilder::new();
+    OpenChannelGenerator.generate(&mut builder, &mut rng);
+
+    let msg_idx = builder.pick_variable(VariableType::OpenChannelMessage, &mut rng);
+    let sent_open_channel = builder.append(Operation::SendOpenChannel, &[msg_idx]);
+    // Add consecutive `RecvAcceptChannel`s.
+    builder.append(Operation::RecvAcceptChannel, &[sent_open_channel]);
+    builder.append(Operation::RecvAcceptChannel, &[sent_open_channel]);
 }
 
 // -- Program::validate tests --
@@ -1373,6 +1428,34 @@ fn validate_rejects_oversized_features() {
             instr: 0,
             len: MAX_MESSAGE_SIZE + 1,
         }),
+    );
+}
+
+#[test]
+fn validate_accepts_unused_affine() {
+    let mut program = generate_open_channel_program(0);
+    // Remove the `RecvAcceptChannel` instruction at the end.
+    program.instructions.pop();
+
+    // Validation should pass despite the unspent `SentOpenChannel`.
+    assert_eq!(program.validate(), Ok(()));
+}
+
+#[test]
+fn validate_rejects_affine_overuse() {
+    let mut program = generate_open_channel_program(0);
+    let sent_open_channel = program.instructions.len() - 2;
+    // Add another `RecvAcceptChannel` in addition to the existing one.
+    program.instructions.push(Instruction {
+        operation: Operation::RecvAcceptChannel,
+        inputs: vec![sent_open_channel],
+    });
+    assert_eq!(
+        program.validate(),
+        Err(ValidateError::AffineOverUse {
+            index: sent_open_channel,
+            var_type: VariableType::SentOpenChannel,
+        })
     );
 }
 
--- a/smite-ir/src/tests.rs
+++ b/smite-ir/src/tests.rs
@@ -956,10 +956,10 @@ fn generated_open_channel_program_structure() {
     let program = generate_open_channel_program(0);
     let ops: Vec<_> = program.instructions.iter().map(|i| &i.operation).collect();
 
-    // Must end with SendMessage, RecvAcceptChannel.
+    // Must end with SendOpenChannel, RecvAcceptChannel.
     assert!(
-        matches!(ops[ops.len() - 2], Operation::SendMessage),
-        "second-to-last instruction should be SendMessage",
+        matches!(ops[ops.len() - 2], Operation::SendOpenChannel),
+        "second-to-last instruction should be SendOpenChannel",
     );
     assert!(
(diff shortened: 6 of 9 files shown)


smite-ir/src/operation.rs
```diff
@@ -457,6 +461,7 @@ pub enum AcceptChannelField {
     FirstPerCommitmentPoint,
     UpfrontShutdownScript,
     ChannelType,
+    ChannelState,
```
Chand-ra: As discussed [here](https://github.com/morehouse/smite/issues/5#issuecomment-4561293465), this is no longer of use. Sending an unexpected message before any `Recv` operation produces a non-fatal error, so it's not something we should make the IR enforce, but rather let the mutators explore freely.

smite-ir/src/builder.rs
```diff
@@ -109,6 +118,27 @@ impl ProgramBuilder {
...
+                    break;
+                }
+            }
+            // If all candidates are spent, grab the absolute last one and let the validator
+            // handle the rest.
+            let candidate_to_resolve = chosen_candidate
+                .unwrap_or_else(|| candidates.last().unwrap())
+                .clone();
+            return self.resolve_candidate(candidate_to_resolve);
+        }
```
Chand-ra: As discussed [here](https://github.com/morehouse/smite/issues/5#issuecomment-4561293465), this is an anti-pattern and we should modify the function's signature to return an `Option<usize>` instead of intentionally generating an invalid instruction.

smite-ir/src/variable.rs
```diff
@@ -42,6 +43,10 @@ pub enum Variable {
     Message(Vec<u8>),
     /// Parsed `accept_channel` response.
     AcceptChannel(AcceptChannel),
+    /// Affine lock: Channel opened, waiting for `accept_channel`.
+    PendingAccept,
```
morehouse: The term "affine lock" is unnecessarily confusing -- I don't think anything is actually being "locked".  Let's just create a section titled "affine variables".

Also I think it would be slightly easier to understand the variable if we called it `SentOpenChannel` instead.  Then `SendOpenChannel` returns a `SentOpenChannel` and `RecvAcceptChannel` requires `SentOpenChannel`.  Both read more naturally.

```suggestion

    // Affine (single-use) variables
    
    /// `open_channel` has been sent, so `accept_channel` may now be received.
    SentOpenChannel,
```

smite-ir/src/variable.rs
```diff
@@ -42,6 +43,10 @@ pub enum Variable {
...
+    /// Affine lock: Channel established. Consumed by all active channel actions.
+    ChannelState,
```
morehouse: I don't think we need this variable.

smite-ir/src/operation.rs
```diff
@@ -113,6 +115,10 @@ pub enum Operation {
     /// Send an encoded message over the connection.
     /// Input: `Message`.
     SendMessage,
+    /// Send an `open_channel` message over the connection.
+    /// Produces a `PendingAccept` variable.
+    /// Input: `Message`.
+    SendOpenChannel,
```
morehouse, line 147: I think we might actually want a specific `OpenChannelMessage` type that is input for `SendOpenChannel`.  That way we can be sure that some other message type is never used as input to `SendOpenChannel`.

Relying on the executor to reject these cases is slow and unnecessary if we add the new variable type.

smite-ir/src/builder.rs
```diff
@@ -73,6 +76,10 @@ impl ProgramBuilder {
             );
         }
 
+        for input in inputs {
+            self.refs_count[*input] += 1;
+        }
+
```
morehouse: `append` also needs to verify that any affine variables referenced haven't already been consumed.

smite-ir/src/builder.rs
```diff
@@ -30,6 +30,8 @@ pub struct ProgramBuilder {
     instructions: Vec<Instruction>,
     /// Candidate variables indexed by output type.
     candidates: HashMap<VariableType, Vec<Candidate>>,
+    /// Reference count for each variable.
+    refs_count: Vec<usize>,
```
morehouse: We don't need a reference count -- really all we need is to know whether affine variables have been consumed yet or not.  So a `Vec<bool>` would suffice, or even better -- just remove affine variables from `candidates` as they're consumed (no new vector necessary).

smite-ir/src/builder.rs
```diff
@@ -73,6 +76,10 @@ impl ProgramBuilder {
             );
         }
 
+        for input in inputs {
+            self.refs_count[*input] += 1;
```
morehouse: We don't need another loop here.  Just move the one-line loop body into the previous loop.

smite-ir/src/builder.rs
```diff
@@ -109,6 +118,27 @@ impl ProgramBuilder {
             return self.generate_fresh(var_type, rng);
         };
 
+        if var_type.is_affine() {
+            let mut chosen_candidate = None;
+            // Iterate backwards to find the last unused candidate.
+            for candidate in candidates.iter().rev() {
```
morehouse: I don't think there's any benefit to choosing the *last* unused candidate.  We could just as effectively choose the first or *any* arbitrary candidate.

Also, if we eliminate `refs_count` entirely and instead ensure that `candidates` contains only *unused* affine variables, then the logic here becomes dead simple -- just pop a value off the candidate list and use it.

smite-ir/src/operation.rs
```diff
@@ -703,10 +720,43 @@ impl Operation {
...
             | Self::BuildOpenChannel
             | Self::SendMessage
+            | Self::SendOpenChannel
             | Self::RecvAcceptChannel => false,
         }
     }
+
+    /// Returns true if this operation is Extract*.
+    #[must_use]
+    pub fn is_extract(&self) -> bool {
```
morehouse: We already have `extractable_fields`, so maybe we don't need this function.

smite-ir/src/program.rs
```diff
@@ -85,6 +86,30 @@ pub enum ValidateError {
...
+        /// Type of the affine variable.
+        var_type: VariableType,
+        /// Number of times the variable is used.
+        ref_count: usize,
+    },
+    /// An affine variable is extracted more than once.
+    #[error(
+        "Instruction {instr}: affine {var_type:?} extracted {extract_count} times from {index} (max 1)"
+    )]
+    AffineOverExtract {
```
morehouse: I think we should just make `InputSwapMutator` skip affine variables.  In general these variables carry no data and are just there for sequencing, so swapping one affine variable with another has no practical effect.

And if extracting affine variables is an issue for other mutators, then perhaps we can outright ban affine variables inside compound variables, so we don't have to deal with it.  At least for now we don't have any need since `Send*Message` only ever has a single return value (the affine variable).

smite-ir/src/operation.rs
```diff
@@ -56,6 +56,8 @@ pub enum Operation {
     LoadTargetPubkeyFromContext,
     /// Load the chain hash from the program context.
     LoadChainHashFromContext,
+    /// Load a state variable from the program context.
+    LoadStateFromContext(VariableType),
```
morehouse: We don't currently need this operation, so let's defer implementing it until there's a need.

smite-scenarios/src/executor.rs
```diff
@@ -190,6 +206,30 @@ pub fn execute(
...
+                log::debug!(
+                    "[{:?}] SendOpenChannel: {} bytes",
+                    start.elapsed(),
+                    bytes.len(),
+                );
+                conn.send_message(bytes)?;
+                Some(Variable::PendingAccept)
+            }
+
             Operation::RecvAcceptChannel => {
```
morehouse, line 251: The executor needs to enforce the affine input type, and should also ensure the affine type hasn't already been consumed.

smite-ir-mutator/src/lib.rs
```diff
@@ -79,14 +79,19 @@ impl MutatorState {
...
             let name = if self.rng.random() {
-                OperationParamMutator.mutate(program, &mut self.rng);
+                OperationParamMutator.mutate(&mut cloned, &mut self.rng);
                 "op-param"
             } else {
-                InputSwapMutator.mutate(program, &mut self.rng);
+                InputSwapMutator.mutate(&mut cloned, &mut self.rng);
                 "input-swap"
             };
+            if cloned.validate().is_err() {
```
morehouse: I don't think we need to do this.  I'd prefer to have the requirement that mutators always produce valid programs.

morehouse (changes requested):

smite-ir/src/operation.rs
Chand-ra: I'm worried about the IR bloat if we do that, we'd eventually need 40+ variants for every LN message. Maybe making the executor reject these is wrong and we should allow `InputSwapMutator` to freely swap different message types?

Chand-ra:
Rebased on top of the latest master to resolve merge conflicts.

smite-ir/src/builder.rs
Chand-ra: Decided to panic instead, because that's in-line with the current behavior. `SpliceMutator` will require this behavior to change as a whole, so I'll modify it if and when we need it instead of front-loading the complexity in this PR.

smite-ir/src/builder.rs
```diff
@@ -71,6 +71,25 @@ impl ProgramBuilder {
                 actual_type, expected_type,
                 "{operation:?} input {i}: expected {expected_type:?}, got {actual_type:?}",
             );
+            if actual_type.is_affine() {
```
morehouse, line 74: There's some trickiness here -- should `is_affine()` return true or false for a compound variable containing an affine field?

If it returns true, we need to ensure below that we only remove the candidate when the affine field is actually extracted -- not when a non-affine field is extracted.

If it returns false, this code will never run for compound variables, and we would fail to detect multiple uses of affine fields.

Or if we are going to ban affine types from being extractable fields, we should add an assertion somewhere to panic on such cases.

smite-ir/src/mutators/input_swap.rs
```diff
@@ -28,6 +28,12 @@ impl Mutator for InputSwapMutator {
         let current_input = program.instructions[instr_idx].inputs[input_pos];
         let expected_type = program.instructions[instr_idx].operation.input_types()[input_pos];
 
+        // Affine variables carry no data, so swapping one affine variable with
+        // another has no practical effect.
+        if expected_type.is_affine() {
+            return false;
+        }
```
morehouse: It would be better to filter out affine inputs from the list above.

smite-scenarios/src/executor.rs
```diff
@@ -149,6 +154,7 @@ pub fn execute(
 ) -> Result<(), ExecuteError> {
     let secp = Secp256k1::new();
     let mut variables: Vec<Option<Variable>> = Vec::with_capacity(program.instructions.len());
+    let mut consumed_affines = vec![false; program.instructions.len()];
```
morehouse: We already have the `variables` vector.  We can simply consume affine variables directly from there, replacing them with `None`.  No need to track `consumed` separately.

smite-scenarios/src/executor.rs
```diff
@@ -159,6 +165,16 @@ pub fn execute(
                 got: instr.inputs.len(),
             });
         }
+        // Before running the operation, mark all affine inputs as consumed.
+        for &input_idx in &instr.inputs {
```
morehouse: Double resolving is non-ideal.  We should mark affine variables as consumed when we do regular resolution later.

smite-ir/src/operation.rs
morehouse: We only need one special message type per special send operation.  i.e. `SendOpenChannel` gets a special `OpenChannelMessage`.  I don't think we need to be concerned about bloat here -- there's a good reason for the new types, just as there's a good reason for the new `Send*` operations.

Letting `InputSwapMutator` swap to a different message type would reintroduce the problem we're trying to solve with this PR.

morehouse (changes requested):

smite-ir/src/builder.rs
Chand-ra: Compound variables containing an affine type should definitely return false on `is_affine()`, to allow multiple extractions from them.

To access an affine variable contained in a compound one, Generators should first manually `Extract*` the affine variable, registering it as an available candidate. We can then detect over-extractions by re-introducing the `OverExtract` logic we just got rid of, and over-use of the extracted affine will be naturally handled by the existing code.

Or, we can introduce dummy operations that accept such compound variables and output a single affine type.

In any case, I don't think we need to deal with this for now. The specifics can be ironed out if and when we encounter such a scenario (I personally have a feeling we won't).

Chand-ra:
Had to rebase again to solve a merge conflict.

smite-ir/src/builder.rs
```diff
@@ -105,10 +124,17 @@ impl ProgramBuilder {
...
+        let Some(candidates) = self.candidates.get_mut(&var_type) else {
             return self.generate_fresh(var_type, rng);
         };
 
+        if var_type.is_affine() {
+            let chosen_candidate = candidates
+                .pop()
+                .unwrap_or_else(|| panic!("no candidates for {var_type:?}"));
+            return self.resolve_candidate(chosen_candidate);
+        }
```
morehouse, line 155: Typical usage of `ProgramBuilder` is like this:

```rust
let v = builder.pick_variable(SentOpenChannel, rng);
builder.append(RecvAcceptChannel, &[v]);
```

So we can't remove from `self.candidates` in both `pick_variable` and `append` or we'll double-consume the affine variable.

I'm pretty sure we want `append` to be the only place where we actually mark the variable as consumed.

smite-ir/src/builder.rs
```diff
@@ -105,10 +124,17 @@ impl ProgramBuilder {
     /// variable selection (75% most recent, 15% any existing, 10% fresh).
     #[allow(clippy::missing_panics_doc)] // candidates is always non-empty
```
morehouse: `candidates` actually *can* be empty now since we delete affine types from the entry's vector in `append` but never fully remove the entry itself.

We should either ensure that empty entries get fully deleted or we should fix the assumption here.

smite-ir/src/builder.rs
morehouse: Then let's ban it and add an assertion in the `operation.extractable_fields()` loop below that extractable fields are never affine.

smite-ir/src/mutators/input_swap.rs
```diff
@@ -19,7 +19,13 @@ impl Mutator for InputSwapMutator {
...
             .iter()
             .enumerate()
-            .flat_map(|(i, instr)| (0..instr.inputs.len()).map(move |j| (i, j)))
+            .flat_map(|(i, instr)| {
+                (0..instr.inputs.len())
+                    // Affine variables carry no data, so swapping one affine variable with
+                    // another has no practical effect.
+                    .filter(move |&j| !instr.operation.input_types()[j].is_affine())
+                    .map(move |j| (i, j))
+            })
```
morehouse, line 31: Nit:  it's a bit awkward/inefficient to call `input_types` once for every input instead of once per instruction

```suggestion
            .flat_map(|(i, instr)| {
                instr
                    .operation
                    .input_types()
                    .into_iter()
                    .enumerate()
                    .filter_map(move |(j, ty)| (!ty.is_affine()).then_some((i, j)))
            })
```

smite-scenarios/src/executor.rs
```diff
@@ -231,7 +236,35 @@ pub fn execute(
                 None
             }
 
+            Operation::SendOpenChannel => {
+                let bytes = resolve_message(&variables, instr.inputs[0])?;
```
morehouse: This should be resolving an `OpenChannelMessage` now instead of `Message`, which also allows us to remove the msg_type check below.

We should also test the case where a different message type is passed as input.

smite-scenarios/src/executor.rs
```diff
@@ -231,7 +236,35 @@ pub fn execute(
...
+                    bytes.len(),
+                );
+                conn.send_message(bytes)?;
+                Some(Variable::SentOpenChannel)
+            }
+
             Operation::RecvAcceptChannel => {
+                let sent_open_channel = resolve_sent_open_channel(&variables, instr.inputs[0])?;
+                // Consume the affine `SentOpenChannel`.
+                variables[sent_open_channel] = None;
```
morehouse: I think it would make sense to rename `resolve_sent_open_channel` to `consume_sent_open_channel` and have it update `variables` instead, so the resolve+consume operations happen in the same place.

smite-scenarios/src/executor.rs
```diff
@@ -422,6 +455,19 @@ fn resolve_funding_transaction(
     }
 }
 
+fn resolve_sent_open_channel(
+    variables: &[Option<Variable>],
+    index: usize,
+) -> Result<usize, ExecuteError> {
+    let Ok(var) = resolve(variables, index) else {
+        return Err(ExecuteError::AffineOverUse { index });
```
morehouse: This mislabels index-out-of-bounds errors as `AffineOverUse`.

smite-scenarios/src/executor.rs
```diff
@@ -1621,6 +1695,38 @@ mod tests {
...
+        conn.queue_recv(ac_bytes);
+        let err = execute(
+            &program,
+            &sample_context(),
+            &mut conn,
+            &mut MockBitcoinCli::default(),
+            std::time::Instant::now(),
+        )
+        .unwrap_err();
+        dbg!(&err);
```
```suggestion
```

smite-scenarios/src/executor.rs
```diff
@@ -422,6 +455,19 @@ fn resolve_funding_transaction(
...
 
+fn resolve_sent_open_channel(
+    variables: &[Option<Variable>],
+    index: usize,
+) -> Result<usize, ExecuteError> {
+    let Ok(var) = resolve(variables, index) else {
+        return Err(ExecuteError::AffineOverUse { index });
+    };
+    match var {
+        Variable::SentOpenChannel => Ok(index),
```
morehouse: Returning the same index that was passed in seems off.  I think we should just return `Ok(())` instead.

If we do change this function to also consume the affine variable, the return value will be unused anyway,

morehouse (changes requested):

smite-ir/src/builder.rs
Chand-ra: > So we can't remove from `self.candidates` in both `pick_variable` and `append` or we'll double-consume the affine variable.
> 
> I'm pretty sure we want `append` to be the only place where we actually mark the variable as consumed.

Ugh, you're right, my bad.

smite-ir/src/builder.rs
Chand-ra: I think `#[allow(clippy::missing_panics_doc)]` doesn't really make sense as a whole. We _do_ panic if we try to `generate_fresh()` a type that cannot be freshly generated.

Let's add panic docs instead.

smite-ir/src/builder.rs
```diff
@@ -103,12 +126,27 @@ impl ProgramBuilder {
...
     /// variable selection (75% most recent, 15% any existing, 10% fresh).
-    #[allow(clippy::missing_panics_doc)] // candidates is always non-empty
+    ///
+    /// # Panics
+    ///
+    /// Panics in the following scenarios:
+    /// * An affine type is requested but its candidate pool is empty (all instances have been consumed).
+    /// * Attempts to generate a fresh variable for a type that cannot be instantiated out-of-thin-air
+    ///   (e.g., `Message`, `AcceptChannel`, or affine types).
+    /// * Resolving the chosen candidate triggers a panic in the underlying `append` operation.
```
morehouse: Nit: line wrap looks different than elsewhere in this file.

smite-scenarios/src/executor.rs
```diff
@@ -422,6 +450,28 @@ fn resolve_funding_transaction(
...
+        Err(e) => return Err(e),
+    };
+    match var {
+        Variable::SentOpenChannel => {
+            // Consume the affine `SentOpenChannel`.
+            variables[index] = None;
+            Ok(())
+        }
+        _ => Err(type_err(VariableType::SentOpenChannel, var)),
+    }
```
morehouse: The two matches can be combined.


```suggestion
    match resolve(variables, index) {
        Ok(Variable::SentOpenChannel) => {
            // Consume the affine `SentOpenChannel`.
            variables[index] = None;
            Ok(())
        }
        Ok(_) => Err(type_err(VariableType::SentOpenChannel, var)),
        // Map `VoidVariable` to `AffineOverUse`
        Err(ExecuteError::VoidVariable { index }) => Err(ExecuteError::AffineOverUse { index }),
        Err(e) => Err(e),
    }
```

smite-ir/src/builder.rs
```diff
@@ -103,12 +126,27 @@ impl ProgramBuilder {
 
     /// Selects or creates a variable of the given type using probabilistic
     /// variable selection (75% most recent, 15% any existing, 10% fresh).
```
morehouse, line 131: We should update the comment to briefly describe how affine variables are selected.


```suggestion
    /// Selects or creates a variable of the given type using probabilistic
    /// variable selection (75% most recent, 15% any existing, 10% fresh).
    /// For affine variables the most recent candidate is always selected.
```

smite-ir/src/builder.rs
```diff
@@ -168,12 +206,18 @@ impl ProgramBuilder {
...
+                panic!("cannot generate fresh OpenChannelMessage: requires composed inputs")
+            }
             VariableType::AcceptChannel => {
                 panic!("cannot generate fresh AcceptChannel: requires protocol interaction")
             }
             VariableType::FundingTransaction => {
                 panic!("cannot generate fresh FundingTransaction: requires composed inputs")
             }
+            VariableType::SentOpenChannel => {
+                panic!("cannot generate fresh affine type")
```
morehouse: Nit:


```suggestion
                panic!("cannot generate fresh SentOpenChannel: affine type")
```

smite-ir/src/builder.rs
```diff
@@ -71,13 +71,36 @@ impl ProgramBuilder {
...
+                let cands = self.candidates.get_mut(&actual_type).unwrap_or_else(|| {
+                    panic!("{operation:?} input {i}: no candidates for {actual_type:?}")
+                });
+
+                // Find the position, or panic if it doesn't exist.
+                let pos = cands
+                    .iter()
+                    .position(|c| match c {
+                        Candidate::Direct(idx) => *idx == input_idx,
+                        Candidate::Extract { compound_idx, .. } => *compound_idx == input_idx,
```
morehouse: The extract arm is banned and should never exist


```suggestion
                        Candidate::Direct(idx) => *idx == input_idx,
                        Candidate::Extract { .. } => panic!("extractable affine variable in candidate list"),
```

smite-ir/src/mutators/input_swap.rs
```diff
@@ -19,7 +19,16 @@ impl Mutator for InputSwapMutator {
...
-            .flat_map(|(i, instr)| (0..instr.inputs.len()).map(move |j| (i, j)))
+            .flat_map(|(i, instr)| {
+                instr
+                    .operation
+                    .input_types()
+                    .into_iter()
+                    .enumerate()
+                    // Affine variables carry no data, so swapping one affine variable with
+                    // another has no practical effect.
+                    .filter_map(move |(j, ty)| (!ty.is_affine()).then_some((i, j)))
```
morehouse, line 30: We should add a test ensuring that affine inputs are never swapped.

smite-ir/src/mutators/operation_param.rs
```diff
@@ -103,7 +103,8 @@ fn mutate_operation(op: &mut Operation, rng: &mut impl Rng) -> bool {
         | Operation::BuildChannelAnnouncement
         | Operation::SendMessage
         | Operation::RecvAcceptChannel
-        | Operation::BroadcastTransaction => {
+        | Operation::BroadcastTransaction
+        | Operation::SendOpenChannel => {
```
morehouse: Nit: `SendOpenChannel` comes immediately after `SendMessage` in the enum order

smite-ir/src/operation.rs
```diff
@@ -615,10 +620,12 @@ impl Operation {
             Self::LoadChainHashFromContext => Some(VariableType::ChainHash),
             Self::ExtractAcceptChannel(field) => Some(field.output_type()),
             Self::CreateFundingTransaction => Some(VariableType::FundingTransaction),
-            Self::BuildOpenChannel
-            | Self::BuildChannelAnnouncement
-            | Self::BuildNodeAnnouncement { .. } => Some(VariableType::Message),
+            Self::BuildChannelAnnouncement | Self::BuildNodeAnnouncement { .. } => {
+                Some(VariableType::Message)
+            }
+            Self::BuildOpenChannel => Some(VariableType::OpenChannelMessage),
```
morehouse: Nit: `BuildOpenChannel` comes immediately after `CreateFundingTransaction` in the enum order

smite-ir/src/operation.rs
```diff
@@ -642,9 +649,9 @@ impl Operation {
             | Self::LoadChannelType(_)
             | Self::LoadTargetPubkeyFromContext
             | Self::LoadChainHashFromContext
-            | Self::MineBlocks(_)
-            | Self::RecvAcceptChannel => vec![],
+            | Self::MineBlocks(_) => vec![],
 
+            Self::RecvAcceptChannel => vec![VariableType::SentOpenChannel],
```
morehouse: Nit: `RecvAcceptChannel` comes immediately after `SendOpenChannel` in the enum order

smite-ir/src/operation.rs
```diff
@@ -656,6 +663,8 @@ impl Operation {
             Self::SendMessage => vec![VariableType::Message],
             Self::BroadcastTransaction => vec![VariableType::FundingTransaction],
 
+            Self::SendOpenChannel => vec![VariableType::OpenChannelMessage],
```
morehouse, line 669: Nit: `SendOpenChannel` comes immediately after `SendMessage` in the enum order

smite-ir/src/operation.rs
```diff
@@ -729,7 +738,8 @@ impl Operation {
             | Self::BuildNodeAnnouncement { .. }
             | Self::SendMessage
             | Self::MineBlocks(_)
-            | Self::BroadcastTransaction => vec![],
+            | Self::BroadcastTransaction
+            | Self::SendOpenChannel => vec![],
```
morehouse: Nit: `SendOpenChannel` comes immediately after `SendMessage` in the enum order

smite-ir/src/operation.rs
```diff
@@ -768,7 +778,8 @@ impl Operation {
             | Self::BuildChannelAnnouncement
             | Self::SendMessage
             | Self::RecvAcceptChannel
-            | Self::BroadcastTransaction => false,
+            | Self::BroadcastTransaction
+            | Self::SendOpenChannel => false,
```
morehouse: Nit: `SendOpenChannel` comes immediately after `SendMessage` in the enum order

smite-scenarios/src/executor.rs
```diff
@@ -1323,9 +1395,15 @@ mod tests {
...
 
-        // Verify a correctly-sized pong was sent.
-        assert_eq!(conn.sent.len(), 1);
-        let pong = Message::decode(&conn.sent[0]).unwrap();
+        // Verify exactly two messages were sent: `open_channel` and `pong`.
+        assert_eq!(conn.sent.len(), 2);
+
+        // Verify the first message was `open_channel` (type 32).
+        let sent_type = u16::from_be_bytes([conn.sent[0][0], conn.sent[0][1]]);
+        assert_eq!(sent_type, msg_type::OPEN_CHANNEL);
```
morehouse: Nit: for consistency let's follow the same decode pattern used for pong below:


```suggestion
        // Verify the first message was `open_channel`.
        let oc = Message::decode(&conn.set[0]).unwrap();
        let Message::OpenChannel(_) = oc else {
            panic!("expected OpenChannel, got {:?}", oc.msg_type());
        };
```

smite-ir/src/tests.rs
```diff
@@ -1095,6 +1095,28 @@ fn pick_variable_reuses_existing() {
...
+    assert_eq!(idx, unspent_sent_open_channel);
+}
+
+#[test]
+#[should_panic(expected = "cannot generate fresh affine type")]
+fn pick_variable_panics_on_empty_affine() {
+    let mut rng = SmallRng::seed_from_u64(0);
+    let mut builder = ProgramBuilder::new();
+    builder.pick_variable(VariableType::SentOpenChannel, &mut rng);
+}
```
morehouse, line 1118: I think it would make sense to add a test that `pick_variable` panics when all affine candidates have been consumed already.  Basically it would just do `OpenChannelGenerator.generate` followed by `pick_variable(SentOpenChannel)`.

smite-scenarios/src/executor.rs
morehouse: This comment wasn't addressed.

smite-scenarios/src/executor.rs
```diff
@@ -1621,6 +1695,73 @@ mod tests {
...
+        };
+
+        let mut conn = MockConnection::new();
+        let err = execute(
+            &program,
+            &sample_context(),
+            &mut conn,
+            &mut MockBitcoinCli::default(),
+            std::time::Instant::now(),
+        )
```
morehouse: Nit: since we don't actually use `conn` we can inline its creation


```suggestion
        let err = execute(
            &program,
            &sample_context(),
            &mut MockConnection::new(),
            &mut MockBitcoinCli::default(),
            std::time::Instant::now(),
        )
```

morehouse (commented):
The code looks correct to me now.  All my remaining comments are nits related to style or missing tests.

Also the commit message for d7f0788c2dbcacc9400af4b4560ca0dd4a312e72 is slightly out-of-date and the first bullet point should be updated to:

```
- Add `SendOpenChannel` operation: Consumes a composed `OpenChannelMessage`
  and outputs a `SentOpenChannel` affine variable.
```

smite-ir/src/builder.rs
Chand-ra: This is required by clippy.

Chand-ra:
Had to rebase (yet again) to resolve a merge conflict.

smite-ir/src/tests.rs
```diff
@@ -1095,6 +1095,39 @@ fn pick_variable_reuses_existing() {
...
+
+#[test]
+#[should_panic(expected = "no candidates for SentOpenChannel")]
+fn pick_variable_panics_on_all_affine_consumed() {
+    let mut rng = SmallRng::seed_from_u64(0);
+    let mut builder = ProgramBuilder::new();
+    OpenChannelGenerator.generate(&mut builder, &mut rng);
+
+    // All `SentOpenChannel` have bee already consumed.
```
morehouse:
```suggestion
    // All `SentOpenChannel` have been already consumed.
```

smite-ir/src/builder.rs
```diff
@@ -71,13 +71,38 @@ impl ProgramBuilder {
...
 
         let idx = self.instructions.len();
 
         if let Some(out_type) = operation.output_type() {
             // Register extractable fields for compound types.
             for (extract_op, field_type) in operation.extractable_fields() {
+                assert!(
+                    !out_type.is_affine(),
+                    "Affine variables cannot be extracted"
+                );
```
morehouse, line 105: I think we actually want to assert also that `field_type` is not affine:


```suggestion
                assert!(!field_type.is_affine(), "affine variables cannot be extracted");
                assert!(!out_type.is_affine(), "compound variables cannot be affine");
```

smite-ir/src/builder.rs
```diff
@@ -103,12 +128,28 @@ impl ProgramBuilder {
...
+    /// # Panics
+    ///
+    /// Panics in the following scenarios:
+    /// * An affine type is requested but its candidate pool is empty (all instances have been consumed).
+    /// * Attempts to generate a fresh variable for a type that cannot be instantiated out-of-thin-air
+    ///   (e.g., `Message`, `AcceptChannel`, or affine types).
+    /// * Resolving the chosen candidate triggers a panic in the underlying `append` operation.
     pub fn pick_variable(&mut self, var_type: VariableType, rng: &mut impl Rng) -> usize {
-        let Some(candidates) = self.candidates.get(&var_type) else {
+        let Some(candidates) = self.candidates.get_mut(&var_type) else {
```
morehouse: We no longer need `get_mut` since the candidate list only gets modified in `append` now.
```suggestion
        let Some(candidates) = self.candidates.get(&var_type) else {
```

smite-ir/src/builder.rs
morehouse: No, I modified it locally to wrap at 80 chars and clippy did not complain.

smite-ir/src/tests.rs
```diff
@@ -1675,3 +1755,35 @@ fn input_swap_preserves_types() {
         }
     }
 }
+
+#[test]
+fn input_swap_preserves_affine() {
```
morehouse, line 1760: The insertion in this test makes it difficult to follow -- the reader must think about input indexes of all the following instructions, which are not updated after the insert.

I think it would be easier to understand if we just rewrote the tail of the program -- pop the `RecvAcceptChannel` instruction, then append another `SendOpenChannel` and a `RecvAcceptChannel` consuming the first `SentOpenChannel`.

morehouse (commented):
The commit message for 9cfec5c still says a `Message` is consumed when in fact it is an `OpenChannelMessage`.

smite-ir/src/builder.rs
Chand-ra: Oh okay, I thought you were talking about the indentation before `(e.g., ...).`.

smite-ir/src/builder.rs
```diff
@@ -71,13 +71,39 @@ impl ProgramBuilder {
...
 
         let idx = self.instructions.len();
 
         if let Some(out_type) = operation.output_type() {
             // Register extractable fields for compound types.
             for (extract_op, field_type) in operation.extractable_fields() {
+                assert!(
+                    !out_type.is_affine(),
+                    "Affine variables cannot be extracted"
+                );
```
morehouse, line 105:
```suggestion
                assert!(!field_type.is_affine(), "affine variables cannot be extracted");
```

morehouse (approved):
LGTM.  Also did a 15 minute smoke test fuzzing LDK, and no issues.

smite-ir/src/operation.rs
NishantBansal2003, line 585: Now that `RecvAcceptChannel` takes `SentOpenChannel` as input, the empty parens should be removed here and the corresponding tests, comments etc should also be updated accordingly

PR #100, opened 2026-05-27
smite-ir: add BroadcastTransaction IR operation

NishantBansal2003:
ref: https://github.com/morehouse/smite/pull/90#issuecomment-4532231730 and https://github.com/morehouse/smite/pull/90#issuecomment-4549524815

Depends-on: https://github.com/morehouse/smite/pull/90

This will be used to broadcast the transaction. Right now, we only take the funding transaction as input, but in the future, once we have more transaction types, we will likely move to accepting a generic transaction input and allow broadcasting at arbitrary points.

Currently, this will be used to broadcast the funding transaction after `funding_signed` is received (or possibly even before that), so that the targets can exchange `channel_ready` messages after some number of blocks are mined.

### DIFF
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -126,6 +126,9 @@ pub enum Operation {
     RecvAcceptChannel,
     /// Mines the given number of blocks on the Bitcoin network.
     MineBlocks(u8),
+    /// Sign wallet inputs of the transaction and broadcast it via `bitcoin-cli`.
+    /// Input: `FundingTransaction`.
+    BroadcastTransaction,
 }
 
 /// A BOLT 2 compliant `upfront_shutdown_script` template.
@@ -553,6 +556,7 @@ impl fmt::Display for Operation {
             Self::LoadChainHashFromContext => write!(f, "LoadChainHashFromContext()"),
             Self::RecvAcceptChannel => write!(f, "RecvAcceptChannel()"),
             Self::MineBlocks(v) => write!(f, "MineBlocks({v})"),
+            Self::BroadcastTransaction => write!(f, "BroadcastTransaction"),
             // Operations with inputs: parens added by Program::Display.
             Self::DerivePoint => write!(f, "DerivePoint"),
             Self::ExtractAcceptChannel(field) => write!(f, "Extract{field}"),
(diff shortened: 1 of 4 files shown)


smite-ir/src/operation.rs
```diff
@@ -724,6 +731,7 @@ impl Operation {
             | Self::CreateFundingTransaction
             | Self::BuildOpenChannel
             | Self::SendMessage
+            | Self::BroadcastTransaction
```
morehouse: Nit: `BroadcastTransaction` comes after `RecvAcceptChannel` in the enum order

smite-ir/src/tests.rs
```diff
@@ -471,8 +481,8 @@ fn create_funding_transaction_with_wrong_input(input: usize, wrong: Operation) -
     // We will take the correct set of instructions for creating the funding
     // transaction, and then at given input position insert an invalid input
     // type that `CreateFundingTransaction` does not expect.
-    let mut instructions = create_funding_transaction_instructions();
-    let producer = instructions.last().unwrap().inputs[input];
+    let mut instructions = create_and_broadcast_tx_instructions();
+    let producer = instructions.iter().nth_back(1).unwrap().inputs[input];
```
morehouse: Nit: this is getting a little convoluted.  Maybe we should just search for the `CreateFundingTransaction` operation directly.

smite-ir/src/operation.rs
```diff
@@ -126,6 +126,10 @@ pub enum Operation {
     RecvAcceptChannel,
     /// Mines the given number of blocks on the Bitcoin network.
     MineBlocks(u8),
+    /// Sign all inputs of the transaction with the wallet and broadcast it via
+    /// `bitcoin-cli`.
+    /// Input: `FundingTransaction`.
```
morehouse: I think this is true today, since all inputs are our own.  But eventually it will not be true (e.g., dual funding, channel closure).


```suggestion
    /// Sign wallet inputs of the transaction and broadcast it via `bitcoin-cli`.
    /// Input: `FundingTransaction`.
```

smite-ir/src/tests.rs
```diff
@@ -369,7 +369,7 @@ fn validate_rejects_mine_blocks_with_wrong_input() {
 }
 
 #[test]
```
morehouse, line 375: We should add `BroadcastTransaction` to `postcard_roundtrip`.

morehouse (approved):

morehouse (approved):
LGTM

PR #59, opened 2026-04-24
smite-scenarios: ignore unknown odd-type messages

morehouse:
BOLT 1 requires nodes to ignore unknown message types if the type is odd.  In particular, Eclair sends various custom informational messages (e.g., `recommended_feerates` as type 39409) that interfere with message processing if we fail to ignore them.

This change is required to actually extract any `accept_channel` response from Eclair.

Ref: #5 (Milestone 1)

PR #101, opened 2026-05-27
workloads: bump eclair to v0.14.0

### LINKED ISSUE #3271: Malformed messages silently dropped instead of closing connection
When `TransportHandler` receives a message that decrypts successfully but fails decoding, it logs a warning and silently drops the message:

https://github.com/ACINQ/eclair/blob/a4d66adce2ec180c94532d1d0014f86f6e74f607/eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala#L114-L115

No warning message is sent to the peer, and the connection remains open.

This makes it much more difficult to debug implementation incompatibilities, while allowing small codec bugs to cascade into worse issues.

### Force closures and obscured bugs

Suppose either Eclair or one of its peers has a codec bug, preventing Eclair from decoding a channel update successfully.  When decoding fails, Eclair ignores the channel update.  As a result, commitments get unsynced and the channel must be force closed.

When examining logs later to debug the force closure, the cause appears to be a signature failure, not a codec problem.  The root cause is obscured.

If Eclair had instead immediately sent a warning message and closed the connection on decode failure, the channel would have remained open.  And when debugging the connection issue later on, the root cause would be immediately obvious to both node operators.

### Suggested fix

On `Attempt.Failure` in `decodeAndSendToListener`, Eclair should send a `Warning` to the remote peer and close the connection.  This would match the pattern Eclair already uses for other protocol violations, and AFAICT would match what every other implementation does on decode failure.

### Discovery

This issue was surfaced while fuzzing the `init` message handshake using [Smite](https://github.com/morehouse/smite).  Eclair would drop invalid `init` messages without closing the connection, effectively turning the connection into a zombie.  Any subsequent messages would be silently dropped with a log message such as:

```
WARN  PeerConnection - acking unhandled message Ping(...) in state INITIALIZING
```

morehouse:
Eclair [v0.14.0](https://github.com/ACINQ/eclair/releases/tag/v0.14.0) fixes https://github.com/ACINQ/eclair/issues/3271, so we don't need to [work around](https://github.com/morehouse/smite/pull/59#issuecomment-4354678028) that anymore when fuzzing Eclair.

Verified with 10+ minute runs of all 4 scenarios against the Eclair target.  No hangs were reported.


ekzyis:
Mhh, unfortunately can't review because I'm currently getting this when fuzzing any eclair scenario on master (cd92510) and here (18dd1e2):

```
[QEMU-NYX] Error: Agent has crashed before initializing the fuzzing loop: exit (code 0)

[!] libnyx failed to initialize QEMU-Nyx: agent abort() ->
        Agent has crashed before initializing the fuzzing loop: exit (code 0)


[-] PROGRAM ABORT : Something went wrong ...
         Location : afl_fsrv_start(), src/afl-forkserver.c:835
```

<details>
<summary>full afl-fuzz -X output</summary>

```
[+] Enabled environment variable AFL_CUSTOM_MUTATOR_LIBRARY with value target/release/libsmite_ir_mutator.so
[+] Enabled environment variable AFL_CUSTOM_MUTATOR_ONLY with value 1
[!] WARNING: Mistyped AFL environment variable: AFL_FRAMESHIFT_DISABLE=1
[+] Enabled environment variable AFL_NO_UI with value 1
afl-fuzz++4.35c based on afl by Michal Zalewski and a large online community
[+] AFL++ is maintained by Marc "van Hauser" Heuse, Dominik Maier, Andrea Fioraldi and Heiko "hexcoder" Eißfeldt
[+] AFL++ is open source, get it at https://github.com/AFLplusplus/AFLplusplus
[+] NOTE: AFL++ >= v3 has changed defaults and behaviours - see README.md
[+] AFL++ Nyx mode is enabled (developed and maintained by Sergej Schumilo)
[+] Nyx is open source, get it at https://github.com/Nyx-Fuzz
[+] No -M/-S set, autoconfiguring for "-S default"
[+] Enabled environment variable AFL_DISABLE_TRIM with value 1
[*] Getting to work...
[+] Using exploration-based constant power schedule (EXPLORE)
[+] Enabled testcache with 50 MB
[+] Generating fuzz data with a length of min=1 max=1048576
[*] Trying to load libnyx.so plugin...
[+] libnyx plugin is ready!
[+] Disabling the UI because AFL_NO_UI is set.
[+] You have 20 CPU cores and 2 runnable tasks (utilization: 10%).
[+] Try parallel jobs - see /nix/store/lzlfx38i7gzdl443kpjcmb0j6v2agsvy-aflplusplus-4.35c/share/doc/afl/fuzzing_in_depth.md#c-using-multiple-cores
[*] Setting up output directories...
[+] Output directory exists but deemed OK to reuse.
[*] Deleting old session data...
[+] Output dir cleanup successful.
[*] Checking CPU core loadout...
[+] Found a free CPU core, try binding to #0.
[*] Loading custom mutator library from 'target/release/libsmite_ir_mutator.so'...
[+] Found 'afl_custom_mutator'.
[*] optional symbol 'afl_custom_fuzz_count' not found.
[*] optional symbol 'afl_custom_post_process' not found.
[*] optional symbol 'afl_custom_init_trim' not found.
[*] optional symbol 'afl_custom_trim' not found.
[*] optional symbol 'afl_custom_post_trim' not found.
[*] optional symbol 'afl_custom_havoc_mutation' not found.
[*] optional symbol 'afl_custom_havoc_mutation_probability' not found.
[*] optional symbol 'afl_custom_queue_get' not found.
[+] Found 'afl_custom_splice_optout'.
[*] optional symbol 'afl_custom_fuzz_send' not found.
[*] optional symbol 'afl_custom_post_run' not found.
[*] optional symbol 'afl_custom_queue_new_entry' not found
[+] Found 'afl_custom_describe'.
[+] Custom mutator 'target/release/libsmite_ir_mutator.so' installed successfully.
[*] Validating target binary...
[*] Scanning '/tmp/smite-seeds'...
[*] Creating hard links for all input files...
[+] Loaded a total of 1 seeds.
[*] Spinning up the NYX backend...
[!] libnyx: spawning qemu with:
 /nix/store/hs9zdphnr4nbxa52pnlr1q8xdh6r0kcw-QEMU-Nyx-ff1c897321-/bin/qemu-system-x86_64 -kernel /nix/store/7fbs2hj90ivhl8m96kk0yximxmmh4gsa-nyx-packer-7967b43-/linux_initramfs/bzImage-linux-4.15-rc7 -initrd /nix/store/7fbs2hj90ivhl8m96kk0
yximxmmh4gsa-nyx-packer-7967b43-/linux_initramfs/init.cpio.gz -append nokaslr oops=panic nopti ignore_rlimit_data -display none -serial none -enable-kvm -net none -k de -m 4096 -chardev socket,server,path=/tmp/smite-out/workdir/interface_0
,id=nyx_interface -device nyx,chardev=nyx_interface,bitmap_size=65536,input_buffer_size=1048576,worker_id=0,workdir=/tmp/smite-out/workdir,sharedir=/tmp/smite-nyx,aux_buffer_size=4096 -machine kAFL64-v1 -cpu kAFL64-Hypervisor-v1 -fast_vm_r
eload path=/tmp/smite-out/workdir/snapshot/,load=off,skip_serialization=on
[QEMU-Nyx] Could not access KVM-PT kernel module!
[QEMU-Nyx] Trying vanilla KVM...
[QEMU-Nyx] NYX runs in fallback mode (no Intel-PT tracing or nested hypercall support)!
[QEMU-NYX] Max Dirty Ring Size -> 1048576 (Entries: 65536)
[QEMU-Nyx] Warning: Attempt to use unsupported CPU model (PT) without KVM-PT (Hint: use '-cpu kAFL64-Hypervisor-v2' instead)
qemu-system-x86_64: warning: host doesn't support requested feature: CPUID.07H:EBX.hle [bit 4]
qemu-system-x86_64: warning: host doesn't support requested feature: CPUID.07H:EBX.rtm [bit 11]
[QEMU-NYX] Dirty ring mmap region located at 0x7fd992f7e000
[QEMU-NYX] Booting VM to start fuzzing...
[!] libnyx: input buffer is write protected
[hget] 16528 bytes received from hypervisor! (hcat_no_pt)
[hget] 16528 bytes received from hypervisor! (habort_no_pt)
[hget] 461742080 bytes received from hypervisor! (container.tar)
[hcat] 1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
[capablities] host_config.bitmap_size: 0x10000
[capablities] host_config.ijon_bitmap_size: 0x1000
[capablities] host_config.payload_buffer_size: 0x100000x
[init] using TARGET_MAP_SIZE: 276813
[init] scenario not compiled with afl instrumentation
[QEMU-NYX] Error: Agent has crashed before initializing the fuzzing loop: exit (code 0)

[!] libnyx failed to initialize QEMU-Nyx: agent abort() ->
        Agent has crashed before initializing the fuzzing loop: exit (code 0)


[-] PROGRAM ABORT : Something went wrong ...
         Location : afl_fsrv_start(), src/afl-forkserver.c:835
```

</details>

(The warning about `AFL_FRAMESHIFT_DISABLE=1` is there because I run v4.35c, not v4.40c. I think it's unrelated.)

I also can't reproduce https://github.com/morehouse/smite/pull/59#pullrequestreview-4197764207 on #59 (55db49f1b4927d37e0f31907571a5008c03f6233) anymore. Same error.

morehouse:
Does #88 help?  It looks like the exact error that motivated that change.

ekzyis:
> Does #88 help? It looks like the exact error that motivated that change.

Ahh yes, it works with #88

ekzyis (approved):
ACK 18dd1e2 after rebase

I'm getting this when fuzzing eclair-init on 18dd1e2:

```
[QEMU-NYX] Error: Agent has crashed before initializing the fuzzing loop: exit (code 0)

[!] libnyx failed to initialize QEMU-Nyx: agent abort() ->
        Agent has crashed before initializing the fuzzing loop: exit (code 0)


[-] PROGRAM ABORT : Something went wrong ...
         Location : afl_fsrv_start(), src/afl-forkserver.c:839

```

and this on master (f870482)

```
[*] Attempting dry run with 'id:000000,time:0,execs:0,orig:seed1'...

[-] The program took more than 1000 ms to process one of the initial test cases.
    This is bad news; raising the limit with the -t option is possible, but
    will probably make the fuzzing process extremely slow.

    If this test case is just a fluke, the other option is to just avoid it
    altogether, and find one that is less of a CPU hog.

[-] PROGRAM ABORT : Test case 'id:000000,time:0,execs:0,orig:seed1' results in a timeout
         Location : perform_dry_run(), src/afl-fuzz-init.c:1014
```

but 18dd1e2 works after rebase on master, which includes #88.

PR #104, opened 2026-05-28
smitebot: Add `smitebot build` command

Ashish-Kumar-Dash:
Adds `smitebot build` to build Smite workload Docker images without starting a fuzzing campaign.

The command selects the correct workload Dockerfile from `--target`, passes `SCENARIO` as the build arg, and tags images using the existing `smite-<target>-<scenario>` convention. It also supports coverage images, custom tags, and `--no-cache` for debugging rebuilds.

Validation:
- cargo fmt --all --check
- cargo clippy -p smitebot --all-targets --offline -- -D warnings
- cargo test -p smitebot --offline
- cargo run -p smitebot --offline -- build --help

Ref. #70

### DIFF
--- a/smitebot/README.md
+++ b/smitebot/README.md
@@ -19,6 +19,32 @@ smitebot doctor --aflpp-path ~/AFLplusplus --json
 
 ## Commands
 
+### smitebot build
+
+`smitebot build` builds Smite workload Docker images for manual rebuilds and debugging.
+
+```bash
+smitebot build --target lnd --scenario encrypted_bytes
+smitebot build --target cln --scenario noise --coverage
+smitebot build --target ldk --scenario init --image local/ldk-init:debug --no-cache
+```
+
+Flags:
+
+- `--target`: Workload implementation to build an image for (`lnd`, `cln`, `ldk`, or `eclair`).
+- `--scenario`: Scenario that the image should run.
+- `--coverage`: Build a coverage-instrumented image.
+- `--image`: Use a custom image tag instead of the default tag used by Smite.
+- `--smite-dir`: Path to the Smite repository root. Defaults to the current directory.
+- `--no-cache`: Perform a clean rebuild without using cached Docker layers.
+
+By default, image tags follow the existing Smite convention:
+
+```text
+smite-<target>-<scenario>
+smite-<target>-<scenario>-coverage
+```
+
 ### smitebot doctor
 
 `smitebot doctor` validates host prerequisites before running Smite campaigns.
--- a/smitebot/src/commands/build.rs
+++ b/smitebot/src/commands/build.rs
@@ -1,0 +1,266 @@
+//! Docker image builds for Smite workloads.
+//! The command keeps Docker's output visible so rebuild failures are easy to debug.
+
+use std::fmt;
+use std::path::PathBuf;
+use std::process::{Command, ExitStatus};
+
+use clap::{Args, ValueEnum};
+
+/// Command handler for `smitebot build`.
+pub struct BuildCommand;
+
+/// CLI arguments for `smitebot build`.
+#[derive(Debug, Args)]
+pub struct BuildArgs {
+    /// Target implementation to build.
+    #[arg(long)]
+    target: WorkloadTarget,
+    /// Scenario binary selected by the workload Dockerfile.
+    #[arg(long)]
+    scenario: String,
+    /// Build the coverage-instrumented Docker image.
+    #[arg(long)]
+    coverage: bool,
+    /// Override the Docker image tag.
+    #[arg(long)]
+    image: Option<String>,
+    /// Path to smite repository root.
+    #[arg(long, default_value = ".")]
+    smite_dir: PathBuf,
+    /// Pass --no-cache to docker build.
+    #[arg(long)]
+    no_cache: bool,
+}
+
+/// Smite workload targets with Dockerfiles under `workloads/`.
+#[derive(Clone, Copy, Debug, ValueEnum)]
+enum WorkloadTarget {
+    /// Lightning Network Daemon workload.
+    Lnd,
+    /// Core Lightning workload.
+    Cln,
+    /// LDK Node workload.
+    Ldk,
+    /// Eclair workload.
+    Eclair,
+}
+
+impl fmt::Display for WorkloadTarget {
+    /// Formats the lowercase target name used in paths and Docker image tags.
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Lnd => write!(f, "lnd"),
+            Self::Cln => write!(f, "cln"),
+            Self::Ldk => write!(f, "ldk"),
+            Self::Eclair => write!(f, "eclair"),
+        }
+    }
+}
+
+/// Fully resolved Docker build inputs.
+#[derive(Debug, PartialEq, Eq)]
+struct BuildInputs {
+    /// Docker image tag to produce.
+    image: String,
+    /// Workload Dockerfile selected from `--target` and `--coverage`.
+    dockerfile: PathBuf,
+    /// Smite repository root used as the Docker build context.
+    smite_dir: PathBuf,
+    /// Scenario passed to Docker as `--build-arg SCENARIO=...`.
+    scenario: String,
+    /// Whether Docker should rebuild without using its layer cache.
+    no_cache: bool,
+}
+
+impl BuildInputs {
+    /// Resolves Docker build inputs from parsed CLI arguments.
+    fn from_args(args: &BuildArgs) -> Self {
+        let dockerfile_name = if args.coverage {
+            "Dockerfile.coverage"
+        } else {
+            "Dockerfile"
+        };
+
+        Self {
+            image: args.image.clone().unwrap_or_else(|| {
+                default_workload_image_tag(args.target, &args.scenario, args.coverage)
+            }),
+            dockerfile: args
+                .smite_dir
+                .join("workloads")
+                .join(args.target.to_string())
+                .join(dockerfile_name),
+            smite_dir: args.smite_dir.clone(),
+            scenario: args.scenario.clone(),
+            no_cache: args.no_cache,
+        }
+    }
+}
+
+impl BuildCommand {
+    /// Builds the requested Smite Docker image and returns whether Docker succeeded.
+    pub fn execute(args: &BuildArgs) -> bool {
+        let inputs = BuildInputs::from_args(args);
+        if !inputs.dockerfile.exists() {
+            log::error!("Dockerfile not found: {}", inputs.dockerfile.display());
+            return false;
+        }
+
+        log::info!(
+            "building {} with {}",
+            inputs.image,
+            inputs.dockerfile.display()
+        );
+
+        let status = match run_docker_build(&inputs) {
+            Ok(status) => status,
+            Err(e) => {
+                log::error!("failed to run docker build: {e}");
+                return false;
+            }
+        };
+
+        if !status.success() {
+            log::error!("docker build failed with {status}");
+            return false;
+        }
+
+        log::info!("built {}", inputs.image);
+        true
+    }
+}
+
+/// Returns the default image tag used by Smite's manual Docker build flow.
+fn default_workload_image_tag(target: WorkloadTarget, scenario: &str, coverage: bool) -> String {
+    let suffix = if coverage { "-coverage" } else { "" };
+    format!("smite-{target}-{scenario}{suffix}")
+}
+
+/// Runs `docker build`, streaming stdout/stderr directly to the terminal.
+fn run_docker_build(inputs: &BuildInputs) -> std::io::Result<ExitStatus> {
+    let mut command = Command::new("docker");
+    command.arg("build");
+    if inputs.no_cache {
+        command.arg("--no-cache");
+    }
+    command
+        .arg("-t")
+        .arg(&inputs.image)
+        .arg("--build-arg")
+        .arg(format!("SCENARIO={}", inputs.scenario))
+        .arg("-f")
+        .arg(&inputs.dockerfile)
+        .arg(&inputs.smite_dir);
+
+    command.status()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::path::Path;
+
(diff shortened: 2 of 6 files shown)


Ashish-Kumar-Dash:
ping @NishantBansal2003

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,267 @@
...
+            Self::Eclair => "eclair",
+        }
+    }
+}
+
+impl fmt::Display for WorkloadTarget {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
```
NishantBansal2003: nit: for consistency with rest of the codebase, for eg in `operation.rs`:
```suggestion
impl fmt::Display for WorkloadTarget {
    /// Formats the lowercase target name used in paths and Docker image tags.
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self {
            Self::Lnd => write!(f, "lnd"),
            Self::Cln => write!(f, "cln"),
            Self::Ldk => write!(f, "ldk"),
            Self::Eclair => write!(f, "eclair"),
        }
    }
}
```

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,267 @@
...
+
+/// Fully resolved Docker build inputs.
+#[derive(Debug, PartialEq, Eq)]
+struct WorkloadImageBuildPlan {
```
NishantBansal2003: nit: struct name is a bit confusing given the surrounding comments. It might be better to use a more aligned name, for eg `BuildInputs`

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,267 @@
...
+    image: String,
+    dockerfile: PathBuf,
+    context: PathBuf,
+    scenario: String,
+    no_cache: bool,
```
NishantBansal2003, line 73: nit: missing field comments here. They would help clarify what each field represents. For eg I don’t immediately understand what `context` refers to without looking at the implementation

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,267 @@
...
+    fn from_args(args: &BuildArgs) -> Self {
+        let dockerfile_name = if args.coverage {
+            "Dockerfile.coverage"
+        } else {
+            "Dockerfile"
+        };
+        let default_image = default_workload_image_tag(args.target, &args.scenario, args.coverage);
+
+        Self {
+            image: args.image.clone().unwrap_or(default_image),
```
NishantBansal2003: nit: we should only resolve the default image tag when an override image tag is not provided
```suggestion
        Self {
            image: args.image.clone().unwrap_or_else(|| {
                default_workload_image_tag(args.target, &args.scenario, args.coverage)
            }),
```

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,267 @@
...
+    ]);
+    args
+}
+
+/// Runs `docker build`, streaming stdout/stderr directly to the terminal.
+fn docker_build_status(plan: &WorkloadImageBuildPlan) -> std::io::Result<bool> {
+    let mut command = Command::new("docker");
+    command.args(docker_build_args(plan));
+    command.status().map(|status| status.success())
+}
```
NishantBansal2003: These two functions could be merged into a single function, which would make the code easier to read. A name like `run_docker_build` might fit well. Also, the existing code (for eg `bitcoind.rs`) uses the `.arg(...)` pattern rather than accumulating arguments in a vector, so I would prefer to follow that style for consistency

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,267 @@
...
+            Ok(true) => {
+                println!("smitebot build: built {}", plan.image);
+                true
+            }
+            Ok(false) => false,
+            Err(e) => {
+                eprintln!("smitebot build: failed to run docker build: {e}");
+                false
+            }
+        }
```
NishantBansal2003: The existing code first runs the cmd and checks whether it was able to execute successfully, something like:
```rust
let status = match run_...(...) {
    Ok(status) => status,
    Err(e) => {
        eprintln!("...: failed to run command: {e}");
        return false;
    }
};
```
It then checks the exit status separately:
```rust
if !status.success() {
    eprintln!("...: command failed");
    return false;
}
```
This avoids having multiple match arms for execution and exit-status handling, and also makes the control flow clearer, so it would be better to stick to it for consistency.

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,267 @@
...
+            scenario: scenario.to_string(),
+            coverage: false,
+            image: None,
+            smite_dir: PathBuf::from("/repo/smite"),
+            no_cache: false,
+        }
+    }
+
+    #[test]
+    fn workload_target_display_is_lowercase() {
```
NishantBansal2003: this test seems unnecessary to me

NishantBansal2003 (changes requested):
Great work! Some commands I tested are:

- `smitebot build --target lnd --scenario ir`

- `smitebot build --target ldk --scenario noise --coverage --smite-dir ./smite`

- Using cache: `smitebot build --target ldk --scenario noise --coverage`

- Without cache: `smitebot build --target ldk --scenario noise --coverage --no-cache`

- `smitebot build --target cln --scenario encrypted_bytes --image cln-smitebot` 

Things I wanted to discuss are:
- In `doctor.rs` we use `println!` because the output is intended to be final and, in the JSON case, machine-parsable. However, `smitebot build` is not in the same situation, so I think we should use logging instead of `println!` and `eprintln!`. Using logs would also give us timestamps, which can be useful later for debugging. WDYT?
- For `target`, we already define the set of valid values, and those are also shown in `smitebot help build`. For `scenario`, however, we only accept a raw string. Would it make sense to define a fixed set of valid scenario values as well? The drawback is that every time a new scenario is added, `smitebot` would also need to be updated

Ashish-Kumar-Dash:
> * In `doctor.rs` we use `println!` because the output is intended to be final and, in the JSON case, machine-parsable. However, `smitebot build` is not in the same situation, so I think we should use logging instead of `println!` and `eprintln!`. Using logs would also give us timestamps, which can be useful later for debugging. WDYT?

Yes, I agree on logging for `smitebot build`. Using logs for smitebot’s own status/error messages seems cleaner. Docker output can still stream directly, and timestamps would help when debugging longer builds.

> * For `target`, we already define the set of valid values, and those are also shown in `smitebot help build`. For `scenario`, however, we only accept a raw string. Would it make sense to define a fixed set of valid scenario values as well? The drawback is that every time a new scenario is added, `smitebot` would also need to be updated

tbh I’d prefer keeping it as a raw string for now. Targets are a fixed workload set, but scenarios are more likely to evolve in future, including IR/custom scenarios. Since the Dockerfiles already consume `SCENARIO` dynamically, hardcoding allowed scenarios in smitebot, like you pointed out as well, would add extra maintenance every time a scenario is added.

smitebot/src/commands/build.rs
Ashish-Kumar-Dash: yes got it

smitebot/README.md
```diff
@@ -19,6 +19,23 @@ smitebot doctor --aflpp-path ~/AFLplusplus --json
...
 
+### smitebot build
+
+`smitebot build` builds Smite workload Docker images for manual rebuilds and debugging.
+
+```bash
+smitebot build --target lnd --scenario encrypted_bytes
+smitebot build --target cln --scenario noise --coverage
+smitebot build --target ldk --scenario init --image local/ldk-init:debug --no-cache
+```
```
NishantBansal2003, line 30: Could you add a bit more documentation for the flags? It's not clear what options like `--image` or `--coverage` do without checking the code

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    command.status()
+}
+
+/// Initializes logging for operational build messages.
+fn init_logger() {
+    match simple_logger::init_with_env() {
+        Ok(()) | Err(_) => {}
+    }
+}
```
NishantBansal2003: This should probably go in `main.rs` and be made consistent with the rest of the code: `simple_logger::init_with_env().expect("Failed to initialize logger");`

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,287 @@
...
+
+/// Fully resolved Docker build inputs.
+#[derive(Debug, PartialEq, Eq)]
+struct BuildInputs {
+    /// Docker image tag to produce.
+    image: String,
+    /// Workload Dockerfile selected from `--target` and `--coverage`.
+    dockerfile: PathBuf,
+    /// Docker build context, normally the smite repository root.
+    context: PathBuf,
```
NishantBansal2003: This `context` name is not very clear to me. Could we use a more descriptive name? If it always refers to the smite directory, perhaps the name should reflect that

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// Scenario passed to Docker as `--build-arg SCENARIO=...`.
+    scenario: String,
+    /// Whether Docker should rebuild without using its layer cache.
+    no_cache: bool,
+}
+
+impl BuildInputs {
+    /// Builds a Docker invocation plan from parsed CLI arguments.
```
NishantBansal2003: stale comment

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,287 @@
...
+        Ok(()) | Err(_) => {}
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::path::Path;
+
+    fn build_args(target: WorkloadTarget, scenario: &str) -> BuildArgs {
```
NishantBansal2003: `sample_build_args` ?

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,287 @@
...
+        );
+        assert_eq!(
+            default_workload_image_tag(WorkloadTarget::Cln, "noise", true),
+            "smite-cln-noise-coverage"
+        );
+    }
+
+    #[test]
+    fn default_image_tag_preserves_custom_scenario_names() {
+        assert_eq!(
```
NishantBansal2003: what's the difference between `default_workload_image_tag_matches_smite_convention` and `default_image_tag_preserves_custom_scenario_names`? They look very similar to me

NishantBansal2003 (changes requested):

smitebot/src/commands/build.rs
Ashish-Kumar-Dash: yes, removed the custom-scenarios tag test, wanted to add more tests

smitebot/src/commands/build.rs
Ashish-Kumar-Dash: using smite_dir now

smitebot/README.md
```diff
@@ -19,6 +19,32 @@ smitebot doctor --aflpp-path ~/AFLplusplus --json
...
+
+Flags:
+
+- `--target`: workload implementation to build (`lnd`, `cln`, `ldk`, or `eclair`).
+- `--scenario`: scenario selected by the workload Dockerfile and passed as `SCENARIO`.
+- `--coverage`: use `Dockerfile.coverage` and tag the image with the `-coverage` suffix.
+- `--image`: override the default Docker image tag.
+- `--smite-dir`: path to the Smite repository root; defaults to the current directory.
+- `--no-cache`: pass `--no-cache` to Docker for a clean rebuild.
```
NishantBansal2003: I think the README should be end-user focused and explain things from the project's perspective rather than the implementation details. Explanations that are specific to the code or internal design would be better placed in inline code comments
```suggestion
- `--target`: Workload implementation to build an image for (`lnd`, `cln`, `ldk`, or `eclair`).
- `--scenario`: Scenario that the image should run.
- `--coverage`: Build a coverage-instrumented image.
- `--image`: Use a custom image tag instead of the default tag used by Smite.
- `--smite-dir`: Path to the Smite repository root. Defaults to the current directory.
- `--no-cache`: Perform a clean rebuild without using cached Docker layers.
```

NishantBansal2003 (approved):

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+}
+
+impl BuildCommand {
+    /// Builds the requested Smite Docker image and returns whether Docker succeeded.
+    pub fn execute(args: &BuildArgs) -> bool {
+        let inputs = BuildInputs::from_args(args);
+        if !inputs.dockerfile.exists() {
+            log::error!("Dockerfile not found: {}", inputs.dockerfile.display());
```
morehouse, line 106: Why do we use `log` for the build command but `println!` for the doctor command?

smitebot/src/commands/build.rs
```diff
@@ -0,0 +1,266 @@
...
+
+/// CLI arguments for `smitebot build`.
+#[derive(Debug, Args)]
+pub struct BuildArgs {
+    /// Target implementation to build.
+    #[arg(long)]
+    target: WorkloadTarget,
+    /// Scenario binary selected by the workload Dockerfile.
+    #[arg(long)]
+    scenario: String,
```
morehouse, line 21: It would be great if we could also validate that `--scenario` is a valid value before attempting a ~20 minute build.

Probably the best way is to validate against the binary names in `smite-scenarios/src/bin/`.  (For a future PR).

morehouse (approved):
Good work!

smitebot/src/commands/build.rs
Ashish-Kumar-Dash: Like Nishant pointed out earlier in the thread, In doctor.rs we use println! because the output is intended to be final and, in the JSON case, machine-parsable. However, smitebot build is not in the same situation, so think we should use logging instead of println! and eprintln!. Using logs would also give us timestamps, which can be useful later for debugging.

NishantBansal2003: That’s true in principle, although we’re not actually getting timestamps from logs today since timestamp output is currently disabled

Ashish-Kumar-Dash: Interesting, I hope when we do flip timestamps on the log callsites will serve its purpose. Why's it disabled rn, is it by default?

Issue #98, opened 2026-05-26
Milestone 4: Co-op Channel Closes


ekzyis:
## Fuzz scenario

`PostChannelReadySetup`: We want to fuzz starting with `shutdown` followed by `closing_complete` and `closing_sig`, so we need to have an open, ready channel and `option_simple_close` must have been negotiated during `init` exchange. Therefore, the scenario setup would be a modified or extended version of `PostInitSetup` which includes `option_simple_close`, opening a channel and waiting for `channel_ready`:

```
  Smite                         Target
    |                              |
    |----------- init ------------>|
    |    *option_simple_close*     |
    |<---------- init -------------|
    |    *option_simple_close*     |
    |                              |
    |-------- open_channel ------->|
    |<------ accept_channel -------|
    |------- funding_created ----->|
    |<------ funding_signed -------|
    |                              |
    |  [tx broadcast+mine blocks]  |
    |                              |
    |------- channel_ready ------->|
    |<------ channel_ready --------|
    |                              |
    |---------- ping ------------->|
    |<--------- pong --------------|
    |                              |
    +---------- SNAPSHOT ----------+
```

During fuzzing, a mutual close is negotiated with a `shutdown` exchange. Since there are no pending HTLCs, an exchange of `closing_complete` and `closing_sig` should immediately follow:

```
  Smite                         Target
    +---------- SNAPSHOT ----------+
    |                              |
    |---------- shutdown --------->|
    |<--------- shutdown ----------|
    |                              |
    |------ closing_complete ----->|
    |<----- closing_complete ------|
    |                              |
    |--------- closing_sig ------->|
    |<-------- closing_sig --------|
```

Advanced scenarios could fuzz additional exchanges of `closing_complete`+`closing_sig` messages to bump fees etc.

### Setup

- [ ] `PostChannelReadySetup`: see above

### Messages

* [x] `init`
* [x] (`ping`, `pong`)
* [x] `open_channel`
* [x] `accept_channel`
* [x] `funding_created`
* [x] `funding_signed`
* [x] `channel_ready`
* [x] `shutdown`
* [x] `closing_complete` (https://github.com/morehouse/smite/pull/105)
* [x] `closing_sig` (https://github.com/morehouse/smite/pull/109)

### Operations

For funding flow:

* [x] `BuildOpenChannel`
* [x] `SendOpenChannel`
* [x] `RecvAcceptChannel`
* [x] `CreateFundingTransaction`
[...]

PR #105, opened 2026-05-28
smite: add `closing_complete` codec

ekzyis:
part of #98

### DIFF
--- a/smite/src/bolt/closing_complete.rs
+++ b/smite/src/bolt/closing_complete.rs
@@ -1,0 +1,366 @@
+//! BOLT 2 `closing_complete` message.
+
+use super::BoltError;
+use super::tlv::TlvStream;
+use super::types::ChannelId;
+use super::wire::WireFormat;
+use bitcoin::secp256k1::ecdsa::Signature;
+
+/// TLV type for closer output only.
+const TLV_CLOSER_OUTPUT_ONLY: u64 = 1;
+/// TLV type for closee output only.
+const TLV_CLOSEE_OUTPUT_ONLY: u64 = 2;
+/// TLV type for closer and closee outputs.
+const TLV_CLOSER_AND_CLOSEE_OUTPUTS: u64 = 3;
+
+/// BOLT 2 `closing_complete` message (type 40).
+///
+/// When closing a channel and shutdown is complete, each peer sends
+/// `closing_complete` with the transaction details.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ClosingComplete {
+    /// The ID of the channel to be closed.
+    pub channel_id: ChannelId,
+    /// Output script for the initiator of the co-op close.
+    pub closer_scriptpubkey: Vec<u8>,
+    /// Output script for the non-initiator of the co-op close.
+    pub closee_scriptpubkey: Vec<u8>,
+    /// Suggested absolute fee for the closing tx.
+    pub fee_satoshis: u64,
+    /// Suggested locktime for the closing tx.
+    pub locktime: u32,
+    /// Optional TLV extensions.
+    pub tlvs: ClosingTlvs,
+}
+
+/// TLV extensions for the `closing_complete` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct ClosingTlvs {
+    /// Signature if closing tx only has local output.
+    pub closer_output_only: Option<Signature>,
+    /// Signature if closing tx only has remote output.
+    pub closee_output_only: Option<Signature>,
+    /// Signature if closing tx has local and remote output.
+    pub closer_and_closee_outputs: Option<Signature>,
+}
+
+impl ClosingComplete {
+    /// Encodes to wire format (without message type prefix).
+    #[must_use]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.channel_id.write(&mut out);
+        self.closer_scriptpubkey.write(&mut out);
+        self.closee_scriptpubkey.write(&mut out);
+        self.fee_satoshis.write(&mut out);
+        self.locktime.write(&mut out);
+
+        // Encode TLVs
+        let mut tlv_stream = TlvStream::new();
+        if let Some(closer_output_only) = &self.tlvs.closer_output_only {
+            tlv_stream.add(
+                TLV_CLOSER_OUTPUT_ONLY,
+                closer_output_only.serialize_compact().to_vec(),
+            );
+        }
+        if let Some(closee_output_only) = &self.tlvs.closee_output_only {
+            tlv_stream.add(
+                TLV_CLOSEE_OUTPUT_ONLY,
+                closee_output_only.serialize_compact().to_vec(),
+            );
+        }
+        if let Some(closer_and_closee_outputs) = &self.tlvs.closer_and_closee_outputs {
+            tlv_stream.add(
+                TLV_CLOSER_AND_CLOSEE_OUTPUTS,
+                closer_and_closee_outputs.serialize_compact().to_vec(),
+            );
+        }
+        out.extend(tlv_stream.encode());
+
+        out
+    }
+
+    /// Decodes from wire format (without message type prefix).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if the payload is too short for any fixed field or
+    /// signature TLV, or `InvalidSignature` if a signature TLV is not a valid
+    /// compact ECDSA signature.
+    #[allow(clippy::similar_names)]
+    pub fn decode(payload: &[u8]) -> Result<Self, BoltError> {
+        let mut cursor = payload;
+
+        let channel_id = WireFormat::read(&mut cursor)?;
+        let closer_scriptpubkey = WireFormat::read(&mut cursor)?;
+        let closee_scriptpubkey = WireFormat::read(&mut cursor)?;
+        let fee_satoshis = WireFormat::read(&mut cursor)?;
+        let locktime = WireFormat::read(&mut cursor)?;
+
+        // Decode TLVs (remaining bytes)
+        let tlv_stream = TlvStream::decode_with_known(cursor, &[TLV_CLOSEE_OUTPUT_ONLY])?;
+        let tlvs = ClosingTlvs::from_stream(&tlv_stream)?;
+
+        Ok(Self {
+            channel_id,
+            closer_scriptpubkey,
+            closee_scriptpubkey,
+            fee_satoshis,
+            locktime,
+            tlvs,
+        })
+    }
+}
+
+impl ClosingTlvs {
+    /// Extracts `closing_complete` TLVs from a parsed TLV stream.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Truncated` if a signature TLV has invalid length, or
+    /// `InvalidSignature` if the bytes are not a valid compact ECDSA signature.
+    fn from_stream(stream: &TlvStream) -> Result<Self, BoltError> {
+        Ok(Self {
+            closer_output_only: stream.get_as::<Signature>(TLV_CLOSER_OUTPUT_ONLY)?,
+            closee_output_only: stream.get_as::<Signature>(TLV_CLOSEE_OUTPUT_ONLY)?,
+            closer_and_closee_outputs: stream.get_as::<Signature>(TLV_CLOSER_AND_CLOSEE_OUTPUTS)?,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use bitcoin::secp256k1::ecdsa::Signature;
+
+    use super::super::{CHANNEL_ID_SIZE, COMPACT_SIGNATURE_SIZE};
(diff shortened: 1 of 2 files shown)


smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,343 @@
...
+/// TLV type for closee output only.
+const TLV_CLOSEE_OUTPUT_ONLY: u64 = 2;
+/// TLV type for closer and closee outputs.
+const TLV_CLOSER_AND_CLOSEE_OUTPUTS: u64 = 3;
+
+// BOLT 2 `closing_complete` message (type 40).
+//
+// When closing a channel and shutdown is complete, each peer sends
+// `closing_complete` with the transaction details.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
```
ekzyis: I added the `Default` trait here, instead of a `sample_closing_complete()` function in the tests (bolt.rs + bolt/\<message\>.rs).

**This is inconsistent with the other messages though**.

I wonder if there was a reason to not use the `Default` trait, or if we should also use the `Default` trait for them. We could get rid of a lot of code in the tests. It does not look to me like the specific values picked for the sample messages matter.

Adding a `Default` trait just for less code in the tests might be a bad argument though.

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,343 @@
...
+            ClosingComplete::decode(&[0x11; 20]),
+            Err(BoltError::Truncated {
+                expected: CHANNEL_ID_SIZE,
+                actual: 20
+            })
+        );
+    }
+
+    #[test]
+    fn decode_truncated_closer_scriptpubkey_length_prefix() {
```
ekzyis, line 170: Testing truncated length prefixes is new. Not sure if needed.

smite/src/bolt/closing_complete.rs
morehouse: `Default` values tend to be all zeroes, which means our tests lose some of their power.  Without distinct values for each field, we can't catch any swapped-field or similar bugs.

I would prefer the other approach to maximize the effectiveness of our tests.

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,335 @@
...
+// When closing a channel and shutdown is complete, each peer sends
+// `closing_complete` with the transaction details.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct ClosingComplete {
+    // The ID of the channel to be closed.
+    pub channel_id: ChannelId,
+    // Output script for local ("closer") output.
+    pub closer_scriptpubkey: Vec<u8>,
+    // Output script for remote ("closee") output.
+    pub closee_scriptpubkey: Vec<u8>,
```
morehouse: Note that it's not necessarily true that `local == closer` and `remote == closee`.  It just happens to be true when we are the initiators of the co-op close.


```suggestion
    // Output script for the initiator of the co-op close.
    pub closer_scriptpubkey: Vec<u8>,
    // Output script for the non-initiator of the co-op close.
    pub closee_scriptpubkey: Vec<u8>,
```

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,335 @@
...
+                Signature::read(&mut cursor)
+            })
+            .transpose()?;
+
+        Ok(Self {
+            closer_output_only,
+            closee_output_only,
+            closer_and_closee_outputs,
+        })
+    }
```
morehouse: We should use the new `stream.get_as` function so that trailing bytes cause an error.  It also simplifies the code and allows us to remove the clippy annotation.


```suggestion
    fn from_stream(stream: &TlvStream) -> Result<Self, BoltError> {
        Ok(Self {
            closer_output_only: stream.get_as::<Signature>(TLV_CLOSER_OUTPUT_ONLY)?,
            closee_output_only: stream.get_as::<Signature>(TLV_CLOSEE_OUTPUT_ONLY)?,
            closer_and_closee_outputs: stream.get_as::<Signature>(TLV_CLOSER_AND_CLOSEE_OUTPUTS)?,
        })
    }
```

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,335 @@
...
+        let mut encoded = cc.encode();
+
+        let sig_start = encoded.len() - COMPACT_SIGNATURE_SIZE;
+        encoded[sig_start..].copy_from_slice(&[0xff; COMPACT_SIGNATURE_SIZE]);
+
+        assert!(matches!(
+            ClosingComplete::decode(&encoded),
+            Err(BoltError::InvalidSignature(_))
+        ));
+    }
```
morehouse, line 317: We should add a trailing TLV bytes test.

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,335 @@
...
+use bitcoin::secp256k1::ecdsa::Signature;
+
+/// TLV type for closer output only.
+const TLV_CLOSER_OUTPUT_ONLY: u64 = 1;
+/// TLV type for closee output only.
+const TLV_CLOSEE_OUTPUT_ONLY: u64 = 2;
+/// TLV type for closer and closee outputs.
+const TLV_CLOSER_AND_CLOSEE_OUTPUTS: u64 = 3;
+
+// BOLT 2 `closing_complete` message (type 40).
```
morehouse: We should use `///` for the public structs and all of their fields.

smite/src/bolt/closing_complete.rs
morehouse: Can't hurt.

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,335 @@
...
+            })
+        );
+    }
+
+    #[test]
+    fn decode_truncated_closer_scriptpubkey() {
+        let mut data = vec![0x11; CHANNEL_ID_SIZE];
+        // 10 as length prefix, but only give 5 bytes
+        data.extend_from_slice(&[0x00, 0x0a]);
+        data.extend_from_slice(&[0x22; 0x05]);
```
morehouse: Nit: for lengths, I think it's more readable like this:


```suggestion
        data.extend_from_slice(&[0x22; 5]);
```

Applies to all tests in this file.

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    #[test]
+    fn roundtrip_with_tlvs() {
+        let sig = Signature::from_compact(&[1u8; 64]).expect("valid signature");
+        let original = ClosingComplete {
+            tlvs: ClosingCompleteTlvs {
+                closer_output_only: Some(sig),
+                closee_output_only: Some(sig),
+                closer_and_closee_outputs: Some(sig),
```
morehouse: It would be good to have different signatures for each TLV to be sure the codec doesn't mix them up.

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,335 @@
...
+        assert!(decoded.tlvs.closer_and_closee_outputs.is_none());
+    }
+
+    #[test]
+    fn default_tlvs_are_none() {
+        let tlvs = ClosingCompleteTlvs::default();
+        assert!(tlvs.closer_output_only.is_none());
+        assert!(tlvs.closee_output_only.is_none());
+        assert!(tlvs.closer_and_closee_outputs.is_none());
+    }
```
morehouse, line 298: These two tests check the same thing.

morehouse (changes requested):

smite/src/bolt/closing_complete.rs
ekzyis: done in 2bcb4a9

smite/src/bolt/closing_complete.rs
ekzyis: Ohh, right, good idea! Done in 2bcb4a9

smite/src/bolt/closing_complete.rs
ekzyis: added a trailing TLV bytes test for each field in 2bcb4a9

smite/src/bolt/closing_complete.rs
ekzyis: done in 2bcb4a9

smite/src/bolt/closing_complete.rs
ekzyis: done in 2bcb4a9

smite/src/bolt/closing_complete.rs
ekzyis: done in 2bcb4a9

smite/src/bolt/closing_complete.rs
ekzyis: Kept only `decode_empty_tlv_values` in 2bcb4a9

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,364 @@
...
+
+    /// Valid `ClosingComplete` message for testing.
+    fn sample_closing_complete(tlvs: Option<ClosingCompleteTlvs>) -> ClosingComplete {
+        ClosingComplete {
+            channel_id: ChannelId::new([0xaa; 32]),
+            closer_scriptpubkey: vec![0x00; 22],
+            closee_scriptpubkey: vec![0x00; 22],
+            fee_satoshis: 1000,
+            locktime: 0,
+            tlvs: tlvs.unwrap_or_default(),
```
morehouse, line 146: Let's use distinct non-zero values for the scriptpubkeys and locktime, so we can detect if the fields were accidentally swapped.

Also applies to `bolt.rs`.

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,364 @@
...
+
+/// TLV extensions for the `closing_complete` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct ClosingCompleteTlvs {
+    // Signature if closing tx only has local output.
+    pub closer_output_only: Option<Signature>,
+    // Signature if closing tx only has remote output.
+    pub closee_output_only: Option<Signature>,
+    // Signature if closing tx has local and remote output.
+    pub closer_and_closee_outputs: Option<Signature>,
```
morehouse: Nit:


```suggestion
    /// Signature if closing tx only has local output.
    pub closer_output_only: Option<Signature>,
    /// Signature if closing tx only has remote output.
    pub closee_output_only: Option<Signature>,
    /// Signature if closing tx has local and remote output.
    pub closer_and_closee_outputs: Option<Signature>,
```

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,364 @@
...
+    pub fee_satoshis: u64,
+    /// Suggested locktime for the closing tx.
+    pub locktime: u32,
+    /// Optional TLV extensions.
+    pub tlvs: ClosingCompleteTlvs,
+}
+
+/// TLV extensions for the `closing_complete` message.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct ClosingCompleteTlvs {
```
morehouse: The spec calls these `closing_tlvs` and shares the definition between `closing_complete` and `closing_sig`.  I think it would make sense for us to do the same.


```suggestion
pub struct ClosingTlvs {
```

smite/src/bolt/closing_complete.rs
```diff
@@ -0,0 +1,364 @@
...
+            ClosingComplete::decode(&encoded),
+            Err(BoltError::InvalidSignature(_))
+        ));
+    }
+
+    #[test]
+    fn decode_closer_output_only_reject_trailing_bytes() {
+        let mut encoded = sample_closing_complete(None).encode();
+        // TLV type 1 (`closer_output_only`), len 65 — signatures must be exactly 64 bytes
+        encoded.extend_from_slice(&[0x01, 0x41]);
```
morehouse: Nit: we can make the code self-documenting by using the named constants for these last 3 tests


```suggestion
        encoded.extend_from_slice(&[TLV_CLOSER_OUTPUT_ONLY as u8, (COMPACT_SIZE + 1) as u8]);
```

smite/src/bolt/closing_complete.rs
ekzyis: Sorry for being more distracted. Fixed in d15772c.

smite/src/bolt/closing_complete.rs
ekzyis: right, done in d15772c

smite/src/bolt/closing_complete.rs
ekzyis: done in d15772c

smite/src/bolt/closing_complete.rs
ekzyis: done in d15772c

ekzyis:
Addressed all feedback in d15772c. Will rebase #109 on this.

morehouse (approved):
LGTM

PR #108, opened 2026-05-29
smite: check if TLV length matches length of known encoding

### LINKED ISSUE #107: bolt01: TLV length must equal length of known encoding for type
**Problem**

While considering tests for #105, I noticed that we test truncated TLVs:

https://github.com/morehouse/smite/blob/78167723a2ac66b6b04c23a12dd40f48a66d6d6f/smite/src/bolt/channel_ready.rs#L208-L224

but we don't always assert that the length matches what we would expect for known types (may they be even or odd):

```md
 - if `type` is known:
   - MUST decode the next `length` bytes using the known encoding for `type`.
   - if `length` is not exactly equal to that required for the known encoding for `type`:
     - MUST fail to parse the `tlv_stream`.
```

-- bolt01, [_Type-Length-Value Format_](https://github.com/lightning/bolts/blob/aff897f8f6ac0a7ba071b3b61fede7dc813be0b5/01-messaging.md?plain=1#L169-L172)

This means that in the case of `channel_ready`, this test would currently fail:

```diff
diff --git a/smite/src/bolt/channel_ready.rs b/smite/src/bolt/channel_ready.rs
index a685b29..37242e3 100644
--- a/smite/src/bolt/channel_ready.rs
+++ b/smite/src/bolt/channel_ready.rs
@@ -223,6 +223,24 @@ mod tests {
         );
     }

+    #[test]
+    // Test constants are known to fit in u8
+    #[allow(clippy::cast_possible_truncation)]
+    fn decode_short_channel_id_reject_trailing_bytes() {
+        let msg = sample_channel_ready(None);
+        let mut encoded = msg.encode();
+
+        // short_channel_id TLV should be 8 bytes, but we push 9 bytes
+        encoded.push(TLV_SHORT_CHANNEL_ID as u8);
+        encoded.push(0x09);
+        encoded.extend_from_slice(&[0xbb; 9]);
+
+        assert!(
+            ChannelReady::decode(&encoded).is_err(),
+            "over-length short_channel_id TLV must be rejected"
+        );
+    }
+
     #[test]
     fn default_tlvs_are_none() {
         let tlvs = ChannelReadyTlvs::default();
```

Currently, if provided 9 as the length, we would read 8 bytes and ignore the trailing byte. Reading 8 bytes is correct, ignoring the trailing byte is not.

We only handle this for the init TLVs by checking if the length is a multiple of 32:

https://github.com/morehouse/smite/blob/78167723a2ac66b6b04c23a12dd40f48a66d6d6f/smite/src/bolt/init.rs#L131-L137

IIUC, TLVs include the length so nodes know how much to _skip_ when encountering unknown TLVs:

```md
 - otherwise, if `type` is unknown:
   - if `type` is even:
     - MUST fail to parse the `tlv_stream`.
   - otherwise, if `type` is odd:
     - MUST discard the next `length` bytes.
```

For known TLVs, how much to read should be based on the known encoding.

**Suggested Solution**

We could fix this with a `TlvStream::get_as<T: WireFormat>(&self, tlv_type: u64)` function:

```rust
/// Gets a record by type and decodes it as a fixed-size `WireFormat` value.
///
/// Returns `None` if the record is absent. Rejects TLV values that are
/// longer than the type's wire encoding (BOLT 1: known TLV length must
/// exactly match the encoding size).
///
/// # Errors
///
/// Returns `Truncated` if the value is too short, or `TlvTrailingBytes`
/// if bytes remain after decoding.
pub fn get_as<T: WireFormat>(&self, tlv_type: u64) -> Result<Option<T>, BoltError> {
    self.get(tlv_type)
        .map(|data| {
            let mut cursor = data;
            let value = T::read(&mut cursor)?;
            if !cursor.is_empty() {
                return Err(BoltError::TlvTrailingBytes { tlv_type });
            }
            Ok(value)
        })
        .transpose()
}
```

WDYT? Do you agree with the problem?

ekzyis:
close #107 

This is the suggestion in #107.

### DIFF
--- a/smite/src/bolt/init.rs
+++ b/smite/src/bolt/init.rs
@@ -125,21 +125,9 @@ impl InitTlvs {
     ///
     /// # Errors
     ///
-    /// Returns `Truncated` if the networks TLV has invalid length.
+    /// Returns a `BoltError` if the networks TLV has invalid length.
     fn from_stream(stream: &TlvStream) -> Result<Self, BoltError> {
-        let networks = if let Some(data) = stream.get(TLV_NETWORKS) {
-            let (chunks, remainder) = data.as_chunks::<CHAIN_HASH_SIZE>();
-            if !remainder.is_empty() {
-                return Err(BoltError::Truncated {
-                    expected: (chunks.len() + 1) * CHAIN_HASH_SIZE,
-                    actual: data.len(),
-                });
-            }
-            Some(chunks.to_vec())
-        } else {
-            None
-        };
-
+        let networks = stream.get_as_many::<[u8; CHAIN_HASH_SIZE]>(TLV_NETWORKS)?;
         let remote_addr = stream.get(TLV_REMOTE_ADDR).map(Vec::from);
 
         Ok(Self {
@@ -389,8 +377,9 @@ mod tests {
 
         assert_eq!(
             Init::decode(&data),
-            Err(BoltError::Truncated {
-                expected: CHAIN_HASH_SIZE * 2, // Next multiple of 32
+            Err(BoltError::TlvTrailingBytes {
+                tlv_type: 1,
+                expected: 32,
                 actual: 33
             })
         );
--- a/smite/src/bolt/tx_init_rbf.rs
+++ b/smite/src/bolt/tx_init_rbf.rs
@@ -3,7 +3,7 @@
 use super::BoltError;
 use super::tlv::TlvStream;
 use super::types::ChannelId;
-use super::wire::WireFormat;
+use super::wire::{EmptyTlv, WireFormat};
 
 /// TLV type for funding output contribution.
 const TLV_FUNDING_OUTPUT_CONTRIBUTION: u64 = 0;
@@ -103,11 +103,12 @@ impl TxInitRbfTlvs {
     ///
     /// # Errors
     ///
-    /// Returns a `BoltError` if `funding_output_contribution` has invalid
-    /// length.
+    /// Returns a `BoltError` if a TLV value has invalid length.
     fn from_stream(stream: &TlvStream) -> Result<Self, BoltError> {
         let funding_output_contribution = stream.get_as::<i64>(TLV_FUNDING_OUTPUT_CONTRIBUTION)?;
-        let require_confirmed_inputs = stream.get(TLV_REQUIRE_CONFIRMED_INPUTS).is_some();
+        let require_confirmed_inputs = stream
+            .get_as::<EmptyTlv>(TLV_REQUIRE_CONFIRMED_INPUTS)?
+            .is_some();
 
         Ok(Self {
             funding_output_contribution,
@@ -206,6 +207,21 @@ mod tests {
             TxInitRbf::decode(&encoded),
             Err(BoltError::TlvUnknownEvenType(4))
         ));
+    }
+
+    #[test]
+    fn decode_reject_require_confirmed_inputs_trailing_bytes() {
+        let mut encoded = sample_msg().encode();
+        // type 2 (`require_confirmed_inputs`), len 1 — must be zero-length
+        encoded.extend_from_slice(&[0x02, 0x01, 0xff]);
+        assert_eq!(
+            TxInitRbf::decode(&encoded),
+            Err(BoltError::TlvTrailingBytes {
+                tlv_type: TLV_REQUIRE_CONFIRMED_INPUTS,
+                expected: 0,
+                actual: 1,
+            })
+        );
     }
 
     #[test]
--- a/smite/src/bolt/tx_init_rbf.rs
+++ b/smite/src/bolt/tx_init_rbf.rs
@@ -103,16 +103,10 @@ impl TxInitRbfTlvs {
     ///
     /// # Errors
     ///
-    /// Returns `Truncated` if `funding_output_contribution` has invalid length.
+    /// Returns a `BoltError` if `funding_output_contribution` has invalid
+    /// length.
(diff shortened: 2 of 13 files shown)


smite/src/bolt/init.rs
```diff
@@ -125,21 +125,9 @@ impl InitTlvs {
...
-                    expected: (chunks.len() + 1) * CHAIN_HASH_SIZE,
-                    actual: data.len(),
-                });
-            }
-            Some(chunks.to_vec())
-        } else {
-            None
-        };
-
+        let networks = stream.get_as_many::<[u8; 32]>(TLV_NETWORKS)?;
```
morehouse:
```suggestion
        let networks = stream.get_as_many::<[u8; CHAIN_HASH_SIZE]>(TLV_NETWORKS)?;
```

smite/src/bolt/wire.rs
```diff
@@ -985,4 +1004,42 @@ mod tests {
...
+                actual: 100
+            })
+        );
+    }
+
+    #[test]
+    fn attribution_data_write_roundtrip() {
+        let original = AttributionData {
+            htlc_hold_times: [42; 20],
+            truncated_hmacs: [TruncatedHmac([0xcc; 4]); 210],
```
morehouse: We should use the named constants for the field lengths.

smite/src/bolt/wire.rs
```diff
@@ -229,10 +230,28 @@ impl WireFormat for sha256::Hash {
     }
 }
 
+impl WireFormat for AttributionData {
+    fn read(data: &mut &[u8]) -> Result<Self, BoltError> {
+        let value = Self::decode(data)?;
```
morehouse: Generally `WireFormat::read` should not return an error if there are trailing bytes, but that can happen here due to a `decode` error.


```suggestion
        if data.len() < Self::SIZE {
            return Err(Truncated { ... });
        }
        let value = Self::decode(&data[..Self::SIZE])?;
```

smite/src/bolt/wire.rs
```diff
@@ -229,10 +230,28 @@ impl WireFormat for sha256::Hash {
     }
 }
 
+impl WireFormat for AttributionData {
```
morehouse: Let's move the `WireFormat` to `attribution_data.rs`.

smite/src/bolt/wire.rs
```diff
@@ -985,4 +1004,42 @@ mod tests {
         );
         assert_eq!(data.len(), 5); // 5 bytes remaining
     }
+
+    #[test]
+    fn attribution_data_read_truncated() {
```
morehouse: We should also add a test that trailing bytes are ignored by `WireFormat::read` (`get_as` is supposed to handle trailing bytes instead.

smite/src/bolt/wire.rs
```diff
@@ -229,10 +230,28 @@ impl WireFormat for sha256::Hash {
...
 }
 
+impl WireFormat for AttributionData {
+    fn read(data: &mut &[u8]) -> Result<Self, BoltError> {
+        let value = Self::decode(data)?;
+        *data = &data[Self::SIZE..];
+        Ok(value)
+    }
+
+    fn write(&self, out: &mut Vec<u8>) {
```
morehouse: Now that we're implementing `write`, the logic from `encode` is duplicated.

I think it would make sense to call `write` from `encode` to deduplicate.  Also we could do something similar with calling `read` from `decode`.

morehouse (changes requested):
Concept ACK.  It would also be good to do this for the zero-length bool TLVs -- currently we don't detect any trailing bytes for those.

smite/src/bolt/init.rs
ekzyis: fixed in abc901a

smite/src/bolt/wire.rs
ekzyis: fixed in 66aa815

smite/src/bolt/wire.rs
ekzyis: done in 66aa815

smite/src/bolt/wire.rs
ekzyis: done in 66aa815

smite/src/bolt/wire.rs
ekzyis: `encode` calls `write` and `decode` calls `read` in 66aa815

smite/src/bolt/wire.rs
ekzyis: `read` doesn't call `Self::decode` anymore in 66aa815 because of https://github.com/morehouse/smite/pull/108#discussion_r3364343954

ekzyis:
> It would also be good to do this for the zero-length bool TLVs -- currently we don't detect any trailing bytes for those.

Added this in bc8f7e7, a separate commit, because it adds a new type `EmptyTlv`

smite/src/bolt/tx_init_rbf.rs
```diff
@@ -214,6 +209,21 @@ mod tests {
         ));
     }
 
+    #[test]
+    fn decode_reject_require_confirmed_inputs_trailing_bytes() {
```
morehouse, line 213: Nit: let's add this test to the other 3 codecs that use `EmptyTlv` for consistency

morehouse (approved):

smite/src/bolt/tx_init_rbf.rs
ekzyis: done in 6dcdc9c

morehouse (approved):
LGTM

PR #99, opened 2026-05-26
smite-ir: add `LoadFee` operation

devvaansh:
Fees are needed for BOLT 7 `channel_update` messages (`fee_base_msat` and `fee_proportional_millionths`).

A fee parameter carries a u32 millisatoshi or millionths value, distinct from the existing `Amount` type which holds u64 satoshis.

Part of #71.

PR #110, opened 2026-05-31
smite-ir: implement `BuildChannelUpdate` operation

devvaansh:
Enables programs that build signed `channel_update` messages and send
 them to the target. The signature is computed internally over the double-SHA256 of the message body, using the supplied node secret key (per BOLT 7).  All other wire fields are passed as variable inputs so they can be reused
and mutated by the existing IR machinery.

Ref: #71

### DIFF
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -136,6 +136,25 @@ pub enum Operation {
         /// 32-byte node alias, zero-padded.
         alias: [u8; 32],
     },
+    /// Build a `channel_update` message (BOLT 7, type 258).
+    ///
+    /// The signature is computed internally over the double-SHA256 of the
+    /// message body following the signature field, using the supplied node
+    /// secret key (per BOLT 7).
+    ///
+    /// Inputs (11):
+    ///   0: `node_sk` (`PrivateKey`) -- signs the body
+    ///   1: `chain_hash` (`ChainHash`)
+    ///   2: `short_channel_id` (`ShortChannelId`)
+    ///   3: `timestamp` (`Timestamp`)
+    ///   4: `message_flags` (`U8`)
+    ///   5: `channel_flags` (`U8`)
+    ///   6: `cltv_expiry_delta` (`U16`)
+    ///   7: `htlc_minimum_msat` (`Amount`)
+    ///   8: `fee_base_msat` (`ForwardingFee`)
+    ///   9: `fee_proportional_millionths` (`ForwardingFee`)
+    ///  10: `htlc_maximum_msat` (`Amount`)
+    BuildChannelUpdate,
 
     // -- Act: side effects against the target --
     /// Send an encoded message over the connection.
@@ -597,6 +616,7 @@ impl fmt::Display for Operation {
                 format_hex(rgb_color),
                 format_hex(alias),
             ),
+            Self::BuildChannelUpdate => write!(f, "BuildChannelUpdate"),
             Self::SendMessage => write!(f, "SendMessage"),
             Self::SendOpenChannel => write!(f, "SendOpenChannel"),
         }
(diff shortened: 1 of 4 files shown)


devvaansh:
Fee fields now use `ForwardingFee` type (post-#99).

ekzyis (commented):
The `ShortChannelId` changes look unrelated. Should they be in 4b04a9d4c71de5ba409e3c2de81085c41c608010, or in a different commit?

devvaansh:
> The ShortChannelId changes look unrelated. Should they be in https://github.com/morehouse/smite/commit/4b04a9d4c71de5ba409e3c2de81085c41c608010, or in a different commit?

 I initially bundled those SCID changes because I was refactoring nearby code while working on BuildChannelUpdate, and it felt like a natural cleanup,I've dropped them from this commit. If we still want to simplify tweak_short_channel_id → tweak_u64, I'll open it as a separate PR so it can be reviewed on its own merits. Force-pushed to 326a60a.

smite-ir/src/tests.rs
```diff
@@ -321,6 +321,89 @@ fn display_build_node_announcement_program() {
...
+            inputs: vec![],
+        },
+        Instruction {
+            operation: Operation::LoadAmount(1_000),
+            inputs: vec![],
+        },
+        Instruction {
+            operation: Operation::LoadAmount(100),
+            inputs: vec![],
+        },
```
morehouse, line 376: `fee_base_msat` and `fee_proportional_millionths` are supposed to be of type `ForwardingFee`, not `Amount`.

smite-scenarios/src/executor.rs
```diff
@@ -1178,6 +1232,107 @@ mod tests {
         assert!(na.verify());
     }
 
+    #[test]
+    fn execute_build_channel_update() {
+        let mut sk_bytes = [0u8; 32];
+        sk_bytes[31] = 0x42;
+        let chain_hash = [0xcc; 32];
+        let scid_packed: u64 = 0x0837_A400_034D_0001;
```
morehouse: We should just create a true scid object here instead.

smite-scenarios/src/executor.rs
```diff
@@ -1178,6 +1232,107 @@ mod tests {
...
+                operation: Operation::SendMessage,
+                inputs: vec![11],
+            },
+        ];
+
+        let program = Program {
+            instructions: instrs,
+        };
+        let mut ctx = sample_context();
+        ctx.chain_hash = chain_hash;
```
morehouse: Nit: the other tests do not override the chain hash and just use `sample_context()` inline where needed

morehouse (commented):
Needs a rebase

smite-scenarios/src/executor.rs
devvaansh: constructed The SCID via `ShortChannelId::new(538_532, 845, 1)`  and round-tripped through `as_u64()` at the load site; the equality check now compares `ShortChannelId` values directly.

smite-ir/src/operation.rs
```diff
@@ -785,6 +820,7 @@ impl Operation {
             | Self::DerivePoint
             | Self::ExtractAcceptChannel(_)
             | Self::BuildOpenChannel
+            | Self::BuildChannelUpdate
```
morehouse: Nit: `BuildChannelUpdate` comes after `BuildNodeAnnouncement` in the enum order

morehouse (approved):

smite-ir/src/operation.rs
devvaansh: `BuildChannelUpdate` now follows `BuildNodeAnnouncement` in the `has_side_effects` arm to match the enum order.

morehouse (approved):
LGTM

PR #113, opened 2026-06-02
smite-ir-mutator: warn on missing environment variables

morehouse:
When our custom mutator library is used, we need to ensure that AFL++ uses *only* our custom mutators and nothing else, or else false positives can occur.  To accomplish this, multiple other environment variables also need to be set, which we now warn about if they're missing.

Unfortunately there's no good way to detect if we fail to set `AFL_CUSTOM_MUTATOR_LIBRARY`, since then AFL++ never actually loads our library at all.

### DIFF
--- a/smite-ir-mutator/src/lib.rs
+++ b/smite-ir-mutator/src/lib.rs
@@ -120,6 +120,35 @@ fn decode_and_validate(bytes: &[u8]) -> Option<Program> {
 
 // -- AFL++ custom mutator ABI -------------------------------------------------
 
+/// Warns about any missing AFL++ environment variables required by this
+/// mutator.
+#[cfg(not(test))]
+fn warn_on_unset_afl_env() {
+    let unset: Vec<&str> = [
+        "AFL_CUSTOM_MUTATOR_ONLY",
+        "AFL_FRAMESHIFT_DISABLE",
+        "AFL_DISABLE_TRIM",
+    ]
+    .into_iter()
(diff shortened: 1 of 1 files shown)


smite-ir-mutator/src/lib.rs
```diff
@@ -120,6 +120,35 @@ fn decode_and_validate(bytes: &[u8]) -> Option<Program> {
...
+    let bar = "=".repeat(72);
+    eprintln!("\n{bar}");
+    eprintln!("[smite-ir-mutator] WARNING: required AFL++ env vars are not set to 1:");
+    for name in &unset {
+        eprintln!("  - {name}");
+    }
+    eprintln!("Fuzzing will be inefficient and may produce false-positive crashes.");
+    eprintln!("{bar}\n");
+
+    std::thread::sleep(std::time::Duration::from_secs(10));
```
ekzyis, line 149: nit: Unfortunately, afl-fuzz is unresponsive during these 10 seconds, so we can't exit with CTRL+C

ekzyis (approved):
LGTM, tested and works. Noticed one nit, but can be ignored

smite-ir-mutator/src/lib.rs
morehouse: *Maybe* we can get around this by invoking `sleep` directly through libc, but I'm not sure it's worth the dependency.  I expect smitebot will soon remove our need for this warning entirely, since all the necessary env vars will get set automatically for us.

PR #112, opened 2026-06-01
smite-ir: implement `ChannelAnnouncementGenerator`

morehouse:
Generates programs that send correctly-signed `channel_announcement` messages to the target.

Verified that all 4 targets successfully decode the `channel_announcement`s we send and pass initial validation, including signature checks.  Once we can actually mine fake funding transactions, we will be able to exercise even deeper code paths.

[ldk-coverage.tar.gz](https://github.com/user-attachments/files/28481561/ldk-coverage.tar.gz)
[lnd-coverage.tar.gz](https://github.com/user-attachments/files/28481565/lnd-coverage.tar.gz)
[cln-coverage.tar.gz](https://github.com/user-attachments/files/28481575/cln-coverage.tar.gz)
[eclair-coverage.tar.gz](https://github.com/user-attachments/files/28481583/eclair-coverage.tar.gz)

Ref: #71

PR #114, opened 2026-06-02
smite-ir: implement `ChannelUpdateGenerator`

devvaansh:
Generates programs that build and send correctly-signed `channel_update` messages to the target.

The signing key, chain hash, and short channel id are drawn via `pick_variable` so they can be shared with other gossip messages (e.g. a preceding `channel_announcement`) when generators are composed in the future.

`message_flags` and `channel_flags` are exposed as plain `U8` variables so that OperationParamMutator can flip individual bits, including must_be_one, direction, and disable.

Ref: #71

### DIFF
--- a/smite-ir/src/tests.rs
+++ b/smite-ir/src/tests.rs
@@ -6,7 +6,10 @@ use rand::rngs::SmallRng;
 use smite::bolt::{MAX_MESSAGE_SIZE, ShortChannelId};
 
 use super::*;
-use generators::{ChannelAnnouncementGenerator, NodeAnnouncementGenerator, OpenChannelGenerator};
+use generators::{
+    ChannelAnnouncementGenerator, ChannelUpdateGenerator, NodeAnnouncementGenerator,
+    OpenChannelGenerator,
+};
 use minimizers::{CommonSubexpressionEliminator, DeadCodeEliminator, Minimizer};
 use mutators::{InputSwapMutator, OperationParamMutator};
 use operation::{AcceptChannelField, ChannelTypeVariant, ShutdownScriptVariant};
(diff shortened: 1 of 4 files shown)


devvaansh:
@morehouse Should I verify these against lnd/cln/eclair/ldk before you review? I can attach coverage reports like #112.

smite-ir/src/tests.rs
```diff
@@ -1006,6 +1009,54 @@ fn generated_node_announcement_program_postcard_roundtrip() {
...
+    assert_eq!(build_count, 1, "expected exactly one BuildChannelUpdate");
+}
+
+#[test]
+fn generated_channel_update_program_postcard_roundtrip() {
+    let program = generate_channel_update_program(42);
+    let bytes = postcard::to_allocvec(&program).expect("postcard serialization");
+    let decoded: Program = postcard::from_bytes(&bytes).expect("postcard deserialization");
+    assert_eq!(program, decoded);
+}
```
morehouse, line 1026: Nit: let's put this test next to the other `*_postcard_roundtrip` tests.

smite-ir/src/tests.rs
```diff
@@ -1006,6 +1009,54 @@ fn generated_node_announcement_program_postcard_roundtrip() {
...
+
+#[test]
+fn validate_accepts_generated_channel_update_program() {
+    for seed in 0..100 {
+        let program = generate_channel_update_program(seed);
+        assert_well_formed(&program);
+    }
+}
```
morehouse, line 1050: This test can be removed.  It tests the exact same thing as `generated_channel_update_program_is_type_correct`.

smite-ir/src/tests.rs
```diff
@@ -1006,6 +1009,54 @@ fn generated_node_announcement_program_postcard_roundtrip() {
     assert_eq!(program, decoded);
 }
 
+fn generate_channel_update_program(seed: u64) -> Program {
```
morehouse, line 988: Nit: let's move the `generated_channel_update_*` tests to just after the `generated_node_announcement_*` tests.

morehouse:
> @morehouse Should I verify these against lnd/cln/eclair/ldk before you review? I can attach coverage reports like #112.

Yes, it would be good for you to fuzz and check the coverage.  I will do the same before merging.

smite-ir/src/tests.rs
devvaansh: done

smite-ir/src/tests.rs
devvaansh: done

smite-ir/src/tests.rs
devvaansh: done

morehouse (commented):
Code LGTM

morehouse (approved):
Fuzzed with this generator overnight.  Coverage looks good -- all targets are able to decode our `channel_update`s, though we never get to signature verification since we never send a matching `channel_announcement` first.

[ldk-coverage.tar.gz](https://github.com/user-attachments/files/28890716/ldk-coverage.tar.gz)
[lnd-coverage.tar.gz](https://github.com/user-attachments/files/28890726/lnd-coverage.tar.gz)
[cln-coverage.tar.gz](https://github.com/user-attachments/files/28890733/cln-coverage.tar.gz)
[eclair-coverage.tar.gz](https://github.com/user-attachments/files/28890737/eclair-coverage.tar.gz)

PR #116, opened 2026-06-04
smitebot: add campaign configuration

Ashish-Kumar-Dash:
## Summary
 
 Adds the campaign configuration layer that `start`/`stop`/`status` will consume. Addresses the sample config request.
 
  - `smitebot/src/config.rs` — `CampaignConfig` struct, `Target` enum, TOML parsing with `deny_unknown_fields`, validation (runners ≥ 1, non-empty scenario)
  - `smitebot/src/commands/config.rs` — `smitebot config <path>` subcommand to validate configs before launching campaigns
  - `smitebot/examples/lnd-encrypted-bytes.toml` — annotated sample config
  - corresponding README updates for `smitebot config` command
 
 Ref. #70

### DIFF
--- a/smitebot/README.md
+++ b/smitebot/README.md
@@ -17,7 +17,33 @@ smitebot doctor --aflpp-path ~/AFLplusplus
 smitebot doctor --aflpp-path ~/AFLplusplus --json
 ```
 
+## Configuration
+
+Campaign settings are stored in a TOML file. See [`sample-campaign.toml`](sample-campaign.toml) for a complete example.
+
+| Field        | Required | Description                                                          |
+| ------------ | -------- | -------------------------------------------------------------------- |
+| `target`     | yes      | Lightning implementation to fuzz (`lnd`, `cln`, `ldk`, or `eclair`). |
+| `scenario`   | yes      | Scenario binary selected by the workload Dockerfile.                 |
+| `aflpp_path` | yes      | Path to the AFL++ source tree.                                       |
+| `smite_dir`  | yes      | Path to the smite repository root.                                   |
+| `runners`    | yes      | Number of parallel AFL++ instances to launch (must be at least 1).   |
+| `seed_dir`   | no       | Directory containing seed inputs; omit to start from an empty corpus.|
+| `output_dir` | yes      | AFL++ output directory for findings and stats.                       |
+| `sharedir`   | yes      | Nyx shared directory path; created automatically by `smitebot start`.|
+| `image`      | no       | Docker image tag override; defaults to `smite-<target>-<scenario>`.  |
+| `afl_env`    | no       | Extra environment variables passed to AFL++ instances.               |
+| `afl_flags`  | no       | Extra CLI flags appended to `afl-fuzz`.                              |
+
 ## Commands
+
+### smitebot config
+
+`smitebot config` validates a campaign configuration file, reports the resolved settings, and checks that referenced paths exist on disk.
+
+```bash
+smitebot config sample-campaign.toml
+```
 
 ### smitebot build
 
--- a/smitebot/sample-campaign.toml
+++ b/smitebot/sample-campaign.toml
@@ -1,0 +1,35 @@
+# Sample smitebot campaign configuration for fuzzing LND with the encrypted_bytes scenario on 8 cores.
+
+# Target Lightning implementation (lnd, ldk, cln, eclair).
+target = "lnd"
+
+# Fuzzing scenario.
+scenario = "encrypted_bytes"
+
+# Path to AFL++ source tree.
+aflpp_path = "/home/user/AFLplusplus"
+
+# Path to smite repository root.
+smite_dir = "."
+
+# Number of parallel AFL++ instances to launch.
+runners = 8
+
+# Seed corpus directory (optional; omit to start from an empty corpus).
+seed_dir = "/tmp/smite-seeds"
+
+# AFL++ output directory.
+output_dir = "/tmp/smite-out"
+
+# Nyx shared directory path; created automatically by `smitebot start`.
+sharedir = "/tmp/smite-nyx"
+
+# Docker image tag override (default: smite-<target>-<scenario>).
+# image = "smite-lnd-encrypted_bytes"
+
+# Extra environment variables passed to AFL++ instances.
+# [afl_env]
+# AFL_KEEP_TIMEOUTS = "1"
+
+# Extra CLI flags appended to afl-fuzz.
+# afl_flags = ["-t", "1000+"]
--- a/smitebot/src/commands/config.rs
+++ b/smitebot/src/commands/config.rs
@@ -1,0 +1,169 @@
+//! Campaign configuration validation.
+
+use std::collections::HashMap;
+use std::path::PathBuf;
+
+use clap::Args;
+use serde::Serialize;
+
+use crate::config::CampaignConfig;
+
+/// Command handler for `smitebot config`.
+pub struct ConfigCommand;
+
+/// CLI arguments for `smitebot config`.
+#[derive(Debug, Args)]
+pub struct ConfigArgs {
+    /// Path to the campaign configuration TOML file.
+    path: PathBuf,
+    /// Emit machine-readable JSON output.
+    #[arg(long)]
+    json: bool,
+}
+
+/// JSON-serializable report for `smitebot config --json`.
+#[derive(Debug, Serialize)]
+struct ConfigReport {
+    target: String,
+    scenario: String,
+    aflpp_path: String,
+    smite_dir: String,
+    runners: u16,
+    seed_dir: Option<String>,
+    output_dir: String,
+    sharedir: String,
+    image: String,
+    afl_env: HashMap<String, String>,
+    afl_flags: Vec<String>,
+    errors: Vec<String>,
+    valid: bool,
+}
+
+/// JSON-serializable error report when config fails to load.
+#[derive(Debug, Serialize)]
+struct ConfigErrorReport {
+    error: String,
+    valid: bool,
+}
+
+impl ConfigCommand {
+    /// Validates a campaign configuration file and reports the result.
+    pub fn execute(args: &ConfigArgs) -> bool {
+        match CampaignConfig::load(&args.path) {
+            Ok(config) => {
+                let errors = config.check_paths();
+                let valid = errors.is_empty();
+
+                if args.json {
+                    let report = ConfigReport {
+                        target: config.target.to_string(),
+                        scenario: config.scenario.clone(),
+                        aflpp_path: config.aflpp_path.display().to_string(),
+                        smite_dir: config.smite_dir.display().to_string(),
+                        runners: config.runners,
+                        seed_dir: config.seed_dir.as_ref().map(|p| p.display().to_string()),
+                        output_dir: config.output_dir.display().to_string(),
+                        sharedir: config.sharedir.display().to_string(),
+                        image: config.image_tag(),
+                        afl_env: config.afl_env.clone(),
+                        afl_flags: config.afl_flags.clone(),
+                        errors,
+                        valid,
+                    };
+                    let json = serde_json::to_string_pretty(&report)
+                        .expect("ConfigReport is always serializable");
+                    println!("{json}");
+                } else {
+                    println!("target:     {}", config.target);
+                    println!("scenario:   {}", config.scenario);
+                    println!("aflpp_path: {}", config.aflpp_path.display());
+                    println!("smite_dir:  {}", config.smite_dir.display());
+                    println!("runners:    {}", config.runners);
+                    match &config.seed_dir {
+                        Some(dir) => println!("seed_dir:   {}", dir.display()),
+                        None => println!("seed_dir:   (empty corpus)"),
+                    }
+                    println!("output_dir: {}", config.output_dir.display());
+                    println!("sharedir:   {}", config.sharedir.display());
+                    println!("image:      {}", config.image_tag());
+                    for (key, val) in &config.afl_env {
+                        println!("afl_env:    {key}={val}");
+                    }
+                    for flag in &config.afl_flags {
+                        println!("afl_flags:  {flag}");
+                    }
+                    for error in &errors {
+                        eprintln!("error: {error}");
+                    }
+                }
+                valid
+            }
+            Err(e) => {
+                if args.json {
+                    let report = ConfigErrorReport {
+                        error: e.to_string(),
+                        valid: false,
+                    };
+                    let json = serde_json::to_string_pretty(&report)
+                        .expect("ConfigErrorReport is always serializable");
+                    println!("{json}");
+                } else {
+                    eprintln!("error: {e}");
+                }
+                false
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn config_report_json_is_machine_readable() {
+        let report = ConfigReport {
+            target: "lnd".to_string(),
+            scenario: "encrypted_bytes".to_string(),
+            aflpp_path: "/home/user/AFLplusplus".to_string(),
+            smite_dir: ".".to_string(),
+            runners: 8,
+            seed_dir: Some("/tmp/seeds".to_string()),
+            output_dir: "/tmp/out".to_string(),
+            sharedir: "/tmp/nyx".to_string(),
+            image: "smite-lnd-encrypted_bytes".to_string(),
+            afl_env: HashMap::from([("AFL_KEEP_TIMEOUTS".to_string(), "1".to_string())]),
+            afl_flags: vec!["-t".to_string(), "1000+".to_string()],
+            errors: vec!["aflpp_path does not exist: /home/user/AFLplusplus".to_string()],
+            valid: false,
+        };
+
+        let json = serde_json::to_string(&report).unwrap();
+        let parsed: serde_json::Value = serde_json::from_str(&json).unwrap();
+
+        assert_eq!(parsed["target"], "lnd");
+        assert_eq!(parsed["runners"], 8);
+        assert_eq!(parsed["valid"], false);
+        assert_eq!(parsed["seed_dir"], "/tmp/seeds");
+        assert_eq!(parsed["afl_env"]["AFL_KEEP_TIMEOUTS"], "1");
+        assert_eq!(parsed["afl_flags"][0], "-t");
+        assert_eq!(
+            parsed["errors"][0],
+            "aflpp_path does not exist: /home/user/AFLplusplus"
+        );
+    }
+
+    #[test]
+    fn config_error_report_json_is_machine_readable() {
+        let report = ConfigErrorReport {
+            error: "failed to parse config.toml".to_string(),
+            valid: false,
+        };
+
+        let json = serde_json::to_string(&report).unwrap();
+        let parsed: serde_json::Value = serde_json::from_str(&json).unwrap();
+
+        assert_eq!(parsed["valid"], false);
+        assert!(parsed["error"].is_string());
+    }
+}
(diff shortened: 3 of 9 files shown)


Ashish-Kumar-Dash:
ping @NishantBansal2003

smitebot/sample-campaign.toml
```diff
@@ -0,0 +1,36 @@
+# Sample smitebot campaign configuration for fuzzing LND with the
```
NishantBansal2003: I’d prefer this config file to be generic, containing defaults and examples (see: https://github.com/go-continuous-fuzz/go-continuous-fuzz/blob/main/sample-go-continuous-fuzz.conf), and to be placed alongside the README.md file.

smitebot/src/commands/config.rs
```diff
@@ -0,0 +1,47 @@
+//! Campaign configuration validation.
```
NishantBansal2003, line 1: Was `smitebot config` cmd part of the implementation plan? I don’t see it mentioned anywhere. It also feels odd that I need to run `smitebot config` before `smitebot start` just to validate the config. Ideally, validation should happen when `smitebot start` is run. Perhaps we should remove the config command and perform validation inline in start/status

smitebot/README.md
```diff
@@ -17,8 +17,34 @@ smitebot doctor --aflpp-path ~/AFLplusplus
...
+| `scenario` | yes | Scenario binary selected by the workload Dockerfile. |
+| `aflpp_path` | yes | Path to the AFL++ source tree. |
+| `smite_dir` | yes | Path to the smite repository root. |
+| `runners` | yes | Number of parallel AFL++ instances to launch (must be at least 1). |
+| `seed_dir` | yes | Directory containing seed inputs for the fuzzing corpus. |
+| `output_dir` | yes | AFL++ output directory for findings and stats. |
+| `sharedir` | yes | Nyx shared directory created by `setup-nyx.sh`. |
+| `image` | no | Docker image tag override; defaults to `smite-<target>-<scenario>`. |
+| `afl_env` | no | Extra environment variables passed to AFL++ instances. |
+| `afl_flags` | no | Extra CLI flags appended to `afl-fuzz`. |
```
NishantBansal2003:
```suggestion
| Field        | Required | Description                                                          |
| ------------ | -------- | -------------------------------------------------------------------- |
| `target`     | yes      | Lightning implementation to fuzz (`lnd`, `cln`, `ldk`, or `eclair`). |
| `scenario`   | yes      | Scenario binary selected by the workload Dockerfile.                 |
| `aflpp_path` | yes      | Path to the AFL++ source tree.                                       |
| `smite_dir`  | yes      | Path to the smite repository root.                                   |
| `runners`    | yes      | Number of parallel AFL++ instances to launch (must be at least 1).   |
| `seed_dir`   | yes      | Directory containing seed inputs for the fuzzing corpus.             |
| `output_dir` | yes      | AFL++ output directory for findings and stats.                       |
| `sharedir`   | yes      | Nyx shared directory created by `setup-nyx.sh`.                      |
| `image`      | no       | Docker image tag override; defaults to `smite-<target>-<scenario>`.  |
| `afl_env`    | no       | Extra environment variables passed to AFL++ instances.               |
| `afl_flags`  | no       | Extra CLI flags appended to `afl-fuzz`.                              |
```

smitebot/README.md
```diff
@@ -17,8 +17,34 @@ smitebot doctor --aflpp-path ~/AFLplusplus
...
 ## Commands
 
+### smitebot config
+
+`smitebot config` validates a campaign configuration file and reports the resolved settings.
+
+```bash
+smitebot config examples/lnd-encrypted-bytes.toml
+```
+
```
NishantBansal2003, line 47: Should be removed based on https://github.com/morehouse/smite/pull/116#discussion_r3368888593

smitebot/src/config.rs
```diff
@@ -0,0 +1,301 @@
...
+    pub afl_env: HashMap<String, String>,
+    /// Extra CLI flags appended to the `afl-fuzz` command.
+    #[serde(default)]
+    pub afl_flags: Vec<String>,
+}
+
+/// Lightning Network implementation to target.
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+pub enum Target {
```
NishantBansal2003, line 44: This is duplicated with `WorkloadTarget` in `build.rs`. Perhaps we should define a shared `Target` type and use it in both places

smitebot/src/config.rs
```diff
@@ -0,0 +1,301 @@
...
+        }
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ConfigError {
+    #[error("failed to read {}: {source}", path.display())]
+    Read {
+        path: PathBuf,
+        #[source]
```
NishantBansal2003:
```suggestion
```

smitebot/src/config.rs
```diff
@@ -0,0 +1,301 @@
...
+        source: std::io::Error,
+    },
+    #[error("failed to parse {}: {source}", path.display())]
+    Parse {
+        path: PathBuf,
+        #[source]
```
NishantBansal2003:
```suggestion
```

smitebot/src/config.rs
```diff
@@ -0,0 +1,301 @@
...
+        config.validate()?;
+        Ok(config)
+    }
+
+    /// Returns the Docker image tag, using the Smite convention as the default.
+    #[must_use]
+    pub fn image_tag(&self) -> String {
+        self.image
+            .clone()
+            .unwrap_or_else(|| format!("smite-{}-{}", self.target, self.scenario))
```
NishantBansal2003, line 106: This is very similar to `default_workload_image_tag`, so perhaps we should reuse that here

smitebot/src/config.rs
```diff
@@ -0,0 +1,301 @@
...
+            source,
+        })?;
+        config.validate()?;
+        Ok(config)
+    }
+
+    /// Returns the Docker image tag, using the Smite convention as the default.
+    #[must_use]
+    pub fn image_tag(&self) -> String {
+        self.image
```
NishantBansal2003, line 104: We should also error on an empty image tag, just as we do for an empty scenario

NishantBansal2003 (changes requested):

smitebot/sample-campaign.toml
Ashish-Kumar-Dash: Moved the sample toml alongside README, however I'd prefer keeping it as a concrete LND example though, something you can copy, paste and actually run is more useful than a template with placeholders imo. Users just swap target/scenario to match their setup..... Smitebot's config is 9 fields only, so the current level of documentation feels right for me without adding Default/Example blocks for every field. Happy to hear if there's a specific field you think needs more context though!

smitebot/src/commands/config.rs
Ashish-Kumar-Dash: Fair, but without it, the entire config module is dead code and the build fails (warnings = "deny"). start doesn't exist yet so nothing else uses these types. It's also handy on its own: you can validate your config before committing to a 20-min campaign startup,  same idea as docker compose config. Once start lands it'll call  CampaignConfig::load internally too, but the standalone validate is still useful for CI/scripting. Hopefully you get my point!

smitebot/README.md
Ashish-Kumar-Dash: Open to discussion

smitebot/src/config.rs
Ashish-Kumar-Dash: #[source] is what gives you Error::source() for error chaining, without it callers get None when trying to introspect the underlying cause. doctor.rs also uses the same pattern on its CheckFailure::Io variant. Removing it would break meaningful error inspection. We should keep it ig, its harmless and allows debugging

smitebot/src/commands/config.rs
Ashish-Kumar-Dash: Also, its an *optional* command for users to verify their TOML before launching a campaign, start will validate internally too

smitebot/src/config.rs
Ashish-Kumar-Dash: They share the same format string but default_workload_image_tag takes a coverage flag that doesn't apply in config context. Now that both use the shared Target type, I could call it here with `coverage: false`, but that's weird cause the config has no concept of coverage! Happy to extract a shared helper if you feel strongly, but for I'd prefer to keep them independent!

smitebot/src/config.rs
NishantBansal2003: `#[source]` is only required when the field name is not source, otherwise, it is optional. Please check: https://docs.rs/thiserror/latest/thiserror
So either change the field name to `error`, or simply remove the `#[source]` attribute

Ashish-Kumar-Dash: Got it, removed the attribute, thanks for the heads up!

smitebot/src/commands/config.rs
NishantBansal2003: Config error would already surface when we run start, so I don’t think we need to wait for any campaign startup. We also already have `smitebot build` for debugging scenario and image tags.
For the warning, we can add a `dead_code` attribute for now. Though, I don’t have a strong opinion here, I’m fine with it if @morehouse  finds it useful

morehouse: As long as the underlying logic can be reused for `smitebot config`, `smitebot start`, etc., I think it's fine to have a separate `smitebot config` command for validating configs.

I can imagine the command being useful in situations where we want to dry-run a config file without actually running it.

smitebot/src/config.rs
```diff
@@ -0,0 +1,315 @@
...
+    #[test]
+    fn load_accepts_optional_image_override() {
+        let dir = tempfile::tempdir().unwrap();
+        let content = format!("{VALID_CONFIG}image = \"custom:tag\"\n");
+        let path = write_config(dir.path(), &content);
+
+        let config = CampaignConfig::load(&path).unwrap();
+
+        assert_eq!(config.image.as_deref(), Some("custom:tag"));
+    }
```
NishantBansal2003: this test looks exactly the same as `image_tag_uses_override_when_present` below

smitebot/src/config.rs
```diff
@@ -0,0 +1,315 @@
...
+        assert_eq!(config.image_tag(), "my-image:latest");
+    }
+
+    #[test]
+    fn target_display_matches_config_values() {
+        assert_eq!(Target::Lnd.to_string(), "lnd");
+        assert_eq!(Target::Cln.to_string(), "cln");
+        assert_eq!(Target::Ldk.to_string(), "ldk");
+        assert_eq!(Target::Eclair.to_string(), "eclair");
+    }
```
NishantBansal2003: Why was this test added again? I remember we agreed to remove it in the last PR

NishantBansal2003 (commented):
Looks fine to me, open discussion point: https://github.com/morehouse/smite/pull/116#discussion_r3368886837

ready for @morehouse  review

smitebot/src/config.rs
Ashish-Kumar-Dash: I thought this would be useful for 'config' in particular to verify the values, I have removed it now

smitebot/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// Scenario binary selected by the workload Dockerfile.
+    pub scenario: String,
+    /// Path to AFL++ source tree.
+    pub aflpp_path: PathBuf,
+    /// Path to the smite repository root.
+    pub smite_dir: PathBuf,
+    /// Number of parallel AFL++ instances to launch.
+    pub runners: u16,
+    /// Directory containing seed inputs for the fuzzing corpus.
+    pub seed_dir: PathBuf,
```
morehouse: I frequently start from an empty corpus.  It may be helpful to make `seed_dir` optional and automatically create a minimal seed tempdir when it is missing.

smitebot/src/config.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// AFL++ output directory for findings and stats.
+    pub output_dir: PathBuf,
+    /// Nyx shared directory created by `setup-nyx.sh`.
+    pub sharedir: PathBuf,
```
morehouse: The comment makes it sound like the user has to create the sharedir manually with `setup-nyx.sh`, but it should be done automatically by smitebot.

smitebot/README.md
```diff
@@ -17,8 +17,34 @@ smitebot doctor --aflpp-path ~/AFLplusplus
...
+| `sharedir`   | yes      | Nyx shared directory created by `setup-nyx.sh`.                      |
+| `image`      | no       | Docker image tag override; defaults to `smite-<target>-<scenario>`.  |
+| `afl_env`    | no       | Extra environment variables passed to AFL++ instances.               |
+| `afl_flags`  | no       | Extra CLI flags appended to `afl-fuzz`.                              |
+
 ## Commands
 
+### smitebot config
+
+`smitebot config` is an optional validation command that checks a campaign configuration file and reports the resolved settings. Useful for catching errors before launching a campaign.
```
morehouse: If a goal is to catch errors early, perhaps we also check that the specified paths exist.

smitebot/src/commands/config.rs
```diff
@@ -0,0 +1,47 @@
...
+/// CLI arguments for `smitebot config`.
+#[derive(Debug, Args)]
+pub struct ConfigArgs {
+    /// Path to the campaign configuration TOML file.
+    path: PathBuf,
+}
+
+impl ConfigCommand {
+    /// Validates a campaign configuration file and reports the result.
+    pub fn execute(args: &ConfigArgs) -> bool {
```
morehouse, line 51: `smitebot doctor` seems similar to `smitebot config`, yet the former has a `--json` flag and the latter does not.  What is the reasoning here?

smitebot/README.md
```diff
@@ -17,8 +17,34 @@ smitebot doctor --aflpp-path ~/AFLplusplus
 smitebot doctor --aflpp-path ~/AFLplusplus --json
 ```
 
+## Configuration
```
morehouse, line 20: Now that we have config files, it seems natural to lean on them for the other existing commands, rather than manually specifying the various flags.

e.g.,

```bash
smitebot doctor campaign.toml
smitebot build campaign.toml
```

or

```bash
smitebot doctor --config campaign.toml
smitebot build --config campaign.toml
```

This would be a good follow-up PR.

smitebot/src/commands/config.rs
Ashish-Kumar-Dash: No specific reasoning, I'll add it here too

smitebot/src/commands/config.rs
```diff
@@ -0,0 +1,118 @@
...
+                    warnings.push(format!(
+                        "smite_dir does not exist: {}",
+                        config.smite_dir.display()
+                    ));
+                }
+                if let Some(dir) = &config.seed_dir
+                    && !dir.exists()
+                {
+                    warnings.push(format!("seed_dir does not exist: {}", dir.display()));
+                }
```
morehouse: If `seed_dir` is present, we should also check that it is non-empty.  AFL++ will fail to start if the seed directory is empty.

smitebot/src/commands/config.rs
```diff
@@ -14,29 +16,98 @@ pub struct ConfigCommand;
...
+                        None => log::info!("seed_dir:   (empty corpus)"),
+                    }
+                    log::info!("output_dir: {}", config.output_dir.display());
+                    log::info!("sharedir:   {}", config.sharedir.display());
+                    log::info!("image:      {}", config.image_tag());
+                    for (key, val) in &config.afl_env {
+                        log::info!("afl_env:    {key}={val}");
+                    }
+                    for flag in &config.afl_flags {
+                        log::info!("afl_flag:   {flag}");
```
morehouse: I think for the `config` command we definitely want `println!` rather than using a logger.

smitebot/src/commands/config.rs
```diff
@@ -14,29 +16,98 @@ pub struct ConfigCommand;
...
+    aflpp_path: String,
+    smite_dir: String,
+    runners: u16,
+    seed_dir: Option<String>,
+    output_dir: String,
+    sharedir: String,
+    image: String,
+    afl_env: HashMap<String, String>,
+    afl_flags: Vec<String>,
+    warnings: Vec<String>,
```
morehouse: These are errors, not warnings because they cause `valid` to be false.

smitebot/src/commands/config.rs
```diff
@@ -14,29 +16,98 @@ pub struct ConfigCommand;
...
+                    }
+                    for warning in &warnings {
+                        log::warn!("{warning}");
+                    }
                 }
-                true
+                valid
             }
             Err(e) => {
                 log::error!("{e}");
```
morehouse: If we're going to support JSON output, we need to support it on errors as well.

smitebot/src/commands/config.rs
```diff
@@ -0,0 +1,118 @@
...
+                    ));
+                }
+                if !config.smite_dir.exists() {
+                    warnings.push(format!(
+                        "smite_dir does not exist: {}",
+                        config.smite_dir.display()
+                    ));
+                }
+                if let Some(dir) = &config.seed_dir
+                    && !dir.exists()
```
morehouse: These checks will also be needed for `smitebot start`.  We should put them in a `CampaignConfig::check_paths` method or similar.

This would also enable us to test that logic more easily.

smitebot/src/config.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    #[test]
+    fn check_paths_reports_missing_seed_dir() {
+        let dir = tempfile::tempdir().unwrap();
+        let path = write_config(dir.path(), VALID_CONFIG);
+        let config = CampaignConfig::load(&path).unwrap();
+
+        let errors = config.check_paths();
+        assert!(errors.iter().any(|e| e.contains("seed_dir")));
```
morehouse, line 351: This test actually fails on my machine since `/tmp/smite-seeds` is present on my machine.

We should really be using a tempdir for all checked paths and written files.

smitebot/src/config.rs
```diff
@@ -0,0 +1,427 @@
...
+        if self.image.as_ref().is_some_and(String::is_empty) {
+            return Err(ConfigError::EmptyImage);
+        }
+        if self
+            .seed_dir
+            .as_ref()
+            .is_some_and(|p| p.as_os_str().is_empty())
+        {
+            return Err(ConfigError::EmptySeedDir);
+        }
```
morehouse: Is there a reason that we disallow empty-string `seed_dir` but allow empty strings for all the other `PathBuf`s?

smitebot/src/config.rs
```diff
@@ -0,0 +1,427 @@
...
+        let path = write_config(dir.path(), &content);
+
+        let err = CampaignConfig::load(&path).unwrap_err();
+        assert!(matches!(err, ConfigError::InvalidRunners));
+    }
+
+    #[test]
+    fn load_rejects_empty_scenario() {
+        let dir = tempfile::tempdir().unwrap();
+        let content = VALID_CONFIG.replace(r#"scenario = "encrypted_bytes""#, r#"scenario = """#);
```
morehouse: Nit: for these short strings I find it more readable to escape the quotes instead


```suggestion
        let content = VALID_CONFIG.replace("scenario = \"encrypted_bytes\"", "scenario = \"\"");
```

smitebot/src/config.rs
Ashish-Kumar-Dash: fixed, using tempdir paths now

smitebot/src/config.rs
Ashish-Kumar-Dash: not really, removed it

morehouse (approved):
LGTM

smitebot/README.md
Ashish-Kumar-Dash: I will raise the corresponding PR after `smitebot start` lands, since there can be merge conflicts as both are editing build.rs

Issue #111, opened 2026-06-01
Milestone 3: HTLC and Commitment Operations Implementation Plan


NishantBansal2003:
## Objective

After the funding flow, we need a scenario for fuzzing normal channel operations. So the main objective of this plan is that we can complete the commitment dance fully, and the design should be extensible enough to support future extensions. That way, once we have more advanced mutators, cases like multiple HTLC adds or many commitment dance operations can happen correctly, helping test deeper code paths and edge cases.

### BOLT messages

- [x] `update_fulfill_htlc`
- [x] `update_fail_htlc`
- [x] `update_fail_malformed_htlc`
- [x] `update_add_htlc`
- [x] `commitment_signed`
- [x] `revoke_and_ack`
- [ ] `update_fee`

### Commitment sign/verify APIs

Currently, smite only supports the initial funding flow, so we need to extend it with HTLC support. The main changes are:

- Add `Htlc` struct and in-flight HTLC tracking to `CommitmentState`.
- Extend channel/holder config with HTLC basepoints and signing keys.
- Extend commitment signing/verification to include HTLC second-stage signatures.
- Introduce a `PendingUpdate` enum covering the commitment-affecting updates (`AddHtlc(Htlc)`, `FulfillHtlc(id)`, `FailHtlc(id)`, `UpdateFee(feerate_per_kw)`).
- Update commitment advancement (`CommitmentState::advance(updates)`) and fee accounting to handle HTLC outputs.

## Variables and VariableType

| `VariableType`       | Payload                     |
| -------------------- | --------------------------- |
| `HtlcId`             | `u64`                       |
| `PaymentHash`        | `[u8; 32]`                  |
| `OnionRoutingPacket` | `[u8; 1366]`                |
| `Htlc`               | `smite::channel_tx::Htlc`   |
| `RevokeAndAck`       | `smite::bolt::RevokeAndAck` |

## Operations

- `CreateHtlc`: We need this operation so that we can tie the same HTLC to `BuildUpdateAddHtlc` and `AdvanceCommitmentStateAddHtlc`. Otherwise, there is a chance that we send one HTLC but use a different HTLC in the commitment state, which would only result in invalid signatures. So, if we want to test that mismatch case, we can omit this operation, otherwise, it makes sense to keep it.
  - Input types: `HtlcId`, `Amount`,`PaymentHash`, `BlockHeight`
  - Output type: `Htlc`

- `BuildUpdateAddHtlc`:
[...]

PR #117, opened 2026-06-05
smite-scenarios: convert execute fn into Executor struct

NishantBansal2003:
ref: https://github.com/morehouse/smite/issues/111#issuecomment-4622410995

Move execution state into an `Executor` struct that owns the resources used during program execution. This prepares for future changes that will add channel state to the executor, maintaining a `ChannelId -> ChannelState` mapping so IR operations can read and mutate channel-specific state throughout execution.

Now, during the funding flow, we can add a `ChannelState` field that is populated during `BuildFundingCreated` and later used to verify the signature received in `RecvFundingSigned` (See: https://github.com/morehouse/smite/pull/103#issuecomment-4625362278).

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -114,167 +114,200 @@ pub enum ExecuteError {
     InsufficientFunds(#[from] smite::channel_tx::InsufficientFunds),
 }
 
-/// Executes an IR program against a target over the given connection.
-///
-/// # Errors
-///
-/// Returns an error on a connection/send/receive failure, a decode failure of
-/// a received message, an unexpected message type from the target, or when
-/// wallet funds are insufficient to perform a channel operation.
-///
-/// # Panics
-///
-/// Panics on any invariant violation of the program:
-/// - input count does not match the operation's expected input count
-/// - input variable index out of bounds
-/// - input variable refers to a void instruction
-/// - input variable has the wrong type
-/// - `MineBlocks(0)` (panics inside `BitcoinCli::mine_blocks`)
-/// - `LoadShutdownScript(AnySegwit { .. })` with an out-of-range version or
-///   program length (panics inside the encoder)
-/// - `LoadBytes` / `LoadFeatures` payload exceeding `MAX_MESSAGE_SIZE` (panics
-///   inside the encoder)
-/// - `LoadPrivateKey` whose bytes are all-zero or >= the secp256k1 curve
-///   order (probability ~2^-128 for uniform random input)
-#[allow(clippy::too_many_lines)]
-pub fn execute(
-    program: &Program,
-    context: &ProgramContext,
-    conn: &mut impl Connection,
-    bitcoin_cli: &mut impl BitcoinRpc,
-    start: std::time::Instant,
-) -> Result<(), ExecuteError> {
-    let secp = Secp256k1::new();
-    let mut variables: Vec<Option<Variable>> = Vec::with_capacity(program.instructions.len());
+/// Executes IR programs against a target over an established connection.
+pub struct Executor<C, B> {
+    /// Connection used to send and receive Lightning messages.
+    conn: C,
+    /// Interface to bitcoind for wallet and chain operations.
+    bitcoin_cli: B,
+    /// Immutable state captured during snapshot setup.
+    context: ProgramContext,
+}
 
-    for instr in &program.instructions {
-        let expected_count = instr.operation.input_types().len();
-        assert_eq!(
-            instr.inputs.len(),
-            expected_count,
-            "{:?}: expected {expected_count} inputs, got {}",
-            instr.operation,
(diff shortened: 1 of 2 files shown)


smite-scenarios/src/scenarios/ir.rs
```diff
@@ -35,6 +35,7 @@ impl<T: Target, S: SnapshotSetup<T>> Scenario for IrScenario<T, S> {
 
     fn run(&mut self, input: &[u8]) -> ScenarioResult {
         let start = std::time::Instant::now();
+        let mut bitcoin_cli = self.target.bitcoin_cli().clone();
```
ekzyis: nit: could move this to where it's needed?

ekzyis (approved):
LGTM

nit: commit message mentions "owns the resources", but I think "borrow" is more accurate

ekzyis (approved):
ACK dc5480c

CI failure looks unrelated, should work on force-push:

> [build (eclair)](https://github.com/morehouse/smite/actions/runs/27187924509/job/80260995626#step:3:139)
ERROR: Error response from daemon: Get "https://registry-1.docker.io/v2/": context deadline exceeded

smite-scenarios/src/executor.rs
```diff
@@ -460,10 +485,10 @@ fn consume_sent_open_channel(variables: &mut [Option<Variable>], index: usize) {
 
 /// Create a funding transaction by querying the bitcoind for UTXOs and a
 /// change address, then calling [`build_funding_transaction`].
-fn create_funding_transaction(
+fn create_funding_transaction<B: BitcoinRpc>(
     variables: &[Option<Variable>],
     inputs: &[usize],
-    cli: &mut impl BitcoinRpc,
+    cli: &mut B,
```
morehouse: Nit: unnecessary change

smite-scenarios/src/executor.rs
```diff
@@ -613,7 +638,7 @@ fn recv_non_ping(conn: &mut impl Connection) -> Result<Message, ExecuteError> {
 }
 
 /// Receives and decodes an `accept_channel` message.
-fn recv_accept_channel(conn: &mut impl Connection) -> Result<AcceptChannel, ExecuteError> {
+fn recv_accept_channel<C: Connection>(conn: &mut C) -> Result<AcceptChannel, ExecuteError> {
```
morehouse: Nit: unnecessary change

smite-scenarios/src/executor.rs
```diff
@@ -114,161 +114,186 @@ pub enum ExecuteError {
...
-            }
+/// Executes IR programs against a target over an established connection.
+pub struct Executor<'a, C, B> {
+    /// Connection used to send and receive Lightning messages.
+    conn: &'a mut C,
+    /// Interface to bitcoind for wallet and chain operations.
+    bitcoin_cli: &'a mut B,
+    /// Immutable state captured during snapshot setup.
+    context: &'a ProgramContext,
+}
```
morehouse: I think we can simplify and slightly optimize by making the `Executor` own these fields.  We'd then create and save the executor in `Scenario::new` so that it doesn't need to be recreated every iteration.  And we wouldn't need lifetimes anymore.

I think the only other change needed would be a way to do `ping_pong` on the `conn` owned by the executor.  We could add a tiny `Executor::sync` method that does that for us.

smite-scenarios/src/executor.rs
NishantBansal2003: Yeah, I did that just to keep it consistent with how the caller defines it, but I don't have a strong opinion on it, so I will revert it

smite-scenarios/src/executor.rs
NishantBansal2003: done in https://github.com/morehouse/smite/pull/117/commits/dcc237df466b1ab1d3e39228a439605b6427cf05, 

There is a bit of duplicated code around `Executor::sync` and `ExecuteError::is_timeout`, which I kept intentionally. I felt the original code was correct as-is and shouldn't be modified, so I added similar logic in the executor instead

smite-scenarios/src/executor.rs
```diff
@@ -114,161 +114,227 @@ pub enum ExecuteError {
...
+    /// matching pong, ignoring other messages.
+    ///
+    /// This ensures the target has done initial processing of any previously
+    /// sent message before we check if it's still alive.
+    ///
+    /// # Errors
+    ///
+    /// Returns the underlying [`ExecuteError`] if the connection is closed or
+    /// times out.
+    pub fn sync(&mut self) -> Result<(), ExecuteError> {
```
morehouse: Perhaps it would be cleaner to add a `conn_mut()` accessor so that we can keep using the existing `Scenario::ping_pong` and `ScenarioError::is_timeout` methods.  Then we can eliminate the duplication here.

morehouse (approved):
LGTM

PR #119, opened 2026-06-08
smitebot: add `smitebot start` command

Ashish-Kumar-Dash:
## Summary

  - Adds `smitebot start` command that orchestrates a full fuzzing campaign- builds the Docker image, prepares the Nyx sharedir, spawns parallel `afl-fuzz` processes in the background, and persists campaign state for future `stop`/`status` commands.
  - Adds `CampaignState` persistence module with atomic JSON writes to `~/.smitebot/runs/<campaign-id>/state.json`.
  - Refactors `run_docker_build` to `pub(crate)` for reuse by `start`.
  - Adds `Serialize` to `Target` for state serialization.

  ### Key context behind implementation

  - **Background mode**: runners are orphaned via `process_group(0)` + `drop(child)` so campaigns outlive the terminal. `stop` will use persisted PIDs to `killpg` later.
  - **Nyx parallel mode (`-Y`)**: AFL++ requires numeric runner names — `-M 0`  for primary, `-S N` for secondaries.
  - **IR auto-injection**: scenarios starting with `ir` automatically get  `AFL_CUSTOM_MUTATOR_LIBRARY`, `AFL_CUSTOM_MUTATOR_ONLY`,  `AFL_FRAMESHIFT_DISABLE`, `AFL_DISABLE_TRIM` - set before `afl_env` so user
  overrides take precedence.
  - **Seed dir**: when `seed_dir` is omitted, creates `<output_dir>/.seeds/seed0` with a null byte (AFL++ requires `-i` pointing to a non-empty directory).
  - **Startup verification**: polls for `fuzzer_stats` files with a 5 minute timeout to confirm runners are alive.
  
  Ref. #70

### DIFF
--- a/smitebot/README.md
+++ b/smitebot/README.md
@@ -37,6 +37,18 @@ Campaign settings are stored in a TOML file. See [`sample-campaign.toml`](sample
 
 ## Commands
 
+### smitebot start
+
+`smitebot start` launches a fuzzing campaign in the background. It builds the Docker image, sets up the Nyx sharedir, and spawns parallel AFL++ instances.
+
+```bash
+smitebot start campaign.toml
+```
+
+For IR scenarios (scenario names starting with `ir`), the required AFL++ custom mutator environment variables are injected automatically.
+
+Campaign state is saved to `~/.smitebot/runs/<campaign-id>/state.json` for use by future `stop` and `status` commands.
+
 ### smitebot config
 
 `smitebot config` validates a campaign configuration file, reports the resolved settings, and checks that referenced paths exist on disk.
--- a/smitebot/src/commands/build.rs
+++ b/smitebot/src/commands/build.rs
@@ -6,7 +6,7 @@ use std::process::{Command, ExitStatus};
 
 use clap::Args;
 
-use crate::config::Target;
+use crate::config::{CampaignConfig, Target};
 
 /// Command handler for `smitebot build`.
 pub struct BuildCommand;
@@ -36,20 +36,35 @@ pub struct BuildArgs {
 
 /// Fully resolved Docker build inputs.
 #[derive(Debug, PartialEq, Eq)]
-struct BuildInputs {
+pub struct BuildInputs {
     /// Docker image tag to produce.
-    image: String,
+    pub image: String,
     /// Workload Dockerfile selected from `--target` and `--coverage`.
-    dockerfile: PathBuf,
+    pub dockerfile: PathBuf,
     /// Smite repository root used as the Docker build context.
-    smite_dir: PathBuf,
+    pub smite_dir: PathBuf,
     /// Scenario passed to Docker as `--build-arg SCENARIO=...`.
-    scenario: String,
+    pub scenario: String,
     /// Whether Docker should rebuild without using its layer cache.
-    no_cache: bool,
+    pub no_cache: bool,
 }
 
 impl BuildInputs {
+    /// Resolves Docker build inputs from a campaign configuration.
+    pub fn from_config(config: &CampaignConfig, image: &str) -> Self {
+        Self {
+            image: image.to_string(),
+            dockerfile: config
+                .smite_dir
+                .join("workloads")
+                .join(config.target.to_string())
+                .join("Dockerfile"),
+            smite_dir: config.smite_dir.clone(),
+            scenario: config.scenario.clone(),
+            no_cache: false,
+        }
+    }
+
     /// Resolves Docker build inputs from parsed CLI arguments.
     fn from_args(args: &BuildArgs) -> Self {
         let dockerfile_name = if args.coverage {
@@ -78,33 +93,39 @@ impl BuildCommand {
     /// Builds the requested Smite Docker image and returns whether Docker succeeded.
     pub fn execute(args: &BuildArgs) -> bool {
         let inputs = BuildInputs::from_args(args);
-        if !inputs.dockerfile.exists() {
-            log::error!("Dockerfile not found: {}", inputs.dockerfile.display());
-            return false;
-        }
-
         log::info!(
             "building {} with {}",
             inputs.image,
             inputs.dockerfile.display()
         );
+        run_build(&inputs)
+    }
+}
 
-        let status = match run_docker_build(&inputs) {
-            Ok(status) => status,
-            Err(e) => {
-                log::error!("failed to run docker build: {e}");
-                return false;
-            }
-        };
+/// Checks that the Dockerfile exists, runs `docker build`, and reports success.
+///
+/// Shared by `smitebot build` and `smitebot start`.
+pub fn run_build(inputs: &BuildInputs) -> bool {
+    if !inputs.dockerfile.exists() {
+        log::error!("Dockerfile not found: {}", inputs.dockerfile.display());
+        return false;
+    }
 
-        if !status.success() {
-            log::error!("docker build failed with {status}");
+    let status = match run_docker_build(inputs) {
+        Ok(status) => status,
+        Err(e) => {
+            log::error!("failed to run docker build: {e}");
             return false;
         }
+    };
 
-        log::info!("built {}", inputs.image);
-        true
+    if !status.success() {
+        log::error!("docker build failed with {status}");
+        return false;
     }
+
+    log::info!("built {}", inputs.image);
+    true
 }
 
 /// Returns the default image tag used by Smite's manual Docker build flow.
--- a/smitebot/src/commands/start.rs
+++ b/smitebot/src/commands/start.rs
@@ -1,0 +1,631 @@
+//! Campaign launch orchestration.
+//!
+//! Builds the Docker image, prepares the Nyx sharedir, spawns parallel
+//! `afl-fuzz` processes in the background, and persists campaign state so
+//! that `stop` and `status` can manage the running campaign later.
+
+use std::fs;
+use std::os::unix::process::CommandExt;
+use std::path::{Path, PathBuf};
+use std::process::{Child, Command};
+use std::thread;
+use std::time::{Duration, Instant};
+
+use clap::Args;
+
+use crate::commands::build::{BuildInputs, run_build};
+use crate::config::CampaignConfig;
+use crate::state::{CampaignState, RunnerState, Status};
+
+/// Maximum time to wait for all runners to report non-zero `execs_per_sec`.
+const STARTUP_TIMEOUT: Duration = Duration::from_mins(5);
+
+/// How often to poll `fuzzer_stats` files during startup verification.
+const STARTUP_POLL_INTERVAL: Duration = Duration::from_secs(5);
+
+/// Command handler for `smitebot start`.
+pub struct StartCommand;
+
+/// CLI arguments for `smitebot start`.
+#[derive(Debug, Args)]
+pub struct StartArgs {
+    /// Path to the campaign configuration TOML file.
+    path: PathBuf,
+}
+
+impl StartCommand {
+    /// Launches a fuzzing campaign from the given configuration.
+    pub fn execute(args: &StartArgs) -> bool {
+        let config = match CampaignConfig::load(&args.path) {
+            Ok(c) => c,
+            Err(e) => {
+                log::error!("{e}");
+                return false;
+            }
+        };
+
+        let path_errors = config.check_paths();
+        if !path_errors.is_empty() {
+            for err in &path_errors {
+                log::error!("{err}");
+            }
+            return false;
+        }
+
+        let image = config.image_tag();
+
+        log::info!("building Docker image {image}");
+        let inputs = BuildInputs::from_config(&config, &image);
+        if !run_build(&inputs) {
+            return false;
+        }
+
+        if config.scenario.starts_with("ir") && !build_ir_mutator(&config.smite_dir) {
+            return false;
+        }
+
+        log::info!("setting up Nyx sharedir at {}", config.sharedir.display());
+        if !setup_nyx(&config, &image) {
+            return false;
+        }
+
+        let seed_dir = match ensure_seed_dir(&config) {
+            Ok(dir) => dir,
+            Err(e) => {
+                log::error!("failed to prepare seed directory: {e}");
+                return false;
+            }
+        };
+
+        let campaign_id = config.campaign_id();
+        let Some(runs_dir) = CampaignState::runs_dir() else {
+            log::error!("unable to determine home directory");
+            return false;
+        };
+
+        let state_path = runs_dir.join(&campaign_id).join("state.json");
+
+        let Some(git_hash) = smite_git_hash(&config.smite_dir) else {
+            log::error!("could not determine smite git hash");
+            return false;
+        };
+        let Some(image_digest) = docker_image_id(&image) else {
+            log::error!("could not determine Docker image digest for {image}");
+            return false;
+        };
+
+        let mut state = CampaignState::new(campaign_id, &config, image, image_digest, git_hash);
+
+        if let Err(e) = state.save(&state_path) {
+            log::error!("{e}");
+            return false;
+        }
+
+        if !launch_runners(&config, &seed_dir, &mut state, &state_path) {
+            return false;
+        }
+        if let Err(e) = state.save(&state_path) {
+            log::error!("{e}");
+            return false;
+        }
+
+        log::info!("campaign {} is running", state.id);
+        log::info!("state saved to {}", state_path.display());
+        true
+    }
+}
+
+/// Runs `scripts/setup-nyx.sh` to prepare the Nyx sharedir.
+fn setup_nyx(config: &CampaignConfig, image: &str) -> bool {
+    let script = config.smite_dir.join("scripts").join("setup-nyx.sh");
+    if !script.exists() {
+        log::error!("setup-nyx.sh not found: {}", script.display());
+        return false;
+    }
+
+    let status = match Command::new(&script)
+        .arg(&config.sharedir)
+        .arg(image)
+        .arg(&config.aflpp_path)
+        .status()
+    {
+        Ok(status) => status,
+        Err(e) => {
+            log::error!("failed to run setup-nyx.sh: {e}");
+            return false;
+        }
+    };
+
+    if !status.success() {
+        log::error!("setup-nyx.sh failed with {status}");
+        return false;
+    }
+
+    log::info!("Nyx sharedir ready at {}", config.sharedir.display());
+    true
+}
+
+/// Spawns all runners, verifies they reach non-zero `execs_per_sec`, and updates
+/// campaign state. Runner PIDs are persisted so `smitebot stop` can terminate them.
+fn launch_runners(
+    config: &CampaignConfig,
+    seed_dir: &Path,
+    state: &mut CampaignState,
+    state_path: &Path,
+) -> bool {
+    log::info!("starting {} runners", config.runners);
+
+    let mut runners = Vec::new();
+    let mut children = Vec::new();
+    for id in 0..config.runners {
+        match spawn_runner(config, id, seed_dir) {
+            Ok((runner, child)) => {
+                log::info!("spawned runner {id} (pid {})", runner.pid);
+                runners.push(runner);
+                children.push(child);
+            }
+            Err(e) => {
+                log::error!("failed to spawn runner {id}: {e}");
+                kill_runners(&runners);
+                state.status = Status::Failed;
+                state.runners = runners;
+                if let Err(e) = state.save(state_path) {
+                    log::warn!("failed to save state: {e}");
+                }
+                return false;
+            }
+        }
+    }
+
+    state.runners = runners;
+
+    // Persist PIDs before the lengthy verify_startup poll so smitebot stop
+    // can find the runners if we crash during verification.
+    if let Err(e) = state.save(state_path) {
+        log::warn!("failed to save state: {e}");
+    }
+
+    log::info!("waiting for runners to initialize");
+    if !verify_startup(&config.output_dir, &state.runners, &mut children) {
+        log::error!("not all runners started within the timeout");
+        kill_runners(&state.runners);
+        state.status = Status::Failed;
+        if let Err(e) = state.save(state_path) {
+            log::warn!("failed to save state: {e}");
+        }
+        return false;
+    }
+
+    state.status = Status::Running;
+    true
+}
+
+/// Ensures a seed directory exists for AFL++, creating a minimal corpus if needed.
+///
+/// When the user omits `seed_dir` from the config, AFL++ still requires `-i`
+/// pointing to a non-empty directory.
+fn ensure_seed_dir(config: &CampaignConfig) -> std::io::Result<PathBuf> {
+    if let Some(dir) = &config.seed_dir {
+        return Ok(dir.clone());
+    }
+    let seed_dir = config.output_dir.join(".seeds");
+    fs::create_dir_all(&seed_dir)?;
+    if seed_dir.read_dir()?.next().is_none() {
+        fs::write(seed_dir.join("seed0"), b"\x00")?;
+    }
+    Ok(seed_dir)
+}
+
+/// Spawns a single `afl-fuzz` process in its own process group.
+///
+/// The child is intentionally orphaned so the campaign outlives the `smitebot`
+/// process. The `stop` command uses the persisted PID to terminate it later.
+fn spawn_runner(
+    config: &CampaignConfig,
+    id: u16,
+    seed_dir: &Path,
+) -> std::io::Result<(RunnerState, Child)> {
+    let afl_fuzz = config.aflpp_path.join("afl-fuzz");
+    let name = id.to_string();
+
+    let mut cmd = Command::new(&afl_fuzz);
+    cmd.arg("-Y");
+    cmd.arg("-i").arg(seed_dir);
+    cmd.arg("-o").arg(&config.output_dir);
+
+    // Runner 0 is the primary (-M), all others are secondaries (-S).
+    if id == 0 {
+        cmd.arg("-M").arg(&name);
+    } else {
+        cmd.arg("-S").arg(&name);
+    }
+
+    for flag in &config.afl_flags {
+        cmd.arg(flag);
+    }
+
+    cmd.arg("--").arg(&config.sharedir);
+
+    // IR mutator defaults first so user afl_env can override them.
+    for (key, val) in ir_mutator_envs(config) {
+        cmd.env(key, val);
+    }
+
+    for (key, val) in &config.afl_env {
+        cmd.env(key, val);
+    }
+
+    cmd.process_group(0);
+
+    let child = cmd.spawn()?;
+    let state = RunnerState {
+        id,
+        pid: child.id(),
+    };
+    Ok((state, child))
+}
+
+/// Polls for `fuzzer_stats` files to confirm all runners have started executing.
+///
+/// AFL++ creates `<output_dir>/<runner_name>/fuzzer_stats` once fuzzing begins.
+/// We verify `execs_per_sec` is non-zero to confirm actual execution, not just
+/// file creation.
+fn verify_startup(output_dir: &Path, runners: &[RunnerState], children: &mut [Child]) -> bool {
+    let deadline = Instant::now() + STARTUP_TIMEOUT;
+
+    while Instant::now() < deadline {
+        let all_started = runners
+            .iter()
+            .all(|r| has_nonzero_execs(&output_dir.join(r.name()).join("fuzzer_stats")));
+
+        if all_started {
+            return true;
+        }
+
+        for (child, runner) in children.iter_mut().zip(runners.iter()) {
+            match child.try_wait() {
+                Ok(Some(_)) => {
+                    log::error!(
+                        "runner {} (pid {}) died during startup",
+                        runner.id,
+                        runner.pid
+                    );
+                    return false;
+                }
+                Ok(None) => {}
+                Err(e) => {
+                    log::error!(
+                        "failed to check runner {} (pid {}): {e}",
+                        runner.id,
+                        runner.pid
+                    );
+                    return false;
+                }
+            }
+        }
+
+        thread::sleep(STARTUP_POLL_INTERVAL);
+    }
+
+    for runner in runners {
+        let stats = output_dir.join(runner.name()).join("fuzzer_stats");
+        if !has_nonzero_execs(&stats) {
+            log::error!(
+                "runner {} (pid {}) did not reach non-zero execs_per_sec",
+                runner.id,
+                runner.pid
+            );
+        }
+    }
+
+    false
+}
+
+/// Returns true if the `fuzzer_stats` file exists and reports non-zero `execs_per_sec`.
+fn has_nonzero_execs(path: &Path) -> bool {
+    let Ok(contents) = fs::read_to_string(path) else {
+        return false;
+    };
+    contents.lines().any(|line| {
+        line.trim_start().starts_with("execs_per_sec")
+            && line
+                .split(':')
+                .nth(1)
+                .and_then(|v| v.trim().parse::<f64>().ok())
+                .is_some_and(|n| n > 0.0)
+    })
+}
+
+/// Kills all runner process groups via SIGKILL.
+fn kill_runners(runners: &[RunnerState]) {
+    for runner in runners {
+        let pgid = format!("-{}", runner.pid);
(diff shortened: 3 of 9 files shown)


Ashish-Kumar-Dash:
I ran `smitebot start` end to end on my machine with lnd/encrypted_bytes, 2 runners.
Sample config I used for testing
```toml
  target = "lnd"
  scenario = "noise"
  aflpp_path = "/home/akd/AFLplusplus"
  smite_dir = "."
  runners = 2
  output_dir = "/tmp/smite-test-out"
  sharedir = "/tmp/smite-test-nyx"
 ```
 
Docker build, Nyx sharedir setup, runner spawning, and startup verification all worked. Both runners hit ~530 exec/sec, 142 corpus entries in under a minute, 99% stability. state.json persisted correctly with PIDs,  image digest, and git hash.
 
Some key findings during testing I fixed later ->
- AFL++ Nyx mode needs -Y (not -X) for parallel -M/-S support ,  -X is  standalone only. Fixed.
- -Y mode requires numeric runner names (-M 0, -S 1) instead of string names  (-M main, -S runner001). Fixed.
 - setup-nyx.sh fails if python isn't symlinked to python3 , separate fix is needed.
 - Scenario reported "not compiled with afl instrumentation" but still fuzzed

PR is ready for review @NishantBansal2003 . Screenshots attached.

Screenshot from 2026-06-13 22-32-17

> **Runner 1 actively fuzzing: 531 exec/sec, 142 corpus entries, 99.36% stability.**

Screenshot from 2026-06-13 22-36-47

  > **Both runners alive with `-Y` parallel Nyx mode: `-M 0` (primary) and `-S 1` (secondary). Persisted campaign state with PIDs, image digest, git hash, and `"status": "running"`.  AFL++ `fuzzer_stats` confirms runner 0 (pid 70235) is executing — matches PID in state.json.**

Screenshot from 2026-06-13 22-38-09

> **Docker build, Nyx sharedir setup, and runner spawning — campaign reaches  `running` state.**

File: smitebot/README.md [hunk 1]
```diff
@@ -37,6 +37,25 @@ Campaign settings are stored in a TOML file. See [`sample-campaign.toml`](sample
...
+```bash
+smitebot start campaign.toml
+smitebot start campaign.toml --skip-build
+smitebot start campaign.toml --skip-build --skip-setup
+```
+
+Flags:
+
+- `--skip-build`: Skip the Docker image build step (use when the image is already built).
+- `--skip-setup`: Skip the Nyx sharedir setup step (use when the sharedir is already prepared).
```
NishantBansal2003: I don’t think we need to skip the docker image build. If the image already exists, I assume the build will just use the cached layers and immediately start fuzzing.

The same applies to the Nyx shared directory. If the shared directory already exists, the script will delete it before creating a new one, so I don’t think there’s a need to skip that either: https://github.com/morehouse/smite/blob/b82be88060678aa092d28fa10813ba649beeeae5/scripts/setup-nyx.sh#L50

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,450 @@
...
+    /// Skip the Nyx sharedir setup step.
+    #[arg(long)]
+    skip_setup: bool,
+}
+
+impl StartCommand {
+    /// Launches a fuzzing campaign from the given configuration.
+    pub fn execute(args: &StartArgs) -> bool {
+        let config = match CampaignConfig::load(&args.path) {
+            Ok(c) => c,
```
NishantBansal2003, line 40: We should check whether the paths actually exist using `check_paths` here.

smitebot/src/commands/build.rs
```diff
@@ -114,20 +120,26 @@ fn default_workload_image_tag(target: Target, scenario: &str, coverage: bool) ->
 }
 
 /// Runs `docker build`, streaming stdout/stderr directly to the terminal.
-fn run_docker_build(inputs: &BuildInputs) -> std::io::Result<ExitStatus> {
+pub(crate) fn run_docker_build(
+    smite_dir: &Path,
+    image: &str,
+    scenario: &str,
+    dockerfile: &Path,
+    no_cache: bool,
```
NishantBansal2003: This segregation of the arguments doesn't look great. Perhaps we should construct a `BuildInputs` struct and use it wherever we need to build the docker image

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,450 @@
...
+        }
+
+        log::info!("campaign {} is running", state.id);
+        log::info!("state saved to {}", state_path.display());
+        true
+    }
+}
+
+/// Builds the Docker image for the campaign.
+fn build_image(config: &CampaignConfig, image: &str) -> bool {
```
NishantBansal2003: This part is exactly the same as: https://github.com/morehouse/smite/blob/b82be88060678aa092d28fa10813ba649beeeae5/smitebot/src/commands/build.rs#L77-L80 Perhaps we should extract it into a helper function and use it in both places

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,450 @@
...
+        return false;
+    }
+
+    let status = Command::new(&script)
+        .arg(&config.sharedir)
+        .arg(image)
+        .arg(&config.aflpp_path)
+        .status();
+
+    match status {
```
NishantBansal2003: existing code follows this pattern: https://github.com/morehouse/smite/blob/b82be88060678aa092d28fa10813ba649beeeae5/smitebot/src/commands/build.rs#L92-L103

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,450 @@
...
+            target: config.target,
+            scenario: config.scenario.clone(),
+            image: image.clone(),
+            image_digest,
+            output_dir: config.output_dir.clone(),
+            sharedir: config.sharedir.clone(),
+            smite_git_hash: git_hash,
+            start_time: run_date("--iso-8601=seconds").unwrap_or_else(|| "unknown".to_string()),
+            runners: Vec::new(),
+        };
```
NishantBansal2003: This should be moved into a separate function for initializing the initial state

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,450 @@
...
+            return false;
+        }
+
+        let seed_dir = match ensure_seed_dir(&config) {
+            Ok(dir) => dir,
+            Err(e) => {
+                log::error!("failed to prepare seed directory: {e}");
+                return false;
+            }
+        };
```
NishantBansal2003, line 78: We should not save the state if we later error out before even starting the runners

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,450 @@
...
+    if config.scenario.starts_with("ir") {
+        let mutator_lib = config
+            .smite_dir
+            .join("target")
+            .join("release")
+            .join("libsmite_ir_mutator.so");
+        cmd.env("AFL_CUSTOM_MUTATOR_LIBRARY", &mutator_lib);
+        cmd.env("AFL_CUSTOM_MUTATOR_ONLY", "1");
+        cmd.env("AFL_FRAMESHIFT_DISABLE", "1");
+        cmd.env("AFL_DISABLE_TRIM", "1");
```
NishantBansal2003: this env variable is not required

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,450 @@
...
+        cmd.env(key, val);
+    }
+
+    cmd.process_group(0);
+
+    let child = cmd.spawn()?;
+    let pid = child.id();
+
+    // Intentionally drop so afl-fuzz outlives smitebot, managed via PID later.
+    drop(child);
```
NishantBansal2003: I don’t think this is really needed. What you described in the comment is, I believe, already handled by `cmd.process_group(0);`

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,450 @@
...
+    let deadline = Instant::now() + timeout;
+
+    while Instant::now() < deadline {
+        let all_started = runners
+            .iter()
+            .all(|r| output_dir.join(&r.name).join("fuzzer_stats").exists());
+
+        if all_started {
+            return true;
+        }
```
NishantBansal2003, line 283: I think we should check whether `fuzzer_stats` exists here and whether it has a non-zero `execs_per_sec` value before concluding that the specific runner has started. WDYT?

NishantBansal2003 (changes requested):
I haven't listed all my suggestions, but let's try to minimize code duplication where possible. I can see several repeated patterns that could be extracted into helper functions and reused. Also, it would be great to structure the code a bit more clearly and add brief comments where needed to explain what is being done and why.

smitebot/README.md
Ashish-Kumar-Dash: Those were UX conveniences, but I don't have a strong opinion here so I will remove it

smitebot/src/commands/start.rs
Ashish-Kumar-Dash: done

smitebot/src/commands/build.rs
```diff
@@ -36,20 +36,35 @@ pub struct BuildArgs {
 
 /// Fully resolved Docker build inputs.
 #[derive(Debug, PartialEq, Eq)]
-struct BuildInputs {
+pub(crate) struct BuildInputs {
```
NishantBansal2003: We don't use the `pub(crate)` pattern in our codebase, Just use `pub`

smitebot/src/commands/build.rs
```diff
@@ -36,20 +36,35 @@ pub struct BuildArgs {
...
-    scenario: String,
+    pub(crate) scenario: String,
     /// Whether Docker should rebuild without using its layer cache.
-    no_cache: bool,
+    pub(crate) no_cache: bool,
 }
 
 impl BuildInputs {
+    /// Resolves Docker build inputs from a campaign configuration.
+    pub(crate) fn from_config(config: &crate::config::CampaignConfig, image: &str) -> Self {
```
NishantBansal2003: nit: since we already import `Target` from `crate::config`, I'd move this there too and use `CampaignConfig` directly

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,529 @@
...
+
+        state.runners = runners;
+
+        log::info!("waiting for runners to initialize");
+        let started = verify_startup(&config.output_dir, &state.runners);
+        if !started {
+            log::error!("not all runners started within the timeout");
+            state.status = Status::Failed;
+            if let Err(e) = state.save(&state_path) {
```
NishantBansal2003: nit: can move the `spawn_runner` loop and `verify_startup` into a separate `launch_runners` func

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,529 @@
...
+        cmd.arg("-S").arg(name);
+    }
+
+    for flag in &config.afl_flags {
+        cmd.arg(flag);
+    }
+
+    cmd.arg("--").arg(&config.sharedir);
+
+    // Environment contract defined in smite-ir-mutator/src/lib.rs.
```
NishantBansal2003:
```suggestion
    // For IR scenarios, enable the smite-ir custom mutator for AFL++ and disable all built-in AFL++ mutators.
```

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,529 @@
...
+}
+
+/// Polls for `fuzzer_stats` files to confirm all runners have started executing.
+///
+/// AFL++ creates `<output_dir>/<runner_name>/fuzzer_stats` once fuzzing begins.
+/// We verify `execs_per_sec` is non-zero to confirm actual execution, not just
+/// file creation.
+fn verify_startup(output_dir: &Path, runners: &[RunnerState]) -> bool {
+    let timeout = Duration::from_mins(5);
+    let poll_interval = Duration::from_secs(5);
```
NishantBansal2003: These should definitely be constants and moved to a global scope with appropriate comments

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,529 @@
...
+    let seed_dir = config.output_dir.join(".seeds");
+    fs::create_dir_all(&seed_dir)?;
+    if seed_dir.read_dir()?.next().is_none() {
+        fs::write(seed_dir.join("seed0"), b"\x00")?;
+    }
+    Ok(seed_dir)
+}
+
+/// Builds the initial campaign state before runners are spawned.
+fn init_campaign_state(config: &CampaignConfig, campaign_id: String, image: &str) -> CampaignState {
```
NishantBansal2003: I think this belongs in `state.rs` as part of init

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,529 @@
...
+    let git_hash = smite_git_hash(&config.smite_dir);
+    let image_digest = docker_image_id(image);
+
+    if git_hash.is_none() {
+        log::warn!("could not determine smite git hash");
+    }
+    if image_digest.is_none() {
+        log::warn!("could not determine Docker image digest for {image}");
+    }
```
NishantBansal2003: Since this happens before startup, I think it makes more sense to return an error if we can’t retrieve these values, rather than logging a warning and making them Option

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,529 @@
...
+                    log::info!("spawned {name} (pid {pid})");
+                    runners.push(RunnerState { id, name, pid });
+                }
+                Err(e) => {
+                    log::error!("failed to spawn {name}: {e}");
+                    state.status = Status::Failed;
+                    state.runners = runners;
+                    if let Err(e) = state.save(&state_path) {
+                        log::warn!("failed to save state: {e}");
+                    }
```
NishantBansal2003: What happens if some runners are spawned successfully and a later startup fails? How are the already-spawned runners terminated?

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,529 @@
...
+            }
+        }
+
+        state.runners = runners;
+
+        log::info!("waiting for runners to initialize");
+        let started = verify_startup(&config.output_dir, &state.runners);
+        if !started {
+            log::error!("not all runners started within the timeout");
+            state.status = Status::Failed;
```
NishantBansal2003: What happens if some runners started successfully and a later runners don't? How are the already-started runners terminated?

smitebot/src/state.rs
```diff
@@ -0,0 +1,199 @@
...
+        fs::write(&tmp, &json).map_err(|source| StateError::Write {
+            path: tmp.clone(),
+            source,
+        })?;
+        fs::rename(&tmp, path).map_err(|source| StateError::Write {
+            path: path.to_path_buf(),
+            source,
+        })?;
+        Ok(())
+    }
```
NishantBansal2003, line 141: This is exactly the kind of place where an inline comment would help, from the rustdoc, I assume the rename is atomic, not the write, otherwise, using a tmp file wouldn’t be necessary

NishantBansal2003 (changes requested):

smitebot/src/commands/start.rs
Ashish-Kumar-Dash: added both the comments to clarify the WHAT and WHY clearly

smitebot/src/commands/start.rs
Ashish-Kumar-Dash: The already-spawned runners are intentionally left running. They're orphaned processes (process_group(0)), that's the design. We save their PIDs to state.json with `status: "failed"`, and `smitebot stop` handles teardown. Putting kill logic in start would duplicate stop and create two teardown  paths.

smitebot/src/commands/start.rs
Ashish-Kumar-Dash: PIDs are persisted, status: "failed", smitebot stop cleans up. This is the core functionality of `smitebot stop` command which lands after `smitebot start` & `smitebot status`

smitebot/src/state.rs
```diff
@@ -0,0 +1,261 @@
...
+            start_time,
+            runners: Vec::new(),
+        }
+    }
+
+    /// Returns the base directory for all campaign state: `~/.smitebot/runs`.
+    ///
+    /// Returns `None` if the home directory cannot be determined.
+    pub fn runs_dir() -> Option<PathBuf> {
+        dirs::home_dir().map(|home| home.join(".smitebot").join("runs"))
```
NishantBansal2003: This will eliminate the new dependency
```suggestion
        std::env::home_dir().map(|home| home.join(".smitebot").join("runs"))
```

smitebot/src/state.rs
```diff
@@ -0,0 +1,261 @@
...
+    Failed,
+}
+
+/// State of a single AFL++ runner process.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct RunnerState {
+    /// Runner index (0 = primary, 1..N = secondary).
+    pub id: u16,
+    /// AFL++ runner name used with -M/-S flag.
+    pub name: String,
```
NishantBansal2003: Since `name` is derivable from `id`, do we need to store it separately?

smitebot/src/state.rs
```diff
@@ -0,0 +1,261 @@
...
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("nested").join("deep").join("state.json");
+        let state = sample_state();
+
+        state.save(&path).unwrap();
+        assert!(path.exists());
+    }
+
+    #[test]
+    fn save_is_atomic() {
```
NishantBansal2003: This test doesn’t test atomicity. To test atomicity, we’d need to interrupt execution during the rename
```suggestion
    fn save_removes_temp_file() {
```

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,540 @@
...
+            return false;
+        }
+
+        let seed_dir = match ensure_seed_dir(&config) {
+            Ok(dir) => dir,
+            Err(e) => {
+                log::error!("failed to prepare seed directory: {e}");
+                return false;
+            }
+        };
```
NishantBansal2003, line 78: This will be needed when starting a new campaign, but if we’re resuming an existing campaign, we don’t need a seed directory. In that case, the AFL++ arguments will differ. I think we need a resume mode in start as well, unless you’re planning to add a separate command for that? Currently, if someone runs start again with the same configuration, AFL++ will fail because the output directory already exists. So either we need a way to resume fuzzing, or we need to delete the existing output before starting a new campaign

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,540 @@
...
+
+        let Some(git_hash) = smite_git_hash(&config.smite_dir) else {
+            log::error!("could not determine smite git hash");
+            return false;
+        };
+        let Some(image_digest) = docker_image_id(&image) else {
+            log::error!("could not determine Docker image digest for {image}");
+            return false;
+        };
+        let start_time = run_date("--iso-8601=seconds").unwrap_or_else(|| "unknown".to_string());
```
NishantBansal2003: Can we use a library/function call to obtain the timestamp here? Shelling out to date doesn’t seem ideal

NishantBansal2003 (changes requested):
Let’s follow OOPs patterns in the code. Wherever we have something like `fn(struct, ...)`, and you think the function is closely related to one of the struct arguments, consider making it a method on that struct instead, i.e., `struct.fn(...)`

smitebot/src/commands/start.rs
Ashish-Kumar-Dash: Resume mode is out of scope for this PR, start is new-campaign-only. Resume needs different AFL++ flags (-i-), skips build/seed setup, and reuses the output_dir, so it's a separate feature. I can add it later in a seperate PR. Added it to the list

smitebot/src/state.rs
Ashish-Kumar-Dash: Not really, we can optimize the code in this regard, fixed in latest commit

smitebot/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license.workspace = true
 workspace = true
 
 [dependencies]
+chrono = "0.4"
```
NishantBansal2003: This feels like a fairly heavy dependency for a small benefit. Is there another way to distinguish campaign IDs? How about just using `std::time` and appending the number of seconds since the Unix epoch?

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,472 @@
...
+        if !run_build(&inputs) {
+            return false;
+        }
+
+        log::info!("setting up Nyx sharedir at {}", config.sharedir.display());
+        if !setup_nyx(&config, &image) {
+            return false;
+        }
+
+        let seed_dir = match ensure_seed_dir(&config) {
```
NishantBansal2003, line 72: `check_paths` already verifies that the seed directory exists and is non-empty. Do we really need the separate `ensure_seed_dir` function? Or is the intent that if the user explicitly provides a seed directory in the config, it must already exist and be non-empty, whereas if no seed directory is provided, we create one ourselves?
https://github.com/morehouse/smite/blob/2ed34ce738fab99ce5a3e4141088c7534b729e82/smitebot/src/config.rs#L128-L134

NishantBansal2003:
When I run this and the campaign starts, does your terminal output also keep wrapping around on the same line?

Also, when I kill the PIDs stored in `state.json`, it doesn’t seem to stop any of the zombie QEMU processes -- only the `afl-fuzz` processes get terminated. Are you planning to handle complete cleanup/termination explicitly in the stop command?

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,472 @@
...
+        };
+        let Some(image_digest) = docker_image_id(&image) else {
+            log::error!("could not determine Docker image digest for {image}");
+            return false;
+        };
+        let start_time = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("system clock is before Unix epoch")
+            .as_secs()
+            .to_string();
```
NishantBansal2003: This can be moved into `CampaignState::new`, since it’s only used there

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,472 @@
...
+                return false;
+            }
+        }
+    }
+
+    state.runners = runners;
+
+    log::info!("waiting for runners to initialize");
+    let started = verify_startup(&config.output_dir, &state.runners);
+    if !started {
```
NishantBansal2003:
```suggestion
    if !verify_startup(&config.output_dir, &state.runners) {
```

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,472 @@
...
+        fs::write(seed_dir.join("seed0"), b"\x00")?;
+    }
+    Ok(seed_dir)
+}
+
+/// Spawns a single `afl-fuzz` process in its own process group and returns its PID.
+///
+/// The child is intentionally orphaned so the campaign outlives the `smitebot`
+/// process. The `stop` command uses the persisted PID to terminate it later.
+fn spawn_runner(config: &CampaignConfig, id: u16, seed_dir: &Path) -> std::io::Result<u32> {
```
NishantBansal2003: This is spawning a runner, so it should return a `RunnerState` rather than just a PID

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,472 @@
...
+    let afl_fuzz = config.aflpp_path.join("afl-fuzz");
+    let name = id.to_string();
+
+    let mut cmd = Command::new(&afl_fuzz);
+    cmd.arg("-Y");
+    cmd.arg("-i").arg(seed_dir);
+    cmd.arg("-o").arg(&config.output_dir);
+
+    if id == 0 {
```
NishantBansal2003, line 237: Needs a comment explaining the logic here

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,472 @@
...
+    let output = Command::new("docker")
+        .arg("inspect")
+        .arg("--format={{.Id}}")
+        .arg(image)
+        .output()
+        .ok()?;
+
+    if output.status.success() {
+        Some(String::from_utf8_lossy(&output.stdout).trim().to_string())
+    } else {
```
NishantBansal2003: nit: Both functions share the same boilerplate, consider extracting it into a helper

smitebot/src/config.rs
```diff
@@ -98,6 +100,16 @@ impl CampaignConfig {
         Ok(config)
     }
 
+    /// Returns a unique campaign ID from the target, scenario, and current time.
+    #[must_use]
+    pub fn campaign_id(&self) -> String {
+        let epoch_secs = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("system clock is before Unix epoch")
+            .as_secs();
```
NishantBansal2003: This function is used in two places with the same code. How about moving it to utils (or another appropriate location) and reusing it in both places?

smitebot/src/state.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// Docker image digest (image ID hash for locally built images).
+    pub image_digest: String,
+    /// AFL++ output directory containing runner findings and stats.
+    pub output_dir: PathBuf,
+    /// Path to the Nyx sharedir.
+    pub sharedir: PathBuf,
+    /// Smite repository git hash at campaign start.
+    pub smite_git_hash: String,
+    /// ISO 8601 timestamp when the campaign started.
+    pub start_time: String,
```
NishantBansal2003: This should be a u64 now instead of a String, since we’re using unix timestamp secs

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,472 @@
...
+
+    // Enable the smite-ir custom mutator and disable built-in AFL++ mutators.
+    // Environment contract defined in smite-ir-mutator/src/lib.rs.
+    // Applied after afl_env so the contract cannot be accidentally overridden.
+    if config.scenario.starts_with("ir") {
+        let mutator_lib = config
+            .smite_dir
+            .join("target")
+            .join("release")
+            .join("libsmite_ir_mutator.so");
```
NishantBansal2003: We should probably check that `libsmite_ir_mutator.so` exists here, ideally at startup, so we can fail early instead of discovering the issue later

NishantBansal2003 (changes requested):

smitebot/src/commands/start.rs
Ashish-Kumar-Dash: Yes, check_paths validates user-provided paths (including seed_dir when specified). ensure_seed_dir handles the case when seed_dir is omitted.  AFL++ requires -i to point to a non-empty directory, so we create a minimal one-byte corpus at output_dir/.seeds/seed0. They're complementary

Ashish-Kumar-Dash:
> When I run this and the campaign starts, does your terminal output also keep wrapping around on the same line?

Yes, AFL++ runners inherit the parent's terminal since we don't redirect stdout/stderr. Multiple runners writing status updates to the same terminal causes the garbled output we're seeing. Redirecting runner output to per-runner log files is a possible improvement we can make later for the next iteration.

> Also, when I kill the PIDs stored in `state.json`, it doesn’t seem to stop any of the zombie QEMU processes -- only the `afl-fuzz` processes get terminated. Are you planning to handle complete cleanup/termination explicitly in the stop command?

Yes, planned for `smitebot stop`, otherwise things will get messy. We set process_group(0) in spawn_runner to enable process group signals. However, @morehouse  has flagged in issue #70 that QEMU instances don't always die when killing AFL++ , he's had to pkill them manually. The stop command will start with killpg and  verify whether a fallback cleanup sweep is needed during implementation. I can provide more details on this when implementing the `stop` command.

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    // Enable the smite-ir custom mutator and disable built-in AFL++ mutators.
+    // Environment contract defined in smite-ir-mutator/src/lib.rs.
+    // Applied after afl_env so the contract cannot be accidentally overridden.
+    if config.scenario.starts_with("ir") {
+        cmd.env("AFL_CUSTOM_MUTATOR_LIBRARY", config.ir_mutator_path());
+        cmd.env("AFL_CUSTOM_MUTATOR_ONLY", "1");
+        cmd.env("AFL_FRAMESHIFT_DISABLE", "1");
+    }
```
morehouse: This should come before we process `config.afl_env` so that user-specified values will override these.

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,452 @@
...
+        let all_started = runners
+            .iter()
+            .all(|r| has_nonzero_execs(&output_dir.join(r.name()).join("fuzzer_stats")));
+
+        if all_started {
+            return true;
+        }
+
+        thread::sleep(STARTUP_POLL_INTERVAL);
+    }
```
morehouse, line 308: This will wait the full 5 minutes if a runner dies early.  We should also check whether any runners have died and return early in that case.

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,452 @@
...
+    // Applied after afl_env so the contract cannot be accidentally overridden.
+    if config.scenario.starts_with("ir") {
+        cmd.env("AFL_CUSTOM_MUTATOR_LIBRARY", config.ir_mutator_path());
+        cmd.env("AFL_CUSTOM_MUTATOR_ONLY", "1");
+        cmd.env("AFL_FRAMESHIFT_DISABLE", "1");
+    }
+
+    cmd.process_group(0);
+
+    let child = cmd.spawn()?;
```
morehouse, line 260: All runners end up sharing the same stdio, which is very hard to read (multiple TUIs constantly overwriting each other).  We should consider spawning tmux windows for each runner like `aflr` does.

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,452 @@
...
+                }
+                return false;
+            }
+        }
+    }
+
+    state.runners = runners;
+
+    log::info!("waiting for runners to initialize");
+    if !verify_startup(&config.output_dir, &state.runners) {
```
morehouse: Currently if some runners failed to start, we end up recording a failed state here, but we don't actually clean up any remaining runners.  The user has to manually pkill them.

We should really clean those up automatically.

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,452 @@
...
+    cmd.arg("--").arg(&config.sharedir);
+
+    for (key, val) in &config.afl_env {
+        cmd.env(key, val);
+    }
+
+    // Enable the smite-ir custom mutator and disable built-in AFL++ mutators.
+    // Environment contract defined in smite-ir-mutator/src/lib.rs.
+    // Applied after afl_env so the contract cannot be accidentally overridden.
+    if config.scenario.starts_with("ir") {
```
morehouse: We should add a test for the auto-custom-mutator logic.

smitebot/src/config.rs
```diff
@@ -132,6 +150,15 @@ impl CampaignConfig {
...
         }
+        if self.scenario.starts_with("ir") {
+            let mutator = self.ir_mutator_path();
+            if !mutator.exists() {
+                errors.push(format!(
+                    "IR mutator library not found: {}",
+                    mutator.display()
+                ));
+            }
+        }
```
morehouse: `smitebot start` should automatically build the mutator library.

morehouse (changes requested):

smitebot/src/commands/start.rs
Ashish-Kumar-Dash: Agreed, shared stdio with multiple TUIs is inconvenient. No tmux integration exists in the codebase yet, so will have to start from scratch and will add tmux session/window spawning in a follow up PR since it spreads across doctor, start, stop, and state. Added it to the list.

smitebot/src/commands/start.rs
```diff
@@ -0,0 +1,592 @@
...
+                .nth(1)
+                .and_then(|v| v.trim().parse::<f64>().ok())
+                .is_some_and(|n| n > 0.0)
+    })
+}
+
+/// Returns true if a process with the given PID is still running.
+fn is_runner_alive(pid: u32) -> bool {
+    Path::new(&format!("/proc/{pid}")).exists()
+}
```
morehouse: This doesn't quite work as expected.  When a child process dies and the parent doesn't read the exit code, it becomes a zombie which still has an entry at `/proc/{pid}`.

I think what we actually want is to save the `Child` handles from spawning the runners and then do `child.try_wait` to check whether they died.

morehouse (approved):
LGTM

PR #120, opened 2026-06-09
smite-ir: implement `BuildFundingCreated` and `SendFundingCreated` operation 

NishantBansal2003:
Depends-on: #117 

This PR adds the `BuildFundingCreated` and `SendFundingCreated` IR operations, which build and send a `funding_created` message to the target peer.

It also adds a new `channel_states` field to the `Executor` struct. This field is used to store the `channel_state` while building the `funding_created` message and is subsequently used to verify the received `funding_signed` signature.

### DIFF
--- a/smite-ir/src/tests.rs
+++ b/smite-ir/src/tests.rs
@@ -461,7 +461,7 @@ fn postcard_roundtrip() {
             },
             Instruction {
                 operation: Operation::DerivePoint,
-                inputs: vec![7],
+                inputs: vec![8],
             },
             Instruction {
                 operation: Operation::LoadFeeratePerKw(15_000),
@@ -469,11 +469,11 @@ fn postcard_roundtrip() {
             },
             Instruction {
                 operation: Operation::CreateFundingTransaction,
-                inputs: vec![1, 8, 4, 9],
+                inputs: vec![1, 9, 5, 10],
             },
             Instruction {
                 operation: Operation::BroadcastTransaction,
-                inputs: vec![10],
+                inputs: vec![11],
             },
             Instruction {
                 operation: Operation::LoadU16(144),
@@ -482,12 +482,13 @@ fn postcard_roundtrip() {
             Instruction {
                 operation: Operation::BuildFundingCreated,
                 inputs: vec![
-                    10, 4, 5, 0, 1, 1, 1, 4, 12, 1, 1, 1, 1, 4, 12, 3, 5, 10, 1, 1,
+                    11, 5, 6, 0, 1, 1, 1, 5, 13, 9, 9, 9, 9, 5, 13, 3, 5, 10, 1, 9,
                 ],
             },
         ],
     };
 
+    assert_well_formed(&program);
     let bytes = postcard::to_allocvec(&program).expect("postcard serialization");
     let decoded: Program = postcard::from_bytes(&bytes).expect("postcard deserialization");
     assert_eq!(program, decoded);
@@ -596,7 +597,7 @@ fn displays_create_and_broadcast_tx_program() {
 }
 
 #[test]
-fn displays_build_funding_created_program() {
+fn displays_build_and_send_funding_created_program() {
     let instructions = vec![
         // Funding transaction.
         Instruction {
@@ -667,6 +668,11 @@ fn displays_build_funding_created_program() {
                 4, 2, 5, 0, 7, 7, 7, 8, 9, 11, 11, 11, 11, 8, 9, 12, 13, 14, 7, 11,
             ],
         },
+        // Send funding_created.
+        Instruction {
+            operation: Operation::SendFundingCreated,
+            inputs: vec![15],
+        },
     ];
 
     let program = Program { instructions };
@@ -693,6 +699,7 @@ fn displays_build_funding_created_program() {
         "v13 = LoadAmount(0)".into(),
         "v14 = LoadFeeratePerKw(253)".into(),
         "v15 = BuildFundingCreated(v4, v2, v5, v0, v7, v7, v7, v8, v9, v11, v11, v11, v11, v8, v9, v12, v13, v14, v7, v11)".into(),
+        "v16 = SendFundingCreated(v15)".into(),
     ];
 
     assert_eq!(lines.len(), expected.len(), "line count mismatch");
--- a/smite-ir/src/tests.rs
+++ b/smite-ir/src/tests.rs
@@ -475,6 +475,16 @@ fn postcard_roundtrip() {
                 operation: Operation::BroadcastTransaction,
                 inputs: vec![10],
             },
+            Instruction {
+                operation: Operation::LoadU16(144),
+                inputs: vec![],
+            },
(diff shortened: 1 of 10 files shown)


smite-ir/src/tests.rs
```diff
@@ -388,6 +388,16 @@ fn postcard_roundtrip() {
...
                 inputs: vec![10],
             },
+            Instruction {
+                operation: Operation::LoadU16(144),
+                inputs: vec![],
+            },
+            Instruction {
+                operation: Operation::BuildFundingCreated,
+                inputs: vec![
+                    10, 4, 5, 0, 1, 1, 1, 4, 12, 1, 1, 1, 1, 4, 12, 3, 5, 10, 1, 1,
```
ekzyis: Here, v4 is passed as the input with index 1 to `BuildFundingCreated`, which is `funding_satoshis`. This is also used by `CreateFundingTransaction`. Similar to checking `funding_satoshis` > `push_msat`, should it also be a mutator error if `funding_satoshis` does not match the funding tx output amount ?

This test should fail then:

<details>
<summary>test</summary>

```diff
diff --git a/smite-scenarios/src/executor.rs b/smite-scenarios/src/executor.rs
index 78cc5f5..82e4f17 100644
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -1923,2 +1923,28 @@ mod tests {

+    #[test]
+    fn execute_build_funding_created_funding_satoshis_tx_mismatch() {
+        let mut instrs = build_and_send_funding_created_instructions();
+        instrs[14] = Instruction {
+            operation: Operation::LoadAmount(5_000_000),
+            inputs: vec![],
+        };
+        instrs[15].inputs[1] = 14;
+        let mut mock_cli = MockBitcoinCli {
+            utxos: vec![sample_utxo()],
+            change_spk: sample_change_spk(),
+            ..Default::default()
+        };
+        let mut mock_conn = MockConnection::new();
+        let ctx = &sample_context();
+        let err = Executor::new(&mut mock_conn, &mut mock_cli, ctx)
+            .execute(
+                &Program {
+                    instructions: instrs,
+                },
+                std::time::Instant::now(),
+            )
+            .unwrap_err();
+        assert!(matches!(err, ExecuteError::Commitment(_)));
+    }
+
     // -- extract_field tests --
```

</details>

NishantBansal2003: Ok, this was intentional. I am explicitly building the `FundingCreated` message with an arbitrary `funding_satoshis` value, which does not necessarily match the output value of the funding transaction. This allows us to test both invalid and valid cases. If this results in a `CommitmentError`, we explicitly ignore it in `ir.rs`, since it is a condition generated by the fuzzer and can be safely ignored.

We could have extracted `funding_satoshis` from the funding transaction instead, but that would largely restrict the operation to valid cases. In the future, I was thinking of splitting the funding transaction fields as well, so that we can provide a raw txid and vout separately and then mutate them while building the `FundingCreated` message.

morehouse: I've looked at the input list 3 times now and I'm pretty sure the indexes are way off. `v10` has type `FeeratePerKw` which does not match the expected input type of `FundingTransaction`.  `v4` has type `ShortChannelId` which does not match the expected input type of `Amount`.  etc.

smite-scenarios/src/executor.rs
```diff
@@ -122,16 +130,22 @@ pub struct Executor<C, B> {
...
 impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
     /// Creates an executor with the given connection, bitcoin-cli handle, and
-    /// program context.
+    /// program context. Channel state starts empty.
     pub fn new(conn: C, bitcoin_cli: B, context: ProgramContext) -> Self {
         Self {
             conn,
             bitcoin_cli,
             context,
+            channel_states: HashMap::new(),
```
morehouse, line 148: We will eventually want a way to start with a non-empty channel state for some `SnapshotSetup`s.  But for now this is fine.

smite-scenarios/src/executor.rs
```diff
@@ -568,6 +611,89 @@ fn build_open_channel(variables: &[Option<Variable>], inputs: &[usize]) -> OpenC
...
+    let channel_id = ChannelId::v1_from_funding_outpoint(config.funding_outpoint);
+
+    channel_states.insert(
+        channel_id,
+        ChannelState {
+            config,
+            holder,
+            commitment: state,
+        },
+    );
```
morehouse: This will overwrite the channel state if two `BuildFundingCreated` operations produce the same `channel_id`, which will cause us to mis-verify signatures if we have a program like this:

```
BuildFundingCreated(var_set_0)
SendFundingCreated(...)
...  // finish funding flow
...  // add an HTLC and send commitment_signed
BuildFundingCreated(var_set_0)  // resets channel state
...  // handle target's commitment_signed.  Signatures will mis-match
```

I think the simplest solution is to only set the channel state if an entry doesn't already exist.  Then the second `BuildFundingCreated` still produces a `funding_created` message that can be sent (which could trigger bugs) but our internal state still matches the actual open channel.

smite-ir/src/tests.rs
ekzyis: > `v4` has type `ShortChannelId` which does not match the expected input type of `Amount`. etc.

Oh, you're right. I meant to comment on the input list at executor.rs:2007:

```rust
            Instruction {
                operation: Operation::LoadAmount(u64::MAX),
                inputs: vec![],
            },
            Instruction {
                operation: Operation::BuildFundingCreated,
                inputs: vec![
                    6, 4, 8, 0, 1, 1, 1, 9, 10, 3, 3, 3, 3, 9, 10, 11, 12, 13, 1, 3,
                ],
            },
            Instruction {
                operation: Operation::SendFundingCreated,
                inputs: vec![15],
            },
```

NishantBansal2003: > I've looked at the input list 3 times now and I'm pretty sure the indexes are way off. `v10` has type `FeeratePerKw` which does not match the expected input type of `FundingTransaction`. `v4` has type `ShortChannelId` which does not match the expected input type of `Amount`. etc.

In https://github.com/morehouse/smite/commit/10497b1a85a338651a9bc0d13f9bd8243bbbbb19,  operations were added in between without the indexes being updated accordingly. After the rebase, this wasn’t flagged, so it slipped through unnoticed. I’ll fix this and add `assert_well_formed(&program);` here so that any similar errors surface immediately



> meant to comment on the input list at executor.rs:2007:

For @ekzyis, I was referring to `executor.rs` only, and I hope that answers your question

smite-scenarios/src/executor.rs
NishantBansal2003: Make sense, Done in https://github.com/morehouse/smite/pull/120/commits/9874ae7485ea379f47ab66225a09d38963ca5a71

morehouse (approved):
Ready to squash

morehouse (approved):
LGTM

PR #121, opened 2026-06-10
smite: add `announcement_signatures` codec

devvaansh:
Implements the BOLT 7 announcement_signatures message (type 259).
A direct (non-gossip) message between the two endpoints of a channel that conveys each peer's node_signature and bitcoin_signature over the channel_announcement body, allowing the counterparty to assemble and broadcast the fully-signed channel_announcement.

Wire layout:
[channel_id:32]
[short_channel_id:8]
[signature:64]    // node_signature
[signature:64]    // bitcoin_signature

Codec only -- encode/decode and roundtrip tests; signing/verification happens at a higher layer (the IR `BuildAnnouncementSignatures` operation, in a follow-up PR) since the digest is over a different message.

Ref: #71

### DIFF
--- a/smite/src/bolt.rs
+++ b/smite/src/bolt.rs
@@ -5,6 +5,7 @@
 
 mod accept_channel;
 mod accept_channel2;
+mod announcement_signatures;
 mod attribution_data;
 mod channel_announcement;
 mod channel_ready;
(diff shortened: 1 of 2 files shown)


morehouse (approved):
LGTM, nice work

PR #123, opened 2026-06-12
smite-ir: implement RecvFundingSigned operation

NishantBansal2003:
This PR adds `RecvFundingSigned` operation, which receives a `funding_signed` message from the target and verifies that it is expected for the given `channel_id`. It also checks that a `funding_signed` message is not received when the opener cannot afford the feerate. Finally, it verifies that the commitment signature sent by the remote party is valid for the local commitment transaction.

I also took this PR as an opportunity to address the comment in: https://github.com/morehouse/smite/pull/97#discussion_r3369197594, since otherwise there would be an inconsistency between the `Display` implementations of `RecvFundingSigned` and `RecvAcceptChannel`

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -8,8 +8,8 @@ use bitcoin::secp256k1::{PublicKey, Secp256k1, SecretKey};
 use bitcoin::{OutPoint, ScriptBuf};
 use smite::bitcoin::{BitcoinCli, Utxo};
 use smite::bolt::{
-    AcceptChannel, ChannelAnnouncement, ChannelId, ChannelUpdate, FundingCreated, Message,
-    NodeAnnouncement, OpenChannel, OpenChannelTlvs, Pong, ShortChannelId, msg_type,
+    AcceptChannel, ChannelAnnouncement, ChannelId, ChannelUpdate, FundingCreated, FundingSigned,
+    Message, NodeAnnouncement, OpenChannel, OpenChannelTlvs, Pong, ShortChannelId, msg_type,
 };
 use smite::channel_tx::{
     ChannelConfig, ChannelPartyConfig, ChannelState, FundingTransaction, HolderIdentity, Side,
@@ -120,6 +120,19 @@ pub enum ExecuteError {
     /// Failed to construct the initial commitment state.
     #[error("commitment: {0}")]
     Commitment(#[from] smite::channel_tx::CommitmentError),
+
+    /// Received a `funding_signed` for a channel with no tracked state.
+    #[error("unknown channel: no tracked state for channel_id {0:?}")]
+    UnknownChannel(ChannelId),
+
+    /// The opener cannot afford the feerate for the commitment transaction.
+    #[error("opener cannot afford commitment fee for channel_id {0:?}")]
+    OpenerCannotAffordFee(ChannelId),
+
+    /// The counterparty's signature in `funding_signed` failed to verify
+    /// against the holder's first commitment transaction.
+    #[error("invalid counterparty signature for channel_id {0:?}")]
+    InvalidCounterpartySignature(ChannelId),
 }
 
 /// Executes IR programs against a target over an established connection.
@@ -158,10 +171,15 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
     ///
     /// # Errors
     ///
-    /// Returns an error on a connection/send/receive failure, a decode failure of
-    /// a received message, an unexpected message type from the target, when
-    /// wallet funds are insufficient to perform a channel operation, or when the
-    /// initial commitment transaction cannot be constructed.
+    /// Returns an error when:
+    /// - a connection/send/receive operation fails
+    /// - a received message fails to decode
+    /// - the target sends an unexpected message type
+    /// - wallet funds are insufficient to perform a channel operation
+    /// - the initial commitment transaction cannot be constructed
+    /// - no channel state exists for a received `funding_signed`
+    /// - the opener cannot afford the commitment feerate
+    /// - the counterparty's signature fails verification
     ///
     /// # Panics
     ///
@@ -316,6 +334,15 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                     let ac = recv_accept_channel(&mut self.conn)?;
                     log::debug!("[{:?}] RecvAcceptChannel: received", start.elapsed());
                     Some(Variable::AcceptChannel(ac))
+                }
+
+                Operation::RecvFundingSigned => {
+                    consume_sent_funding_created(&mut variables, instr.inputs[0]);
+                    log::debug!("[{:?}] RecvFundingSigned: waiting", start.elapsed());
+                    let fs = recv_funding_signed(&mut self.conn)?;
+                    log::debug!("[{:?}] RecvFundingSigned: received", start.elapsed());
+                    verify_funding_signed(&fs, &self.channel_states)?;
+                    Some(Variable::ChannelId(fs.channel_id))
                 }
 
                 Operation::MineBlocks(v) => {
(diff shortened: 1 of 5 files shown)


smite-scenarios/src/scenarios/ir.rs
```diff
@@ -91,6 +91,23 @@ impl<T: Target, S: SnapshotSetup<T>> Scenario for IrScenario<T, S> {
...
+                // The opener cannot afford the commitment feerate. This is a
+                // protocol violation by the target (it should not have sent
+                // funding_signed if the opener cannot afford the fee).
+                return ScenarioResult::Fail(format!("opener cannot afford fee: {id:?}"));
+            }
+            Err(ExecuteError::InvalidCounterpartySignature(id)) => {
+                // The target's funding_signed signature did not verify against
+                // the holder's commitment transaction. This is a protocol
+                // violation by the target.
+                return ScenarioResult::Fail(format!("invalid counterparty signature: {id:?}"));
```
morehouse: For these two failure messages it's not immediately clear that bytes at the end are the channel ID.


```suggestion
            Err(ExecuteError::OpenerCannotAffordFee(id)) => {
                // The opener cannot afford the commitment feerate. This is a
                // protocol violation by the target (it should not have sent
                // funding_signed if the opener cannot afford the fee).
                return ScenarioResult::Fail(format!("opener cannot afford fee for channel {id:?}"));
            }
            Err(ExecuteError::InvalidCounterpartySignature(id)) => {
                // The target's funding_signed signature did not verify against
                // the holder's commitment transaction. This is a protocol
                // violation by the target.
                return ScenarioResult::Fail(format!("invalid counterparty signature for channel {id:?}"));
```

smite-scenarios/src/executor.rs
```diff
@@ -160,8 +173,10 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
     ///
     /// Returns an error on a connection/send/receive failure, a decode failure of
     /// a received message, an unexpected message type from the target, when
-    /// wallet funds are insufficient to perform a channel operation, or when the
-    /// initial commitment transaction cannot be constructed.
+    /// wallet funds are insufficient to perform a channel operation, when the
+    /// initial commitment transaction cannot be constructed, when no channel state
+    /// exists for a received `funding_signed`, when the opener cannot afford the
+    /// commitment feerate, or when the counterparty's signature fails verification.
```
morehouse: Perhaps we should turn this paragraph into a bulleted list.

smite-scenarios/src/executor.rs
```diff
@@ -120,6 +120,19 @@ pub enum ExecuteError {
...
     #[error("commitment: {0}")]
     Commitment(#[from] smite::channel_tx::CommitmentError),
+
+    /// Received a `funding_signed` for a channel with no tracked state.
+    #[error("unknown channel: no tracked state for channel_id {0:?}")]
+    UnknownChannel(ChannelId),
+
+    /// The opener cannot afford the feerate for the commitment transaction.
+    #[error("opener cannot afford commitment fee for channel_id {0:?}")]
+    OpenerCannotAffordFee(ChannelId),
```
morehouse, line 130: According to the spec, a correct target should reject this case immediately after receiving the invalid `open_channel`.  So it seems that this check should really live in the handling of `RecvAcceptChannel`.

I think it would be a good future PR to implement an `open_channel`/`accept_channel` oracle that checks whether the target properly validates the channel parameters according to the spec.

smite-scenarios/src/executor.rs
```diff
@@ -2012,22 +2090,42 @@ mod tests {
...
+        // the opener's commitment.
+        let channel_id = ChannelId::v1_from_funding_outpoint(OutPoint {
+            txid: "09b0549b35f14ee862f63bd75811c6c27963c4dea6766ec6836952ec78df1e7e"
+                .parse()
+                .unwrap(),
+            vout: 0,
+        });
+        let fs_bytes = Message::FundingSigned(FundingSigned {
+            channel_id,
+            signature: "304402203dbf3dbf337b042a72576488c1fb019086089d8d790a47f652346cff2511b6e70220395fdf700cb82b0abfcfe8e0b7c822181f2ee72409c82c3ff8e04e36593662c7".parse().unwrap(),
```
morehouse, line 2125: How was this signature generated?  Would it make sense to compute the signature in this test itself for clarity?

smite-scenarios/src/executor.rs
```diff
@@ -2120,6 +2214,113 @@ mod tests {
...
+
+        let channel_id = ChannelId::v1_from_funding_outpoint(OutPoint {
+            txid: "09b0549b35f14ee862f63bd75811c6c27963c4dea6766ec6836952ec78df1e7e"
+                .parse()
+                .unwrap(),
+            vout: 0,
+        });
+        let fs_bytes = Message::FundingSigned(FundingSigned {
+            channel_id,
+            signature: "304402203dbf3dbf337b042a72576488c1fb019086089d8d790a47f652346cff2511b6e70220395fdf700cb82b0abfcfe8e0b7c822181f2ee72409c82c3ff8e04e36593662c7".parse().unwrap(),
```
morehouse: I believe this signature is also invalid now due to the changed `push_msat` below.  The only reason the test passes is because the fee check happens before the signature check.

It would be more robust to use a valid signature here.

smite-scenarios/src/executor.rs
NishantBansal2003: Yeah, ideally that’s the correct place to perform the check. However, we currently key the state by `ChannelId`, which is derived from the funding outpoint, and we don’t know the funding outpoint when processing `RecvAcceptChannel`. Even if we were to verify it manually at that stage, we’d still need `funding_satoshis` from the corresponding `open_channel` message to fully validate it. Because of that, we first need a way to map an `accept_channel` response back to its corresponding `open_channel` message. The verification in `funding_signed` is effectively doing that mapping indirectly, which is why I placed the check there.

But yeah, we could definitely move this validation to immediately after `RecvAcceptChannel` if we maintain that mapping explicitly. I think we can revisit this in the future and keep the current approach for the MVP?

smite-scenarios/src/executor.rs
NishantBansal2003: All the inputs used in the executor tests are taken directly from `funding.rs` and `commitment.rs`. For `funding_signed`, I also used the same LDK test oracle that is used in `commitment.rs`.

We could definitely compute the signature directly in this test, but I didn’t do that initially because it would introduce a fair amount of redundant code. We’d have to reconstruct the `channel_config` and `commitment_state` again, even though the executor already does that internally. Instead, I used the LDK oracle and hardcoded the expected signature. So, I’m fine with computing the signature in the test instead of hardcoding it if you think that’s the better approach.

smite-scenarios/src/executor.rs
morehouse: Sounds good.  Maybe we could create a `OpenAcceptChannelOracle` object that we inform about every `open_channel` and `accept_channel`, and the oracle could manage the mapping itself via `temporary_channel_id`...

smite-scenarios/src/executor.rs
morehouse: I think using the LDK oracle is fine.  We should just make sure we add comments explaining how the values were computed.

smite-scenarios/src/executor.rs
```diff
@@ -2012,22 +2093,45 @@ mod tests {
...
+        // the opener's commitment.
+        let channel_id = ChannelId::v1_from_funding_outpoint(OutPoint {
+            txid: "09b0549b35f14ee862f63bd75811c6c27963c4dea6766ec6836952ec78df1e7e"
+                .parse()
+                .unwrap(),
+            vout: 0,
+        });
+
+        // The expected signature for the `funding_signed` message sent by the
+        // acceptor is generated by the LDK program.
```
morehouse: Nit:  the reader will wonder what "the LDK program" is.  Maybe wording like this is clearer?


```suggestion
        // The expected signature here was computed using LDK as the source of
        // truth.
```

Also can be applied to the other expected signatures.

morehouse (commented):
Looks good, ready to squash.

morehouse (approved):
LGTM

PR #124, opened 2026-06-13
smite-ir: implement `BuildAnnouncementSignatures` operation

devvaansh:
Adds the IR operation that builds a signed `announcement_signatures` message (BOLT 7, type 259).
Per BOLT 7, the two signatures cover the `channel_announcement` body for the same channel, so the op takes the same body-defining inputs as `BuildChannelAnnouncement` plus a `channel_id` for the message itself. The executor reconstructs the body, signs it with all four keys, and emits the slot-1 signatures (treating slot-1 keys as the "local" side — consistent with our agreed-upon "no sort, ~50% wrong order on purpose" approach from #71).

Inputs (8): `channel_id`, `features`, `chain_hash`, `short_channel_id`, `node_sk_1`,` node_sk_2`,` bitcoin_sk_1`, `bitcoin_sk_2`.

Builds on #121 (`announcement_signatures` codec). Will rebase onto fresh master once that lands.

Ref: #71

### DIFF
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -179,6 +179,23 @@ pub enum Operation {
     ///   9: `fee_proportional_millionths` (`ForwardingFee`)
     ///  10: `htlc_maximum_msat` (`Amount`)
     BuildChannelUpdate,
+    /// Build an `announcement_signatures` message (BOLT 7, type 259).
+    ///
+    /// Signs the `channel_announcement` body with our node and bitcoin secret
+    /// keys. The body is constructed with pubkeys sorted in the lexicographic
+    /// order required by BOLT 7, using the target's public keys (inputs 5 and
+    /// 7) directly — we do not need the target's secret keys.
+    ///
+    /// Inputs (8):
+    ///   0: `channel_id`       (`ChannelId`)
+    ///   1: `features`         (`Features`)
+    ///   2: `chain_hash`       (`ChainHash`)
+    ///   3: `short_channel_id` (`ShortChannelId`)
+    ///   4: `node_sk_1`        (`PrivateKey`) -- our node signing key
+    ///   5: `node_id_2`        (`Point`)      -- target's node public key
+    ///   6: `bitcoin_sk_1`     (`PrivateKey`) -- our bitcoin signing key
+    ///   7: `bitcoin_key_2`    (`Point`)      -- target's bitcoin public key
+    BuildAnnouncementSignatures,
 
     // -- Act: side effects against the target --
     /// Send an encoded message over the connection.
@@ -648,6 +665,7 @@ impl fmt::Display for Operation {
                 format_hex(alias),
             ),
             Self::BuildChannelUpdate => write!(f, "BuildChannelUpdate"),
+            Self::BuildAnnouncementSignatures => write!(f, "BuildAnnouncementSignatures"),
             Self::SendMessage => write!(f, "SendMessage"),
             Self::SendOpenChannel => write!(f, "SendOpenChannel"),
             Self::SendFundingCreated => write!(f, "SendFundingCreated"),
@@ -684,7 +702,8 @@ impl Operation {
             Self::BuildFundingCreated => Some(VariableType::FundingCreatedMessage),
             Self::BuildChannelAnnouncement
             | Self::BuildNodeAnnouncement { .. }
-            | Self::BuildChannelUpdate => Some(VariableType::Message),
+            | Self::BuildChannelUpdate
+            | Self::BuildAnnouncementSignatures => Some(VariableType::Message),
             Self::SendMessage | Self::MineBlocks(_) | Self::BroadcastTransaction => None,
             Self::SendOpenChannel => Some(VariableType::SentOpenChannel),
             Self::SendFundingCreated => Some(VariableType::SentFundingCreated),
@@ -806,6 +825,17 @@ impl Operation {
                 VariableType::ForwardingFee,  // fee_proportional_millionths
                 VariableType::Amount,         // htlc_maximum_msat
             ],
+
+            Self::BuildAnnouncementSignatures => vec![
+                VariableType::ChannelId,      // channel_id
+                VariableType::Features,       // features
+                VariableType::ChainHash,      // chain_hash
+                VariableType::ShortChannelId, // short_channel_id
+                VariableType::PrivateKey,     // node_sk_1 (our node signing key)
+                VariableType::Point,          // node_id_2 (target's node public key)
+                VariableType::PrivateKey,     // bitcoin_sk_1 (our bitcoin signing key)
+                VariableType::Point,          // bitcoin_key_2 (target's bitcoin public key)
+            ],
         }
     }
 
@@ -841,6 +871,7 @@ impl Operation {
             | Self::BuildChannelAnnouncement
             | Self::BuildNodeAnnouncement { .. }
             | Self::BuildChannelUpdate
+            | Self::BuildAnnouncementSignatures
             | Self::SendMessage
             | Self::SendOpenChannel
             | Self::SendFundingCreated
@@ -890,7 +921,8 @@ impl Operation {
             | Self::BuildOpenChannel
             | Self::BuildFundingCreated
             | Self::BuildNodeAnnouncement { .. }
-            | Self::BuildChannelUpdate => false,
+            | Self::BuildChannelUpdate
+            | Self::BuildAnnouncementSignatures => false,
         }
     }
 
@@ -925,6 +957,7 @@ impl Operation {
             | Self::BuildFundingCreated
             | Self::BuildChannelAnnouncement
             | Self::BuildChannelUpdate
+            | Self::BuildAnnouncementSignatures
             | Self::SendMessage
             | Self::SendOpenChannel
             | Self::SendFundingCreated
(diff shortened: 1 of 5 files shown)


smite-ir/src/operation.rs
```diff
@@ -179,6 +179,25 @@ pub enum Operation {
...
+    ///
+    /// Inputs (8):
+    ///   0: `channel_id`       (`ChannelId`)
+    ///   1: `features`         (`Features`)
+    ///   2: `chain_hash`       (`ChainHash`)
+    ///   3: `short_channel_id` (`ShortChannelId`)
+    ///   4: `node_sk_1`        (`PrivateKey`) -- signs `node_signature`
+    ///   5: `node_sk_2`        (`PrivateKey`) -- derives `node_id_2`
+    ///   6: `bitcoin_sk_1`     (`PrivateKey`) -- signs `bitcoin_signature`
+    ///   7: `bitcoin_sk_2`     (`PrivateKey`) -- derives `bitcoin_key_2`
```
morehouse: When we are actually fuzzing the funding flow, we won't know the target's secret keys.  We should change this to use the target's public keys instead.

smite-scenarios/src/executor.rs
```diff
@@ -738,6 +745,64 @@ fn build_channel_announcement(
...
+    let node_sk_1 = SecretKey::from_slice(&node_sk_1_bytes).expect("valid private key");
+    let node_sk_2 = SecretKey::from_slice(&node_sk_2_bytes).expect("valid private key");
+    let bitcoin_sk_1 = SecretKey::from_slice(&bitcoin_sk_1_bytes).expect("valid private key");
+    let bitcoin_sk_2 = SecretKey::from_slice(&bitcoin_sk_2_bytes).expect("valid private key");
+
+    let secp = Secp256k1::new();
+    let node_id_1 = PublicKey::from_secret_key(&secp, &node_sk_1);
+    let node_id_2 = PublicKey::from_secret_key(&secp, &node_sk_2);
+    let bitcoin_key_1 = PublicKey::from_secret_key(&secp, &bitcoin_sk_1);
+    let bitcoin_key_2 = PublicKey::from_secret_key(&secp, &bitcoin_sk_2);
```
morehouse: For `announcement_signatures`, when we construct the `channel_announcement` I think we actually want to sort the pubkeys as described in BOLT 7, so we get valid signatures.

smite-scenarios/src/executor.rs
```diff
@@ -738,6 +745,64 @@ fn build_channel_announcement(
...
+        features,
+        chain_hash,
+        short_channel_id,
+        node_id_1,
+        node_id_2,
+        bitcoin_key_1,
+        bitcoin_key_2,
+        extra: Vec::new(),
+    };
+    ca.sign(&node_sk_1, &node_sk_2, &bitcoin_sk_1, &bitcoin_sk_2);
```
morehouse: We probably want a separate API for signing the `channel_announcement` with only *our* keys.

I think a `ChannelAnnouncement::signing_digest(&self) -> secp256k1::Message` or similar is probably best.  We can then sign the digest with whatever keys we need to.  Additionally we can probably deduplicate a bit of code in the existing `sign` and `verify` methods.

morehouse (changes requested):

smite-ir/src/operation.rs
devvaansh: Changed inputs 5 and 7 of `BuildAnnouncementSignatures` from `VariableType::PrivateKey` to `VariableType::Point`  also Updated the doc comment.

smite-scenarios/src/executor.rs
devvaansh: Totally agree - duplicating the hash logic across `sign()` and `verify()` was a little messy. I added `pub fn signing_digest(&self) -> secp256k1::Message` to `ChannelAnnouncement` so we have a single source of truth for the pre-image.
Refactored both `sign()` and `verify()` to call it internally - cleaned up the repeated `write_body()` + double-SHA256 pattern. Now `build_announcement_signatures()` just calls `ca.signing_digest()` directly and signs with our two keys without any code duplication.

 this also makes the API more flexible - callers can now get the digest and sign it however they need to, which is exactly what we needed here anyway.

smite-scenarios/src/executor.rs
devvaansh: makes sense - now `build_announcement_signatures()` sorts the pubkeys lexicographically (smaller into node_id_1) to match BOLT 7, so the signatures actually validate.

smite-ir/src/tests.rs
```diff
@@ -417,6 +417,83 @@ fn display_build_channel_update_program() {
...
+        },
+        Instruction {
+            operation: Operation::LoadPrivateKey(key(2)),
+            inputs: vec![],
+        },
+        // Target's bitcoin public key (input 7 to BuildAnnouncementSignatures).
+        Instruction {
+            operation: Operation::LoadTargetPubkeyFromContext,
+            inputs: vec![],
+        },
```
morehouse: Note this isn't actually the target's bitcoin pubkey, which would actually need to come from the `accept_channel` they send.  It would be good to add a comment mentioning this.

smite-scenarios/src/executor.rs
```diff
@@ -1545,6 +1614,153 @@ mod tests {
...
+        //  v1 = LoadFeatures
+        //  v2 = LoadChainHashFromContext
+        //  v3 = LoadShortChannelId
+        //  v4 = LoadPrivateKey(node_sk_1)     -- our node signing key
+        //  v5 = LoadPrivateKey(node_sk_2)     -- target's node key (derive pubkey from)
+        //  v6 = DerivePoint(v5)               -- node_id_2 (target's node pubkey)
+        //  v7 = LoadPrivateKey(bitcoin_sk_1)  -- our bitcoin signing key
+        //  v8 = LoadPrivateKey(bitcoin_sk_2)  -- target's bitcoin key (derive pubkey from)
+        //  v9 = DerivePoint(v8)               -- bitcoin_key_2 (target's bitcoin pubkey)
+        // v10 = BuildAnnouncementSignatures(v0,v1,v2,v3, v4,v6, v7,v9)
```
```suggestion
        // v10 = BuildAnnouncementSignatures(v0, v1, v2, v3, v4, v6, v7, v9)
```

smite-scenarios/src/executor.rs
```diff
@@ -1545,6 +1614,153 @@ mod tests {
...
+            node_id_2: n2,
+            bitcoin_key_1: bk1,
+            bitcoin_key_2: bk2,
+            extra: Vec::new(),
+        };
+        let digest = expected_ca.signing_digest();
+        let expected_node_sig = secp.sign_ecdsa(&digest, &node_sk_1);
+        let expected_bitcoin_sig = secp.sign_ecdsa(&digest, &bitcoin_sk_1);
+        assert_eq!(ann_sigs.node_signature, expected_node_sig);
+        assert_eq!(ann_sigs.bitcoin_signature, expected_bitcoin_sig);
```
morehouse: Rather than recomputing signatures and checking equality, we should construct the `ChannelAnnouncement` using the sigs from `announcement_signatures` plus the two other signatures and call `ChannelAnnouncement::verify`.

devvaansh: test now assembles the complete message with our sigs in the right slots and the counterparty's sigs (which we can compute since we control all keys here), then validates the whole thing.

smite-ir/src/tests.rs
```diff
@@ -417,6 +417,86 @@ fn display_build_channel_update_program() {
...
+        Instruction {
+            operation: Operation::LoadPrivateKey(key(2)),
+            inputs: vec![],
+        },
+        // A placeholder for the target's bitcoin pubkey (input 7 to
+        // BuildAnnouncementSignatures).  In real usage this would come from the
+        // `bitcoin_key` field in the target's `accept_channel` message, not from
+        // context; we use LoadTargetPubkeyFromContext here purely for illustration.
```
morehouse, line 456: `bitcoin_key` is not a field in `accept_channel`.


```suggestion
        // A placeholder for the target's bitcoin pubkey (input 7 to
        // BuildAnnouncementSignatures).  In real usage this would come from the
        // `funding_pubkey` field in the target's `accept_channel` message, not from
        // context; we use LoadTargetPubkeyFromContext here purely for illustration.
```

smite-scenarios/src/executor.rs
```diff
@@ -1545,6 +1614,174 @@ mod tests {
...
+                ann_sigs.node_signature,
+                secp.sign_ecdsa(&digest, &bitcoin_sk_2),
+                ann_sigs.bitcoin_signature,
+            )
+        };
+        ca.node_signature_1 = ns1;
+        ca.node_signature_2 = ns2;
+        ca.bitcoin_signature_1 = bs1;
+        ca.bitcoin_signature_2 = bs2;
+        assert!(ca.verify(), "channel_announcement signatures must be valid");
```
morehouse: Looking at this now, I think it would be cleaner to just directly verify the signatures we care about, rather than figuring out what locations to copy them to so that we can use `verify()`.


```suggestion
        let digest = ca.signing_digest();
        assert!(secp.verify_ecdsa(&digest, &ann_sigs.node_signature, &node_id_ours).is_ok());
        assert!(secp.verify_ecdsa(&digest, &ann_sigs.bitcoin_signature, &node_id_ours).is_ok());
```

morehouse (approved):

morehouse (approved):
LGTM

Issue #115, opened 2026-06-04
Effectiveness Evaluation Framework for Different Mutator Configurations


Chand-ra:
Use this statistical framework to rigorously compare two IR‑mutator configurations (e.g., “full stack” vs. “full stack + new operator”) across Lightning implementations. The primary metric is **edge coverage growth** over 24‑hour campaigns.

The framework is largely inspired by [*Evaluating Fuzz Testing (Klees et al., 2018)*](https://arxiv.org/abs/1808.09700).

## 1. What You Compare

Both configurations run the **IR fuzzing scenario** against the same Lightning targets, starting from the same snapshot. The only difference is the set of mutators.

| | Configuration A | Configuration B |
|---|---|---|
| **Mutators** | e.g., current Full Stack | e.g., Full Stack + your new operator |
| **Scenario** | IR scenario (`IrScenario`) | Same |
| **Snapshot** | `PostInitSetup` (or advanced, but identical for both) | Same |
| **Generators** | Existing suite | Same |
| **Seeds** | Empty corpus, or a small set of simple programs (see Section 2) | Same |
| **Engine** | AFL++ (latest stable) + Nyx, `fast` schedule | Same |

> **Why everything except the mutators is identical?**  
>
> We want to isolate the effect of the mutator changes. If we also changed the scenario, snapshot, or seeds, any observed difference in coverage growth could be caused by those factors rather than the mutators themselves. Keeping all other variables constant lets us attribute differences directly to the mutation strategy.

## 2. Seed Corpus
  
   The IR fuzzer includes generators that produce valid programs at runtime, an empty initial corpus is acceptable and often simplest. The generators will populate the queue with structurally sound seeds.

   _Alternatively_, you may start with a small set of simple IR programs (e.g., ping/pong, a minimal open_channel). These give a slight head start but are not strictly necessary.

>   **Important**: If the IR format changes (e.g., a new instruction set or a new postcard schema), all seeds must be regenerated from the new IR. Never mix seeds produced by different IR versions in the same comparison: that would introduce a confounding variable.

- **Same seeds for all trials:** Use the identical seed set for every trial (empty or not) and every configuration to ensure fairness.

## 3. Trial Structure

- **Number of trials:** 30 independent trials per configuration, per target. This number is recommended by *Klees et al. (2018)* to achieve reasonable statistical power for detecting meaningful differences.
- **Duration:** Each trial runs for 24 hours (wall‑clock). This is long enough to observe substantial coverage growth and plateaus.
[...]

PR #125, opened 2026-06-16
scripts: Add automated statistical evaluation script

Chand-ra:
Introduce `smite-evaluation.py` to handle the rigorous statistical evaluation of an experimental fuzzer configuration against a baseline.

The script automatically handles time-series interpolation, computes Mann-Whitney U tests and Vargha-Delaney A12 effect sizes for final coverage and AUC, and controls for family-wise error rates using the Holm-Bonferroni method. It outputs a Markdown report, CSV metrics, and visual diagnostics.

See issue #115 for more details.

### DIFF
--- a/scripts/smite-evaluation.py
+++ b/scripts/smite-evaluation.py
@@ -1,0 +1,661 @@
+#!/usr/bin/env python3
+"""
+Smite Fuzzing Evaluation Script
+
+This script automates the rigorous statistical evaluation of an experimental fuzzer
+configuration against a baseline, inspired by the framework from Klees et al. (2018).
+
+It is designed to evaluate any feature that impacts coverage-over-time capabilities
+(e.g., new mutators, custom schedulers, dictionary extraction, or seed selection).
+The only prerequisite is variable isolation: every configuration parameter must be
+identical between the baseline and the experimental configuration except for the
+specific feature being tested.
+
+Reference: https://github.com/morehouse/smite/issues/115
+
+Requirements:
+    pip install numpy pandas matplotlib seaborn scipy statsmodels tabulate
+
+Expected Directory Structure:
+The script expects a root directory containing the baseline and experimental subdirectories.
+Each configuration must contain identical target subdirectories, which in turn hold
+the trial directories containing AFL++ output.
+
+<root_dir>/
+├── configuration_1/                   # e.g., 'baseline'
+│   ├── target_1/                      # e.g., 'cln'
+│   │   ├── trial-01/
+│   │   │   └── afl-out/default/       # Contains fuzzer_stats, plot_data, fuzz_bitmap
+│   │   ├── trial-02/
+│   │   └── ...                        # Any number of trials (automatically counted)
+│   └── target_2/                      # e.g., 'lnd'
+└── configuration_2/                   # e.g., 'experimental'
+    ├── target_1/
+    └── target_2/
+
+Expected Output:
+The script will create a `results/` directory inside the provided root directory containing:
+- evaluation_report.md: A detailed Markdown report with summary statistics, adjusted
+  p-values, effect sizes, an interpretation guide, and embedded visualizations.
+- evaluation_metrics.csv: A complete data table of the summary statistics for external analysis.
+- <target>_boxplot.png: Box plots comparing the final edge coverage distributions.
+- <target>_auc_boxplot.png: Box plots comparing the Area Under Curve (exploration speed).
+- <target>_time_series.png: Line charts showing median coverage over time with IQR bands.
+
+Usage:
+    python smite-evaluation.py <path_to_smite_evaluation_dir> <baseline_dir_name> <experimental_dir_name> [--hours <hours>]
+
+Examples:
+    # Run strictly, ensuring all trials completed within 5% of each other
+    python smite-evaluation.py ./smite-evaluation baseline experimental
+
+    # Evaluate strictly up to 12 hours (fails if any trial ended before ~11.4h)
+    python smite-evaluation.py ./smite-evaluation baseline experimental --hours 12
+
+    # Force evaluation at the shortest common timeframe
+    python smite-evaluation.py ./smite-evaluation baseline experimental --hours min
+"""
+
+import os
+import argparse
+import numpy as np
+import pandas as pd
+import matplotlib.pyplot as plt
+import seaborn as sns
+from scipy import stats
+from statsmodels.stats.multitest import multipletests
+
+# Apply clean visual styling for all generated plots
+sns.set_style("whitegrid")
+
+
+def validate_and_find_data(root_dir, config_a, config_b):
+    """
+    Scans the root directory to find configs, targets, and trial paths.
+    Throws errors if the structure doesn't match the expected layout.
+    """
+    if not os.path.exists(root_dir):
+        raise FileNotFoundError(f"Root directory '{root_dir}' does not exist.")
+
+    path_a = os.path.join(root_dir, config_a)
+    path_b = os.path.join(root_dir, config_b)
+
+    if not os.path.isdir(path_a):
+        raise FileNotFoundError(
+            f"Baseline configuration directory '{config_a}' not found in {root_dir}"
+        )
+    if not os.path.isdir(path_b):
+        raise FileNotFoundError(
+            f"Experimental configuration directory '{config_b}' not found in {root_dir}"
+        )
+
+    print(
+        f"[*] Configurations: Baseline (A) = {config_a}, Experimental (B) = {config_b}"
+    )
+
+    # Find targets
+    targets_a = set(os.listdir(path_a))
+    targets_b = set(os.listdir(path_b))
+
+    if targets_a != targets_b:
+        raise ValueError(
+            f"Target mismatch! {config_a} has {targets_a}, but {config_b} has {targets_b}"
+        )
+
+    targets = sorted(list(targets_a))
+    print(f"[*] Detected Targets: {targets}")
+
+    data_paths = {config_a: {}, config_b: {}}
+
+    # Validate trials and files
+    for config in [config_a, config_b]:
+        for target in targets:
+            target_path = os.path.join(root_dir, config, target)
+            trials = [
+                d
+                for d in os.listdir(target_path)
+                if os.path.isdir(os.path.join(target_path, d))
+            ]
+
+            valid_trials = []
+            for trial in trials:
+                base_out = os.path.join(target_path, trial, "afl-out", "default")
+                if not os.path.exists(base_out):
+                    base_out = os.path.join(target_path, trial)
+
+                req_files = ["plot_data", "fuzzer_stats", "fuzz_bitmap"]
+                if all(os.path.exists(os.path.join(base_out, f)) for f in req_files):
+                    valid_trials.append(base_out)
+                else:
+                    print(
+                        f"[!] Warning: Missing required files in {os.path.join(target_path, trial)}. Skipping."
+                    )
+
+            data_paths[config][target] = valid_trials
+            print(f"    - {config}/{target}: found {len(valid_trials)} valid trials")
+
+    return targets, data_paths
+
+
+def parse_plot_data(filepath):
+    """Parses plot_data and returns (times_in_hours, coverage)."""
+    with open(filepath, "r") as f:
+        header_line = f.readline().strip()
+
+    if not header_line:
+        return np.array([0.0]), np.array([0.0])
+
+    header = [col.strip() for col in header_line.lstrip("#").split(",")]
+    df = pd.read_csv(
+        filepath, comment="#", sep=",", header=None, names=header, skipinitialspace=True
+    )
+
+    if len(df) == 0:
+        return np.array([0.0]), np.array([0.0])
+
+    if "relative_time" not in header or "edges_found" not in header:
+        raise ValueError(
+            f"AFL++ >= 3.10c required. plot_data must contain 'relative_time' and 'edges_found'. Found: {header}"
+        )
+
+    # Coerce to float to prevent string errors
+    times = pd.to_numeric(df["relative_time"], errors="coerce").fillna(0).values
+    coverage = pd.to_numeric(df["edges_found"], errors="coerce").fillna(0).values
+
+    # relative_time is already absolute seconds since fuzzer start. Convert directly to hours.
+    relative_times_hrs = times / 3600.0
+
+    return relative_times_hrs, coverage
+
+
+def parse_fuzzer_stats(filepath):
+    """Extracts execs_per_sec from fuzzer_stats."""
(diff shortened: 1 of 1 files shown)


Chand-ra:
Here is an example of what the final evaluation report looks like:

# Fuzzing Evaluation Report

**Configuration B (Experimental):** `ir-full-stack`

## 1. Summary Statistics

| cln      |             24 |             10 |         10 |                   1544   |               1773.5 |           0.000544953 |            1 |                 32339   |             37145.5 |          0.000548015 |           1 |                    4268 |                6760 |              245.48  |          236.215 |
| ldk      |             24 |             10 |         10 |                   2457.5 |               2838.5 |           0.000544953 |            1 |                 52379.2 |             59763.8 |          0.000548015 |           1 |                    6538 |                9883 |              250.87  |          236.815 |
| lnd      |             24 |             10 |         10 |                   3013   |               3291.5 |           0.000544953 |            1 |                 64685.5 |             70097.6 |          0.000548015 |           1 |                    8301 |               10811 |              250.605 |          237.88  |

## 2. Interpretation Guide

### Key Metrics

- **Union Coverage**: OR of all trial bitmaps; the coverage ceiling for a multi-core deployment. Descriptive only, cannot be statistically tested.

### Reading the Results

|---|---|---|

## 3. Visualizations

### Target: cln

cln_time_series

#### Distribution Comparisons

|:---:|:---:|
| cln_boxplot | cln_auc_boxplot |

---

### Target: ldk

ldk_time_series

#### Distribution Comparisons

|:---:|:---:|


---

### Target: lnd

lnd_time_series

#### Distribution Comparisons

|:---:|:---:|
| lnd_boxplot | lnd_auc_boxplot |

---

scripts/smite-evaluation.py
```diff
@@ -0,0 +1,539 @@
...
+    time_col = (
+        "relative_time"
+        if "relative_time" in header
+        else ("unix_time" if "unix_time" in header else None)
+    )
+    cov_col = (
+        "edges_found"
+        if "edges_found" in header
+        else ("map_size" if "map_size" in header else None)
+    )
```
morehouse: My `plot_data` with AFL++ v5.00c has `relative_time` and `edges_found`.  It also has a `map_size` column that gives a percentage value.

What versions of AFL++ have the other column names?  If it's only versions that are "old enough", perhaps we don't need to support them.

scripts/smite-evaluation.py
```diff
@@ -0,0 +1,539 @@
...
+    # Coerce to float to prevent string errors
+    times = pd.to_numeric(df[time_col], errors="coerce").fillna(0).values
+
+    # Strip '%' if it exists (for map_size fallbacks) and coerce to float
+    if df[cov_col].dtype == object:
+        df[cov_col] = df[cov_col].astype(str).str.rstrip("%")
+    coverage = pd.to_numeric(df[cov_col], errors="coerce").fillna(0).values
+
+    # Times are normalized to start at 0, then converted to hours
+    relative_times_hrs = (times - times[0]) / 3600.0
```
morehouse: Maybe this is correct for `unix_time`, but my `plot_data` with `relative_time` appears to already start the timer at time 0, though the first data point isn't recorded until `relative_time=60`.

morehouse: If `plot_data` is empty, the `times[0]` deref will trigger an exception.  We check that `len(times) > 0` in `process_data` but not here.

scripts/smite-evaluation.py
```diff
@@ -0,0 +1,539 @@
...
+    """Extracts execs_per_sec from fuzzer_stats."""
+    with open(filepath, "r") as f:
+        for line in f:
+            if "execs_per_sec" in line:
+                return float(line.split(":")[1].strip())
+    return 0.0
+
+
+def calculate_union_coverage(trial_paths):
+    """Performs bitwise-AND across all trial bitmaps to calculate multi-core union coverage."""
```
morehouse: A comment explaining the inverted meaning of the bitmap bytes would be good here, so the reader doesn't think the union is calculated incorrectly.

scripts/smite-evaluation.py
```diff
@@ -0,0 +1,539 @@
...
+        if os.path.isdir(os.path.join(root_dir, d)) and d != "results"
+    ]
+    if len(items) != 2:
+        raise ValueError(
+            f"Expected exactly 2 configurations in {root_dir}, found {len(items)}: {items}"
+        )
+
+    config_a, config_b = sorted(items)
+    print(f"[*] Detected Configurations: A = {config_a}, B = {config_b}")
+
```
morehouse, line 95: `config_a` is always the baseline and `config_b` is always experimental, but it appears these are assigned solely based on the lexicographic sorting of the names...

This will surprise the user at some point.  We probably want a more user-friendly way to designate which config is the baseline.

scripts/smite-evaluation.py
```diff
@@ -0,0 +1,539 @@
...
+                target_data[config]["execs"].append(execs)
+                raw_time_series[config].append((times, covs))
+
+            target_data[config]["union"] = calculate_union_coverage(
+                data_paths[config][target]
+            )
+
+        # Resolve the evaluation horizon for this specific target
+        eval_hours = global_max_hrs
```
morehouse: Shouldn't we be using the global *minimum* time instead, since that is the only time duration for which we have truly comparable data?  Beyond that point, we have to extrapolate to compare, which kind of defeats the purpose.

morehouse:
The sample report looks great!

Just to be sure though -- the report contains fake data, right?  The edges-found over time graphs don't look like what I observe locally -- a 10-minute run of the IR scenario against LDK finds over 10k edges, and the first data point starts at 8k, not 0.

scripts/smite-evaluation.py
```diff
@@ -0,0 +1,539 @@
+#!/usr/bin/env python3
```
morehouse, line 1: This script should also be marked as executable.

Chand-ra:
> The sample report looks great!
> 
> Just to be sure though -- the report contains fake data, right? The edges-found over time graphs don't look like what I observe locally -- a 10-minute run of the IR scenario against LDK finds over 10k edges, and the first data point starts at 8k, not 0.

Oh yes, I generated some fake data locally using a script for all the six (configuration, target) pairs. It is definitely too clean compared to what you'd get with a real run.

scripts/smite-evaluation.py
Chand-ra: I don't think that would be right.

Say we run 60 trials (30 baseline, 30 experimental) for 24 hours, and one of those trials terminates at hour 2 due to an OOM error or disk-full issue, the global minimum time becomes 2 hours. If we clamp the evaluation to the minimum, we throw away 22 hours of perfectly good data across the other 59 trials just to appease the worst-performing run.

What we're doing instead is [Last Observation Carried Forward](https://methods.sagepub.com/ency/edvol/encyc-of-research-design/chpt/last-observation-carried-forward#_). If a fuzzer dies at hour 14, its coverage growth plateaus at hour 14. We extend that flat line to 24 hours. This accurately punishes the crashed fuzzer without destroying the data of the surviving runs. It is a conservative bias; that is, it works against the hypothesis that the crashing configuration is better.

There _are_ better methods for handling missing data, like regression analysis, but that would be over-engineering for our use-case.

scripts/smite-evaluation.py
Chand-ra: The `relative_time` change was made in AFL++ v3.10c ([Changelog](https://github.com/AFLplusplus/AFLplusplus/blob/stable/docs/Changelog.md#version-313c-release)). I remember reading somewhere that older versions of AFL++ did not output `edges_found`, only `map_size`, but cannot find a concrete changelog online.

Since 3.10c was released in 2021 (5 years ago), I guess it's okay to remove support for versions older than that.

scripts/smite-evaluation.py
Chand-ra: Right, that was added for `unix_time` support. Since we're no longer supporting those versions, this should simplify to `relative_times_hrs = times / 3600.0`.

scripts/smite-evaluation.py
```diff
@@ -0,0 +1,552 @@
...
+The script will create a `results/` directory inside the provided root directory containing:
+- evaluation_report.md: A detailed Markdown report with summary statistics, adjusted
+  p-values, effect sizes, an interpretation guide, and embedded visualizations.
+- evaluation_metrics.csv: A complete data table of the summary statistics for external analysis.
+- <target>_boxplot.png: Box plots comparing the final edge coverage distributions.
+- <target>_auc_boxplot.png: Box plots comparing the Area Under Curve (exploration speed).
+- <target>_time_series.png: Line charts showing median coverage over time with IQR bands.
+
+Usage:
+    python3 evaluate.py <path_to_smite_evaluation_dir> <baseline_dir> <experimental_dir>
```
morehouse: We just want the name of the dir, not the full path.

```suggestion
    python3 evaluate.py <path_to_smite_evaluation_dir> <baseline_dir_name> <experimental_dir_name>
```

scripts/smite-evaluation.py
```diff
@@ -0,0 +1,552 @@
...
+    parser.add_argument(
+        "baseline_dir",
+        type=str,
+        help="Name of the baseline configuration directory (e.g., 'baseline').",
+    )
+    parser.add_argument(
+        "experimental_dir",
+        type=str,
+        help="Name of the experimental configuration directory (e.g., 'experimental').",
+    )
```
morehouse, line 641:
```suggestion
    parser.add_argument(
        "baseline_dir_name",
        type=str,
        help="Name of the baseline configuration directory (e.g., 'baseline').",
    )
    parser.add_argument(
        "experimental_dir_name",
        type=str,
        help="Name of the experimental configuration directory (e.g., 'experimental').",
    )
```

scripts/smite-evaluation.py
morehouse: If one trial crashes after just two hours, there's something going on that we need to figure out.  Hiding the crash by extrapolating data may actually prevent us from finding serious issues in our setup.

We are not "throwing away" the other data (it still exists) -- we're just excluding it from the generated report because we can't fairly compare a 24h run to a 2h run.

It's easy to imagine situations where we save fuzzing data that was generated over X hours and later run a different configuration over a shorter period of say 0.5X hours.  If we were to compare those configurations using the global max runtime X, the first config would pretty much always win by construction (it got to run for twice as long).  Obviously that's not a fair comparison.  But if compare them across the first 0.5X hours, that *is* fair.

TBH I've never seen a fuzzing paper that extrapolates fuzzer performance to "fill in the gaps".  Comparisons are only ever made across equal timeframes.

scripts/smite-evaluation.py
```diff
@@ -0,0 +1,552 @@
...
+- evaluation_metrics.csv: A complete data table of the summary statistics for external analysis.
+- <target>_boxplot.png: Box plots comparing the final edge coverage distributions.
+- <target>_auc_boxplot.png: Box plots comparing the Area Under Curve (exploration speed).
+- <target>_time_series.png: Line charts showing median coverage over time with IQR bands.
+
+Usage:
+    python3 evaluate.py <path_to_smite_evaluation_dir> <baseline_dir> <experimental_dir>
+
+Example:
+    python3 evaluate.py ./smite-evaluation baseline experimental
```
morehouse: Wrong script filename here and above.

scripts/smite-evaluation.py
```diff
@@ -0,0 +1,552 @@
...
+        f.write(
+            "- **`IQR`**: Spread of the middle 50% of trials. A much larger IQR in B suggests "
+            "a few outlier runs may be inflating the median.\n"
+        )
+        f.write(
+            "- **`AUC`**: Coverage *speed* — how much was discovered and how early. "
+            "Useful when final coverage is similar between configurations.\n"
+        )
+        f.write(
+            "- **Union Coverage**: OR of all trial bitmaps; the coverage ceiling for a "
```
morehouse:
```suggestion
            "- **Union Coverage**: Union of all trial bitmaps; the coverage ceiling for a "
```

scripts/smite-evaluation.py
Chand-ra: > If one trial crashes after just two hours, there's something going on that we need to figure out. Hiding the crash by extrapolating data may actually prevent us from finding serious issues in our setup.

If we want to do this, I say we add a check that ensures every trial has run for around the same amount of time (exactly equal, or within X% of global maximum run-time).

> We are not "throwing away" the other data (it still exists) -- we're just excluding it from the generated report because we can't fairly compare a 24h run to a 2h run.

That is true, the way we're doing it right now (with LOCF) provides a conservative bias against the shorter-runtime configuration. Whether that bias is right to assume or not, is definitely arguable.

> It's easy to imagine situations where we save fuzzing data that was generated over X hours and later run a different configuration over a shorter period of say 0.5X hours. If we were to compare those configurations using the global max runtime X, the first config would pretty much always win by construction (it got to run for twice as long). Obviously that's not a fair comparison. But if compare them across the first 0.5X hours, that _is_ fair.

Not really, some configuration might show a lot of advantage after seeing X amounts of data. This is especially true for complex configurations that naturally have low throughput (`execs/sec`).

For instance, some mutator scheduling algorithm might perform magnitudes better after it has seen enough data from a 15-16 hour long run, making it's final 24-h edge coverage exponentially higher than our baseline. Before that, it performs similar to the baseline scheduling configuration. Truncating the run-time of both configuration to < 15 hours then isn't really fair for the experimental configuration.

> TBH I've never seen a fuzzing paper that extrapolates fuzzer performance to "fill in the gaps". Comparisons are only ever made across equal timeframes.

True, every fuzzing paper that I've encountered doesn't even mention this as a potential problem. I guess they ensure the duration is uniform across all trials before even performing the statistics.

Maybe we should error out early on mismatched time-frames after all? We can also add a `--hours` flag in addition to this erroring out logic, which should allow the user to explicitly test shorter runtimes.

morehouse: > Not really, some configuration might show a lot of advantage after seeing X amounts of data. This is especially true for complex configurations that naturally have low throughput (execs/sec).

It is certainly true that some configurations will underperform on shorter time frames and outperform on longer time frames.  But I don't see how comparing across a 2x different timeframe with LOCF helps with that at all.  The only solution to that issue is to run *all* trials for a longer period.

> Maybe we should error out early on mismatched time-frames after all? We can also add a --hours flag in addition to this erroring out logic, which should allow the user to explicitly test shorter runtimes.

Sure, I think erroring out is a good default, with the user having the option to override it with an additional flag.  When overridden, I think it makes the most sense to compare across the common (minimum) time frame though, without LOCF.

morehouse (approved):
LGTM.  Used this to evaluate `GeneratorInsertionMutator`: https://github.com/morehouse/smite/pull/126#issuecomment-4780834528.

PR #126, opened 2026-06-19
smite-ir: Add `GeneratorInsertionMutator`

Chand-ra:
Add `GeneratorInsertionMutator` for smite-IR. Mutates a given program by inserting a generated program at a random point in the said program.

### DIFF
--- a/smite-ir-mutator/src/lib.rs
+++ b/smite-ir-mutator/src/lib.rs
@@ -37,15 +37,16 @@ use std::os::raw::{c_char, c_uint, c_void};
 use std::slice;
 
 use rand::rngs::SmallRng;
-use rand::{RngExt, SeedableRng};
+use rand::{RngExt, SeedableRng, seq::IteratorRandom};
 
 use smite_ir::generators::{
-    ChannelAnnouncementGenerator, ChannelUpdateGenerator, NodeAnnouncementGenerator,
+    AnyGenerator, ChannelAnnouncementGenerator, ChannelUpdateGenerator, NodeAnnouncementGenerator,
     OpenChannelGenerator,
 };
 use smite_ir::minimizers::{CommonSubexpressionEliminator, DeadCodeEliminator, Minimizer};
 use smite_ir::mutators::{
-    InputSwapMutator, InstructionDeleteMutator, InstructionReorderMutator, OperationParamMutator,
+    GeneratorInsertionMutator, InputSwapMutator, InstructionDeleteMutator,
+    InstructionReorderMutator, OperationParamMutator,
 };
 use smite_ir::{Generator, Mutator, Program, ProgramBuilder};
 
@@ -100,7 +101,7 @@ impl MutatorState {
         let stack = 1u32 << self.rng.random_range(0..=4);
         for _ in 0..stack {
             // Uniform pick between the available mutators.
-            let name = match self.rng.random_range(0..4) {
+            let name = match self.rng.random_range(0..5) {
                 0 => {
                     OperationParamMutator.mutate(program, &mut self.rng);
                     "op-param"
@@ -116,6 +117,15 @@ impl MutatorState {
                 3 => {
                     InstructionReorderMutator.mutate(program, &mut self.rng);
                     "instr-reorder"
+                }
+                4 => {
+                    let generator = *AnyGenerator::ALL
+                        .iter()
+                        .choose(&mut self.rng)
+                        .expect("AnyGenerator::ALL is non-empty");
+                    let mutator = GeneratorInsertionMutator::new(generator);
+                    mutator.mutate(program, &mut self.rng);
+                    "gen-insert"
                 }
                 _ => unreachable!("random_range() bound out of sync with match arms"),
             };
@@ -557,7 +567,8 @@ mod tests {
                     name == "op-param"
                         || name == "input-swap"
                         || name == "instr-delete"
-                        || name == "instr-reorder",
+                        || name == "instr-reorder"
+                        || name == "gen-insert",
                     "unexpected mutator name in description: {name:?} (full: {s:?})",
                 );
             }
--- a/smite-ir-mutator/src/lib.rs
+++ b/smite-ir-mutator/src/lib.rs
@@ -39,10 +39,7 @@ use std::slice;
 use rand::rngs::SmallRng;
 use rand::{RngExt, SeedableRng, seq::IteratorRandom};
 
-use smite_ir::generators::{
-    AnyGenerator, ChannelAnnouncementGenerator, ChannelUpdateGenerator, NodeAnnouncementGenerator,
-    OpenChannelGenerator,
-};
+use smite_ir::generators::AnyGenerator;
 use smite_ir::minimizers::{CommonSubexpressionEliminator, DeadCodeEliminator, Minimizer};
 use smite_ir::mutators::{
     GeneratorInsertionMutator, InputSwapMutator, InstructionDeleteMutator,
@@ -79,13 +76,11 @@ impl MutatorState {
     /// the registered generators.
     fn generate_fresh(&mut self) -> Program {
         let mut builder = ProgramBuilder::new();
-        match self.rng.random_range(0..4) {
-            0 => OpenChannelGenerator.generate(&mut builder, &mut self.rng),
-            1 => ChannelAnnouncementGenerator.generate(&mut builder, &mut self.rng),
-            2 => NodeAnnouncementGenerator.generate(&mut builder, &mut self.rng),
-            3 => ChannelUpdateGenerator.generate(&mut builder, &mut self.rng),
-            _ => unreachable!("random_range() bound out of sync with match arms"),
-        }
+        AnyGenerator::ALL
+            .iter()
+            .choose(&mut self.rng)
+            .expect("AnyGenerator::ALL is non-empty")
+            .generate(&mut builder, &mut self.rng);
         self.last_sequence.clear();
         self.last_sequence.push("fresh");
         builder.build()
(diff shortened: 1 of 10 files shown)


smite-ir/src/builder.rs
```diff
@@ -256,6 +256,10 @@ impl ProgramBuilder {
             } => self.append(operation, &[compound_idx]),
         }
     }
+
+    pub fn instructions_len(&mut self) -> usize {
```
morehouse: Needs a doc comment.  Also does not mutate self.


```suggestion
    /// Returns the current number of instructions.
    pub fn instructions_len(&self) -> usize {
```

smite-ir/src/mutators/generator_insert.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+}
+
+impl<G: Generator> Mutator for GeneratorInsertionMutator<G> {
+    fn mutate(&self, program: &mut Program, rng: &mut impl Rng) -> bool {
+        let insert_idx = if program.instructions.is_empty() {
+            0
+        } else {
+            rng.random_range(0..=program.instructions.len())
+        };
```
morehouse: The `is_empty` check is unnecessary.


```suggestion
        let insert_idx = rng.random_range(0..=program.instructions.len());
```

smite-ir/src/generators.rs
```diff
@@ -24,3 +24,34 @@ pub trait Generator {
     /// Emits instructions for this generator's protocol interaction.
     fn generate(&self, builder: &mut ProgramBuilder, rng: &mut impl Rng);
 }
+
+// A list of all the available generators. Must be included here
+// to be scheduled by `GeneratorInsertionMutator`.
+#[derive(Clone, Copy)]
+pub enum AvailableGenerators {
```
morehouse: Nit: I think the name `AnyGenerator` better describes the fact that this type encapsulates any generator implementation


```suggestion
pub enum AnyGenerator {
```

smite-ir/src/generators.rs
```diff
@@ -24,3 +24,34 @@ pub trait Generator {
...
+    Cannounce(ChannelAnnouncementGenerator),
+    Cupdate(ChannelUpdateGenerator),
+    Nannounce(NodeAnnouncementGenerator),
+    Open(OpenChannelGenerator),
```
morehouse: Nit: Let's spell out the variants for clarity


```suggestion
    ChannelAnnouncement(ChannelAnnouncementGenerator),
    ChannelUpdate(ChannelUpdateGenerator),
    NodeAnnouncement(NodeAnnouncementGenerator),
    OpenChannel(OpenChannelGenerator),
```

smite-ir-mutator/src/lib.rs
morehouse: We now might be able to simplify this code with `AnyGenerator::ALL.iter().choose()`.

smite-ir/src/tests.rs
```diff
@@ -2156,6 +2149,144 @@ fn instr_reorder_preserves_well_formedness() {
...
+    let cannounce_program_len = cannounce_program.instructions.len();
+
+    let mutator = GeneratorInsertionMutator::new(ChannelAnnouncementGenerator);
+    let mut rng = SmallRng::seed_from_u64(0);
+
+    mutator.mutate(&mut program, &mut rng);
+    assert_eq!(
+        original_program_len + cannounce_program_len,
+        program.instructions.len()
+    );
```
morehouse, line 2206: The new program length is not expected to always be the sum of the other two program lengths.  When `generate()` is called in the middle of the program rewrite it can `pick_variable`s from the existing program prefix, which ends up producing fewer instructions.

smite-ir/src/tests.rs
```diff
@@ -2156,6 +2149,144 @@ fn instr_reorder_preserves_well_formedness() {
...
+
+    mutator.mutate(&mut program, &mut rng);
+    assert_eq!(
+        program, open_channel_program,
+        "GeneratorInsertionMutator didn't insert generated program"
+    );
+}
+
+#[test]
+fn generator_insertion_preseves_generated() {
```
morehouse:
```suggestion
fn generator_insertion_preserves_generated() {
```

morehouse:
I ran an overnight experiment comparing a configuration with `GeneratorInsertionMutator` (gim) to a configuration without it (baseline).  The experiment wasn't rigorous (all trials were executed simultaneously on a machine that contained both P cores and E cores), but it does seem to support the hypothesis that this mutator is useful.

# Fuzzing Evaluation Report

**Configuration B (Experimental):** `gim`

## 1. Summary Statistics

| ldk      |        14.6636 |             13 |         13 |                    13190 |                14199 |            0.00766076 |     0.810651 |                  190769 |              200973 |            0.0402395 |    0.739645 |                   13544 |               15269 |               363.63 |           157.57 |

## 2. Interpretation Guide

### Key Metrics

- **Union Coverage**: Union of all trial bitmaps; the coverage ceiling for a multi-core deployment. Descriptive only, cannot be statistically tested.

### Reading the Results

|---|---|---|

## 3. Visualizations

### Target: ldk

ldk_time_series

#### Distribution Comparisons

|:---:|:---:|

---

smite-ir/src/generators.rs
```diff
@@ -24,3 +24,34 @@ pub trait Generator {
     /// Emits instructions for this generator's protocol interaction.
     fn generate(&self, builder: &mut ProgramBuilder, rng: &mut impl Rng);
 }
+
+// A list of all the available generators. Must be included here
+// to be scheduled by the IR mutator.
```
morehouse:
```suggestion
/// A list of all the available generators. Any generators included
/// here may be used by the custom mutator library.
```

smite-ir/src/generators.rs
```diff
@@ -24,3 +24,34 @@ pub trait Generator {
...
+    ChannelAnnouncement(ChannelAnnouncementGenerator),
+    ChannelUpdate(ChannelUpdateGenerator),
+    NodeAnnouncement(NodeAnnouncementGenerator),
+    OpenChannel(OpenChannelGenerator),
+}
+
+impl AnyGenerator {
+    /// All variants. Keep in sync with the enum definition.
+    pub const ALL: &[Self] = &[
+        Self::ChannelAnnouncement(ChannelAnnouncementGenerator),
```
morehouse, line 41: We should add an `any_generator_all_is_complete` test patterned after the other `*_all_is_complete` tests.

morehouse (approved):

Chand-ra:
> ## 1. Summary Statistics

> | Target   |   Duration (h) |   n (Baseline) |   n (Exp.) |   Median Cov. (Baseline) |   Median Cov. (Exp.) |   Adj. p-value (Cov.) |   Â12 (Cov.) |   Median AUC (Baseline) |   Median AUC (Exp.) |   Adj. p-value (AUC) |   Â12 (AUC) |   Union Cov. (Baseline) |   Union Cov. (Exp.) |   Execs/s (Baseline) |   Execs/s (Exp.) |
| ldk      |        14.6636 |             13 |         13 |                    13190 |                14199 |            0.00766076 |     0.810651 |                  190769 |              200973 |            0.0402395 |    0.739645 |                   13544 |               15269 |               363.63 |           157.57 |

Some thoughts:

- A p-value of ~0.007 means there's a 0.7% chance this result could appear just by luck and a Â12 of ~0.81 means GIM is better in 81% of random pairings.

  Pretty strong results, but `execs/sec` has very obviously dwindled quite a bit (by ~56%). However, we don't really statistically control `execs/sec` (AFL++ only writes a single value for execs/sec so it's not easily doable) so it's worth taking this with a grain of salt.

- The coverage graph tells a pretty interesting story as well. GIM was lagging quite a bit before the 4 hour mark (most likely due to it's low throughput) but then explodes past after that point. A longer trial might have shown even a greater magnitude of difference.

- Perhaps the most surprising point to me here is that the Union Coverage jumped from 13,544 to 15,269 (a gain of over 1,700 unique edges). It is most likely unlocking entirely new states in LDK that the baseline stack never reached across any of its 13 instances.

  This leads me to believe that inserting fresh programs/flows is paramount to the IR scenario's performance as a whole, and it might be worth reconsidering the probability with which we generate fresh programs ([currently at 5%](https://github.com/morehouse/smite/blob/master/smite-ir-mutator/src/lib.rs#L209-L211)).

- The only thing that bothers me is the high IQR spread for GIM (especially for AUC), although I suspect this has quite a lot to do with the performance vs. efficiency cores debacle that you just mentioned.

morehouse (approved):
LGTM

morehouse:
>     * Perhaps the most surprising point to me here is that the Union Coverage jumped from 13,544 to 15,269 (a gain of over 1,700 unique edges). It is most likely unlocking entirely new states in LDK that the baseline stack never reached across any of its 13 instances.
>       This leads me to believe that inserting fresh programs/flows is paramount to the IR scenario's performance as a whole, and it might be worth reconsidering the probability with which we generate fresh programs ([currently at 5%](https://github.com/morehouse/smite/blob/master/smite-ir-mutator/src/lib.rs#L209-L211)).

Could be a good experiment to run.  I suspect tuning the percentage won't make much difference, but I'm happy to be proven wrong.

I looked into the actual coverage diffs a bit.  The one big improvement I saw was that multiple gossip messages could be sent in the same program, which enabled the fuzzer to exercise more of the interactions between `channel_announcement`, `node_announcement`, and `channel_update`.  Implementations generally only process `node_announcement` and `channel_update` messages if they've seen a previous `channel_announcement` for the same node/channel.  We still don't pass the on-chain validation though, which will require further [improvements](https://github.com/morehouse/smite/issues/71#issuecomment-4793245346).

I didn't look closely at the `open_channel -> accept_channel` flow, but it's possible that we hit some new paths there as well, perhaps related to multiple in-flight channel opens and duplicate channel opens.

>     * The only thing that bothers me is the high IQR spread for GIM (especially for AUC), although I suspect this has quite a lot to do with the performance vs. efficiency cores debacle that you just mentioned.

The differing cores likely had an impact.  But the mutator also increases the program space that the fuzzer can explore, which I suspect also increases the variance.  Reduced execs/sec also might contribute to increased variance.

PR #127, opened 2026-06-19
smite: reject funding amounts exceeding max supply

NishantBansal2003:
A `funding_satoshis` value near `u64::MAX` would overflow when fees are added to compute the required amount, panicking instead of returning an error. Guard against this by rejecting amounts greater than `Amount::MAX_MONEY` up front, returning `InsufficientFunds`.

### DIFF
--- a/smite/src/channel_tx/funding.rs
+++ b/smite/src/channel_tx/funding.rs
@@ -54,7 +54,16 @@ pub fn build_funding_transaction(
     utxos: Vec<Utxo>,
     change_spk: ScriptBuf,
 ) -> Result<FundingTransaction, InsufficientFunds> {
+    // Amounts exceeding Bitcoin's maximum supply can never be funded. The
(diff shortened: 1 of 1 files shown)


morehouse:
Will this change prevent smite from finding issues like https://github.com/ElementsProject/lightning/issues/9225?

NishantBansal2003:
> Will this change prevent smite from finding issues like [ElementsProject/lightning#9225](https://github.com/ElementsProject/lightning/issues/9225)?

Nope. `BuildOpenChannel` and `CreateFundingTransaction` resolve `funding_satoshis` independently, so it’s possible to send one value in `open_channel` and use another when creating the funding transaction

Values exceeding 21M BTC should be rejected by `build_funding_transaction` anyway, but rejecting them earlier avoids overflow issues

morehouse (approved):
LGTM

PR #128, opened 2026-06-19
smite-ir: implement BuildChannelReady operation

NishantBansal2003:
This PR adds:

* Optional `opener_next_per_commitment_point` and `acceptor_next_per_commitment_point` fields to `ChannelState`, initialized from `channel_ready` and updated on each `revoke_and_ack`. When a `commitment_signed` is sent or received, the stored next point is absorbed into the current commitment state and cleared. This lays the groundwork for draining messages only when no next per-commitment point is available (ref: https://github.com/morehouse/smite/issues/111#issuecomment-4702876811).
* A `BuildChannelReady` operation that constructs `channel_ready` once the funding transaction reaches the required confirmation depth. It also adds an `include_alias` flag to control alias TLV inclusion. To avoid unnecessary updates, the next per-commitment point is only recorded when the commitment number is 0 and no point is already known.

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -8,9 +8,9 @@ use bitcoin::secp256k1::{PublicKey, Secp256k1, SecretKey};
 use bitcoin::{OutPoint, ScriptBuf};
 use smite::bitcoin::{BitcoinCli, Utxo};
 use smite::bolt::{
-    AcceptChannel, AnnouncementSignatures, ChannelAnnouncement, ChannelId, ChannelUpdate,
-    FundingCreated, FundingSigned, Message, NodeAnnouncement, OpenChannel, OpenChannelTlvs, Pong,
-    ShortChannelId, msg_type,
+    AcceptChannel, AnnouncementSignatures, ChannelAnnouncement, ChannelId, ChannelReady,
+    ChannelReadyTlvs, ChannelUpdate, FundingCreated, FundingSigned, Message, NodeAnnouncement,
+    OpenChannel, OpenChannelTlvs, Pong, ShortChannelId, msg_type,
 };
 use smite::channel_tx::{
     ChannelConfig, ChannelPartyConfig, ChannelState, FundingTransaction, HolderIdentity, Side,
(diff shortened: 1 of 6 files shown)


smite/src/channel_tx/commitment.rs
```diff
@@ -142,6 +150,37 @@ impl HolderIdentity {
...
+            acceptor_next_per_commitment_point: None,
+        }
+    }
+
+    /// Returns a mutable reference to the next per-commitment point, resolved
+    /// against the holder's side.
+    ///
+    /// `is_holder` selects whose point is returned: the holder's when `true`,
+    /// the counterparty's when `false`.
+    pub fn next_per_commitment_point_mut(&mut self, is_holder: bool) -> &mut Option<PublicKey> {
```
morehouse: Nit: it would be more readable to remove the `is_holder` bool and instead have two methods: `next_holder_per_commitment_point_mut` and `next_counterparty_per_commitment_point_mut`.  The code would be slightly simpler as well.

smite-scenarios/src/executor.rs
```diff
@@ -742,6 +749,39 @@ fn build_funding_created(
...
+    // redundant ones, so recording a resend would leave us with the wrong point
+    // and make us reject a valid received commitment signature as invalid.
+    if let Some(state) = channel_states.get_mut(&channel_id)
+        && state.commitment.commitment_number == 0
+    {
+        let next_point = state.next_per_commitment_point_mut(true);
+        if next_point.is_none() {
+            *next_point = Some(second_per_commitment_point);
+        }
+    }
```
morehouse, line 776: I think we'll want to only update `state` when we actually *send* `channel_ready`.  Otherwise we could build multiple different `channel_ready` messages and only send one of them.  And if the one we send doesn't match the one we recorded in `state`, we'll get unsynced with the target's state.  See https://github.com/morehouse/smite/issues/130#issuecomment-4791384996.

But since we also will need to do this for `funding_created`, it's probably better to do in follow-up PRs that directly address #130.

morehouse (approved):

morehouse (approved):
LGTM

PR #129, opened 2026-06-21
smite-ir: implement RecvChannelReady operation

NishantBansal2003:
Depends-on: #128 

Adds a `RecvChannelReady` operation. Receiving `channel_ready` from the target is different from other channel-opening messages: it is independent of whether we have sent our own `channel_ready`, and the target will only send it once the funding transaction has been broadcast and reached the `minimum_depth` requested in `accept_channel`.

To handle this, `RecvChannelReady` is made conditional. When executed, it may either receive a `channel_ready` or do nothing. A `channel_ready` is expected when a tracked channel is still at commitment number 0, the counterparty’s next per-commitment point is unknown, and the funding transaction has at least 8 confirmations. Since Lightning implementations typically use a default maximum `minimum_depth` of 8 in `accept_channel`, reaching 8 confirmations guarantees the target has sent its `channel_ready`

### DIFF
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -234,6 +234,13 @@ pub enum Operation {
     /// Produces the `ChannelId` carried in the message.
     /// TODO: Add `ExtractFundingSigned` when implementing force-close scenarios.
     RecvFundingSigned,
+    /// Receive and parse a `channel_ready` response.
+    ///
+    /// This is a no-op unless some tracked channel is awaiting `channel_ready`
+    /// (still at commitment number 0 with the counterparty's next per-commitment
+    /// point unknown) and its funding transaction has enough confirmations for
+    /// the target to have sent `channel_ready`.
+    RecvChannelReady,
     /// Mines the given number of blocks on the Bitcoin network.
     MineBlocks(u8),
     /// Sign wallet inputs of the transaction and broadcast it via `bitcoin-cli`.
@@ -692,6 +699,7 @@ impl fmt::Display for Operation {
             Self::SendFundingCreated => write!(f, "SendFundingCreated"),
             Self::RecvAcceptChannel => write!(f, "RecvAcceptChannel"),
             Self::RecvFundingSigned => write!(f, "RecvFundingSigned"),
+            Self::RecvChannelReady => write!(f, "RecvChannelReady()"),
             Self::BroadcastTransaction => write!(f, "BroadcastTransaction"),
         }
     }
@@ -726,7 +734,10 @@ impl Operation {
             | Self::BuildNodeAnnouncement { .. }
             | Self::BuildChannelUpdate
             | Self::BuildAnnouncementSignatures => Some(VariableType::Message),
-            Self::SendMessage | Self::MineBlocks(_) | Self::BroadcastTransaction => None,
+            Self::SendMessage
+            | Self::RecvChannelReady
+            | Self::MineBlocks(_)
+            | Self::BroadcastTransaction => None,
             Self::SendOpenChannel => Some(VariableType::SentOpenChannel),
             Self::SendFundingCreated => Some(VariableType::SentFundingCreated),
             Self::RecvAcceptChannel => Some(VariableType::AcceptChannel),
@@ -754,6 +765,7 @@ impl Operation {
             | Self::LoadChannelType(_)
             | Self::LoadTargetPubkeyFromContext
             | Self::LoadChainHashFromContext
+            | Self::RecvChannelReady
             | Self::MineBlocks(_) => vec![],
 
             Self::DerivePoint => vec![VariableType::PrivateKey],
@@ -905,6 +917,7 @@ impl Operation {
             | Self::SendOpenChannel
             | Self::SendFundingCreated
             | Self::RecvFundingSigned
+            | Self::RecvChannelReady
             | Self::MineBlocks(_)
             | Self::BroadcastTransaction => vec![],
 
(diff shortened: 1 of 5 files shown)


smite-ir/src/operation.rs
```diff
@@ -216,6 +234,13 @@ pub enum Operation {
     /// Produces the `ChannelId` carried in the message.
     /// TODO: Add `ExtractFundingSigned` when implementing force-close scenarios.
     RecvFundingSigned,
+    /// Receive and parse a `channel_ready` response.
+    ///
+    /// This is a no-op unless some tracked channel is awaiting `channel_ready`
+    /// (still at commitment number 0 with the counterparty's next per-commitment
+    /// point unknown) and its funding transaction has enough confirmations for
+    /// the target to have sent `channel_ready`.
+    RecvChannelReady,
```
morehouse, line 243: The executor automatically handles the `second_per_commitment_point` internally, but would it still be valuable to output the `ChannelReady` object, or maybe just the `alias` field?

smite-ir/src/operation.rs
```diff
@@ -671,6 +699,7 @@ impl fmt::Display for Operation {
             Self::SendFundingCreated => write!(f, "SendFundingCreated"),
             Self::RecvAcceptChannel => write!(f, "RecvAcceptChannel"),
             Self::RecvFundingSigned => write!(f, "RecvFundingSigned"),
+            Self::RecvChannelReady => write!(f, "RecvChannelReady"),
```
morehouse:
```suggestion
            Self::RecvChannelReady => write!(f, "RecvChannelReady()"),
```

smite-scenarios/src/executor.rs
```diff
@@ -954,6 +1011,57 @@ fn recv_funding_signed(conn: &mut impl Connection) -> Result<FundingSigned, Exec
...
+/// A `channel_ready` is expected when a tracked channel is still at commitment
+/// number 0, the counterparty's next per-commitment point is unknown, and the
+/// funding transaction has at least 8 confirmations. Since Lightning
+/// implementations typically use a default maximum `minimum_depth` of 8 in
+/// `accept_channel`, reaching 8 confirmations guarantees the target has sent
+/// its `channel_ready`.
+fn is_channel_ready_expected(
+    channel_states: &mut HashMap<ChannelId, ChannelState>,
+    bitcoin_cli: &mut impl BitcoinRpc,
+) -> bool {
```
morehouse, line 1059: `channel_states` isn't modified.


```suggestion
fn is_channel_ready_expected(
    channel_states: &HashMap<ChannelId, ChannelState>,
    bitcoin_cli: &mut impl BitcoinRpc,
) -> bool {
```

smite-scenarios/src/executor.rs
```diff
@@ -954,6 +1011,57 @@ fn recv_funding_signed(conn: &mut impl Connection) -> Result<FundingSigned, Exec
...
+    channel_states.values_mut().any(|state| {
+        state.commitment.commitment_number == 0
+            && state.next_per_commitment_point_mut(false).is_none()
```
morehouse: Ah, so this is why we use `&mut` reference to `channel_states` above.  Probably we want an immutable accessor as well, so we can avoid unnecessary `&mut` references.

smite-scenarios/src/executor.rs
```diff
@@ -954,6 +1011,57 @@ fn recv_funding_signed(conn: &mut impl Connection) -> Result<FundingSigned, Exec
...
+            && bitcoin_cli.get_transaction_confirmations(state.config.funding_outpoint.txid) >= 8
```
morehouse: Nit: 8 blocks should be a names constant


```suggestion
            && bitcoin_cli.get_transaction_confirmations(state.config.funding_outpoint.txid) >= MAX_MINIMUM_DEPTH
```

smite-ir/src/operation.rs
NishantBansal2003: I think we can output the `alias` field from the peer, but there are two issues:
- it is a tlv, so it may be absent
- the `RecvChannelReady` operation itself can be a no-op in some cases if no `channel_ready` is expected

So, if we want the `output_type` to be `VariableType::ShortChannelId`, we would need to output a default `alias` value whenever we are unable to extract one. Perhaps we could use the SCID 0 in those cases?

smite-scenarios/src/executor.rs
```diff
@@ -994,6 +1016,55 @@ fn recv_funding_signed(conn: &mut impl Connection) -> Result<FundingSigned, Exec
...
+/// funding transaction has at least [`MAX_MINIMUM_DEPTH`] confirmations.
+fn is_channel_ready_expected(
+    channel_states: &HashMap<ChannelId, ChannelState>,
+    bitcoin_cli: &mut impl BitcoinRpc,
+) -> bool {
+    channel_states.values().any(|state| {
+        state.commitment.commitment_number == 0
+            && state.next_counterparty_per_commitment_point().is_none()
+            && bitcoin_cli.get_transaction_confirmations(state.config.funding_outpoint.txid)
+                >= MAX_MINIMUM_DEPTH
```
morehouse, line 1064: In the future, once we start saving `open_channel` and `accept_channel` data, we could use the actual `minimum_depth` value from `accept_channel` instead.

smite-ir/src/operation.rs
morehouse: Ah, yeah that's a bit ugly.  We can figure it out later if we need it.

morehouse (approved):
Ready to merge after squashing the fixup

PR #131, opened 2026-06-23
smitebot: accept config file support to `doctor` and `build`

Ashish-Kumar-Dash:
## Summary

Follow-up to #116 , doctor and build now accept a campaign config file as a positional argument, replacing the need to pass paths manually via CLI flags.

CLI flags optionally override individual config values when provided:
```
 smitebot build campaign.toml
 smitebot build campaign.toml --target cln --coverage
 smitebot doctor campaign.toml
 smitebot doctor campaign.toml --aflpp-path ~/other
```
## Overview 
 - Makes the campaign config file optional for doctor and build,  users can run these commands with just CLI flags (e.g. `smitebot build --target lnd   --scenario encrypted_bytes`) or with a config file where CLI flags override individual values
  - When no config file is provided, `--target` and `--scenario` are required for build, `--aflpp-path` is required for doctor, and `--smite-dir` defaults to .
  - Updated README examples to show both standalone and config-based usage
  
  Ref. #70

### DIFF
--- a/smitebot/src/commands/build.rs
+++ b/smitebot/src/commands/build.rs
@@ -14,21 +14,24 @@ pub struct BuildCommand;
 /// CLI arguments for `smitebot build`.
 #[derive(Debug, Args)]
 pub struct BuildArgs {
-    /// Target implementation to build.
-    #[arg(long)]
-    target: Target,
-    /// Scenario binary selected by the workload Dockerfile.
-    #[arg(long)]
-    scenario: String,
+    /// Path to a campaign configuration TOML file. When provided, settings are
+    /// read from the file and CLI flags override individual values.
+    config: Option<PathBuf>,
+    /// Target implementation to build; overrides the config value.
+    #[arg(long, required_unless_present = "config")]
+    target: Option<Target>,
+    /// Scenario binary selected by the workload Dockerfile; overrides the config value.
+    #[arg(long, required_unless_present = "config")]
+    scenario: Option<String>,
     /// Build the coverage-instrumented Docker image.
     #[arg(long)]
     coverage: bool,
-    /// Override the Docker image tag.
+    /// Docker image tag; overrides the config value and the default smite naming convention.
     #[arg(long)]
     image: Option<String>,
-    /// Path to smite repository root.
-    #[arg(long, default_value = ".")]
-    smite_dir: PathBuf,
+    /// Path to smite repository root; overrides the config value.
+    #[arg(long)]
+    smite_dir: Option<PathBuf>,
     /// Pass --no-cache to docker build.
     #[arg(long)]
     no_cache: bool,
@@ -39,7 +42,7 @@ pub struct BuildArgs {
 pub struct BuildInputs {
     /// Docker image tag to produce.
     pub image: String,
-    /// Workload Dockerfile selected from `--target` and `--coverage`.
+    /// Workload Dockerfile selected from the resolved target and the coverage flag.
     pub dockerfile: PathBuf,
     /// Smite repository root used as the Docker build context.
     pub smite_dir: PathBuf,
@@ -65,25 +68,46 @@ impl BuildInputs {
         }
     }
 
-    /// Resolves Docker build inputs from parsed CLI arguments.
-    fn from_args(args: &BuildArgs) -> Self {
+    /// Resolves Docker build inputs from an optional config, overriding with CLI args.
+    ///
+    /// Clap enforces that `--target` and `--scenario` are present when no config
+    /// file is provided; `--smite-dir` defaults to the current directory.
+    fn resolve(config: Option<&CampaignConfig>, args: &BuildArgs) -> Self {
+        let target = args
+            .target
+            .or(config.map(|c| c.target))
+            .expect("clap ensures --target is present when config is absent");
+        let scenario = args
+            .scenario
+            .clone()
+            .or_else(|| config.map(|c| c.scenario.clone()))
+            .expect("clap ensures --scenario is present when config is absent");
+        let smite_dir = args
+            .smite_dir
+            .clone()
+            .or_else(|| config.map(|c| c.smite_dir.clone()))
+            .unwrap_or_else(|| PathBuf::from("."));
+
         let dockerfile_name = if args.coverage {
             "Dockerfile.coverage"
         } else {
             "Dockerfile"
         };
 
+        let image = args.image.clone().unwrap_or_else(|| {
+            config
+                .and_then(|c| c.image.clone())
+                .unwrap_or_else(|| default_workload_image_tag(target, &scenario, args.coverage))
+        });
+
         Self {
-            image: args.image.clone().unwrap_or_else(|| {
-                default_workload_image_tag(args.target, &args.scenario, args.coverage)
-            }),
-            dockerfile: args
-                .smite_dir
+            image,
+            dockerfile: smite_dir
                 .join("workloads")
-                .join(args.target.to_string())
+                .join(target.to_string())
                 .join(dockerfile_name),
-            smite_dir: args.smite_dir.clone(),
-            scenario: args.scenario.clone(),
+            smite_dir,
+            scenario,
             no_cache: args.no_cache,
         }
     }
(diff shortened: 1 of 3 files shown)


smitebot/src/commands/build.rs
```diff
@@ -14,19 +14,21 @@ pub struct BuildCommand;
 /// CLI arguments for `smitebot build`.
 #[derive(Debug, Args)]
 pub struct BuildArgs {
-    /// Target implementation to build.
+    /// Campaign configuration file or `--target --scenario`.
+    config: Option<PathBuf>,
+    /// Target implementation to build (required when no config file is given).
     #[arg(long)]
```
NishantBansal2003: `#[arg(long, conflicts_with = "config", required_unless_present = "config")]` I think this may reduce a lot of boilerplate regarding the exclusivity and required flags

smitebot/src/commands/build.rs
```diff
@@ -14,19 +14,21 @@ pub struct BuildCommand;
...
     /// Build the coverage-instrumented Docker image.
     #[arg(long)]
     coverage: bool,
     /// Override the Docker image tag.
     #[arg(long)]
     image: Option<String>,
-    /// Path to smite repository root.
+    /// Path to smite repository root (used with `--target --scenario`).
     #[arg(long, default_value = ".")]
     smite_dir: PathBuf,
```
NishantBansal2003: This is also present in the config file

smitebot/src/commands/build.rs
```diff
@@ -51,7 +53,7 @@ pub struct BuildInputs {
 
 impl BuildInputs {
     /// Resolves Docker build inputs from a campaign configuration.
-    pub fn from_config(config: &CampaignConfig, image: &str) -> Self {
+    pub(crate) fn from_config(config: &CampaignConfig, image: &str) -> Self {
```
NishantBansal2003: This was mentioned in the previous review as well. We generally don’t follow this pattern in the codebase, we either use `pub fn` or keep the function private

smitebot/src/commands/doctor.rs
```diff
@@ -43,13 +44,15 @@ pub struct DoctorCommand;
...
     json: bool,
-    /// Path to AFL++ source tree used for fuzzing.
+    /// Path to AFL++ source tree (required when no config file is given).
     #[arg(long)]
-    aflpp_path: PathBuf,
-    /// Path to smite repository root.
+    aflpp_path: Option<PathBuf>,
+    /// Path to smite repository root (used with `--aflpp-path`).
     #[arg(long, default_value = ".")]
     smite_dir: PathBuf,
```
NishantBansal2003, line 66: This is also present in the config file

NishantBansal2003 (changes requested):
I think that if we want to preserve the flags, they should override the corresponding fields in the config. This would make the config required while allowing flags to override specific values when provided. It simplifies the model, aligns with many established CLIs, and avoids the overhead of determining which options are required or mutually exclusive, WDYT?

smitebot/src/commands/build.rs
```diff
@@ -14,21 +14,23 @@ pub struct BuildCommand;
...
+    target: Option<Target>,
+    /// Scenario binary selected by the workload Dockerfile; overrides the config value.
     #[arg(long)]
-    scenario: String,
+    scenario: Option<String>,
     /// Build the coverage-instrumented Docker image.
     #[arg(long)]
     coverage: bool,
-    /// Override the Docker image tag.
+    /// Docker image tag; overrides the config value.
```
NishantBansal2003:
```suggestion
    /// Docker image tag; overrides the config value and the default smite naming convention.
```

smitebot/src/commands/build.rs
```diff
@@ -92,7 +94,18 @@ impl BuildInputs {
...
+            Err(e) => {
+                log::error!("{e}");
+                return false;
+            }
+        };
+        let target = args.target.unwrap_or(config.target);
+        let scenario = args.scenario.clone().unwrap_or(config.scenario);
+        let smite_dir = args.smite_dir.clone().unwrap_or(config.smite_dir);
+
+        let inputs = BuildInputs::from_parts(target, &scenario, &smite_dir, args);
```
NishantBansal2003: We can add `BuildInputs::resolve` to resolve the config first, then override fields with CLI args when provided. The image should also be overridden by the corresponding flag, if provided in config. This would also allow us to add unit tests verifying that config values are correctly overridden when args are present

smitebot/src/commands/doctor.rs
```diff
@@ -127,8 +130,17 @@ impl Serialize for CheckFailure {
...
+            Ok(c) => c,
+            Err(e) => {
+                log::error!("{e}");
+                return false;
+            }
+        };
+        let aflpp_root = args.aflpp_path.clone().unwrap_or(config.aflpp_path);
+        let smite_dir = args.smite_dir.clone().unwrap_or(config.smite_dir);
+        let aflpp_root = aflpp_root.as_path();
+        let smite_dir = &smite_dir;
```
NishantBansal2003: This can be removed and used explicitly wherever a borrow is required

NishantBansal2003 (changes requested):

smitebot/src/commands/build.rs
```diff
@@ -75,15 +87,17 @@ impl BuildInputs {
 
         Self {
             image: args.image.clone().unwrap_or_else(|| {
-                default_workload_image_tag(args.target, &args.scenario, args.coverage)
+                config
+                    .image
+                    .clone()
+                    .unwrap_or_else(|| default_workload_image_tag(target, &scenario, args.coverage))
```
NishantBansal2003: nit: Move this alongside target and scenario for consistency

smitebot/src/commands/build.rs
NishantBansal2003:
```suggestion
    /// Workload Dockerfile selected from the resolved target and the coverage flag.
```

smitebot/src/commands/build.rs
```diff
@@ -158,13 +179,33 @@ mod tests {
     use super::*;
     use std::path::Path;
 
-    fn sample_build_args(target: Target, scenario: &str) -> BuildArgs {
+    use crate::config::CampaignConfig;
```
NishantBansal2003: Not required, also can remove the blank line between imports

NishantBansal2003 (approved):

smitebot/src/commands/build.rs
```diff
@@ -14,21 +14,23 @@ pub struct BuildCommand;
 /// CLI arguments for `smitebot build`.
 #[derive(Debug, Args)]
 pub struct BuildArgs {
-    /// Target implementation to build.
+    /// Path to the campaign configuration TOML file.
+    config: PathBuf,
```
morehouse: `config` should be an optional arg.  The user should still be able to do `smitebot build --target lnd --scenario encrypted_bytes` without a config file.

smitebot/src/commands/doctor.rs
```diff
@@ -43,15 +44,17 @@ pub struct DoctorCommand;
 /// CLI arguments for `smitebot doctor`.
 #[derive(Debug, Args)]
 pub struct DoctorArgs {
+    /// Path to the campaign configuration TOML file.
+    config: PathBuf,
```
morehouse: `config` should also be optional for `doctor`

morehouse (changes requested):

smitebot/src/commands/doctor.rs
```diff
@@ -127,8 +132,29 @@ impl Serialize for CheckFailure {
...
+            c.aflpp_path.clone()
+        } else {
+            log::error!("--aflpp-path is required when no config file is provided");
+            return false;
+        };
+        let smite_dir = args
+            .smite_dir
+            .clone()
+            .or_else(|| config.as_ref().map(|c| c.smite_dir.clone()))
+            .unwrap_or_else(|| PathBuf::from("."));
```
morehouse: Maybe we should use the same `resolve` pattern that `build.rs` uses.  This would allow easy testing and make the pattern consistent.  We also wouldn't need to use the if-else chain for `aflpp_root` any more.

smitebot/src/commands/build.rs
```diff
@@ -14,21 +14,25 @@ pub struct BuildCommand;
 /// CLI arguments for `smitebot build`.
 #[derive(Debug, Args)]
 pub struct BuildArgs {
-    /// Target implementation to build.
+    /// Path to a campaign configuration TOML file. When provided, settings are
+    /// read from the file and CLI flags override individual values. When
+    /// omitted, --target and --scenario are required.
+    config: Option<PathBuf>,
+    /// Target implementation to build; overrides the config value.
     #[arg(long)]
```
morehouse: I think `required_unless_present = "config"` should probably be used for the various `build` and `doctor` flags.

morehouse (changes requested):

morehouse (approved):
Ready to merge after squash

PR #132, opened 2026-06-25
workloads: patch CLN to remove 5s delay on disconnect

morehouse:
While testing the new `GeneratorInsertionMutator`, I discovered that CLN would "hang" whenever a program was executed that attempted to open two channels simultaneously on the same connection.  It turns out that CLN doesn't support multiple in-flight channel opens and will send an error message and close the connection if a peer attempts to do this. But before disconnection, CLN waits up to 5s for the subdaemon to finish sending any final messages and shut down gracefully. Because openingd refuses to shut down when it's in the middle of a funding flow, we would have to wait the full 5s, which would register as a hang.

The workaround is to remove CLN's wait entirely, since we don't actually care about any messages after the error message.

For reference, the 5s timeout locations patched by this change are here:

https://github.com/ElementsProject/lightning/blob/daa3b06b95deb3aac717bca501e0c19aba9c3858/connectd/multiplex.c#L130-L132

https://github.com/ElementsProject/lightning/blob/daa3b06b95deb3aac717bca501e0c19aba9c3858/connectd/multiplex.c#L1340-L1342

https://github.com/ElementsProject/lightning/blob/daa3b06b95deb3aac717bca501e0c19aba9c3858/connectd/multiplex.c#L1417-L1419

The one location we specifically avoid patching is here:

https://github.com/ElementsProject/lightning/blob/daa3b06b95deb3aac717bca501e0c19aba9c3858/connectd/multiplex.c#L112-L114

### DIFF
--- a/workloads/cln/Dockerfile
+++ b/workloads/cln/Dockerfile
@@ -73,6 +73,33 @@ RUN git clone --depth 1 --branch ${CLN_VERSION} --recurse-submodules \
     https://github.com/ElementsProject/lightning.git /cln
 WORKDIR /cln
 RUN pip3 install --break-system-packages mako
+# Patch connectd to remove a 5s delay that occurs when disconnecting from peers.
+# This delay causes CLN to become unresponsive on the connection for 5s, which
+# kills our fuzzing speed when it occurs.
+#
+# The 5s timeout we need to remove occurs before connectd calls
+# close_subd_timeout, which is the function that force-closes the subdaemon fd
+# after an error message has been queued to the peer. In some cases the
+# subdaemon does not exit on its own (e.g., openingd will not exit if it's in
+# the middle of a funding flow), and we have to wait for the full 5s timeout
+# before close_subd_timeout is called. The only benefit of waiting is to give
+# the subdaemon time to queue any additional messages it wants to send after the
+# error message. But for our fuzzing purposes, we don't care about messages
+# after the error message -- once we receive the error we stop execution anyway.
+#
+# There's a separate 5s timeout that occurs before connectd force-closes the
+# connection socket via close_peer_io_timeout. We do *not* want to remove that
+# timeout since then the connection might get closed before we have a chance to
+# read the messages leading up to the error message. The sed command below is
+# designed to remove all of the relevant 5s timeouts while skipping the
+# close_peer_io_timeout one that we want to retain.
+#
+# If CLN changes this code significantly in the future, we will need to update
+# these commands accordingly.
+RUN sed -i '/peer->to_peer/!s/time_from_sec(5)/time_from_sec(0)/' connectd/multiplex.c && \
+    test "$(grep -c 'time_from_sec(0)' connectd/multiplex.c)" -eq 3 && \
+    test "$(grep -c 'time_from_sec(5)' connectd/multiplex.c)" -eq 1 && \
+    grep -A1 'time_from_sec(5)' connectd/multiplex.c | grep -q close_peer_io_timeout
 # AFL_UBSAN_VERBOSE=1 gives us helpful UBSan reports instead of simply SIGILL.
 ENV AFL_UBSAN_VERBOSE=1
 RUN CC=afl-clang-lto ./configure --prefix=/usr/local --disable-rust \


morehouse:
More context... This is what happens before this PR:

1. Executor sends `open_channel`.
2. lightningd spins up an openingd to handle the funding flow.  openingd sends `accept_channel`.
3. Executor waits for and receives `accept_channel`.
4. Executor sends `open_channel`.
6. lightningd detects that there's already a funding flow underway for this peer and tells connectd to send an error and disconnect
7. connectd queues the error for sending and starts a 5s timer after which it will close the openingd fd
8. Executor waits for `accept_channel` but receives the error instead and bails out of the program.
9. The connection is still open.  The scenario goes into the `ping_pong` loop to test whether CLN is still responsive.  Scenario sends `ping` and begins waiting for the `pong` (which never arrives).
10. openingd has no clue it's supposed to be shutting down (it only knows about the first `open_channel`)
11. After 5s, connectd closes the openingd fd
12. openingd shuts down
13. connectd starts another 5s timer while waiting for any remaining outgoing messages to be sent.  connectd closes the connection when the outgoing queue is drained, or when the 5s timer goes off.
14. Scenario detects the connection was closed and returns.  A hang is recorded.

This patch turns the 5s timer in (7) into a 0s timer that fires immediately, so that connectd immediately tells all subdaemons to shutdown.  This eliminates the 5s unresponsive period before the TCP connection is closed.  The second 5s wait in (13) is left alone so that CLN's outgoing queue is still properly drained before disconnect.

So this is what happens after this PR:

1. Executor sends `open_channel`.
2. lightningd spins up an openingd to handle the funding flow.  openingd sends `accept_channel`.
3. Executor waits for and receives `accept_channel`.
4. Executor sends `open_channel`.
6. lightningd detects that there's already a funding flow underway for this peer and tells connectd to send an error and disconnect
7. connectd queues the error for sending and immediately closes the openingd fd
8. openingd shuts down
9. Executor waits for `accept_channel` but receives the error instead and bails out of the program.
10. The scenario goes into the `ping_pong` loop to test whether CLN is still responsive.  Scenario sends `ping` and begins waiting for the `pong` (which never arrives).
11. connectd starts a 5s timer while waiting for any remaining outgoing messages to be sent.  Since there are no remaining messages to send, connectd closes the connection immediately.
12. Scenario detects the connection was closed and returns.  No hang

workloads/cln/Dockerfile
```diff
@@ -73,6 +73,33 @@ RUN git clone --depth 1 --branch ${CLN_VERSION} --recurse-submodules \
...
+# read the messages leading up to the error message. The sed command below is
+# designed to remove all of the relevant 5s timeouts while skipping the
+# close_peer_io_timeout one that we want to retain.
+#
+# If CLN changes this code significantly in the future, we will need to update
+# these commands accordingly.
+RUN sed -i '/peer->to_peer/!s/time_from_sec(5)/time_from_sec(0)/' connectd/multiplex.c && \
+    test "$(grep -c 'time_from_sec(0)' connectd/multiplex.c)" -eq 3 && \
+    test "$(grep -c 'time_from_sec(5)' connectd/multiplex.c)" -eq 1 && \
+    grep -A1 'time_from_sec(5)' connectd/multiplex.c | grep -q close_peer_io_timeout
```
ekzyis, line 102: Would it make sense to use `patch` instead of `sed`? The `test` and `grep` invocations seem to be checking whether the patch was applied cleanly.

morehouse: The idea was to increase the odds of the change withstanding updates to CLN (reduce required maintenance), since a patch will be invalidated by any changes to line numbers.

But I don't feel too strongly about this approach and can change it if you prefer a patch file.

FWIW there's also a place in the CLN coverage Dockerfile that we use `sed` like this:

https://github.com/morehouse/smite/blob/bf52f5fa81881812ed88fcaf02f4ae31bc0b9e4c/workloads/cln/Dockerfile.coverage#L68-L75

ekzyis: > But I don't feel too strongly about this approach and can change it if you prefer a patch file.

Nah, I think this is good enough. Was mostly just curious about the decision

NishantBansal2003 (approved):
Makes sense 👍

PR #133, opened 2026-06-26
smite: add NyxLogger for debugging in Nyx mode

morehouse:
Enables scenario output to be printed in Nyx mode when `SMITE_NYX_LOG` is enabled.

This was helpful for debugging a timing difference when running in Nyx vs Docker.

### DIFF
--- a/smite/src/nyx_log.rs
+++ b/smite/src/nyx_log.rs
@@ -1,0 +1,60 @@
+//! Logger that forwards `log` records to the Nyx host via `nyx_println`.
+//!
+//! In Nyx mode the guest's stdout/stderr is discarded, so `simple_logger`
+//! output never reaches the host. This logger instead routes each record
+//! through `nyx_println` for debugging. Honors the `RUST_LOG` log level.
+//!
+//! This logger is only installed when the `nyx` feature is active and both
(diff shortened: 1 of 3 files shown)


smite/src/nyx_log.rs
```diff
@@ -0,0 +1,67 @@
...
+        Some(v) => match v.trim().to_ascii_lowercase().as_str() {
+            "trace" => LevelFilter::Trace,
+            "debug" => LevelFilter::Debug,
+            "warn" => LevelFilter::Warn,
+            "error" => LevelFilter::Error,
+            "off" => LevelFilter::Off,
+            _ => LevelFilter::Info,
+        },
+        None => LevelFilter::Info,
+    }
```
NishantBansal2003: nit: I think we can reuse the built-in parser here, like `SimpleLogger` does:

```suggestion
    std::env::var("RUST_LOG")
        .ok()
        .and_then(|v| v.trim().parse().ok())
        .unwrap_or(LevelFilter::Info)
```

NishantBansal2003 (approved):

smite/src/nyx_log.rs
morehouse: Nice, thanks!  Done.

PR #134, opened 2026-06-28
multi: record channel negotiation state and fold build/send operations

### LINKED ISSUE #130: `invalid counterparty signature` false-positive crashes in funding flow generator
While fuzzing smite with the funding flow generator in commit: https://github.com/NishantBansal2003/smite/tree/funding-signed-generator, a large number of `invalid counterparty signature` crashes were observed.

Upon inspection, it appears that smite may falsely report a valid counterparty signature as invalid because signing and verification depend on different sets of parameters:

| Field | `sign_counterparty_commitment` | `verify_counterparty_signature` |
|-------|:------------------------------:|:-------------------------------:|
| OPENER `revocation_basepoint` | YES | |
| OPENER `delayed_payment_basepoint` | | YES |
| OPENER `per_commitment_point` | | YES |
| ACCEPTOR `revocation_basepoint` | | YES |
| ACCEPTOR `delayed_payment_basepoint` | YES | |
| ACCEPTOR `per_commitment_point` | YES | |

So, smite may generate a valid signature that the target successfully verifies, causing it to send a `funding_signed` message. However, if the parameters used by `verify_counterparty_signature` are not populated correctly, smite may incorrectly reject the incoming `funding_signed` signature as invalid, resulting in a false-positive crash.

The same issue applies to `to_self_delay` and `dust_limit_satoshis`, which are also split across the sign and verify paths.

```sh
INFO  [smite_scenarios::targets::bitcoind] Preserving data directory: /data
INFO  [smite_scenarios::targets::bitcoind] Starting bitcoind...
INFO  [smite_scenarios::targets::bitcoind] Waiting for bitcoind to be ready...
INFO  [smite_scenarios::targets::bitcoind] bitcoind is ready
INFO  [smite_scenarios::targets::ldk] Starting ldk-node-wrapper...
INFO  [smite_scenarios::targets::ldk] LDK identity pubkey: 034a3f7cb929a1ba1276b9aab5385aa3fb71083b1eff4e630b33600e19763ac8c2
INFO  [smite_scenarios::targets::ldk] ldk-node-wrapper is ready and synced
INFO  [smite_scenarios::targets::ldk] Both daemons are running, ready to fuzz
DEBUG [smite_scenarios::scenarios] Handshake complete, received target init
DEBUG [smite_scenarios::scenarios::ir] [6.428µs] Executing IR program (30 instructions, 231 input bytes)
DEBUG [smite_scenarios::executor] [74.309µs] SendOpenChannel: 350 bytes
DEBUG [smite_scenarios::executor] [101.135µs] RecvAcceptChannel: waiting
DEBUG [smite_scenarios::executor] [6.514909ms] RecvAcceptChannel: received
DEBUG [smite_scenarios::executor] [11.901285ms] SendFundingCreated: 132 bytes
DEBUG [smite_scenarios::executor] [11.920391ms] RecvFundingSigned: waiting
DEBUG [smite_scenarios::executor] [16.841031ms] RecvFundingSigned: received
ERROR [smite::runners] Test case failed: invalid counterparty signature for channel ChannelId([220, 234, 183, 205, 46, 28, 108, 142, 107, 119, 148, 82, 193, 4, 177, 244, 127, 168, 210, 167, 182, 242, 208, 242, 199, 155, 96, 185, 148, 194, 161, 8])
DEBUG [smite::process] ldk-node-wrapper: dropping running process, attempting shutdown
DEBUG [smite::process] ldk-node-wrapper: sending SIGTERM to process group 52
DEBUG [smite::process] ldk-node-wrapper: exited with exit status: 0
DEBUG [smite::process] bitcoind: dropping running process, attempting shutdown
DEBUG [smite::process] bitcoind: sending SIGTERM to process group 6
DEBUG [smite::process] bitcoind: exited with exit status: 0
```

NishantBansal2003:
Fixes: #130 

Store negotiated `open_channel`/`accept_channel` messages during execution so later commitment steps use the parameters actually exchanged on the wire.

This change is accompanied by `BuildFundingCreated` using the negotiated state to build the channel state used for signing and verification. Also, all the states are now recorded on send, so there is no state corruption between the target and smite if we store a different state from what was actually sent to the peer.

Mostly, this change is done so that we can extend this approach to add `open_channel <-> accept_channel` oracles, which require a mapping between these two messages. This change lays the groundwork so we can add those oracles in subsequent PRs.

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -299,17 +299,6 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                     Some(Variable::OpenChannelMessage(oc))
                 }
 
-                Operation::BuildChannelReady { include_alias } => {
-                    let cr = build_channel_ready(
-                        &variables,
-                        &instr.inputs,
-                        *include_alias,
-                        &mut self.channel_states,
-                    );
-                    let encoded = Message::ChannelReady(cr).encode();
-                    Some(Variable::Message(encoded))
-                }
-
                 Operation::BuildChannelAnnouncement => {
                     let ca = build_channel_announcement(&variables, &instr.inputs);
                     let encoded = Message::ChannelAnnouncement(ca).encode();
@@ -375,6 +364,23 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                     );
                     self.conn.send_message(&encoded)?;
                     Some(Variable::SentFundingCreated)
+                }
+
+                Operation::SendChannelReady { include_alias } => {
+                    let cr = build_channel_ready(
+                        &variables,
+                        &instr.inputs,
+                        *include_alias,
+                        &mut self.channel_states,
+                    );
+                    let encoded = Message::ChannelReady(cr).encode();
+                    log::debug!(
+                        "[{:?}] SendChannelReady: {} bytes",
+                        start.elapsed(),
+                        encoded.len(),
+                    );
+                    self.conn.send_message(&encoded)?;
+                    None
                 }
 
                 Operation::RecvAcceptChannel => {
@@ -3106,7 +3112,7 @@ mod tests {
     }
 
     #[test]
-    fn execute_build_channel_ready() {
+    fn execute_send_channel_ready() {
         let channel_id = ChannelId::v1_from_funding_outpoint(OutPoint {
             txid: "09b0549b35f14ee862f63bd75811c6c27963c4dea6766ec6836952ec78df1e7e"
                 .parse()
@@ -3127,24 +3133,16 @@ mod tests {
                 inputs: vec![],
             },
             Instruction {
-                operation: Operation::BuildChannelReady {
+                operation: Operation::SendChannelReady {
                     include_alias: false,
                 },
                 inputs: vec![10, 1, 11],
             },
             Instruction {
-                operation: Operation::SendMessage,
-                inputs: vec![12],
-            },
-            Instruction {
-                operation: Operation::BuildChannelReady {
+                operation: Operation::SendChannelReady {
                     include_alias: true,
                 },
                 inputs: vec![10, 3, 11],
-            },
-            Instruction {
-                operation: Operation::SendMessage,
-                inputs: vec![14],
             },
         ]);
 
@@ -3173,7 +3171,7 @@ mod tests {
         // The instructions send 1 `funding_created` and 2 `channel_ready` messages.
         assert_eq!(executor.conn.sent.len(), 3);
 
-        // The first channel_ready was built with include_alias = false, so it must
+        // The first channel_ready was sent with include_alias = false, so it must
         // not carry the short_channel_id TLV.
         let cr1 = match Message::decode(&executor.conn.sent[1]).expect("valid message") {
             Message::ChannelReady(cr) => cr,
@@ -3187,7 +3185,7 @@ mod tests {
         assert_eq!(cr1.second_per_commitment_point, expected_pcp1);
         assert!(cr1.tlvs.short_channel_id.is_none());
 
-        // The second channel_ready was built with include_alias = true, so it must
+        // The second channel_ready was sent with include_alias = true, so it must
         // carry the alias SCID we loaded in its short_channel_id TLV.
         let cr2 = match Message::decode(&executor.conn.sent[2]).expect("valid message") {
             Message::ChannelReady(cr) => cr,
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -300,8 +300,12 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                 }
 
                 Operation::BuildFundingCreated => {
-                    let fc =
-                        build_funding_created(&variables, &instr.inputs, &mut self.channel_states)?;
+                    let fc = build_funding_created(
+                        &variables,
+                        &instr.inputs,
+                        &mut self.channel_states,
+                        &mut self.negotiations,
+                    )?;
                     let encoded = Message::FundingCreated(fc).encode();
                     Some(Variable::FundingCreatedMessage(encoded))
                 }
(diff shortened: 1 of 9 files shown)


smite-scenarios/src/executor.rs
```diff
@@ -1095,6 +1121,49 @@ fn verify_funding_signed(
...
+///
+/// # Panics
+///
+/// Panics if the bytes do not decode to an `open_channel`, which would indicate
+/// a bug in the `SendOpenChannel` path.
+fn record_send_open_channel(negotiations: &mut HashMap<ChannelId, PendingChannel>, bytes: &[u8]) {
+    let open_channel = match Message::decode(bytes) {
+        Ok(Message::OpenChannel(oc)) => oc,
+        other => panic!("SendOpenChannel did not produce an open_channel message: {other:?}"),
+    };
```
morehouse: Nit: we now do an extra `open_channel` decode.  Perhaps we should make the `OpenChannelMessage` type contain an `OpenChannel` instead of a `Vec<u8>` to avoid the unnecessary work.

smite-scenarios/src/executor.rs
```diff
@@ -1095,6 +1121,49 @@ fn verify_funding_signed(
...
+fn record_recv_accept_channel(
+    negotiations: &mut HashMap<ChannelId, PendingChannel>,
+    accept_channel: &AcceptChannel,
+) -> Result<(), ExecuteError> {
+    let pending = negotiations
+        .get_mut(&accept_channel.temporary_channel_id)
+        .ok_or(ExecuteError::UnknownChannel(
+            accept_channel.temporary_channel_id,
+        ))?;
+    pending.accept_channel = Some(accept_channel.clone());
```
morehouse, line 1198: Theoretically, if we sent two `open_channel` messages with the same `temporary_channel_id`, we would record the *first* one.  But if the target responded to both, we would record the *second* `accept_channel`, which would not align with the recorded `open_channel` message.

I would be surprised if any implementation actually responds to the second `open_channel`, and I'm not sure what the best action is in that case, but just wanted to flag this for discussion.

smite-scenarios/src/executor.rs
```diff
@@ -2107,6 +2176,130 @@ mod tests {
...
+            )
+            .unwrap();
+
+        let pending = executor.negotiations.get(&temporary_channel_id).unwrap();
+        assert_eq!(
+            pending.open_channel.temporary_channel_id,
+            temporary_channel_id
+        );
+        let accept_channel = pending.accept_channel.as_ref().unwrap();
+        assert_eq!(accept_channel.temporary_channel_id, temporary_channel_id);
```
morehouse: Nit: we could check that the entire `accept_channel` message is equal to `sample_accept_channel()`.

smite-scenarios/src/executor.rs
```diff
@@ -691,61 +703,75 @@ fn build_open_channel(variables: &[Option<Variable>], inputs: &[usize]) -> OpenC
...
+
+    let pending = negotiations
+        .get(&temporary_channel_id)
+        .ok_or(ExecuteError::IncompleteNegotiation(temporary_channel_id))?;
+    let open_channel = &pending.open_channel;
+    let accept_channel = pending
+        .accept_channel
+        .as_ref()
+        .ok_or(ExecuteError::IncompleteNegotiation(temporary_channel_id))?;
+
```
morehouse: By returning `IncompleteNegotiation`, we lose all ability to send `funding_created` out of order.  I think it would be better to send some default-constructed `funding_created` when we're missing the data we need for a valid signature.

Something like this:
- `temporary_channel_id`, `funding_txid`, and `funding_output_index` determined from inputs
- `signature` is all zeroes
- nothing saved to `channel_states`

smite-scenarios/src/executor.rs
morehouse, line 804: I think we also want to either delete or mark the entry in `negotiations` so that we're able to open another channel using the same `temporary_channel_id` and record the parameters for that channel.

The simplest approach is to delete the entry from `negotiations` once the channel "graduates" to a real `channel_id`.  But if we want to keep the ability to easily send duplicate `funding_created` messages, we can also keep the entry around and just set a flag on the `negotiations` entry to mark the channel as "graduated".  Graduated entries can still be used for subsequent `funding_created` messages but also get overwritten by the next `SendOpenChannel` that uses the same `temporary_channel_id`.

NishantBansal2003:
I was also concerned about `temporary_channel_id` reuse, but [BOLT 2](https://github.com/lightning/bolts/blob/master/02-peer-protocol.md#requirements-8) says:

> The sending node:
> 
> - MUST ensure the chain_hash value identifies the chain it wishes to open the channel within.
> - MUST ensure temporary_channel_id is unique from any other channel ID with the same peer.

Based on that, I expected implementations to reject any reused `temporary_channel_id`, even if the previous channel had already transitioned to the active state.

But, I checked the code in CLN and LDK (and also verified it by running https://github.com/NishantBansal2003/smite/tree/funding-created-generator), and it turns out that once `funding_created` is sent, the entry for the `temporary_channel_id` is removed. So, the same peer can send another `open_channel` with the same `temporary_channel_id`, and both implementations accept it

CLN: https://github.com/ElementsProject/lightning/blob/58f0c16a57e324529d8ce9f40e6511d31b004c0e/connectd/multiplex.c#L73-L78
LDK: https://github.com/lightningdevkit/rust-lightning/blob/759378c912c3a9ca86d312bbf7929550644ca174/lightning/src/ln/channelmanager.rs#L11755-L11758


> if we want to keep the ability to easily send duplicate `funding_created` messages, we can also keep the entry around and just set a flag on the `negotiations` entry to mark the channel as "graduated". Graduated entries can still be used for subsequent `funding_created` messages but also get overwritten by the next `SendOpenChannel` that uses the same `temporary_channel_id`.

I think this is the better way to handle such cases

smite-scenarios/src/executor.rs
NishantBansal2003: `temporary_channel_id` reuse should be rejected until the previous negotiation reaches `funding_created`. So if the target sends a second `accept_channel` reusing the same `temporary_channel_id`, I think we should return an error

morehouse (approved):
LGTM

PR #135, opened 2026-06-29
smite-ir/mutators: Add `SpliceMutator`

Chand-ra:
Add `SpliceMutator` for smite-IR. Mutates a given program by inserting a spliced input at a random point in the said program.

### DIFF
--- a/smite-ir-mutator/src/lib.rs
+++ b/smite-ir-mutator/src/lib.rs
@@ -621,6 +621,31 @@ mod tests {
         }
     }
 
+    #[test]
+    fn fuzz_with_splice_utilizes_splice_mutator() {
+        let state = State::new(0);
+        let input = seed_program_bytes();
+        let splice_input = seed_program_bytes();
+        let mut splice_used = false;
+        // The mutator sequence is determined by a RNG. Loop until the RNG
+        // rolls `SpliceInsertionMutator` to guarantee the FFI correctly
+        // handed the payload to `mutate_stacked()`.
+        for _ in 0..100 {
+            let _ = fuzz_via_ffi(&state, input.clone(), splice_input.clone(), 1 << 16);
+            let ptr = unsafe { afl_custom_describe(state.0, 256) };
+            let s = unsafe { CStr::from_ptr(ptr) }
+                .to_str()
+                .expect("valid utf-8");
+            if s.contains("splice") {
+                splice_used = true;
+            }
+        }
+        assert!(
+            splice_used,
+            "SpliceInsertionMutator was never chosen despite a valid add_buf"
+        );
+    }
+
     // -- Trim tests --
 
     fn init_trim_via_ffi(state: &State, mut input: Vec<u8>) -> i32 {
--- a/smite-ir-mutator/src/lib.rs
+++ b/smite-ir-mutator/src/lib.rs
@@ -89,8 +89,13 @@ impl MutatorState {
     /// Applies a random number of stacked mutations to `program`. The stack
     /// count is a power of two in `[1, 16]` to give a mix of small tweaks and
     /// larger leaps per execution, similar in spirit to AFL++'s havoc stage.
+    ///
+    /// If `splice` is `Some`, `SpliceInsertionMutator` is added to the pool of
+    /// available mutators, allowing the `splice` program to be randomly inserted
+    /// into the target `program` during stack execution.
+    ///
     /// Records the ordered list of mutator names in `last_sequence`.
-    fn mutate_stacked(&mut self, program: &mut Program) {
+    fn mutate_stacked(&mut self, program: &mut Program, splice: Option<Program>) {
         self.last_sequence.clear();
         // Power-of-two stack count: 1, 2, 4, 8, or 16 mutations.
         let stack = 1u32 << self.rng.random_range(0..=4);
@@ -224,16 +229,16 @@ pub unsafe extern "C" fn afl_custom_deinit(data: *mut c_void) {
 /// - `data` must be a pointer returned by [`afl_custom_init`].
 /// - `buf` must point to `buf_size` readable bytes.
 /// - `out_buf` must be a valid, writable pointer to a `*const u8` slot.
-/// - `_add_buf` is unused; we export `afl_custom_splice_optout` so AFL++ never
-///   populates it.
+/// - `add_buf` must point to `add_buf_size` readable bytes, or is null with
+///   `add_buf_size == 0` (no splice input available).
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn afl_custom_fuzz(
     data: *mut c_void,
     buf: *mut u8,
     buf_size: usize,
     out_buf: *mut *const u8,
-    _add_buf: *mut u8,
-    _add_buf_size: usize,
+    add_buf: *mut u8,
+    add_buf_size: usize,
     max_size: usize,
 ) -> usize {
     let state = unsafe { &mut *data.cast::<MutatorState>() };
@@ -245,7 +250,13 @@ pub unsafe extern "C" fn afl_custom_fuzz(
         let input = unsafe { slice::from_raw_parts(buf, buf_size) };
         match postcard::from_bytes::<Program>(input) {
             Ok(mut p) => {
-                state.mutate_stacked(&mut p);
+                let splice = if !add_buf.is_null() && add_buf_size > 0 {
+                    let splice_input = unsafe { slice::from_raw_parts(add_buf, add_buf_size) };
+                    postcard::from_bytes::<Program>(splice_input).ok()
+                } else {
+                    None
+                };
+                state.mutate_stacked(&mut p, splice);
                 p
             }
             Err(_) => state.generate_fresh(),
@@ -369,18 +380,6 @@ pub unsafe extern "C" fn afl_custom_post_trim(_data: *mut c_void, _success: u8) 
     1
 }
 
-/// Marker symbol that tells AFL++ not to populate `add_buf` for
-/// [`afl_custom_fuzz`]. AFL++ never actually calls this function -- it only
-/// checks for the symbol's presence via `dlsym` and, if found, skips picking a
-/// splice input before each fuzz call.
-///
-/// # Safety
-///
-/// Never invoked; the signature exists only so the symbol has the correct
-/// linkage.
-#[unsafe(no_mangle)]
-pub unsafe extern "C" fn afl_custom_splice_optout(_data: *mut c_void) {}
-
 /// Returns a null-terminated description of the most recently applied mutation.
 /// AFL++ uses this when naming queue entries.
 ///
@@ -441,7 +440,12 @@ mod tests {
     /// Calls `afl_custom_fuzz` and returns `(out, len)`. The returned pointer
     /// is borrowed from the state's internal buffer and is only valid until the
     /// next call into the library.
-    fn fuzz_via_ffi(state: &State, mut input: Vec<u8>, max_size: usize) -> (*const u8, usize) {
+    fn fuzz_via_ffi(
+        state: &State,
+        mut input: Vec<u8>,
+        mut splice: Vec<u8>,
+        max_size: usize,
+    ) -> (*const u8, usize) {
         let mut out: *const u8 = ptr::null();
         let len = unsafe {
             afl_custom_fuzz(
@@ -449,8 +453,8 @@ mod tests {
                 input.as_mut_ptr(),
                 input.len(),
                 &raw mut out,
-                ptr::null_mut(),
-                0,
+                splice.as_mut_ptr(),
+                splice.len(),
                 max_size,
             )
         };
@@ -500,7 +504,7 @@ mod tests {
     #[test]
     fn fuzz_with_empty_input_generates_fresh() {
         let state = State::new(1);
-        let (out, len) = fuzz_via_ffi(&state, Vec::new(), 1 << 16);
+        let (out, len) = fuzz_via_ffi(&state, Vec::new(), Vec::new(), 1 << 16);
         assert!(len > 0);
         decode(out, len);
         verify_fresh_generation(&state);
@@ -511,7 +515,7 @@ mod tests {
         let state = State::new(2);
         let mut current = seed_program_bytes();
         for _ in 0..50 {
-            let (out, len) = fuzz_via_ffi(&state, current, 1 << 16);
+            let (out, len) = fuzz_via_ffi(&state, current, Vec::new(), 1 << 16);
             assert!(len > 0);
             // Copy before the next call, which will overwrite the state's
             // out_buf.
@@ -524,7 +528,7 @@ mod tests {
     #[test]
     fn fuzz_with_garbage_input_generates_fresh() {
         let state = State::new(3);
-        let (out, len) = fuzz_via_ffi(&state, vec![0xFFu8; 16], 1 << 16);
+        let (out, len) = fuzz_via_ffi(&state, vec![0xFFu8; 16], Vec::new(), 1 << 16);
         assert!(len > 0);
         decode(out, len);
         verify_fresh_generation(&state);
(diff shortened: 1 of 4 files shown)


smite-ir/src/mutators/splice_insert.rs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        // Insert the spliced program.
+        for instr in &self.splice.instructions {
+            let shifted_inputs: Vec<usize> = instr
+                .inputs
+                .iter()
+                .map(|&input| input + insert_idx)
+                .collect();
+            builder.append(instr.operation.clone(), &shifted_inputs);
```
Chand-ra, line 38: This logic is the only logical difference between this mutator and `GeneratorInsertion`. It could also be implemented as:
```suggestion
        for instr in &self.splice.instructions {
            let mut inputs = vec![];
            for var_type in instr.operation.input_types() {
                inputs.push(builder.pick_variable(var_type, rng));
            }
            builder.append(instr.operation.clone(), &inputs);
        }
```
which I guess would "wire it stronger" to the rest of the program, but I'm not sure if that's a worthwhile tradeoff for the simplicity of the current approach.

smite-ir-mutator/src/lib.rs
```diff
@@ -122,6 +124,11 @@ impl MutatorState {
                     mutator.mutate(program, &mut self.rng);
                     "gen-insert"
                 }
+                5 => {
+                    let mutator = SpliceMutator::new(splice.expect("splice is non-empty").clone());
```
morehouse: Nit: the splice program is expected to be *present*, but the program itself could be empty


```suggestion
                    let mutator = SpliceMutator::new(splice.expect("splice is present").clone());
```

smite-ir/src/mutators/splice.rs
```diff
@@ -0,0 +1,54 @@
+//! Mutator that inserts spliced programs.
+
+use rand::{Rng, RngExt};
+
+use super::Mutator;
+use crate::{Program, ProgramBuilder};
+
+/// Inserts a spliced program at a random point in the given program.
+pub struct SpliceMutator {
```
morehouse: Nit: perhaps `SpliceInsertionMutator` would be a more descriptive name.  It would also make it clear that this mutator parallels `GeneratorInsertionMutator`.

smite-ir/src/mutators/splice_insert.rs
morehouse: It's an interesting idea, but I think the current simpler approach may be better.  By doing `pick_variable`, we end up losing the specific values that the splice program used (which were selected as interesting after many random mutations), which probably hurts fuzzing effectiveness.  We'd also need to figure out how to handle variable types for which `pick_variable` currently panics.

We also can get some of the same behavior already if `InputSwapMutator` is stacked after the splice.

smite-ir/src/mutators/splice_insert.rs
```diff
@@ -0,0 +1,54 @@
...
+        let mut builder = ProgramBuilder::new();
+        let mut iter = std::mem::take(&mut program.instructions).into_iter();
+
+        // Consume and append the pre-insertion instructions.
+        for instr in iter.by_ref().take(insert_idx) {
+            builder.append(instr.operation, &instr.inputs);
+        }
+
+        // Insert the spliced program.
+        for instr in &self.splice.instructions {
```
morehouse, line 32: I think it would be interesting to do an experiment where we compare full-program insertion with only inserting a random subset (prefix) of the spliced program.

Full-program insertion will tend to create longer (and slower) programs, so it's possible that prefix insertion actually performs better.

smite-ir-mutator/src/lib.rs
```diff
@@ -90,13 +90,15 @@ impl MutatorState {
     /// count is a power of two in `[1, 16]` to give a mix of small tweaks and
     /// larger leaps per execution, similar in spirit to AFL++'s havoc stage.
     /// Records the ordered list of mutator names in `last_sequence`.
-    fn mutate_stacked(&mut self, program: &mut Program) {
+    fn mutate_stacked(&mut self, program: &mut Program, splice: Option<&Program>) {
```
morehouse: Nit: doc comment needs to be updated to explain the new `splice` parameter

smite-ir-mutator/src/lib.rs
```diff
@@ -586,13 +588,6 @@ mod tests {
         );
     }
 
-    #[test]
-    fn splice_optout_symbol_exists() {
```
morehouse, line 590: Now that splicing is enabled, we should add an FFI test for it.  Currently all tests set `add_buf=NULL`.

smite-ir-mutator/src/lib.rs
morehouse: Nit: we can eliminate the program clone by hoisting splice mutator creation out of the loop and having `mutate_stacked` take ownership of the splice.

```rust
fn mutate_stacked(..., splice: Option<Program>) {
  ...
  let splice_mutator = splice.map(SpliceMutator::new);
  let upper_bound = if splice_mutator.is_some() { 6 } else { 5 };
  for _ in 0..stack {
    let name = match self.rng.random_range(0..upper_bound) {
      ...
      5 => {
        splice_mutator.as_ref().expect("splice present").mutate(program, &mut self.rng);
        "splice"
      }
      ...
```

smite-ir/src/tests.rs
```diff
@@ -2313,56 +2364,95 @@ fn generator_insertion_shifts_correctly() {
...
+
+    let mutator = SpliceMutator::new(splice_program.clone());
+    let mut rng = SmallRng::seed_from_u64(0);
+    mutator.mutate(&mut program, &mut rng);
+
+    assert_graph_shifted_correctly(
+        &original,
+        &splice_program,
+        splice_program.instructions.len(),
+    );
```
morehouse: This is testing the wrong thing.  I'm surprised it actually passes.


```suggestion
    assert_graph_shifted_correctly(
        &original,
        &program,
        splice_program.instructions.len(),
    );
```

smite-ir/src/mutators/splice_insert.rs
Chand-ra: > We'd also need to figure out how to handle variable types for which `pick_variable` currently panics.

Would we? `pick_variable()` currently panics if we try to `generate_fresh()` a type that cannot be generated (messages, affine types, etc.).

If the corpus always has valid programs and every mutation preserves validity, won't both the input programs to `SpliceMutator` always be valid? That is, wouldn't every panicking type already have at least one matching candidate?

smite-ir/src/mutators/splice.rs
Chand-ra: Doesn't splice already imply insertion? Apparently this is the [definition of splice](https://www.merriam-webster.com/dictionary/splice):
```
to join the ends of two pieces of film, tape, rope, etc. so that they form one continuous long piece
```
And I like it as is because it conveys similarity with splice mutators in generic fuzzing, which is more or less what we're trying to replicate here.

smite-ir/src/mutators/splice_insert.rs
Chand-ra: > I think it would be interesting to do an experiment where we compare full-program insertion with only inserting a random subset (prefix) of the spliced program.

Hmm, we're inserting the prefix (instead of a random slice) so that we don't have to worry about variable dependencies. Should be easily implementable but I'm not fully sold on the "performs better" part, although I agree that it is an interesting experiment.

I think a more interesting (and perhaps advanced) version of this would be `FlowInsertionMutator` that selects a random Act instruction in the spliced program, yanks out its entire lineage, and inserts it at a random point in the given program.

smite-ir/src/tests.rs
Chand-ra: Passes because the assertion loop is never executed. We do this in `assert_graph_shifted_correctly()`:
```
for mut_idx in (insert_point + offset)..mutated.instructions.len()
```
But `insert_point` + `offset` = 0 + 10 = 10 here, so the loop ultimately evaluates to `for mut_idx in 10..10`.

We should add a length assertion in `assert_graph_shifted_correctly()` to prevent this from happening:
```
assert_eq!(
    mutated.instructions.len(),
    original.instructions.len() + offset,
    "Mutated program length does not match original + offset"
);
```

smite-ir/src/mutators/splice_insert.rs
morehouse: Yes, right -- we only have panic issues with `generate_fresh`, which shouldn't be called in this case.

We should do an experiment with the following configs to evaluate: (1) baseline, (2) this mutator, (3) a `pick_variable` splice mutator, (4) this mutator + a `pick_variable` splice mutator.

smite-ir/src/mutators/splice.rs
morehouse: Typically a "splice" mutator takes the head of one input and splices it to the tail of another input.  This mutator doesn't do that; rather it *inserts* an entire input into another input.  Hence `SpliceInsertionMutator` would be a more accurate name, especially since it does essentially the same thing as `GeneratorInsertionMutator`.

smite-ir-mutator/src/lib.rs
erickcestari, line 148: nit: We could refactor to avoid hardcoding numbers that would need changing if we added or removed mutators.

```suggestion

/// Mutators available to [`MutatorState::mutate_stacked`]. The pool of
/// candidates is built per call, so optional entries (splice) are included by
/// pushing a variant.
#[derive(Clone, Copy)]
enum StackedMutator {
    OpParam,
    InputSwap,
    InstrDelete,
    InstrReorder,
    GenInsert,
    Splice,
}

fn mutate_stacked(&mut self, program: &mut Program, splice: Option<Program>) {
        self.last_sequence.clear();
        // Power-of-two stack count: 1, 2, 4, 8, or 16 mutations.
        let stack = 1u32 << self.rng.random_range(0..=4);
        let splice_mutator = splice.map(SpliceMutator::new);
        let mut pool = vec![
            StackedMutator::OpParam,
            StackedMutator::InputSwap,
            StackedMutator::InstrDelete,
            StackedMutator::InstrReorder,
            StackedMutator::GenInsert,
        ];
        if splice_mutator.is_some() {
            pool.push(StackedMutator::Splice);
        }
        for _ in 0..stack {
            // Uniform pick between the available mutators.
            let choice = *pool
                .iter()
                .choose(&mut self.rng)
                .expect("pool is non-empty");
            let name = match choice {
                StackedMutator::OpParam => {
                    OperationParamMutator.mutate(program, &mut self.rng);
                    "op-param"
                }
                StackedMutator::InputSwap => {
                    InputSwapMutator.mutate(program, &mut self.rng);
                    "input-swap"
                }
                StackedMutator::InstrDelete => {
                    InstructionDeleteMutator.mutate(program, &mut self.rng);
                    "instr-delete"
                }
                StackedMutator::InstrReorder => {
                    InstructionReorderMutator.mutate(program, &mut self.rng);
                    "instr-reorder"
                }
                StackedMutator::GenInsert => {
                    let generator = *AnyGenerator::ALL
                        .iter()
                        .choose(&mut self.rng)
                        .expect("AnyGenerator::ALL is non-empty");
                    let mutator = GeneratorInsertionMutator::new(generator);
                    mutator.mutate(program, &mut self.rng);
                    "gen-insert"
                }
                StackedMutator::Splice => {
                    splice_mutator
                        .as_ref()
                        .expect("Splice is only pooled when a splice input exists")
                        .mutate(program, &mut self.rng);
                    "splice"
                }
            };
            self.last_sequence.push(name);
        }
    }
```

smite-ir/src/mutators/splice_insert.rs
morehouse: I suspect a prefix insertion mutator and a flow insertion mutator will perform similarly in practice, since we have the `DeadCodeEliminator` that will also periodically do the dependency analysis and dead instruction removal that `FlowInsertionMutator` would.  So either one would be interesting to evaluate IMO.

It may also be interesting to combine this `FlowInsertionMutator` idea with the `pick_variable` rewiring you suggested as an experiment.

morehouse (commented):
Code LGTM.  This is ready for experimental evaluation.

smite-ir-mutator/src/lib.rs
Chand-ra: So basically the `AnyGenerator` variant for mutators. It _does_ improve readability, but I feel like the change doesn't fit well within this PR. I'll follow up with a separate one implementing it.

Chand-ra:
Finally got to evaluating this mutator (thanks to the [new orchestration script](https://github.com/morehouse/smite/blob/master/scripts/smite-orchestrator.py)). Here are the results:

# Fuzzing Evaluation Report

**Configuration B (Experimental):** `baseline-plus-splice`

## 1. Summary Statistics

| cln      |             12 |             20 |         20 |                   3617.5 |               3641.5 |             0.0494718 |      0.7225  |                 42704.7 |             43344.8 |          0.000666608 |      0.8425 |                    5750 |                5071 |               49.025 |           43.57  |
| ldk      |             12 |             20 |         20 |                  12611.5 |              12628.5 |             1         |      0.50125 |                144049   |            139023   |          0.523233    |      0.395  |                   14552 |               14983 |              170.48  |           97.04  |
| lnd      |             12 |             20 |         20 |                  25675.5 |              25675   |             0.771048  |      0.41875 |                307312   |            307591   |          0.635945    |      0.455  |                   25753 |               25764 |               88.195 |           81.885 |

## 2. Interpretation Guide

### Key Metrics

- **Union Coverage**: Union of all trial bitmaps; the coverage ceiling for a multi-core deployment. Descriptive only, cannot be statistically tested.

### Reading the Results

|---|---|---|

## 3. Visualizations

### Target: cln

cln_time_series

#### Distribution Comparisons

|:---:|:---:|
| cln_boxplot | cln_auc_boxplot |

---

### Target: ldk

ldk_time_series

#### Distribution Comparisons

|:---:|:---:|
| ldk_boxplot | lnd_auc_boxplot |

---

### Target: lnd

lnd_time_series

#### Distribution Comparisons

|:---:|:---:|
| lnd_boxplot | lnd_auc_boxplot |

---

By the way, I had also included Eclair in the evaluation but I think I messed up the orchestrator for it. Here is the summary table including Eclair:

## 1. Summary Statistics

| cln      |             12 |             20 |         20 |                   3617.5 |               3641.5 |             0.0659624 |      0.7225  |                 42704.7 |             43344.8 |           0.00088881 |      0.8425 |                    5750 |                5071 |               49.025 |           43.57  |
| eclair   |             12 |             20 |         20 |                   7285.5 |               7241   |             1         |      0.5625  |                 81812.3 |             83814.8 |           0.272722   |      0.6575 |                    8455 |                8541 |               12.72  |           17.82  |
| ldk      |             12 |             20 |         20 |                  12611.5 |              12628.5 |             1         |      0.50125 |                144049   |            139023   |           0.523233   |      0.395  |                   14552 |               14983 |              170.48  |           97.04  |
| lnd      |             12 |             20 |         20 |                  25675.5 |              25675   |             1         |      0.41875 |                307312   |            307591   |           0.635945   |      0.455  |                   25753 |               25764 |               88.195 |           81.885 |

erickcestari:
Hey @Chand-ra, did you run the orchestration script with AFL++'s `-t` flag set to 2000+? If not, I think that could have impacted the fuzzing evaluation.

I also noticed that the orchestration script doesn't have an option to configure the `-t` value and instead uses AFL++'s default. Running without an increased timeout can cause the fuzzer to time out on inputs, especially the longer ones that are capable of opening channels.

```
  -t msec       - timeout for each run (auto-scaled, default 1000 ms). Add a '+'
                  to auto-calculate the timeout, the value being the maximum.
```

Chand-ra:
> Hey @Chand-ra, did you run the orchestration script with AFL++'s `-t` flag set to 2000+? If not, I think that could have impacted the fuzzing evaluation.
> 
> I also noticed that the orchestration script doesn't have an option to configure the `-t` value and instead uses AFL++'s default. Running without an increased timeout can cause the fuzzer to time out on inputs, especially the longer ones that are capable of opening channels.
> 
> ```
>   -t msec       - timeout for each run (auto-scaled, default 1000 ms). Add a '+'
>                   to auto-calculate the timeout, the value being the maximum.
> ```

Just opened PR #206 to address this, although the timeout isn't configurable right now. And while the report in this PR wasn't calculated with the increased timeout, [this](https://github.com/lnfuzz/smite/issues/115#issuecomment-5207972222) one was.

PR #137, opened 2026-06-29
smite: mine consensus-valid transactions that violate mempool policy

NishantBansal2003:
There were a couple of crashes while I was running the funding-flow generator, where we were getting panics in smite because of the restrictive assert statements we have set. These commits fix those issues in smite.

So now we can’t create two transactions with overlapping inputs. We return early when signing or broadcasting a transaction if it has already been broadcast and confirmed, since the wallet can’t sign a transaction whose inputs have already been spent. There were also a couple of crashes around broadcasting, where either `funding_satoshis` itself was below the dust limit, or the feerate used to calculate the transaction fee was too low compared to `bitcoind` default policy. So now we reject those inputs as well.

These cases do not really contribute to fuzzing, because in practice, if `bitcoind` default policy rejects them, Lightning nodes should also reject those `open_channel` messages before sending `accept_channel`. But let’s say they are not rejected, then I think those cases can be caught once we add the `open_channel` and `accept_channel` oracle. So in either case, we don’t need to relax `bitcoind` default policy to exercise the full funding flow

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -62,17 +62,24 @@ const MAX_MINIMUM_DEPTH: u32 = 8;
 
 /// Abstraction over bitcoin-cli operations, allowing mock implementations in tests.
 pub trait BitcoinRpc {
-    /// Mines the given number of blocks.
-    fn mine_blocks(&mut self, num_blocks: u8);
+    /// Mines the given number of blocks, including any transactions in the
+    /// `private_mempool` in the first block.
+    fn mine_blocks(&mut self, num_blocks: u8, private_mempool: &[String]);
 
     /// Returns the wallet's spendable UTXOs.
+    #[must_use]
     fn get_utxos(&mut self) -> Vec<Utxo>;
 
     /// Returns the scriptPubKey for a newly generated wallet address.
+    #[must_use]
     fn get_new_address_script_pubkey(&mut self) -> ScriptBuf;
 
-    /// Signs and broadcasts a transaction
-    fn sign_and_broadcast_tx(&mut self, tx: &bitcoin::Transaction);
+    /// Signs and broadcasts a transaction. Returns hex-encoded raw transaction
+    /// if it is consensus-valid but rejected by mempool policy, so it can be
+    /// added to the `private_mempool`; returns `None` if it was broadcast or is
+    /// already confirmed.
+    #[must_use]
+    fn sign_and_broadcast_tx(&mut self, tx: &bitcoin::Transaction) -> Option<String>;
 
     /// Locks the given outpoints so subsequent [`get_utxos`](Self::get_utxos)
     /// calls exclude them, preventing independently built transactions from
@@ -81,12 +88,13 @@ pub trait BitcoinRpc {
 
     /// Returns the number of confirmations for the transaction with the given
     /// txid, or `0` if it is unconfirmed or unknown to the node.
+    #[must_use]
     fn get_transaction_confirmations(&mut self, txid: Txid) -> u32;
 }
 
 impl BitcoinRpc for BitcoinCli {
-    fn mine_blocks(&mut self, num_blocks: u8) {
-        BitcoinCli::mine_blocks(self, num_blocks);
+    fn mine_blocks(&mut self, num_blocks: u8, private_mempool: &[String]) {
+        BitcoinCli::mine_blocks(self, num_blocks, private_mempool);
     }
 
     fn get_utxos(&mut self) -> Vec<Utxo> {
@@ -97,8 +105,8 @@ impl BitcoinRpc for BitcoinCli {
         BitcoinCli::get_new_address_script_pubkey(self)
     }
 
-    fn sign_and_broadcast_tx(&mut self, tx: &bitcoin::Transaction) {
-        BitcoinCli::sign_and_broadcast_tx(self, tx);
+    fn sign_and_broadcast_tx(&mut self, tx: &bitcoin::Transaction) -> Option<String> {
+        BitcoinCli::sign_and_broadcast_tx(self, tx)
     }
 
     fn lock_utxos(&mut self, outpoints: &[OutPoint]) {
@@ -244,6 +252,12 @@ pub struct Executor<C, B> {
     /// `temporary_channel_id`, so the funding flow can build commitments from
     /// the parameters actually sent on the wire.
     negotiations: HashMap<ChannelId, PendingChannel>,
+    /// Transactions stored outside Bitcoin Core's mempool, typically because they
+    /// were rejected by mempool policy, to be included in the next `MineBlocks`
+    /// operation. Each is stored as `(txid, raw_hex)`: re-signing the same
+    /// transaction can change its raw hex, but the txid stays the same, so
+    /// deduplication keys on the txid while the raw hex is what gets mined.
+    private_mempool: Vec<(Txid, String)>,
 }
 
 impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
@@ -256,6 +270,7 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
             context,
             channel_states: HashMap::new(),
             negotiations: HashMap::new(),
+            private_mempool: Vec::new(),
         }
     }
 
@@ -476,19 +491,34 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                 }
 
                 Operation::MineBlocks(v) => {
-                    self.bitcoin_cli.mine_blocks(*v);
+                    // Clear the private mempool and mine the requested blocks,
+                    // adding those transactions to the first block.
+                    let private_mempool: Vec<String> = std::mem::take(&mut self.private_mempool)
+                        .into_iter()
+                        .map(|(_, hex)| hex)
+                        .collect();
+                    self.bitcoin_cli.mine_blocks(*v, &private_mempool);
                     log::debug!("[{:?}] MineBlocks: mined {} block(s)", start.elapsed(), v);
                     None
                 }
 
                 Operation::BroadcastTransaction => {
                     let ft = resolve_funding_transaction(&variables, instr.inputs[0]);
+                    let txid = ft.tx.compute_txid();
                     log::debug!(
                         "[{:?}] BroadcastTransaction: txid={}",
                         start.elapsed(),
-                        ft.tx.compute_txid(),
+                        txid
                     );
-                    self.bitcoin_cli.sign_and_broadcast_tx(&ft.tx);
+                    // Queue transactions rejected by the mempool in the private
+                    // mempool so they can be mined later. Dedup on txid so the
+                    // same transaction broadcast again before then is queued
+                    // once, regardless of any change to its signed hex.
+                    if let Some(hex) = self.bitcoin_cli.sign_and_broadcast_tx(&ft.tx)
+                        && !self.private_mempool.iter().any(|(t, _)| *t == txid)
+                    {
+                        self.private_mempool.push((txid, hex));
+                    }
                     None
                 }
             };
@@ -1392,6 +1422,7 @@ mod tests {
     #[derive(Default)]
     struct MockBitcoinCli {
         mine_blocks_calls: Vec<u8>,
+        mined_private_mempool: Vec<String>,
         broadcast_calls: Vec<Transaction>,
         utxos: Vec<Utxo>,
         change_spk: ScriptBuf,
@@ -1399,8 +1430,9 @@ mod tests {
     }
 
     impl BitcoinRpc for MockBitcoinCli {
-        fn mine_blocks(&mut self, num_blocks: u8) {
+        fn mine_blocks(&mut self, num_blocks: u8, private_mempool: &[String]) {
             self.mine_blocks_calls.push(num_blocks);
+            self.mined_private_mempool = private_mempool.to_vec();
             self.confirmations += u32::from(num_blocks);
         }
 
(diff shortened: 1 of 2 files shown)


ekzyis (commented):
Only looked at the first commit (5644b83).

I wonder if we also need to unlock the utxos in some cases, like when we fail to broadcast for some reason. Or is that not needed because of snapshot fuzzing, so for every run, `bitcoind` is also in the same initial state with the utxos unlocked?

But then I wonder why we even need to lock utxos: when would we create two funding transactions with overlapping inputs? Aren't we only creating one funding transaction per run?

erickcestari:
> There were a couple of crashes while I was running the funding-flow generator, where we were getting panics in smite because of the restrictive assert statements we have set. These commits fix those issues in smite.

My guess is that a crash can still happens when broadcasting the transaction on chain, because it might try to broadcast the same transaction that was mined before which would make the `sendrawtransaction` rpc call fail and crash the program.

We could ignore such error and add a special case in the minimizer to remove a `BroadcastTransaction` operation if it broadcast the same transaction twice, since that would be essentially dead code.

edit: Nevermind, I just saw the second commit that avoid broadcasting the same transaction twice. Which makes a lot of sense, and since is very lightweight operation we maybe don't need to minimize it?

> But then I wonder why we even need to lock utxos: when would we create two funding transactions with overlapping inputs? Aren't we only creating one funding transaction per run?

I think smite will be able to open multiple channels against the target. Right now (in current master) Smite cannot open a channel, since it misses the funding and broadcast operations of the open channel flow in the generators.

NishantBansal2003:
> I wonder if we also need to unlock the utxos in some cases, like when we fail to broadcast for some reason. Or is that not needed because of snapshot fuzzing, so for every run, `bitcoind` is also in the same initial state with the utxos unlocked?

Yep, I think `bitcoind` will return to the same initial state with a block height of 101, as captured in the snapshot



> But then I wonder why we even need to lock utxos: when would we create two funding transactions with overlapping inputs? Aren't we only creating one funding transaction per run?

We have `GeneratorInsertionMutator`, and eventually have `SpliceMutator`, which will merge two programs. At that point, we’ll have more than one create funding tx operations

erickcestari:
> smite: reject non-relayable funding transactions
> Reject funding transactions that would not be accepted
> under Bitcoin's default relay policy. Funding construction
> now fails for feerates below the minimum relay feerate and
> funding outputs below the dust threshold, avoiding
> transactions that cannot be relayed or mined by default

Should we care whether the transaction follows the bitcoin core relayable Policy? I think that, if the transaction is consensus-valid, Smite should be able to include it in a block, which would help us cover more edge cases.

NishantBansal2003:
> since is very lightweight operation we maybe don't need to minimize it?

I think any operation that makes redundant or unnecessary  RPC calls, should be minimized, since they are time-consuming and actually hurt fuzzing effectiveness



> Should we care whether the transaction follows the bitcoin core relayable Policy? I think that, if the transaction is consensus-valid, Smite should be able to include it in a block, which would help us cover more edge cases.

Some reasons I thought against it are:

* funding tx below the dust limit or with a very low fee rate will not actually be mined on mainnet
* Another thing is that currently the `bitcoind` backend is shared across smite and the target, so changing the policy could mean the target consults `bitcoind` to check whether the tx is broadcastable, and `bitcoind` could say yes. I was thinking of using the default rules for both the target and bitcoind
* In the future, we’re going to have an `open_channel` <-> `accept_channel` oracle, which will itself flag the case if it isn't rejected by the target, so there is no need to go beyond that point to find those issues for now

But I don’t have a strong opinion, so I’m fine if we want to relax the `bitcoind` rules instead. Let me know

ekzyis:
> funding tx below the dust limit or with a very low fee rate will not actually be mined on mainnet

It could get mined with out-of-band payments, or because it's the miner's own tx etc.

They are non-standard according to Bitcoin Core, but they can get mined.

I agree with @erickcestari, I think never broadcasting txs below the dust limit would unnecessarily limit fuzzing effectiveness, or at least I’m not convinced yet why we shouldn’t broadcast them.

erickcestari:
> * In the future, we’re going to have an open_channel <-> accept_channel oracle, which will itself flag the case if it isn't rejected by the target, so there is no need to go beyond that point to find those issues for now

I think I got what @NishantBansal2003 said. If a lightning node is building transactions (funding transactions, commitment transactions, HTLC transactions) that aren't accepted by mempool (dust output, nonstandard script, below static minrelay) this is already a bug that should be reported and the fuzzer should crash with help of oracles.

smite/src/bitcoin.rs
```diff
@@ -182,6 +182,13 @@ impl BitcoinCli {
             complete: bool,
         }
 
+        // A confirmed transaction may be broadcast again by the fuzzer. Its
+        // inputs are spent, so the wallet can no longer fully sign it, skip
+        // signing and broadcasting it again.
+        if self.get_transaction_confirmations(tx.compute_txid()) > 0 {
```
morehouse: Nit: txid is now computed twice in this function.  We could save it as a variable instead.

morehouse (commented):
The first three commits LGTM.  I'd be happy to merge them today as a separate PR.

I'm less sure about the last commit.  I agree that we don't want to change the default mempool policy used by the target node, but I think it could be quite profitable to preserve the ability to mine non-standard transactions that may trigger edge cases on the target node.

One way to reconcile the two is to use separate `bitcoind`s for smite and the target.  Of course then the `bitcoind` nodes communicating with each other would add a new source of latency.

Alternatively, we could keep the shared `bitcoind` and then if `sendrawtransaction` fails due to some mempool policy rejection, we could fallback to mining the transaction directly with `generateblock` instead.

smite/src/bitcoin.rs
NishantBansal2003: Done in https://github.com/morehouse/smite/pull/144

NishantBansal2003:
Updated the last commit to mine the tx directly if it violates mempool policy while still being consensus-valid in: https://github.com/morehouse/smite/pull/137/commits/c19300336c8f68bcf8a8ce78bb2ac0a55ab196d7

smite/src/bitcoin.rs
```diff
@@ -211,6 +227,17 @@ impl BitcoinCli {
             "signrawtransactionwithwallet returned complete=false"
         );
 
+        // If the transaction follows mempool policy, broadcast it normally.
+        // Otherwise, mine it directly into a block.
+        if !self.accepted_by_mempool(&signed_tx.hex) {
```
morehouse: Rather than doing another RPC call, can we just check the output from `sendrawtransaction` to save some overhead?

smite/src/bitcoin.rs
```diff
@@ -211,6 +227,17 @@ impl BitcoinCli {
...
+            self.generate_block_with_tx(&signed_tx.hex);
```
morehouse: So now our `BroadcastTransaction` operation may also mine blocks.  Maybe this is okay as a practicality, though it would be nice to only do the mining in the actual `MineBlocks` operation...

One unexpected example: previously if we broadcast the same transaction twice it was a no-op.  Now if we do that the transaction gets automatically mined on the second broadcast.

smite/src/bitcoin.rs
```diff
@@ -237,6 +264,68 @@ impl BitcoinCli {
...
+        struct TestMempoolAcceptResult {
+            allowed: bool,
+        }
+
+        let txs = serde_json::to_string(&[tx_hex]).expect("tx hex serializes to valid JSON");
+        let test_out = self
+            .run()
+            .arg("testmempoolaccept")
+            .arg(&txs)
+            .arg("0")
```
morehouse:
```suggestion
            // Disable the high-feerate cap and accept any fee rate for broadcast.
            .arg("0")
```

smite/src/bitcoin.rs
```diff
@@ -211,6 +227,17 @@ impl BitcoinCli {
...
         );
 
+        // If the transaction follows mempool policy, broadcast it normally.
+        // Otherwise, mine it directly into a block.
+        if !self.accepted_by_mempool(&signed_tx.hex) {
+            self.generate_block_with_tx(&signed_tx.hex);
+            assert!(
+                self.get_transaction_confirmations(txid) > 0,
+                "generateblock did not mine transaction {txid}"
+            );
```
morehouse: Is this check necessary -- is there any situation where `generateblock` succeeds but fails to mine our transaction?

If not, let's drop this check to save an RPC call.

morehouse (commented):
My main concern is that now the `BroadcastTransaction` operation may silently mine the transaction as well instead of waiting for `MineBlocks` to do that.

WDYT about an approach like this?

- If the broadcast fails due to non-standardness, the executor saves the raw tx in it's own "private mempool"
- On the next `MineBlocks` operation, if the private mempool contains anything, the executor appends its private mempool to `bitcoind`'s mempool from `getrawmempool` and mines the combined mempool in the first block with `generateblock`, then mines the remaining blocks normally.

NishantBansal2003:
> My main concern is that now the `BroadcastTransaction` operation may silently mine the transaction as well instead of waiting for `MineBlocks` to do that.
> 
> WDYT about an approach like this?
> 
> * If the broadcast fails due to non-standardness, the executor saves the raw tx in it's own "private mempool"
> * On the next `MineBlocks` operation, if the private mempool contains anything, the executor appends its private mempool to `bitcoind`'s mempool from `getrawmempool` and mines the combined mempool in the first block with `generateblock`, then mines the remaining blocks normally.

Yeah, I think mining blocks only in the `MineBlocks` operation makes sense for consistency. Now, in  https://github.com/morehouse/smite/pull/137/commits/daa8ebbaaa8dde1365e524760c2a4c6879881e1d, I’ve made sure there’s a clear separation between the two operations based on what they do.

smite/src/bitcoin.rs
```diff
@@ -84,6 +104,60 @@ impl BitcoinCli {
...
+    /// # Panics
+    ///
+    /// - If `bitcoin-cli getrawmempool`, `getnewaddress`, or `generateblock`
+    ///   fails to execute or exits non-zero.
+    /// - If `getrawmempool` does not return valid JSON.
+    /// - If `getnewaddress` does not return a valid regtest address.
+    /// - If any transaction in `private_mempool` is consensus-invalid.
+    fn mine_block_including(&self, private_mempool: &[String]) {
+        let mut txs = self.get_raw_mempool();
+        txs.extend(private_mempool.iter().cloned());
```
morehouse: Nit:

```suggestion
        txs.extend_from_slice(private_mempool);
```

smite/src/bitcoin.rs
```diff
@@ -84,6 +104,60 @@ impl BitcoinCli {
...
+
+        let address = self.get_new_address();
+        let txs_json = serde_json::to_string(&txs).expect("tx list serializes to valid JSON");
```
morehouse: Nit: the `txs_json` line would fit better in the block above.

smite/src/bitcoin.rs
```diff
@@ -84,6 +104,60 @@ impl BitcoinCli {
...
+        let gen_out = self
+            .run()
+            .arg("generateblock")
+            .arg(address.to_string())
+            .arg(&txs_json)
+            .output()
+            .expect("bitcoin-cli generateblock should not fail");
+        assert!(
```
morehouse, line 157: I think this will panic if there's a duplicate tx or if the list isn't topologically ordered.  I think we can't really hit that case currently since the private mempool comes *after* the public one and we deduplicate on txid in the executor.

But we should still probably call out these scenarios in the "Panics" section above.

smite-scenarios/src/executor.rs
```diff
@@ -1389,7 +1415,7 @@ mod tests {
     }
 
     impl BitcoinRpc for MockBitcoinCli {
-        fn mine_blocks(&mut self, num_blocks: u8) {
+        fn mine_blocks(&mut self, num_blocks: u8, _private_mempool: &[String]) {
```
morehouse: We should be able to easily mock a mempool rejection to test the new private mempool behavior.

smite/src/bitcoin.rs
```diff
@@ -219,11 +313,12 @@ impl BitcoinCli {
...
-            broadcast_out.status.success(),
-            "bitcoin-cli sendrawtransaction failed: {}",
-            String::from_utf8_lossy(&broadcast_out.stderr)
-        );
+
+        // The mempool rejects non-standard or low-feerate transactions. Return
+        // the raw hex so the caller can mine it directly later.
+        if !broadcast_out.status.success() {
+            return Some(signed_tx.hex);
+        }
```
morehouse: This considers all failures to be due to non-standardness.  I'm not sure if we can get any other kinds of failures currently, but in the future maybe we could?

I suppose we'll get a panic later when mining if the failure here is due to something else.  Probably that's okay, though debugging may be slightly trickier in that case.

NishantBansal2003: > I'm not sure if we can get any other kinds of failures currently, but in the future maybe we could?

Updated to only check for dust or low feerate related errors (those are the only ones I’ve observed so far). All other errors now cause panic in `sendrawtransaction`

smite-scenarios/src/executor.rs
```diff
@@ -62,17 +62,21 @@ const MAX_MINIMUM_DEPTH: u32 = 8;
...
     /// Returns the scriptPubKey for a newly generated wallet address.
     fn get_new_address_script_pubkey(&mut self) -> ScriptBuf;
 
-    /// Signs and broadcasts a transaction
-    fn sign_and_broadcast_tx(&mut self, tx: &bitcoin::Transaction);
+    /// Signs and broadcasts a transaction. Returns hex-encoded raw transaction
+    /// if it is consensus-valid but rejected by mempool policy, so it can be
+    /// added to the `private_mempool`; returns `None` if it was broadcast or is
+    /// already confirmed.
+    fn sign_and_broadcast_tx(&mut self, tx: &bitcoin::Transaction) -> Option<String>;
```
morehouse, line 82:
```suggestion
    #[must_use]
    fn sign_and_broadcast_tx(&mut self, tx: &bitcoin::Transaction) -> Option<String>;
```

morehouse (approved):
Ready to squash

morehouse (approved):
LGTM

Issue #138, opened 2026-06-29
tmux integration + strategy distribution for `smitebot start`


Ashish-Kumar-Dash:
## Problem

`smitebot start` currently spawns `afl-fuzz` processes as invisible background orphans. This means:

1. **No visibility** — runners overwrite each other's terminal output; no way to see individual AFL++ TUIs
2. **No strategy distribution** — all runners get identical flags. AFL_Runner distributes power schedules, mutation modes, and env vars across runners for better campaign coverage and fuzzing outputs, making `smitebot start` look meek compared to afl-r.
3. **No attach/detach** — no way to reconnect to a running campaign after disconnecting from SSH

## Scope

This issue proposes to modify `smitebot start` to:
- Spawn runners inside tmux windows (one window per runner)
- Distribute AFL++ strategies and env vars across runners (port from AFL_Runner's `MultipleCores` mode)
- Persist the tmux session name in `state.json` for future `stop` command
- Add optional `tmux_session` field to `campaign.toml` as just an extra feature

### tmux session layout

```
tmux session: "smite-lnd-encrypted_bytes-1749465600"  (or custom name from campaign.toml)
  ├── window 0: "runner-0"  →  afl-fuzz -Y -M 0 -i ... -o ... -- /sharedir
  ├── window 1: "runner-1"  →  afl-fuzz -Y -S 1 -i ... -o ... -- /sharedir
  ├── window 2: "runner-2"  →  afl-fuzz -Y -S 2 -i ... -o ... -- /sharedir
  └── ...
```

Each runner gets a dedicated tmux window showing its full AFL++ TUI. One window per runner (not panes), we can add a common pane for all runner in `smitebot status`.

### tmux integration approach

**Direct `std::process::Command` calls to the `tmux` CLI.** 

A small `tmux` module (`smitebot/src/tmux.rs`) with these functions:

```rust
fn is_available() -> bool;              // `tmux -V` succeeds
fn session_exists(name: &str) -> bool;  // `tmux has-session -t <name>`
fn create_session(name: &str, cmd: &[String], envs: &[(String, String)]) -> io::Result<()>;
    // `tmux new-session -d -s <name> -n runner-0 <cmd>`
fn add_window(session: &str, window: &str, cmd: &[String], envs: &[(String, String)]) -> io::Result<()>;
    // `tmux new-window -t <session> -n <window> <cmd>`
fn attach(session: &str) -> io::Result<()>;
[...]

PR #139, opened 2026-07-02
smitebot: tmux integration + strategy distribution for start

Ashish-Kumar-Dash:
## Summary

  - Spawn each `afl-fuzz` runner in its own tmux window instead of invisible background processes, users see individual AFL++ TUIs, can attach/detach via `Ctrl-b d`, and reconnect over SSH
  - Distribute AFL++ power schedules, format hints, and env vars across runners (ported from AFL_Runner's `MultipleCores` mode) for campaign diversity instead of running N identical fuzzers
  - A `fuzzer_stats` older than launch (leftover in a reused `output_dir`) is ignored, so a stale PID is never recorded as a live runner.
  - On failure, preserve tmux session for inspection (`remain-on-exit`) instead of killing it

 ### Strategy distribution

 Round-robin power schedules; `-a binary` on ~70% of secondaries; `AFL_DISABLE_TRIM` on ~60%; `AFL_IMPORT_FIRST` when < 16 runners; `AFL_FINAL_SYNC` on the primary; testcache auto-sized from RAM. IR scenarios auto-inject the custom-mutator env vars. Optional `tmux_session` config key, validated on `CampaignConfig`.

  ## tmux session layout

  tmux session: "lnd-encrypted_bytes-1749465600"  (or custom name from campaign.toml)
    ├── window 0: "runner-0"  →  afl-fuzz -Y -M 0 ...
    ├── window 1: "runner-1"  →  afl-fuzz -Y -S 1 ...
    ├── window 2: "runner-2"  →  afl-fuzz -Y -S 2 ...
    └── ...

  Each runner command is built as a shell string with `exec` so afl-fuzz replaces the shell process directly. Environment variables are prepended as `KEY='val'` assignments, and paths are single-quote escaped.

  ## Features 

  - **One window per runner**, not panes — each runner gets the full AFL++ TUI
  - **Deterministic** flag distribution by runner index (no `rand` dependency)
  - Verification now polls two signals:
   **fresh `fuzzer_stats`** → runner started
   **dead tmux window** (`pane_dead`) → runner failed to launch, reported at once
   neither yet → still calibrating; keep polling to a generous 15-min ceiling that only bounds an alive-but-hung runner
  - **Don't kill session on failure** — `remain-on-exit` preserves error output for debugging;,user cleans up manually or via future `stop` command
  - **Session name collision** doubles as duplicate campaign guard  - Nested tmux handled: `switch-client` when `$TMUX` is set, `attach-session` otherwise
  - **No new dependencies** — tmux CLI called via `std::process::Command`
  
  Ref. #138  #70

### DIFF
--- a/smitebot/README.md
+++ b/smitebot/README.md
@@ -1,65 +1,72 @@
 # smitebot
 
-`smitebot` is the Smite automation CLI. It is intended to orchestrate common fuzzing workflows and reduce manual setup/operations.
+`smitebot` is the Smite automation CLI. It orchestrates fuzzing campaigns against Lightning Network implementations using AFL++ and Nyx, reducing multi-step manual workflows to single commands.
 
 ## Install
 
-Install `smitebot` once from this repository:
-
 ```bash
 cargo install --path smitebot
-```
-
-After install, run it directly:
-
-```bash
-smitebot doctor --aflpp-path ~/AFLplusplus
-smitebot doctor campaign.toml
 ```
 
 ## Configuration
 
 Campaign settings are stored in a TOML file. See [`sample-campaign.toml`](sample-campaign.toml) for a complete example.
 
-| Field        | Required | Description                                                          |
-| ------------ | -------- | -------------------------------------------------------------------- |
-| `target`     | yes      | Lightning implementation to fuzz (`lnd`, `cln`, `ldk`, or `eclair`). |
-| `scenario`   | yes      | Scenario binary selected by the workload Dockerfile.                 |
-| `aflpp_path` | yes      | Path to the AFL++ source tree.                                       |
-| `smite_dir`  | yes      | Path to the smite repository root.                                   |
-| `runners`    | yes      | Number of parallel AFL++ instances to launch (must be at least 1).   |
-| `seed_dir`   | no       | Directory containing seed inputs; omit to start from an empty corpus.|
-| `output_dir` | yes      | AFL++ output directory for findings and stats.                       |
-| `sharedir`   | yes      | Nyx shared directory path; created automatically by `smitebot start`.|
-| `image`      | no       | Docker image tag override; defaults to `smite-<target>-<scenario>`.  |
-| `afl_env`    | no       | Extra environment variables passed to AFL++ instances.               |
-| `afl_flags`  | no       | Extra CLI flags appended to `afl-fuzz`.                              |
+| Field          | Required | Description                                                                               |
+| -------------- | -------- | ----------------------------------------------------------------------------------------- |
+| `target`       | yes      | Lightning implementation to fuzz (`lnd`, `cln`, `ldk`, or `eclair`).                      |
+| `scenario`     | yes      | Scenario binary selected by the workload Dockerfile.                                      |
+| `aflpp_path`   | yes      | Path to the AFL++ source tree.                                                            |
+| `smite_dir`    | yes      | Path to the smite repository root.                                                        |
+| `runners`      | yes      | Number of parallel AFL++ instances to launch (must be at least 1).                        |
+| `seed_dir`     | no       | Directory containing seed inputs; omit to start from an empty corpus.                     |
+| `output_dir`   | yes      | AFL++ output directory for findings and stats.                                            |
+| `sharedir`     | yes      | Nyx shared directory path; created automatically by `smitebot start`.                     |
+| `image`        | no       | Docker image tag override; defaults to `smite-<target>-<scenario>`.                       |
+| `tmux_session` | no       | Custom tmux session name; defaults to the campaign ID. Must not contain `:`, `.`, or `#`. |
+| `afl_env`      | no       | Extra environment variables passed to AFL++ instances.                                    |
+| `afl_flags`    | no       | Extra CLI flags appended to `afl-fuzz`.                                                   |
 
 ## Commands
 
 ### smitebot start
 
-`smitebot start` launches a fuzzing campaign in the background. It builds the Docker image, sets up the Nyx sharedir, and spawns parallel AFL++ instances.
+Launches a fuzzing campaign. Builds the Docker image, sets up the Nyx sharedir, spawns parallel AFL++ instances inside a tmux session (one window per runner), and attaches to the session.
 
 ```bash
 smitebot start campaign.toml
 ```
 
-For IR scenarios (scenario names starting with `ir`), the required AFL++ custom mutator environment variables are injected automatically.
+Each runner gets a deterministic strategy distribution:
+- Power schedule (`-p`): round-robin across `fast`, `explore`, `coe`, `lin`, `quad`, `exploit`, `rare`
+- `-a binary`: ~70% of secondary runners
+- `AFL_DISABLE_TRIM`: ~60% of secondary runners
+- `AFL_FINAL_SYNC`: primary runner only
+- `AFL_IMPORT_FIRST`: enabled when runner count < 16
+- `AFL_TESTCACHE_SIZE`: auto-sized from available RAM
+
+For IR scenarios (scenario names starting with `ir`), the required AFL++ custom mutator environment variables are injected automatically. User `afl_env` values override strategy defaults.
+
+`start` begins fresh campaigns only. If `output_dir` already holds a prior run's `fuzzer_stats`, it exits with an error instead of resuming (resume is not yet supported).
+
+After spawning, startup is verified by polling for `fuzzer_stats` files. Because AFL++ writes `fuzzer_stats` only after calibrating every seed (minutes under Nyx), a runner is reported as failed the moment its tmux window exits rather than after a fixed timeout; the poll otherwise waits up to a generous ceiling (10 min) for a runner that stays alive but never starts. On failure, the tmux session is preserved with `remain-on-exit` so error output can be inspected.
 
 Campaign state is saved to `~/.smitebot/runs/<campaign-id>/state.json` for use by future `stop` and `status` commands.
 
 ### smitebot config
 
-`smitebot config` validates a campaign configuration file, reports the resolved settings, and checks that referenced paths exist on disk.
+Validates a campaign configuration file, reports the resolved settings, and checks that referenced paths exist on disk.
 
 ```bash
-smitebot config sample-campaign.toml
+smitebot config campaign.toml
+smitebot config campaign.toml --json
 ```
+
+- `--json`: Emit machine-readable JSON output. Both success and error paths produce valid JSON.
 
 ### smitebot build
 
-`smitebot build` builds Smite workload Docker images. It can be used standalone with CLI flags or with a campaign config file. When a config file is provided, CLI flags override individual values.
+Builds Smite workload Docker images. Accepts a campaign config file or standalone CLI flags. When both are provided, CLI flags override config values.
 
 ```bash
 smitebot build --target lnd --scenario encrypted_bytes
@@ -68,25 +75,18 @@ smitebot build campaign.toml --target cln
 smitebot build campaign.toml --coverage --no-cache
 ```
 
-Flags:
-
 - `--target`: Target implementation to build. Required when no config file is provided.
 - `--scenario`: Scenario binary for the workload Dockerfile. Required when no config file is provided.
 - `--smite-dir`: Path to the smite repository root; defaults to `.` when no config file is provided.
 - `--coverage`: Build a coverage-instrumented image.
-- `--image`: Docker image tag; overrides the config value and the default smite naming convention.
+- `--image`: Docker image tag; overrides the config value and the default naming convention.
 - `--no-cache`: Perform a clean rebuild without using cached Docker layers.
 
-By default, image tags follow the existing Smite convention:
-
-```text
-smite-<target>-<scenario>
-smite-<target>-<scenario>-coverage
-```
+Image tags follow the Smite convention: `smite-<target>-<scenario>` or `smite-<target>-<scenario>-coverage`.
 
 ### smitebot doctor
 
-`smitebot doctor` validates host prerequisites before running Smite campaigns. It can be used standalone with CLI flags or with a campaign config file. When a config file is provided, CLI flags override individual values.
+Validates host prerequisites before running Smite campaigns. Accepts a campaign config file or standalone CLI flags. When both are provided, CLI flags override config values.
 
 ```bash
 smitebot doctor --aflpp-path ~/AFLplusplus
@@ -95,7 +95,11 @@ smitebot doctor campaign.toml --json
 smitebot doctor campaign.toml --aflpp-path ~/other-aflpp
 ```
 
-## Checks
+- `--aflpp-path`: Path to AFL++ source tree. Required when no config file is provided.
+- `--smite-dir`: Path to the smite repository root; overrides the config value.
+- `--json`: Emit machine-readable JSON output.
+
+Checks performed:
 
 - `x86_64` architecture
 - CPU virtualization enabled (`vmx` or `svm`)
@@ -103,14 +107,12 @@ smitebot doctor campaign.toml --aflpp-path ~/other-aflpp
 - Docker daemon is reachable (`docker version`)
 - AFL++ built with Nyx support (`libnyx.so` under `--aflpp-path`)
 - VMware backdoor is enabled
-- AFL++ tools (`afl-fuzz`, `afl-cmin`, `afl-tmin`, `afl-whatsup`) are executable under `--aflpp-path`
-- Required host tools (`bash`, `python`, `python3`)
+- AFL++ tools (`afl-fuzz`, `afl-cmin`, `afl-tmin`, `afl-whatsup`) are executable
+- Required host tools (`bash`, `python`, `python3`, `tmux`)
 - Required Smite scripts are present and executable
 - Required workload Dockerfiles are present
 
-## JSON output
-
-By default, output is in a human readable format. The `--json` flag changes output to structured JSON:
+JSON output example:
 
 ```json
 {
--- a/smitebot/src/commands/start.rs
+++ b/smitebot/src/commands/start.rs
@@ -1,13 +1,12 @@
 //! Campaign launch orchestration.
 //!
 //! Builds the Docker image, prepares the Nyx sharedir, spawns parallel
-//! `afl-fuzz` processes in the background, and persists campaign state so
+//! `afl-fuzz` processes inside a tmux session, and persists campaign state so
 //! that `stop` and `status` can manage the running campaign later.
 
 use std::fs;
-use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
-use std::process::{Child, Command};
+use std::process::Command;
 use std::thread;
 use std::time::{Duration, Instant};
 
@@ -16,12 +15,23 @@ use clap::Args;
 use crate::commands::build::{BuildInputs, run_build};
 use crate::config::CampaignConfig;
 use crate::state::{CampaignState, RunnerState, Status};
+use crate::tmux;
 
-/// Maximum time to wait for all runners to report non-zero `execs_per_sec`.
-const STARTUP_TIMEOUT: Duration = Duration::from_mins(5);
+/// Safety ceiling for startup verification.
+///
+/// Genuine launch failures are caught immediately when a runner's tmux window
+/// dies (see `verify_startup`), so this only bounds an alive-but-hung runner.
+/// Nyx seed calibration can take several minutes because each exec restores a
+/// VM snapshot, so the ceiling is deliberately generous; its exact value is not
+/// load-bearing — it only prevents an indefinite hang. Kept comfortably above a
+/// measured fresh start (~45s) so a large seed corpus is not falsely failed.
+const VERIFY_TIMEOUT: Duration = Duration::from_mins(10);
 
 /// How often to poll `fuzzer_stats` files during startup verification.
-const STARTUP_POLL_INTERVAL: Duration = Duration::from_secs(5);
+const VERIFY_POLL_INTERVAL: Duration = Duration::from_secs(5);
+
+/// AFL++ power schedules for round-robin distribution across runners.
+const POWER_SCHEDULES: &[&str] = &["fast", "explore", "coe", "lin", "quad", "exploit", "rare"];
 
 /// Command handler for `smitebot start`.
 pub struct StartCommand;
@@ -52,6 +62,34 @@ impl StartCommand {
             return false;
         }
 
+        if !tmux::is_available() {
+            log::error!("tmux is not available; install it or run `smitebot doctor` for details");
+            return false;
+        }
+
+        let campaign_id = config.campaign_id();
+        let tmux_session = config
+            .tmux_session
+            .clone()
+            .unwrap_or_else(|| campaign_id.clone());
+
+        if tmux::session_exists(&tmux_session) {
+            log::error!(
+                "tmux session '{tmux_session}' already exists — is another campaign running?"
+            );
+            return false;
+        }
+
+        if let Some(stats) = existing_campaign_runner(&config.output_dir, config.runners) {
+            log::error!(
+                "output_dir already holds a campaign ({}); smitebot start only \
+                 begins fresh campaigns and does not resume — remove the directory \
+                 or set a different output_dir",
+                stats.display()
+            );
+            return false;
+        }
+
         let image = config.image_tag();
 
         log::info!("building Docker image {image}");
@@ -77,7 +115,6 @@ impl StartCommand {
             }
         };
 
-        let campaign_id = config.campaign_id();
         let Some(runs_dir) = CampaignState::runs_dir() else {
             log::error!("unable to determine home directory");
             return false;
@@ -94,7 +131,14 @@ impl StartCommand {
             return false;
         };
 
-        let mut state = CampaignState::new(campaign_id, &config, image, image_digest, git_hash);
+        let mut state = CampaignState::new(
+            campaign_id,
+            &config,
+            image,
+            image_digest,
+            git_hash,
+            tmux_session,
+        );
 
         if let Err(e) = state.save(&state_path) {
             log::error!("{e}");
@@ -111,6 +155,12 @@ impl StartCommand {
 
         log::info!("campaign {} is running", state.id);
         log::info!("state saved to {}", state_path.display());
+
+        log::info!("attaching to tmux session '{}'", state.tmux_session);
+        if let Err(e) = tmux::attach(&state.tmux_session) {
+            log::warn!("failed to attach to tmux session: {e}");
+        }
+
         true
     }
 }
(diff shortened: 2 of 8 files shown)


Ashish-Kumar-Dash:
Tested locally with 2 runners on lnd/encrypted_bytes:
 
Screenshot from 2026-07-03 23-50-39

Build + Nyx sharedir setup + tmux session launch logs

Screenshot from 2026-07-03 23-52-16

Runner 0 ([fast], primary -M), ~440 execs/sec, 119 corpus, trim disabled

Screenshot from 2026-07-03 23-52-22

Runner 1 ([explore], secondary -S), ~449 execs/sec, 108 corpus, 7 imported

 Strategy distribution, window naming, and Nyx mode all working as expected.  Killed the session with tmux kill-session -t <session>, stop command will automate this.

smitebot/src/config.rs
```diff
@@ -83,6 +85,8 @@ pub enum ConfigError {
     EmptyScenario,
     #[error("image must not be empty when specified")]
     EmptyImage,
+    #[error("tmux_session must not be empty when specified")]
+    EmptyTmuxSession,
```
NishantBansal2003: tmux 3.6b and earlier silently replace : and . in session names, so I think we should reject session names containing those characters. Although this restriction was removed in tmux 3.7, that change is relatively recent, so I think it’s still better to error out when they are used.
ref:
- https://github.com/tmux/tmux/blob/3.6b/session.c#L232-L248
- https://github.com/tmux/tmux/commit/65a032b2
- https://github.com/tmux/tmux/commit/d339ab51#diff-26fff670d68e7fdb1db431ab939f5a20ae8b697540b249843fce97dd349fe8e7

When you update this, please also reference the relevant tmux commits in the code comments so it’s clear why this validation exists.

Also, I think there may be other characters with the same behavior, It’s worth checking, otherwise we might end up behaving weirdly without noticing

smitebot/src/tmux.rs
```diff
@@ -0,0 +1,69 @@
...
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .status()
+        .is_ok_and(|s| s.success())
+}
+
+/// Returns `true` if a tmux session with the given name exists.
+pub fn session_exists(name: &str) -> bool {
+    Command::new("tmux")
+        .args(["has-session", "-t", name])
```
NishantBansal2003: tmux does prefix matching by default, I think we should force exact matching for session names

smitebot/src/tmux.rs
```diff
@@ -0,0 +1,69 @@
...
+        name,
+        "-n",
+        window_name,
+        shell_cmd,
+    ])
+}
+
+/// Adds a new window running `shell_cmd` to an existing tmux session.
+pub fn add_window(session: &str, window_name: &str, shell_cmd: &str) -> io::Result<()> {
+    run_tmux(&["new-window", "-t", session, "-n", window_name, shell_cmd])
```
NishantBansal2003: tmux does prefix matching by default, I think we should force exact matching for session names

smitebot/src/tmux.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/// Attaches to a session, using `switch-client` when already inside tmux.
+pub fn attach(session: &str) -> io::Result<()> {
+    let cmd = if std::env::var_os("TMUX").is_some() {
+        "switch-client"
+    } else {
+        "attach-session"
+    };
+    run_tmux(&[cmd, "-t", session])
```
NishantBansal2003: tmux does prefix matching by default, I think we should force exact matching for session names

smitebot/src/tmux.rs
```diff
@@ -0,0 +1,69 @@
...
+    ])
+}
+
+/// Adds a new window running `shell_cmd` to an existing tmux session.
+pub fn add_window(session: &str, window_name: &str, shell_cmd: &str) -> io::Result<()> {
+    run_tmux(&["new-window", "-t", session, "-n", window_name, shell_cmd])
+}
+
+/// Enables `remain-on-exit` so dead windows preserve their error output.
+pub fn set_remain_on_exit(session: &str) -> io::Result<()> {
```
NishantBansal2003: We should arm this in the same cmd that creates or attaches to the window. Otherwise, we could hit a race where a cmd that exits immediately takes the window down before the option is applied, causing us to lose its error output

smitebot/README.md
```diff
@@ -1,65 +1,70 @@
...
+| `target`       | yes      | Lightning implementation to fuzz (`lnd`, `cln`, `ldk`, or `eclair`).  |
+| `scenario`     | yes      | Scenario binary selected by the workload Dockerfile.                  |
+| `aflpp_path`   | yes      | Path to the AFL++ source tree.                                        |
+| `smite_dir`    | yes      | Path to the smite repository root.                                    |
+| `runners`      | yes      | Number of parallel AFL++ instances to launch (must be at least 1).    |
+| `seed_dir`     | no       | Directory containing seed inputs; omit to start from an empty corpus. |
+| `output_dir`   | yes      | AFL++ output directory for findings and stats.                        |
+| `sharedir`     | yes      | Nyx shared directory path; created automatically by `smitebot start`. |
+| `image`        | no       | Docker image tag override; defaults to `smite-<target>-<scenario>`.   |
+| `tmux_session` | no       | Custom tmux session name; defaults to the campaign ID.                |
```
NishantBansal2003: Invalid session names should be documented here as well

smitebot/README.md
```diff
@@ -1,65 +1,70 @@
 # smitebot
```
NishantBansal2003, line 1: Changes unrelated to this PR/commit should go in a separate PR or commit

smitebot/src/commands/start.rs
```diff
@@ -52,6 +55,11 @@ impl StartCommand {
             return false;
         }
 
+        if !tmux::is_available() {
+            log::error!("tmux is not installed; run `smitebot doctor` for details");
```
NishantBansal2003: Here we should say tmux is not available and to install it, since that’s what smitebot doctor says

smitebot/src/commands/start.rs
```diff
@@ -78,6 +86,18 @@ impl StartCommand {
...
+            .tmux_session
+            .clone()
+            .unwrap_or_else(|| campaign_id.clone());
+
+        if tmux::session_exists(&tmux_session) {
+            log::error!(
+                "tmux session '{tmux_session}' already exists — is another campaign running?"
+            );
+            return false;
+        }
```
NishantBansal2003: I think we should error out early if this happens, so maybe move this to the docker image build step?

smitebot/src/commands/start.rs
```diff
@@ -216,139 +268,166 @@ fn ensure_seed_dir(config: &CampaignConfig) -> std::io::Result<PathBuf> {
...
+        .and_then(|v| v.parse::<u64>().ok())
+        .map(|kb| kb / 1024)?;
+
+    Some(if free_mb > 32_000 {
+        500
+    } else if free_mb > 8_000 {
+        250
+    } else {
+        50
     })
```
NishantBansal2003, line 466: I know 50–500 MB is recommended, but is there any reason behind a threshold like >8_000/32_000?

smitebot/src/commands/start.rs
```diff
@@ -216,139 +268,166 @@ fn ensure_seed_dir(config: &CampaignConfig) -> std::io::Result<PathBuf> {
...
+    // AFL_DISABLE_TRIM on ~60% of runners.
+    let trim_count = (runner_count as usize * 6).div_ceil(10);
+    if (runner_id as usize) < trim_count {
+        envs.push(("AFL_DISABLE_TRIM".to_string(), "1".to_string()));
+    }
 
-        thread::sleep(STARTUP_POLL_INTERVAL);
+    if runner_count < 16 {
+        envs.push(("AFL_IMPORT_FIRST".to_string(), "1".to_string()));
     }
```
NishantBansal2003, line 439: What’s the reasoning behind 16 here? I think we should add a comment explaining it so readers know the rationale behind this choice

smitebot/src/commands/start.rs
```diff
@@ -145,50 +178,69 @@ fn setup_nyx(config: &CampaignConfig, image: &str) -> bool {
...
+            tmux::add_window(session, &window_name, &cmd)
+        };
+
+        if let Err(e) = result {
+            log::error!("failed to create tmux window for runner {id}: {e}");
+            if !runners.is_empty() {
+                log::info!(
+                    "inspect tmux session '{session}' for error output, \
+                     then: tmux kill-session -t {session}"
+                );
```
NishantBansal2003: We can refactor this into `fail_campaign`, since both fail the campaign in the same way

smitebot/src/commands/start.rs
```diff
@@ -428,75 +507,194 @@ mod tests {
...
+    }
+
+    #[test]
+    fn read_fuzzer_pid_returns_none_for_missing_file() {
+        let dir = tempfile::tempdir().unwrap();
+        assert_eq!(read_fuzzer_pid(&dir.path().join("missing")), None);
+    }
+
+    #[test]
+    fn verify_startup_reads_pids_from_fuzzer_stats() {
```
NishantBansal2003, line 628: Can add test like `verify_startup` fails when stats missing

NishantBansal2003 (changes requested):

smitebot/README.md
Ashish-Kumar-Dash: I feel its a minor change so better to push it in this PR itself, than carrying it for the next PR & starting the review cycle again

smitebot/src/commands/start.rs
Ashish-Kumar-Dash: added comment

smitebot/src/commands/start.rs
Ashish-Kumar-Dash: They're arbitrary heuristics, if there's a specific threshold you want me to check for then do inform

smitebot/src/config.rs
Ashish-Kumar-Dash: Yes, I also found that # is silently replaced by tmux 3.6b (in addition to : and  .), confirmed from session_check_name in tmux. Added validation for all three.

smitebot/src/tmux.rs
```diff
@@ -33,27 +37,30 @@ pub fn create_session(name: &str, window_name: &str, shell_cmd: &str) -> io::Res
...
 /// Adds a new window running `shell_cmd` to an existing tmux session.
 pub fn add_window(session: &str, window_name: &str, shell_cmd: &str) -> io::Result<()> {
-    run_tmux(&["new-window", "-t", session, "-n", window_name, shell_cmd])
-}
-
-/// Enables `remain-on-exit` so dead windows preserve their error output.
-pub fn set_remain_on_exit(session: &str) -> io::Result<()> {
-    run_tmux(&["set-option", "-t", session, "remain-on-exit", "on"])
+    let target = format!("={session}");
+    run_tmux(&["new-window", "-t", &target, "-n", window_name, shell_cmd])
```
NishantBansal2003: I’m not entirely sure, but I think `remain-on-exit` does not inherit from earlier windows, so we need to set this option here as well?

smitebot/src/commands/start.rs
```diff
@@ -216,139 +265,171 @@ fn ensure_seed_dir(config: &CampaignConfig) -> std::io::Result<PathBuf> {
...
-                        runner.pid
-                    );
-                    return false;
-                }
-            }
-        }
+    // AFL_DISABLE_TRIM on ~60% of runners.
+    let trim_count = (runner_count as usize * 6).div_ceil(10);
+    if (runner_id as usize) < trim_count {
+        envs.push(("AFL_DISABLE_TRIM".to_string(), "1".to_string()));
```
NishantBansal2003: I think AFL++ suggests applying this (`AFL_DISABLE_TRIM`) env only to secondary instances, not the primary? see: https://aflplus.plus/docs/fuzzing_in_depth/#c-using-multiple-cores

I think the same applies to the `-a binary` flag as well?

smitebot/src/config.rs
```diff
@@ -87,6 +87,11 @@ pub enum ConfigError {
     EmptyImage,
     #[error("tmux_session must not be empty when specified")]
     EmptyTmuxSession,
+    #[error(
+        "tmux_session contains invalid character '{0}': tmux 3.6b and earlier silently \
+             replace these with underscores (see tmux/tmux@65a032b2)"
+    )]
+    InvalidTmuxSessionChar(char),
```
NishantBansal2003: We could use `InvalidTmuxSessionName` and merge both errors. Also, I meant to add a comment in the code for tmux commit (like: `https://github.com/tmux/tmux/commit/d339ab51`) so future readers understand the reasoning

smitebot/README.md
```diff
@@ -1,65 +1,70 @@
...
+| `aflpp_path`   | yes      | Path to the AFL++ source tree.                                        |
+| `smite_dir`    | yes      | Path to the smite repository root.                                    |
+| `runners`      | yes      | Number of parallel AFL++ instances to launch (must be at least 1).    |
+| `seed_dir`     | no       | Directory containing seed inputs; omit to start from an empty corpus. |
+| `output_dir`   | yes      | AFL++ output directory for findings and stats.                        |
+| `sharedir`     | yes      | Nyx shared directory path; created automatically by `smitebot start`. |
+| `image`        | no       | Docker image tag override; defaults to `smite-<target>-<scenario>`.   |
+| `tmux_session` | no       | Custom tmux session name; defaults to the campaign ID. Must not contain `:`, `.`, or `#`. |
+| `afl_env`      | no       | Extra environment variables passed to AFL++ instances.                |
+| `afl_flags`    | no       | Extra CLI flags appended to `afl-fuzz`.                               |
```
NishantBansal2003:
```suggestion
| Field          | Required | Description                                                                               |
| -------------- | -------- | ----------------------------------------------------------------------------------------- |
| `target`       | yes      | Lightning implementation to fuzz (`lnd`, `cln`, `ldk`, or `eclair`).                      |
| `scenario`     | yes      | Scenario binary selected by the workload Dockerfile.                                      |
| `aflpp_path`   | yes      | Path to the AFL++ source tree.                                                            |
| `smite_dir`    | yes      | Path to the smite repository root.                                                        |
| `runners`      | yes      | Number of parallel AFL++ instances to launch (must be at least 1).                        |
| `seed_dir`     | no       | Directory containing seed inputs; omit to start from an empty corpus.                     |
| `output_dir`   | yes      | AFL++ output directory for findings and stats.                                            |
| `sharedir`     | yes      | Nyx shared directory path; created automatically by `smitebot start`.                     |
| `image`        | no       | Docker image tag override; defaults to `smite-<target>-<scenario>`.                       |
| `tmux_session` | no       | Custom tmux session name; defaults to the campaign ID. Must not contain `:`, `.`, or `#`. |
| `afl_env`      | no       | Extra environment variables passed to AFL++ instances.                                    |
| `afl_flags`    | no       | Extra CLI flags appended to `afl-fuzz`.                                                   |
```

NishantBansal2003 (changes requested):
- I think we need a longer timeout for verification. When I first tried running smitebot, it reported an error, but when I checked the tmux windows, both fuzzers were running normally. I haven’t seen this issue again since then, though:
```
INFO  [smitebot::commands::start] Nyx sharedir ready at /tmp/smitebot/smite-nyx
INFO  [smitebot::commands::start] starting 2 runners in tmux session 'lnd-encrypted_bytes-1783323599'
INFO  [smitebot::commands::start] verifying runners started
ERROR [smitebot::commands::start] runner 0 did not produce fuzzer_stats within timeout
ERROR [smitebot::commands::start] runner 1 did not produce fuzzer_stats within timeout
ERROR [smitebot::commands::start] not all runners produced fuzzer_stats within timeout
INFO  [smitebot::commands::start] inspect tmux session 'lnd-encrypted_bytes-1783323599' for error output, then: tmux kill-session -t lnd-encrypted_bytes-1783323599
```
- Also, when startup succeeds, tmux takes over the terminal (with last runner). Does it make sense to keep printing the smitebot logs normally and run tmux in the background instead? Then, once something like `attaching to tmux session 'lnd-encrypted_bytes-1783324069'` is printed, we could inspect the session ourselves if needed

smitebot/src/tmux.rs
Ashish-Kumar-Dash: Fixed by chaining, set-option -w  remain-on-exit on in add_window, same pattern as create_session.

smitebot/src/commands/start.rs
Ashish-Kumar-Dash: Yes, fixed, both -a binary (~70%) and AFL_DISABLE_TRIM (~60%) now only apply to runner_id > 0, with percentages calculated against the secondary count.

Ashish-Kumar-Dash:
> * I think we need a longer timeout for verification. When I first tried running smitebot, it reported an error, but when I checked the tmux windows, both fuzzers were running normally. I haven’t seen this issue again since then, though:
> 
> ```
> INFO  [smitebot::commands::start] Nyx sharedir ready at /tmp/smitebot/smite-nyx
> INFO  [smitebot::commands::start] starting 2 runners in tmux session 'lnd-encrypted_bytes-1783323599'
> INFO  [smitebot::commands::start] verifying runners started
> ERROR [smitebot::commands::start] runner 0 did not produce fuzzer_stats within timeout
> ERROR [smitebot::commands::start] runner 1 did not produce fuzzer_stats within timeout
> ERROR [smitebot::commands::start] not all runners produced fuzzer_stats within timeout
> INFO  [smitebot::commands::start] inspect tmux session 'lnd-encrypted_bytes-1783323599' for error output, then: tmux kill-session -t lnd-encrypted_bytes-1783323599
> ```
> 
> * Also, when startup succeeds, tmux takes over the terminal (with last runner). Does it make sense to keep printing the smitebot logs normally and run tmux in the background instead? Then, once something like `attaching to tmux session 'lnd-encrypted_bytes-1783324069'` is printed, we could inspect the session ourselves if needed

The old `verify_startup`  in start.rs used a fixed wall-clock timeout (30s, then 60s) to decide if a runner came up. But `fuzzer_stats` , the file it polls, is only written *after* AFL++ calibrates every seed, and under Nyx each exec restores a VM snapshot, so first stats routinely lands minutes in. The timeout was racing a startup cost- it passed warm and failed cold or on larger corpora. When it lost, it didn't just error, it reported a healthy campaign as failed and left the Nyx/QEMU VMs running orphaned (the zombie-process problem). Reproduced live- 30s timeout, first stats at ~46s, panes alive the whole time.

The fix stops timing the outcome. It now decides on process state: a fresh `fuzzer_stats` means started, a dead tmux window means the launch failed (reported immediately, not after a wait); anything else means still calibrating, so keep polling. Cold vs warm vs corpus size no longer changes the verdict. The 15-min ceiling is a backstop for an alive-but-hung runner.

Also closed an adjacent latent bug: on a reused `output_dir`, a stale `fuzzer_stats` from a prior run could be read as a fresh start and record a dead PID (which `stop` would later signal, orphaning the real VMs). Startup now ignores any `fuzzer_stats` older than launch. Regression test included.

NishantBansal2003:
IIUC, if the `output_dir/seed` (25+ min of fuzzing) already exists, `smitebot start` currently just returns an error unless `AFL_AUTORESUME=1` is set. So, were you getting the startup failed error with an existing `output_dir` and `AFL_AUTORESUME=1` enabled? If so, there's also a chance that 15 min wouldn't be enough if the seed directory is large (see: https://github.com/AFLplusplus/AFLplusplus/blob/011cd189801830253c66ecd3cd6919ec01b46c34/src/afl-fuzz-init.c#L1995-L2014)

From what I understood from your reply in https://github.com/morehouse/smite/pull/119#discussion_r3426455662, `smitebot start` is only meant for starting a new campaign, so the `output_dir` should not already exist. Based on that, if the `output_dir` is present, we could instead treat it as a resume mode (which doesn’t currently exist), or handle resuming separately.

Leaving that case aside, we're only left with starting a fresh campaign, either as a cold or warm start, which I believe should be covered by simply increasing the timeout to 60s (or appropriate). WDYT?

Ashish-Kumar-Dash:
> From what I understood from your reply in [#119 (comment)](https://github.com/morehouse/smite/pull/119#discussion_r3426455662), `smitebot start` is only meant for starting a new campaign, so the `output_dir` should not already exist. Based on that, if the `output_dir` is present, we could instead treat it as a resume mode (which doesn’t currently exist), or handle resuming separately.

Yes, that should be covered in `smitebot resume`, I'll make the changes for start to be new campaign only, I'll add resume mode in later PR. Apologies for the confusion

> IIUC, if the `output_dir/seed` (25+ min of fuzzing) already exists, `smitebot start` currently just returns an error unless `AFL_AUTORESUME=1` is set. So, were you getting the startup failed error with an existing `output_dir` and `AFL_AUTORESUME=1` enabled? 

no, It was a plain fresh start.

> Leaving that case aside, we're only left with starting a fresh campaign, either as a cold or warm start, which I believe should be covered by simply increasing the timeout to 60s (or appropriate). WDYT?

I'd push back on a fixed 60s. Liveness doesn't depend on guessing how long calibration takes. It fails fast when a runner's window dies, and waits while it's alive, so a healthy-but-slow runner is never reported as failed, regardless of corpus size or cold/warm machine. A fixed 60s is a bet on calibration finishing in time, my measured run was already ~45s, so 60s has ~15s of margin and maybe loses it on a colder machine or larger corpus. The 15-min ceiling is only a checkpost for a truly hung runner, not the success criteria. That's the reason why I made the fix, 60s can wrongly fail a working campaign, liveness only "fails" a runner that has actually died or hung.  

If the liveness check ever feels like too much, we can always fall back to 60s/fixed timeout, it's a small change. But since this PR's real weight is the tmux integration + strategy distribution, I'd happy to revisit the timeout approach in a follow-up. When people start using `smitebot start` regularly, which is what this PR aims, we'd have more data to make an informed decision for later.

smitebot/src/config.rs
```diff
@@ -163,6 +167,23 @@ impl CampaignConfig {
...
+            // tmux 3.6b silently replaces ':', '.', and '#' with '_' in session
+            // names (tmux/tmux@d339ab51). Reject them to avoid silent renaming.
+            for c in [':', '.', '#'] {
+                if session.contains(c) {
+                    return Err(ConfigError::InvalidTmuxSessionName(format!(
+                        "contains '{c}': tmux 3.6b silently replaces these \
+                         with underscores (see tmux/tmux@d339ab51)"
+                    )));
+                }
+            }
```
NishantBansal2003:
```suggestion
            // tmux 3.6b and earlier silently replace ':', '.', and '#' with '_'
            // in session names (https://github.com/tmux/tmux/commit/d339ab51).
            // Reject these characters to avoid unexpected session names.
            for c in [':', '.', '#'] {
                if session.contains(c) {
                    return Err(ConfigError::InvalidTmuxSessionName(format!(
                        "contains '{c}': tmux 3.6b and earlier replace this character with '_' in session names"
                    )));
                }
            }
```

smitebot/src/config.rs
```diff
@@ -323,6 +345,56 @@ sharedir = "/tmp/smite-nyx"
...
+    }
+
+    #[test]
+    fn load_rejects_empty_tmux_session() {
+        let dir = tempfile::tempdir().unwrap();
+        let content = format!("{VALID_CONFIG}tmux_session = \"\"\n");
+        let path = write_config(dir.path(), &content);
+
+        let err = CampaignConfig::load(&path).unwrap_err();
+        assert!(matches!(err, ConfigError::InvalidTmuxSessionName(_)));
```
NishantBansal2003: nit: could assert the complete error string, same applies to the tests below as well

smitebot/src/commands/start.rs
```diff
@@ -216,139 +278,221 @@ fn ensure_seed_dir(config: &CampaignConfig) -> std::io::Result<PathBuf> {
...
+            if runner.pid.is_some() {
+                continue;
+            }
+            let stats_path = output_dir.join(runner.name()).join("fuzzer_stats");
+            if let Some(pid) = read_fuzzer_pid(&stats_path) {
+                runner.pid = Some(pid);
+            } else {
+                all_ready = false;
+            }
+        }
```
NishantBansal2003, line 342: Since resuming isn’t supported yet, let’s return an explicit error. Otherwise, when resuming, this command will just see the existing `fuzzer_stats` file and incorrectly assume everything is ready

smitebot/src/commands/start.rs
```diff
@@ -16,12 +15,22 @@ use clap::Args;
...
+/// Genuine launch failures are caught immediately when a runner's tmux window
+/// dies (see `verify_startup`), so this only bounds an alive-but-hung runner.
+/// Nyx seed calibration can take several minutes because each exec restores a
+/// VM snapshot, so the ceiling is deliberately generous; its exact value is not
+/// load-bearing — it only prevents an indefinite hang.
+const VERIFY_TIMEOUT: Duration = Duration::from_mins(15);
 
 /// How often to poll `fuzzer_stats` files during startup verification.
-const STARTUP_POLL_INTERVAL: Duration = Duration::from_secs(5);
+const VERIFY_POLL_INTERVAL: Duration = Duration::from_secs(2);
```
NishantBansal2003: Should we increase this to, say, 5s? Now that the upper bound on the timeout is much larger, we probably don’t need to poll every 2s?

smitebot/src/commands/start.rs
```diff
@@ -16,12 +15,22 @@ use clap::Args;
...
-/// Maximum time to wait for all runners to report non-zero `execs_per_sec`.
-const STARTUP_TIMEOUT: Duration = Duration::from_mins(5);
+/// Safety ceiling for startup verification.
+///
+/// Genuine launch failures are caught immediately when a runner's tmux window
+/// dies (see `verify_startup`), so this only bounds an alive-but-hung runner.
+/// Nyx seed calibration can take several minutes because each exec restores a
+/// VM snapshot, so the ceiling is deliberately generous; its exact value is not
+/// load-bearing — it only prevents an indefinite hang.
+const VERIFY_TIMEOUT: Duration = Duration::from_mins(15);
```
NishantBansal2003: 15min timeout is probably a bit much, and I don’t think we need to wait that long for a fresh campaign, Maybe we could reduce it a little, though I don’t have a strong opinion on the exact value

smitebot/src/commands/start.rs
```diff
@@ -216,139 +289,235 @@ fn ensure_seed_dir(config: &CampaignConfig) -> std::io::Result<PathBuf> {
...
+    let schedule = POWER_SCHEDULES[runner_id as usize % POWER_SCHEDULES.len()];
+    flags.extend(["-p".to_string(), schedule.to_string()]);
+
+    if runner_id == 0 {
+        envs.push(("AFL_FINAL_SYNC".to_string(), "1".to_string()));
+    }
+
+    // AFL++ docs: the strategy knobs below apply to "the other secondaries",
+    // not the primary. See fuzzing_in_depth.md §c "Using multiple cores".
+    if runner_id > 0 {
```
morehouse, line 419: What about the `-P`, `-L`, and `-Z` options described in #138?

smitebot/src/commands/start.rs
```diff
@@ -216,139 +289,235 @@ fn ensure_seed_dir(config: &CampaignConfig) -> std::io::Result<PathBuf> {
...
+fn runner_strategy(
+    runner_id: u16,
+    runner_count: u16,
+    testcache_mb: Option<u64>,
+) -> (Vec<String>, Vec<(String, String)>) {
+    let mut flags = Vec::new();
+    let mut envs = Vec::new();
+
+    let schedule = POWER_SCHEDULES[runner_id as usize % POWER_SCHEDULES.len()];
+    flags.extend(["-p".to_string(), schedule.to_string()]);
```
morehouse, line 411: IIUC, the AFL doc recommends only setting `-p` for secondaries, not the master.

smitebot/src/commands/start.rs
```diff
@@ -428,75 +597,248 @@ mod tests {
...
-        let runners = vec![RunnerState { id: 0, pid: 100 }];
+        // Empty output dir (and one with only the synthesized .seeds) is fresh.
+        assert_eq!(existing_campaign_runner(dir.path(), 4), None);
+    }
+
+    #[test]
+    fn runner_strategy_primary_gets_final_sync() {
+        let (flags, envs) = runner_strategy(0, 8, None);
+        assert!(envs.iter().any(|(k, v)| k == "AFL_FINAL_SYNC" && v == "1"));
+        assert!(flags.contains(&"-p".to_string()));
```
morehouse, line 696: The second assert doesn't look related to the test name.


```suggestion
```

smitebot/src/config.rs
```diff
@@ -38,6 +38,8 @@ pub struct CampaignConfig {
     /// Extra CLI flags appended to the `afl-fuzz` command.
     #[serde(default)]
     pub afl_flags: Vec<String>,
+    /// Custom tmux session name; defaults to the campaign ID when absent.
+    pub tmux_session: Option<String>,
```
morehouse, line 42: `sample-campaign.toml` needs to be updated with this option.

smitebot/README.md
morehouse: Part of the reason review cycles take so long is that the PRs themselves are large and may require multiple hours of the reviewer's time to properly review.  I know I won't even think about sitting down to fully review a complex 1k line PR unless I have a couple hours of time available for it.  If many changes are needed, that time investment multiplies on each round of review (code review is quadratic).  

I think it's generally more efficient to split any unrelated changes into their own small PRs.  In fact, I would merge the unrelated README changes immediately if they were in a separate PR.  But since they're bundled in here it takes much longer for those changes to get in.

smitebot/src/commands/start.rs
```diff
@@ -216,139 +289,235 @@ fn ensure_seed_dir(config: &CampaignConfig) -> std::io::Result<PathBuf> {
...
-                        runner.pid
-                    );
-                    return false;
-                }
-            }
+        // AFL_DISABLE_TRIM on ~60% of secondary runners.
+        let trim_count = (secondary_count * 6).div_ceil(10);
+        if ((runner_id - 1) as usize) < trim_count {
+            envs.push(("AFL_DISABLE_TRIM".to_string(), "1".to_string()));
         }
```
morehouse, line 431: I think we want to shuffle the flag assignments.  Currently the first 60% of secondaries disable trim and the first 70% set `-a binary`.  That means there are very few runners that have only one of those options and not the other -- pretty much all runners have both or neither.

Something like this might work well:
- create a list of option "modifiers" that are assigned some probability of being applied between 0 and 100.
- for each runner, for each modifier, compute `X = hash(runner_id || modifier_name) % 100` and apply the modifier to that runner if `X < modifier_probability`.

This would still be deterministic and reproducible while giving more variety to the flag combinations.

morehouse (approved):
This is a nice improvement.  I think we still want to tweak this a bit, but let's do it in small follow-up PRs instead of review-looping on this big PR.

PR #140, opened 2026-07-07
smite: add transaction block position lookup to BitcoinCli

devvaansh:
Adds `BitcoinCli::get_transaction_block_position`, which returns the `(block_height, tx_index)` pair for a confirmed transaction by issuing `getrawtransaction` followed by `getblock`. Returns `None` for unconfirmed or unknown transactions.

This is the missing piece needed to derive a BOLT 7 `short_channel_id` from a mined 2-of-2 P2WSH funding output, and is a prerequisite for the `LookupShortChannelId` IR operation that will let `ChannelAnnouncementGenerator` pass on-chain UTXO validation in CLN and LND.

Ref: #71

### DIFF
--- a/smite/src/bitcoin.rs
+++ b/smite/src/bitcoin.rs
@@ -39,6 +39,27 @@ impl PartialOrd for Utxo {
     }
 }
 
+/// The confirmed position of a transaction within the block chain, used to
+/// derive a BOLT 7 `short_channel_id`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct TxBlockPosition {
+    /// The height of the block containing the transaction.
+    pub block_height: u32,
+    /// The transaction's index within its block.
+    pub tx_index: u32,
+}
+
+/// Parsed response from `getrawtransaction <txid> 1`.
+#[derive(Deserialize)]
+struct RawTransactionInfo {
+    /// Omitted while the transaction is unconfirmed (in the mempool), so
+    /// defaults to zero.
+    #[serde(default)]
+    confirmations: u32,
+    /// Omitted while the transaction is unconfirmed (in the mempool).
+    blockhash: Option<String>,
+}
+
 /// Connection info for invoking `bitcoin-cli` against the regtest `bitcoind`
 /// started by a target.
 #[derive(Debug, Clone)]
@@ -285,6 +306,33 @@ impl BitcoinCli {
         );
     }
 
+    /// Calls `getrawtransaction <txid> 1` and returns the parsed response, or
+    /// `None` if the transaction is unknown to the node (non-zero exit code).
+    ///
+    /// # Panics
+    ///
+    /// - If the command fails to execute.
+    /// - If the command succeeds but its output is not valid JSON.
+    fn get_raw_transaction_info(&self, txid: Txid) -> Option<RawTransactionInfo> {
+        let tx_out = self
+            .run()
+            .arg("getrawtransaction")
+            .arg(txid.to_string())
+            .arg("1")
+            .output()
+            .expect("bitcoin-cli getrawtransaction should not fail");
+
+        // A non-zero exit means the transaction is unknown to the node.
+        if !tx_out.status.success() {
+            return None;
+        }
+
+        let tx_info: RawTransactionInfo = serde_json::from_slice(&tx_out.stdout)
+            .expect("getrawtransaction should return valid JSON");
+
+        Some(tx_info)
+    }
+
     /// Returns the number of confirmations for the transaction with the given
     /// txid, or `0` if it is unconfirmed (in the mempool) or unknown to the node
     /// (e.g. not broadcast yet).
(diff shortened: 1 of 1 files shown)


smite/src/bitcoin.rs
```diff
@@ -264,4 +274,86 @@ impl BitcoinCli {
...
+    /// - If either command succeeds but its output is not valid JSON.
+    /// - If `getblock` returns a block whose transaction list does not contain
+    ///   the queried txid (would indicate an inconsistent bitcoind state).
+    #[must_use]
+    pub fn get_transaction_block_position(&self, txid: Txid) -> Option<TxBlockPosition> {
+        #[derive(Deserialize)]
+        struct GetRawTransactionResponse {
+            // Omitted by `getrawtransaction` while the transaction is unconfirmed
+            // (in the mempool).
+            #[serde(default)]
```
NishantBansal2003: nit: not needed, since the field type is `Option`

NishantBansal2003 (commented):
ACK

I assume that, with this, we’re only left with operation like `ComputeShortChannelId`, which resolve a funding tx, retrieve its block height and tx index, and compute the SCID from them? In that case, we also need to handle the case where the funding tx hasn't been confirmed yet

devvaansh:
>ACK

>I assume that, with this, we’re only left with operation like ComputeShortChannelId, which resolve a funding tx, retrieve its block height and tx index, and compute the SCID from them? In that case, we also need to handle the case where the funding tx hasn't been confirmed yet

Yes, next PR adds `LookupShortChannelId` (input: `FundingTransaction` → output: `ShortChannelId`). The generator always emits `BroadcastTransaction` → `MineBlocks(6)` → `LookupShortChannelId` in sequence, so the tx is confirmed by the time the op executes. If a mutator deletes MineBlocks, `get_transaction_block_position` returns `None` and the executor panics, same behaviour as any other broken program (type mismatch, bad variable ref, etc.). No special handling needed.

NishantBansal2003:
> Yes, next PR adds `LookupShortChannelId` (input: `FundingTransaction` → output: `ShortChannelId`). The generator always emits `BroadcastTransaction` → `MineBlocks(6)` → `LookupShortChannelId` in sequence, so the tx is confirmed by the time the op executes. If a mutator deletes MineBlocks, `get_transaction_block_position` returns `None` and the executor panics, same behaviour as any other broken program (type mismatch, bad variable ref, etc.). No special handling needed.

I don’t think we should make the generator mine a specific number of blocks. I think we should keep it as a range, like `rng.random_range(1..=16)`, in the generator as well

If the tx is not confirmed, that's still a valid program, so it shouldn't panic. smite should only panic when the program (generated by smite) is invalid, in which case it’s a bug that should be fixed in `smite-ir` itself

smite/src/bitcoin.rs
```diff
@@ -264,4 +274,85 @@ impl BitcoinCli {
...
+    /// - If `getblock` returns a block whose transaction list does not contain
+    ///   the queried txid (would indicate an inconsistent bitcoind state).
+    #[must_use]
+    pub fn get_transaction_block_position(&self, txid: Txid) -> Option<TxBlockPosition> {
+        #[derive(Deserialize)]
+        struct GetRawTransactionResponse {
+            // Omitted by `getrawtransaction` while the transaction is unconfirmed
+            // (in the mempool).
+            blockhash: Option<String>,
+        }
```
morehouse: We now have two functions that call `getrawtransaction`.  It would seem natural to have a private helper that returns the `GetRawTransactionResponse`.

smite/src/bitcoin.rs
```diff
@@ -264,4 +274,85 @@ impl BitcoinCli {
...
+        // No `blockhash` means the transaction is in the mempool but not yet
+        // confirmed.
+        let blockhash = tx_info.blockhash?;
+
+        let block_out = self
+            .run()
+            .arg("getblock")
+            .arg(&blockhash)
+            .output()
+            .expect("bitcoin-cli getblock should not fail");
```
morehouse, line 383: It might be good to explicitly set `verbosity=1` so that this doesn't break if the default verbosity ever changes.

morehouse (approved):
Please squash to a single commit.

devvaansh:
>I don’t think we should make the generator mine a specific number of blocks. I think we should keep it as a range, like rng.random_range(1..=16), in the generator as well If the tx is not confirmed, that's still a valid program, so it shouldn't panic. smite should only panic when the program (generated by smite) is invalid, in which case it’s a bug that should be fixed in smite-ir itself

 agreed, rng.gen_range(1..=16) is the right call for MineBlocks in the generator. On the unconfirmed case: fair point, None shouldn't cause a panic in the executor — that's a valid program state if a mutator drops MineBlocks. I'll handle it properly in the LookupShortChannelId PR by returning an error/propagating None up instead of unwrapping.

smite/src/bitcoin.rs
devvaansh: Now done

morehouse:
Comments still haven't been addressed.

devvaansh:
>Comments still haven't been addressed.

Sorry forgot to push, now done

morehouse (approved):
Thanks for keeping the merge separate -- made it easy to review the changes.

Please rebase now and drop the merge commit.

devvaansh:
>Thanks for keeping the merge separate -- made it easy to review the changes.
Please rebase now and drop the merge commit.

now rebased cleanly onto upstream/master

PR #141, opened 2026-07-07
smite-ir: implement funding flow generator

NishantBansal2003:
Adds `funding_created` and `channel_ready` generators.
Depends-on: https://github.com/morehouse/smite/pull/137

Verified that all targets correctly send `funding_signed` and `channel_ready` once the commitment transaction is signed and the funding transaction confirms. For CLN, we need to reduce the blockchain polling interval to receive `channel_ready` promptly. We should also bump the CLN version, since I observed crashes that have been fixed in its latest release.

All other targets were able to reach `channel_ready` without any issues, but for CLN I had to (mentioning this so we can discuss a solution and add a fix in smite):
- patch the CLN polling interval from 30 seconds to 2 seconds (I arbitrarily chose the same value as LDK)
- patch `recv_non_ping` to also reject gossip messages, since CLN was apparently sending `channel_update` before sending `channel_ready`

[cln-coverage.tar.gz](https://github.com/user-attachments/files/29754676/cln-coverage.tar.gz)
[eclair-coverage.tar.gz](https://github.com/user-attachments/files/29754698/eclair-coverage.tar.gz)
[ldk-coverage.tar.gz](https://github.com/user-attachments/files/29754703/ldk-coverage.tar.gz)
[lnd-coverage.tar.gz](https://github.com/user-attachments/files/29754719/lnd-coverage.tar.gz)

### DIFF
--- a/smite-ir/src/generators/channel_ready.rs
+++ b/smite-ir/src/generators/channel_ready.rs
@@ -1,0 +1,38 @@
+//! Generator for `channel_ready` message flow.
+
+use rand::{Rng, RngExt};
+
+use super::Generator;
+use crate::builder::ProgramBuilder;
+use crate::{Operation, VariableType};
+
+/// Generates a `channel_ready` -> `channel_ready` flow.
+///
+/// Emits instructions to:
+/// 1. Mine blocks to confirm the funding transaction.
+/// 2. Build and send `channel_ready`
+/// 3. Receive and parse counterparty's `channel_ready`
+#[derive(Clone, Copy)]
+pub struct ChannelReadyGenerator;
+
+impl Generator for ChannelReadyGenerator {
+    fn generate(&self, builder: &mut ProgramBuilder, rng: &mut impl Rng) {
+        // Mine blocks to confirm the funding transaction.
+        builder.append(Operation::MineBlocks(rng.random_range(1..=16)), &[]);
+
+        // Channel parameters.
+        let channel_id = builder.pick_variable(VariableType::ChannelId, rng);
+        let second_per_commitment_point = builder.generate_fresh(VariableType::Point, rng);
+        let short_channel_id = builder.generate_fresh(VariableType::ShortChannelId, rng);
+        let include_alias = rng.random();
+
+        // Build and send channel_ready.
+        builder.append(
+            Operation::SendChannelReady { include_alias },
+            &[channel_id, second_per_commitment_point, short_channel_id],
+        );
+
+        // Receive channel_ready.
+        builder.append(Operation::RecvChannelReady, &[]);
+    }
+}
--- a/smite-ir/src/generators/funding_flow.rs
+++ b/smite-ir/src/generators/funding_flow.rs
@@ -1,0 +1,129 @@
+//! Generator for the complete v1 outbound channel funding flow.
+
+use rand::seq::IndexedRandom;
+use rand::{Rng, RngExt};
+
+use super::Generator;
+use crate::builder::ProgramBuilder;
+use crate::operation::AcceptChannelField;
+use crate::operation::{ChannelTypeVariant, ShutdownScriptVariant};
+use crate::{Operation, VariableType};
+
+/// Generates the complete v1 outbound channel funding flow.
+///
+/// Emits instructions to:
+/// 1. Build and send `open_channel`, then receive `accept_channel`
+/// 2. Build and send `funding_created`, then receive `funding_signed`
+/// 3. Broadcast and mine blocks to confirm the funding transaction
+/// 4. Complete the `channel_ready` exchange
+#[derive(Clone, Copy)]
+pub struct FundingFlowGenerator;
+
+impl Generator for FundingFlowGenerator {
+    fn generate(&self, builder: &mut ProgramBuilder, rng: &mut impl Rng) {
+        // Private/Public keys are generated fresh to ensure they're distinct.
+        let funding_privkey = builder.generate_fresh(VariableType::PrivateKey, rng);
+        let funding_pubkey = builder.append(Operation::DerivePoint, &[funding_privkey]);
+        let revocation_basepoint = builder.generate_fresh(VariableType::Point, rng);
+        let payment_basepoint = builder.generate_fresh(VariableType::Point, rng);
+        let delayed_payment_basepoint = builder.generate_fresh(VariableType::Point, rng);
+        let htlc_basepoint = builder.generate_fresh(VariableType::Point, rng);
+        let first_per_commitment_point = builder.generate_fresh(VariableType::Point, rng);
+
+        // Channel parameters.
+        let chain_hash = builder.pick_variable(VariableType::ChainHash, rng);
+        let temporary_channel_id = builder.pick_variable(VariableType::ChannelId, rng);
+        let funding_satoshis = builder.pick_variable(VariableType::Amount, rng);
+        let push_msat = builder.pick_variable(VariableType::Amount, rng);
+        let dust_limit_satoshis = builder.pick_variable(VariableType::Amount, rng);
+        let max_htlc_value_in_flight_msat = builder.pick_variable(VariableType::Amount, rng);
+        let channel_reserve_satoshis = builder.pick_variable(VariableType::Amount, rng);
+        let htlc_minimum_msat = builder.pick_variable(VariableType::Amount, rng);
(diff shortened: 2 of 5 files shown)


smite-ir/src/generators/channel_ready.rs
```diff
@@ -0,0 +1,38 @@
...
+
+impl Generator for ChannelReadyGenerator {
+    fn generate(&self, builder: &mut ProgramBuilder, rng: &mut impl Rng) {
+        // Mine blocks to confirm the funding transaction.
+        builder.append(Operation::MineBlocks(rng.random_range(1..=16)), &[]);
+
+        // Channel parameters.
+        let channel_id = builder.pick_variable(VariableType::ChannelId, rng);
+        let second_per_commitment_point = builder.generate_fresh(VariableType::Point, rng);
+        let short_channel_id = builder.generate_fresh(VariableType::ShortChannelId, rng);
```
morehouse, line 26: Why are `second_per_commitment_point` and `short_channel_id` generated fresh instead of doing `pick_variable`?

smite-ir/src/tests.rs
```diff
@@ -1128,6 +1130,94 @@ fn generated_open_channel_program_structure() {
...
+    // At least 1 DerivePoint instructions.
+    let derive_count = program
+        .instructions
+        .iter()
+        .filter(|i| matches!(i.operation, Operation::DerivePoint))
+        .count();
+    assert!(
+        derive_count >= 1,
+        "expected at least one DerivePoint, got {derive_count}"
+    );
```
morehouse, line 1219: This check seems odd.  I think it's true, there will always be at least one `DerivePoint`, but is it really a fundamental part of the structure that we want to verify?

morehouse (commented):
Wow, I'm impressed at the coverage achieved already with disjoint generators for each message exchange.  Do you still think it would we be worth creating a combined generator that threads all the correct variables through the whole flow?

> We should also bump the CLN version, since I observed crashes that have been fixed in its latest release.

I have a PR to bump versions for CLN and LND.  Once I've finished testing against LND I'll upload the PR.

> patch the CLN polling interval from 30 seconds to 2 seconds (I arbitrarily chose the same value as LDK)

Yes, let's do this.  If you've already verified that 2 seconds works, let's do that right now and tune it down after further experiments.

> patch recv_non_ping to also reject gossip messages, since CLN was apparently sending channel_update before sending channel_ready

Yes I think in general we want to skip gossip when receiving messages, since peers are free to send gossip at any time.  We can revisit the strategy here if/when we actually want to extract fields from the gossip.

smite-ir/src/generators/channel_ready.rs
NishantBansal2003: Because in the normal flow, both of these values are new/unique, rather than being picked from an existing variable that’s already shared with the peer

I’m not sure I understand why these values should already exist instead of being freshly generated?

smite-ir/src/tests.rs
NishantBansal2003: I think this question is coming from https://github.com/morehouse/smite/pull/141#discussion_r3561519830, but IIUC, atleast `second_per_commitment_point` should be freshly generated/unique in the normal flow. So yes, we should have at least one `DerivePoint`, as that’s how the generator is structured

NishantBansal2003:
> Do you still think it would we be worth creating a combined generator that threads all the correct variables through the whole flow?

We could thread the three generators to get the funding flow generator, which might provide a some benefit in reaching the full flow more easily now. However, in the future, when we have multiple generators, I think a full-flow generator will be necessary, since the probability of all the generators being threaded together in the correct flow will decrease

morehouse:
I think we definitely still need a full funding flow generator.

After 400+ CPU hours of fuzzing with these generators against LND, I wasn't able to get a single input that exercised the full funding flow.  Only 40 inputs made it through all the `open_channel` checks, and only one input successfully strung together an `open_channel -> accept_channel -> funding_created` flow but then failed LND's signature check on `funding_created`.

I'm not sure why my LND coverage is so much worse than what you achieved.  Did you start with a seed corpus or run the campaign much longer than I did?

erickcestari:
> I think we definitely still need a full funding flow generator.
> 
> After 400+ CPU hours of fuzzing with these generators against LND, I wasn't able to get a single input that exercised the full funding flow. Only 40 inputs made it through all the `open_channel` checks, and only one input successfully strung together an `open_channel -> accept_channel -> funding_created` flow but then failed LND's signature check on `funding_created`.
> 
> I'm not sure why my LND coverage is so much worse than what you achieved. Did you start with a seed corpus or run the campaign much longer than I did?

While running the LND target for five minutes with logging enabled (`RUST_LOG=debug` and `SMITE_NYX_LOG=1`), I noticed that most operations requiring an `Amount(u64)` use values exceeding Bitcoin's maximum supply of 21 million BTC. For example, out of 12,450 `SendFundingCreated` messages, 12,335 (99%) failed because the funding amount was greater than 21 million BTC, while only 120 `SendFundingCreated` operations succeeded. Since `open_channel` contains two `Amount` fields, this issue is likely even more pronounced for that message type.

We might want to improve `generate_fresh` and `tweak_u64` so they bias toward valid, consensus-compliant values.

morehouse:
Same issue for LDK.  400+ CPU hours and no valid funding flows.

> While running the LND target for five minutes with logging enabled (`RUST_LOG=debug` and `SMITE_NYX_LOG=1`), I noticed that most operations requiring an `Amount(u64)` use values exceeding Bitcoin's maximum supply of 21 million BTC. For example, out of 12,450 `SendFundingCreated` messages, 12,335 (99%) failed because the funding amount was greater than 21 million BTC, while only 120 `SendFundingCreated` operations succeeded. Since `open_channel` contains two `Amount` fields, this issue is likely even more pronounced for that message type.

Yes the amount fields are definitely bottlenecks we could fix.  I think a full funding flow generator would also be very helpful.

> We might want to improve `generate_fresh` and `tweak_u64` so they bias toward valid, consensus-compliant values.

I think it would make sense to have the generator always choose amounts below 21M or even below the wumbo limit.  Then we could leave `tweak_u64` as is to exercise the invalid cases.

Also, wumbo is the default for CLN and Eclair -- do we want to enable it in LDK and LND?  On one hand it's nice to fuzz the default values, on the other hand wumbo makes it easier to get our amounts accepted.

NishantBansal2003:
> I think we definitely still need a full funding flow generator.
> 
> After 400+ CPU hours of fuzzing with these generators against LND, I wasn't able to get a single input that exercised the full funding flow. Only 40 inputs made it through all the `open_channel` checks, and only one input successfully strung together an `open_channel -> accept_channel -> funding_created` flow but then failed LND's signature check on `funding_created`.
> 
> I'm not sure why my LND coverage is so much worse than what you achieved. Did you start with a seed corpus or run the campaign much longer than I did?

I ran the campaign continuously for nearly a week across 10 runners, first for CLN and Eclair. I then used the resulting corpus to seed LDK and LND and ran those campaigns for another week. That likely explains why LDK and LND were able to complete the funding flow much faster. CLN and Eclair were able to discover the entire funding flow on their own, so I’m not entirely sure what made LDK and LND slower.

With the funding flow generator in place now, all targets should be able to reach the complete funding flow much more easily.

smite-ir/src/generators/funding_flow.rs
```diff
@@ -0,0 +1,20 @@
...
+/// [`ChannelReadyGenerator`] in sequence.
+#[derive(Clone, Copy)]
+pub struct FundingFlowGenerator;
+
+impl Generator for FundingFlowGenerator {
+    fn generate(&self, builder: &mut ProgramBuilder, rng: &mut impl Rng) {
+        OpenChannelGenerator.generate(builder, rng);
+        FundingCreatedGenerator.generate(builder, rng);
+        ChannelReadyGenerator.generate(builder, rng);
+    }
```
morehouse, line 128: I think it would be much better to actually thread the correct variables through in this generator.  e.g., extract the funding pubkey from `accept_channel` and use it when creating the funding transaction.

NishantBansal2003: Great, I completely missed that point. Fixed in: https://github.com/morehouse/smite/pull/141/commits/ffa44e46c302cecc45fc08ffb1307fad4d3af282

morehouse (approved):
LGTM

PR #142, opened 2026-07-08
smite: extract Violation enum for target-bug findings

erickcestari:
Carve target-misbehavior findings out of `ExecuteError` into a dedicated `Violation` enum, and replace raw finding strings in scenarios with its variants. Centralizes the finding vocabulary so new oracles route through a single match arm and each message lives in one place.

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -18,6 +18,7 @@ use smite::channel_tx::{
 };
 use smite::noise::{ConnectionError, NoiseConnection};
 use smite::pending_channel::PendingChannel;
+use smite::violation::Violation;
 use smite_ir::operation::AcceptChannelField;
 use smite_ir::{Operation, Program, Variable};
 use std::collections::HashMap;
@@ -136,30 +137,10 @@ pub enum ExecuteError {
     #[error("commitment: {0}")]
     Commitment(#[from] smite::channel_tx::CommitmentError),
 
-    /// Received a message referencing a channel id we have no tracked state
-    /// for. This covers:
-    /// - a `funding_signed` for a `channel_id` we never opened, or
-    /// - an `accept_channel` for a `temporary_channel_id` we never sent
-    ///   `open_channel` for.
-    #[error("unknown channel: no tracked state for channel id {0:?}")]
-    UnknownChannel(ChannelId),
-
-    /// Received a second `accept_channel` for a `temporary_channel_id` whose
-    /// in-progress negotiation already has one, i.e. the id was reused before
-    /// its negotiation reached `funding_created`.
-    #[error(
-        "temporary_channel_id reuse: previous negotiation for {0:?} has not yet reached funding_created"
-    )]
-    TempChannelIdReuse(ChannelId),
-
-    /// The opener cannot afford the feerate for the commitment transaction.
-    #[error("opener cannot afford commitment fee for channel_id {0:?}")]
-    OpenerCannotAffordFee(ChannelId),
-
-    /// The counterparty's signature in `funding_signed` failed to verify
-    /// against the holder's first commitment transaction.
-    #[error("invalid counterparty signature for channel_id {0:?}")]
-    InvalidCounterpartySignature(ChannelId),
+    /// The target broke a protocol invariant. Surfaced to the scenario as a
+    /// failure; see [`Violation`] for the full catalog of target-bug findings.
+    #[error(transparent)]
+    Violation(#[from] Violation),
 }
 
 /// Executes IR programs against a target over an established connection.
@@ -209,9 +190,9 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
     /// - the target sends an unexpected message type
     /// - wallet funds are insufficient to perform a channel operation
     /// - the initial commitment transaction cannot be constructed
-    /// - no channel state exists for a received `funding_signed`
-    /// - the opener cannot afford the commitment feerate
-    /// - the counterparty's signature fails verification
+    /// - the target commits a [`Violation`] (unknown channel, temporary
+    ///   channel id reuse, opener cannot afford the commitment feerate, or
+    ///   invalid counterparty signature)
     ///
     /// # Panics
     ///
@@ -1072,8 +1053,8 @@ fn recv_funding_signed(conn: &mut impl Connection) -> Result<FundingSigned, Exec
 /// # Errors
 ///
 /// Returns [`ExecuteError::UnexpectedMessage`] if the received message is not a
-/// `channel_ready`, or [`ExecuteError::UnknownChannel`] if no channel state
-/// exists for the message's `channel_id`.
+/// `channel_ready`, or [`Violation::UnknownChannel`] if no channel state exists
+/// for the message's `channel_id`.
 fn recv_channel_ready(
     conn: &mut impl Connection,
     channel_states: &mut HashMap<ChannelId, ChannelState>,
@@ -1090,7 +1071,7 @@ fn recv_channel_ready(
 
     let state = channel_states
         .get_mut(&cr.channel_id)
-        .ok_or(ExecuteError::UnknownChannel(cr.channel_id))?;
+        .ok_or(Violation::UnknownChannel(cr.channel_id))?;
     *state.next_counterparty_per_commitment_point_mut() = Some(cr.second_per_commitment_point);
 
     Ok(())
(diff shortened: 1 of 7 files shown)


smite-scenarios/src/executor.rs
```diff
@@ -136,30 +137,10 @@ pub enum ExecuteError {
...
-    OpenerCannotAffordFee(ChannelId),
-
-    /// The counterparty's signature in `funding_signed` failed to verify
-    /// against the holder's first commitment transaction.
-    #[error("invalid counterparty signature for channel_id {0:?}")]
-    InvalidCounterpartySignature(ChannelId),
+    /// The target broke a protocol invariant. Surfaced to the scenario as a
+    /// failure; see [`Violation`] for the full catalog of target-bug findings.
+    #[error(transparent)]
+    Violation(#[from] Violation),
```
ekzyis, line 143: nit: I think `Violation` as a name is too generic. It doesn't say what was violated. Could rename to `ProtocolViolation`? I assume this is only meant for protocol violations, as per the documentation and current usage?

smite/src/violation.rs
```diff
@@ -0,0 +1,48 @@
+//! Target-misbehavior findings: the fuzzer's oracles.
+//!
+//! A [`Violation`] is a way a run is judged a target bug.
```
ekzyis: nit:


```suggestion
//! A [`Violation`] is a way a run is judged to be a target bug.
```

smite/src/violation.rs
```diff
@@ -0,0 +1,48 @@
...
+/// A detected misbehavior of the target under test.
+#[derive(Debug, thiserror::Error)]
+pub enum Violation {
+    /// The target process died during or after processing the input.
+    #[error("target crashed")]
+    Crashed,
+
+    /// The target stopped responding to the post-input ping-pong sync.
+    #[error("target hung (ping timeout)")]
+    Hung,
```
ekzyis, line 22: Regarding my suggestion to rename to `ProtocolViolation`: these two aren't as clearly _protocol_ violations as the others

smite/src/violation.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// whose in-progress negotiation already has one, i.e. the id was reused
+    /// before its negotiation reached `funding_created`.
+    #[error(
+        "temporary_channel_id reuse: previous negotiation for {0:?} has not yet reached funding_created"
+    )]
+    TempChannelIdReuse(ChannelId),
+
+    /// The target sent `funding_signed` even though the opener cannot afford the
+    /// commitment feerate.
+    #[error("opener cannot afford commitment fee for channel_id {0:?}")]
```
ekzyis, line 48: nit: `UnknownChannel` uses "channel id", but here and for `InvalidCounterPartySignature`, we use "channel_id". Could use only `channel` as before.

smite-scenarios/src/scenarios/noise.rs
```diff
@@ -313,15 +314,15 @@ impl<T: Target> Scenario for NoiseScenario<T> {
             log::debug!("[{:?}] Syncing via ping-pong", start.elapsed());
             if let Err(e) = ping_pong(&mut self.sync_conn) {
                 log::debug!("[{:?}] Sync ping-pong failed: {e}", start.elapsed());
-                return ScenarioResult::Fail("target unresponsive".into());
+                return ScenarioResult::Fail(Violation::Hung.to_string());
```
ekzyis: nit: We don't check `e.is_timeout()` here like in smite-scenarios/src/scenarios/ir.rs:105, which is why I think we used slightly different error messages here before.

However, we log the error above, so I think this is fine.

smite/src/violation.rs
```diff
@@ -0,0 +1,48 @@
...
+    #[error("target crashed")]
+    Crashed,
+
+    /// The target stopped responding to the post-input ping-pong sync.
+    #[error("target hung (ping timeout)")]
+    Hung,
+
+    /// The target referenced a channel for which no state was ever established.
+    /// This covers:
+    /// - a `funding_signed` or `channel_ready` for a `channel_id` we never
```
ekzyis, line 31: nice catch including `channel_ready` here; it wasn't there before

smite-scenarios/src/executor.rs
```diff
@@ -209,9 +190,8 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
     /// - the target sends an unexpected message type
     /// - wallet funds are insufficient to perform a channel operation
     /// - the initial commitment transaction cannot be constructed
-    /// - no channel state exists for a received `funding_signed`
-    /// - the opener cannot afford the commitment feerate
-    /// - the counterparty's signature fails verification
+    /// - the target commits a [`Violation`] (unknown channel, opener cannot
+    ///   afford the commitment feerate, or invalid counterparty signature)
```
ekzyis: nit: This is almost complete, so I think it's meant to be complete: could mention temporary channel id reuse, too.

smite-scenarios/src/executor.rs
```diff
@@ -1118,29 +1098,29 @@ fn is_channel_ready_expected(
...
-/// given `channel_id`, [`ExecuteError::OpenerCannotAffordFee`] if the opener
-/// cannot afford the commitment feerate, or [`ExecuteError::InvalidCounterpartySignature`]
+/// Returns [`Violation::UnknownChannel`] if no channel state exists for the
+/// given `channel_id`, [`Violation::OpenerCannotAffordFee`] if the opener
+/// cannot afford the commitment feerate, or [`Violation::InvalidCounterpartySignature`]
 /// if the signature is invalid for the holder's initial commitment transaction.
 fn verify_funding_signed(
     fs: &FundingSigned,
     channel_states: &HashMap<ChannelId, ChannelState>,
 ) -> Result<(), ExecuteError> {
```
ekzyis: nit: we could narrow this to `Result<(), Violation>`, then we don't need `.into()`, and return type would match documentation (only `Violation` errors):

```diff
diff --git a/smite-scenarios/src/executor.rs b/smite-scenarios/src/executor.rs
index 7baba88..e1fcbac 100644
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -1105,7 +1105,7 @@ fn is_channel_ready_expected(
 fn verify_funding_signed(
     fs: &FundingSigned,
     channel_states: &HashMap<ChannelId, ChannelState>,
-) -> Result<(), ExecuteError> {
+) -> Result<(), Violation> {
     let state = channel_states
         .get(&fs.channel_id)
         .ok_or(Violation::UnknownChannel(fs.channel_id))?;
@@ -1113,14 +1113,14 @@ fn verify_funding_signed(
     // The opener cannot afford the fee, so the acceptor must not send
     // `funding_signed`. Receiving one is a protocol violation.
     if !state.config.can_opener_afford_feerate(&state.commitment) {
-        return Err(Violation::OpenerCannotAffordFee(fs.channel_id).into());
+        return Err(Violation::OpenerCannotAffordFee(fs.channel_id));
     }
 
     state
         .config
         .verify_counterparty_signature(&state.commitment, &state.holder, &fs.signature)
         .then_some(())
-        .ok_or(Violation::InvalidCounterpartySignature(fs.channel_id).into())
+        .ok_or(Violation::InvalidCounterpartySignature(fs.channel_id))
 }
 
 /// Records a sent `open_channel`, keyed by `temporary_channel_id`, so the
@@ -1164,14 +1164,14 @@ fn record_send_open_channel(
 fn record_recv_accept_channel(
     negotiations: &mut HashMap<ChannelId, PendingChannel>,
     accept_channel: &AcceptChannel,
-) -> Result<(), ExecuteError> {
+) -> Result<(), Violation> {
     let pending = negotiations
         .get_mut(&accept_channel.temporary_channel_id)
         .ok_or(Violation::UnknownChannel(
             accept_channel.temporary_channel_id,
         ))?;
     if pending.accept_channel.is_some() && !pending.funding_built {
-        return Err(Violation::TempChannelIdReuse(accept_channel.temporary_channel_id).into());
+        return Err(Violation::TempChannelIdReuse(accept_channel.temporary_channel_id));
     }
     pending.accept_channel = Some(accept_channel.clone());
     Ok(())
```

No changes in other files needed.

ekzyis (commented):
Concept ACK

This replaces four very similar branches in ir.rs with one

smite/src/violation.rs
erickcestari: `Violation` is intentionally generic: it's the catalog of everything the fuzzer judges to be a target bug, not just protocol violations. A crash or a hang isn't a protocol violation, but it's still a violation of what a correct target should do, so `ProtocolViolation` would be too narrow for the enum as a whole. That's also why I kept `Crashed/Hung` in it. Note that the doc on `ExecuteError::Violation` does say "protocol invariant". That's accurate there, since the executor only ever produces the protocol-shaped variants; `Crashed/Hung` are raised directly by the scenarios and never go through `ExecuteError`.

smite-scenarios/src/executor.rs
erickcestari: See my comment below

smite/src/violation.rs
erickcestari: Updated!

smite/src/violation.rs
erickcestari: Updated!

smite-scenarios/src/scenarios/noise.rs
erickcestari: True. I kept as it's for now.

smite-scenarios/src/executor.rs
erickcestari: Nice, forgot that one. Thanks!

smite-scenarios/src/executor.rs
erickcestari: Nice! Addressed!

erickcestari:
Thanks for the review @ekzyis! I've addressed most of your comments and added a comment about `Violation`. I don't have a strong opinion there, but Idk if it's valid another enum to separate the Hungs/Crashed case from the protocol violations.

smite/src/violation.rs
ekzyis: Thanks for the explanation. Makes sense to me!

ekzyis (approved):
ACK 8628e828c10e483d81269055d0c183aa3a2470bb

smite/src/violation.rs
```diff
@@ -0,0 +1,48 @@
+//! Target-misbehavior findings: the fuzzer's oracles.
+//!
+//! A [`Violation`] is a way a run is judged to be a target bug.
+//!
+//! Conditions that are *not* target bugs (transport failures, insufficient
+//! wallet funds, mutator-produced invalid commitments, undecodable harness
+//! input) are represented as ordinary errors and never as a `Violation`.
```
morehouse: I find the wording a bit inaccurate and confusing.  Also these are not oracles, they're findings.  


```suggestion
//! Target-misbehavior findings.
//!
//! Each [`Violation`] variant names a buggy target behavior (e.g., crashing,
//! hanging, breaking a protocol invariant). This is how the fuzzer reports bugs
//! in the target.
//!
//! Conditions that are *not* the target's fault should be ordinary errors and
//! never a `Violation` (e.g., transport failures, insufficient wallet funds,
//! mutator-produced invalid commitments, undecodable harness input).
```

smite-scenarios/src/scenarios/noise.rs
morehouse: Maybe we need another violation type: `UnexpectedDisconnect`.  Then if `is_timeout()` we report `Hung`, otherwise `UnexpectedDisconnect`.

smite/src/violation.rs
erickcestari: Way better. Thanks!

smite-scenarios/src/scenarios/noise.rs
erickcestari: Updated!

morehouse (approved):
LGTM

PR #144, opened 2026-07-09
smite: fix funding transaction signing and broadcast

NishantBansal2003:
ref: https://github.com/morehouse/smite/pull/137#pullrequestreview-4658385631

Now we can’t create two transactions with overlapping inputs. We return early when signing or broadcasting a transaction if it has already been broadcast and confirmed, since the wallet can’t sign a transaction whose inputs have already been spent


erickcestari (approved):
LGTM!

morehouse (approved):
LGTM

PR #145, opened 2026-07-10
smitebot: implement `smitebot stop` command

Ashish-Kumar-Dash:
## Summary

Adds `smitebot stop <campaign-id>`: tears down a running campaign and reaps  **all** of its processes, including the Nyx QEMU children that a plain `tmux kill-session` leaks.

  ## The problem

In Nyx mode each `afl-fuzz` runner spawns a QEMU child, and `tmux kill-session` only sends SIGHUP to the `afl-fuzz` panes, the QEMU children **orphan (reparented away from tmux) and keep running**.

  ## How it works

  Verified against the live process tree and AFL++ source:

  - Each `afl-fuzz` pane is its own **process-group leader**. Its QEMU child is in the **same process group** and *keeps that group id even after it orphans*. So one **group-directed kill reaps the QEMU too**.
  - In testing, an orphaned QEMU did not exit on SIGTERM, so the kill escalates **SIGTERM → SIGKILL**.
  - A killed `afl-fuzz` lingers as a **zombie** under tmux `remain-on-exit` until the session is torn down, so the "did everything die?" check ignores zombie (`Z`) states.

  `stop`:
  1. Loads `state.json` for the campaign (errors name the path + how to list campaigns).
  2. Reads the **live** runner PIDs from the tmux session (`tmux list-panes`) — *not* from `state.json`, whose PIDs can be missing (slow startup) or stale/recycled (after a reboot), killing a recycled PID's group could hit an unrelated process.
  3. `pkill -TERM -g <pgid>` each group → waits (polling, non-zombie only) → `pkill -KILL -g <pgid>` any straggler.
  4. `tmux kill-session` to remove windows and reap the zombie leaders.
  5. Records `status = "stopped"` + `stop_time` in `state.json`.

  ## state.json changes

  - New status `stopped`.  New field stop_time: Option<u64> (absent in older state files → deserializes to None, so they still load).

  **Live end-to-end test I ran** (2-runner lnd campaign):

  _Before — 2 afl-fuzz, their 2 QEMU, 1 session:_
  ```
  $ pgrep -a afl-fuzz ; pgrep -a qemu ; tmux ls
  12122 .../afl-fuzz -Y ... -M 0 ...
  12128 .../afl-fuzz -Y ... -S 1 ...
  12131 .../qemu-system-x86_64 ... worker_id=0 ... load=off
  12158 .../qemu-system-x86_64 ... worker_id=1 ... load=on
  lnd-encrypted_bytes-1783687998: 2 windows
  ```

  _Stop, then re-check — zero survivors:_
  ```
  $ smitebot stop lnd-encrypted_bytes-1783687998
  INFO  [smitebot::commands::stop] campaign lnd-encrypted_bytes-1783687998
  stopped

  $ pgrep -a afl-fuzz ; pgrep -a qemu ; tmux ls
  no server running on /tmp/tmux-1000/default

  $ grep -E 'status|stop_time'
  ~/.smitebot/runs/lnd-encrypted_bytes-1783687998/state.json
    "status": "stopped",
    "stop_time": 1783688225,
  ```
## Workflow 
Each runner group is sent SIGTERM, then SIGKILL after a grace period for anything still alive (Nyx QEMU in fast-reload mode did not exit on SIGTERM in testing). Runner PIDs are read from the live tmux session rather than `state.json`: the recorded PIDs can be stale (e.g. after a reboot) and may have been recycled by the OS, so group-killing them could hit an unrelated process. Reading from tmux also means stop works even when startup verification left the PIDs unrecorded in `state.json`.

Ref. #70

### DIFF
--- a/smitebot/README.md
+++ b/smitebot/README.md
@@ -53,6 +53,16 @@ After spawning, startup is verified by polling for `fuzzer_stats` files. Because
 
 Campaign state is saved to `~/.smitebot/runs/<campaign-id>/state.json` for use by future `stop` and `status` commands.
 
+### smitebot stop
+
+Stops a running campaign: reaps every runner's process group — afl-fuzz and its Nyx QEMU child, which shares the group — tears down the tmux session, and records the stop time in `state.json`.
+
+```bash
+smitebot stop <campaign-id>
+```
+
+`<campaign-id>` is the directory name under `~/.smitebot/runs` (printed by `smitebot start`). 
+
 ### smitebot config
 
 Validates a campaign configuration file, reports the resolved settings, and checks that referenced paths exist on disk.
--- a/smitebot/src/commands/stop.rs
+++ b/smitebot/src/commands/stop.rs
@@ -1,0 +1,270 @@
+//! Campaign teardown: stop a running campaign and reap all of its processes.
+//!
+//! Each afl-fuzz runner is a process-group leader, and its Nyx QEMU child shares
+//! that group and keeps the group id even after it orphans. So a group-directed
+//! kill (`pkill -g <pgid>`) reaps the QEMU too — which a plain `tmux kill-session`
+//! does not (it leaks QEMU). Nyx QEMU did not exit on SIGTERM in testing, so the
+//! kill escalates SIGTERM -> SIGKILL.
+
+use std::process::Command;
+use std::thread;
+use std::time::{Duration, Instant};
+
+use clap::Args;
+
+use crate::state::{CampaignState, Status};
+use crate::tmux;
+use crate::utils;
+
+/// How long to wait for a signalled group to exit before giving up on that
+/// stage. Conservative headroom for afl-fuzz to finish `AFL_FINAL_SYNC` on a
+/// large corpus and exit after SIGTERM; the poll returns as soon as every group
+/// is gone, so this full duration is only spent when something ignores the
+/// signal (SIGTERM, or SIGKILL against a process wedged in uninterruptible I/O).
+const GRACEFUL_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Poll interval while waiting for process groups to exit, and the delay before
+/// the first poll: freshly-signalled runners won't have exited any sooner.
+const POLL_INTERVAL: Duration = Duration::from_secs(1);
+
+/// Command handler for `smitebot stop`.
+pub struct StopCommand;
+
+/// CLI arguments for `smitebot stop`.
+#[derive(Debug, Args)]
+pub struct StopArgs {
+    /// Campaign ID to stop (a directory name under `~/.smitebot/runs`).
+    campaign_id: String,
+}
+
+impl StopCommand {
+    /// Stops a campaign: reaps its runner process groups, tears down the tmux
+    /// session, and records the stop in state.json.
+    pub fn execute(args: &StopArgs) -> bool {
+        let Some(runs_dir) = CampaignState::runs_dir() else {
+            log::error!("unable to determine home directory");
+            return false;
+        };
+        let state_path = runs_dir.join(&args.campaign_id).join("state.json");
+
+        let mut state = match CampaignState::load(&state_path) {
+            Ok(state) => state,
+            Err(e) => {
+                log::error!("{e}");
+                log::error!(
+                    "campaign '{}' not found; list campaigns with: ls {}",
+                    args.campaign_id,
+                    runs_dir.display()
+                );
+                return false;
+            }
+        };
+
+        if state.status == Status::Stopped {
+            log::info!("campaign {} is already stopped", state.id);
+            return true;
+        }
+
+        let clean = terminate_runners(&state);
+
+        if let Err(e) = tmux::kill_session(&state.tmux_session) {
+            // The session may already be gone (e.g. killed manually); not fatal.
+            log::debug!("kill-session '{}': {e}", state.tmux_session);
+        }
+
+        state.status = Status::Stopped;
+        state.stop_time = Some(utils::epoch_secs());
+        if let Err(e) = state.save(&state_path) {
+            log::error!(
+                "runners were reaped but recording the stop failed: {e}; \
+                 campaign {} will still show as running in state.json",
+                state.id
+            );
+            return false;
+        }
+
+        if clean {
+            log::info!("campaign {} stopped", state.id);
+        } else {
+            log::error!(
+                "campaign {} stopped, but some processes survived SIGKILL; \
+                 inspect with `pgrep -a qemu` and kill them manually",
+                state.id
+            );
+        }
+        clean
+    }
+}
+
+/// Terminates every runner process group for a campaign.
+///
+/// PIDs come only from the live tmux session, never from state.json: a recorded
+/// PID whose session is gone (e.g. after a reboot) may have been recycled by the
+/// OS, and group-killing it could hit an unrelated process. When the session is
+/// alive, its pane PIDs are the afl-fuzz leaders, and each shares its process
+/// group with its Nyx QEMU child, so the group kill reaps the QEMU too. Returns
+/// `true` if no process survived.
+fn terminate_runners(state: &CampaignState) -> bool {
+    if !tmux::session_exists(&state.tmux_session) {
+        log::warn!(
+            "no live tmux session for campaign {}; nothing to reap \
+             (if runners leaked — e.g. the session was killed manually — check `pgrep -a qemu`)",
+            state.id,
+        );
+        return true;
+    }
+
+    let pgids = match tmux::list_pane_pids(&state.tmux_session) {
+        Ok(pgids) => pgids,
+        Err(e) => {
+            log::error!(
+                "could not read runner PIDs from tmux for campaign {}: {e}; \
+                 not reaping — runners may still be alive, check `pgrep -a qemu`",
+                state.id,
+            );
+            return false;
+        }
+    };
+    if pgids.is_empty() {
+        // Session exists but has no panes left; nothing to reap.
+        return true;
+    }
+
+    kill_groups(&pgids)
+}
+
+/// Signals each process group, escalating SIGTERM -> SIGKILL, until all exit.
+fn kill_groups(pgids: &[u32]) -> bool {
+    if signal_and_wait(pgids, "TERM", GRACEFUL_TIMEOUT) {
+        return true;
+    }
+    if signal_and_wait(pgids, "KILL", GRACEFUL_TIMEOUT) {
+        return true;
+    }
+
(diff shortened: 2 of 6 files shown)


smitebot/src/state.rs
```diff
@@ -35,6 +35,9 @@ pub struct CampaignState {
     pub smite_git_hash: String,
     /// Unix timestamp (seconds since epoch) when the campaign started.
     pub start_time: u64,
+    /// Unix timestamp when the campaign was stopped; `None` while running.
+    #[serde(default)]
```
NishantBansal2003: nit: not needed, since the field type is `Option`

smitebot/src/commands/stop.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        if let Err(e) = tmux::kill_session(&state.tmux_session) {
+            // The session may already be gone (e.g. killed manually); not fatal.
+            log::debug!("kill-session '{}': {e}", state.tmux_session);
+        }
+
+        state.status = Status::Stopped;
+        state.stop_time = Some(utils::epoch_secs());
+        if let Err(e) = state.save(&state_path) {
+            log::error!("{e}");
```
NishantBansal2003: nit: can add a better error message here

smitebot/src/commands/stop.rs
```diff
@@ -0,0 +1,218 @@
...
+        signal_group(pgid, "TERM");
+    }
+
+    let deadline = Instant::now() + GRACEFUL_TIMEOUT;
+    while Instant::now() < deadline {
+        if pgids.iter().all(|&pgid| !group_alive(pgid)) {
+            return true;
+        }
+        thread::sleep(POLL_INTERVAL);
+    }
```
NishantBansal2003: We could extract this into a separate function that takes the signal and timeout, then use it here for both TERM and KILL, so we wait and poll for KILL in the same way as well

smitebot/src/commands/stop.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    #[test]
+    fn kill_groups_reaps_whole_process_group() {
+        let mut child = spawn_group();
+        let pgid = child.id();
+        assert!(kill_groups(&[pgid]));
+        assert!(!group_alive(pgid));
+        let _ = child.wait(); // reap the leader zombie left by the group kill
+    }
```
NishantBansal2003, line 252: Can add a test like: kill group escalates to SIGKILL when SIGTERM is ignored

smitebot/src/commands/stop.rs
```diff
@@ -0,0 +1,218 @@
...
+
+use crate::state::{CampaignState, Status};
+use crate::tmux;
+use crate::utils;
+
+/// How long to wait for runners to exit after SIGTERM before escalating to
+/// SIGKILL. Generous headroom for afl-fuzz to finish `AFL_FINAL_SYNC` and exit;
+/// the poll returns as soon as every group is gone, so the full wait is only hit
+/// when something is ignoring SIGTERM.
+const GRACEFUL_TIMEOUT: Duration = Duration::from_secs(10);
```
NishantBansal2003: This looks a lot less, I think we should use a slightly longer timeout to avoid any issues in worst case

smitebot/src/commands/stop.rs
```diff
@@ -0,0 +1,218 @@
...
+
+/// Poll interval while waiting for process groups to exit.
+const POLL_INTERVAL: Duration = Duration::from_millis(200);
```
NishantBansal2003: Based on: https://github.com/morehouse/smite/pull/145/changes#r3568922716, I think we should start polling after 1-2s

NishantBansal2003 (changes requested):

smitebot/src/commands/stop.rs
Ashish-Kumar-Dash: raised it to 30s

smitebot/src/tmux.rs
```diff
@@ -99,6 +99,32 @@ pub fn dead_windows(session: &str) -> io::Result<Vec<String>> {
...
+        .args(["list-panes", "-s", "-t", &target, "-F", "#{pane_pid}"])
+        .stderr(Stdio::null())
+        .output();
+    match output {
+        Ok(o) if o.status.success() => String::from_utf8_lossy(&o.stdout)
+            .lines()
+            .filter_map(|l| l.trim().parse().ok())
+            .collect(),
+        _ => Vec::new(),
+    }
```
NishantBansal2003: Cmd errors should not be ignored, also update the comment to match this change

```suggestion
pub fn list_pane_pids(session: &str) -> io::Result<Vec<u32>> {
    let target = format!("={session}");
    let output = Command::new("tmux")
        .args(["list-panes", "-s", "-t", &target, "-F", "#{pane_pid}"])
        .stderr(Stdio::null())
        .output()?;
    if !output.status.success() {
        return Err(io::Error::other(format!(
            "tmux list-panes failed for session {session}: {}",
            output.status
        )));
    }
    Ok(String::from_utf8_lossy(&output.stdout)
        .lines()
        .filter_map(|line| line.trim().parse().ok())
        .collect())
```

smitebot/README.md
```diff
@@ -53,6 +53,16 @@ After spawning, startup is verified by polling for `fuzzer_stats` files. Because
...
 
+### smitebot stop
+
+Stops a running campaign: reaps every runner's process group — afl-fuzz and its Nyx QEMU child, which shares the group — tears down the tmux session, and records the stop time in `state.json`.
+
+```bash
+smitebot stop <campaign-id>
+```
+
+`<campaign-id>` is the directory name under `~/.smitebot/runs` (printed by `smitebot start`). Each runner group is sent SIGTERM, then SIGKILL after a grace period for anything still alive (Nyx QEMU in fast-reload mode did not exit on SIGTERM in testing). Runner PIDs are read from the live tmux session, so stop works even when startup verification left them unrecorded in `state.json`.
```
NishantBansal2003: I think reading PIDs from tmux instead of states.json makes sense for worst-case scenarios, so I think we should document that as well  (if a tmux session is gone, the PIDs stored in states.json may be stale, and killing them could terminate an unrelated process)

smitebot/src/commands/stop.rs
```diff
@@ -0,0 +1,253 @@
...
+///
+/// PIDs come only from the live tmux session, never from state.json: a recorded
+/// PID whose session is gone (e.g. after a reboot) may have been recycled by the
+/// OS, and group-killing it could hit an unrelated process. When the session is
+/// alive, its pane PIDs are the afl-fuzz leaders, and each shares its process
+/// group with its Nyx QEMU child, so the group kill reaps the QEMU too. Returns
+/// `true` if no process survived.
+fn terminate_runners(state: &CampaignState) -> bool {
+    let pgids = tmux::list_pane_pids(&state.tmux_session);
+    if pgids.is_empty() {
```
NishantBansal2003:
```suggestion
    if !tmux::session_exists(&state.tmux_session) {
```

smitebot/src/commands/stop.rs
```diff
@@ -0,0 +1,253 @@
...
+        log::warn!(
+            "no live tmux session for campaign {}; nothing to reap \
+             (if runners leaked — e.g. the session was killed manually — check `pgrep -a qemu`)",
+            state.id,
+        );
+        return true;
+    }
+
```
NishantBansal2003, line 116: Then, based on the suggestions above, I think we can add the following

```suggestion
    let pgids = match tmux::list_pane_pids(&state.tmux_session) {
        Ok(pgids) => pgids,
        Err(e) => {
            log::error!(
                "could not read runner PIDs from tmux for campaign {}: {e}; \
                 not reaping -- runners may still be alive, check `pgrep -a qemu`",
                state.id,
            );
            return false;
        }
    };

    if pgids.is_empty() {
        // Session exists but has no panes left; nothing to reap.
        return true;
    }
```

smitebot/src/commands/stop.rs
```diff
@@ -0,0 +1,253 @@
...
+        if group_alive(pgid) {
+            log::error!("process group {pgid} still alive after SIGKILL");
+            clean = false;
+        }
+    }
+    clean
+}
+
+/// Sends `signal` to every still-alive group, then polls until all exit or
+/// `timeout` elapses. Returns `true` if every group is gone.
```
NishantBansal2003: nit:
```suggestion
/// Sends `signal` to every still-alive group, then polls at `POLL_INTERVAL`
/// until all exit or `timeout` elapses. Returns `true` if every group is gone.
```

smitebot/src/commands/stop.rs
```diff
@@ -0,0 +1,253 @@
...
+    if signal_and_wait(pgids, "KILL", GRACEFUL_TIMEOUT) {
+        return true;
+    }
+
+    let mut clean = true;
+    for &pgid in pgids {
+        if group_alive(pgid) {
+            log::error!("process group {pgid} still alive after SIGKILL");
+            clean = false;
+        }
```
NishantBansal2003, line 152: nit: Can add an inline comment explaining why we don’t exit early and log all pgid's

NishantBansal2003 (commented):
Tested this, and it works fine, I think any unexpected issues can be fixed as they come up, so it looks fine to me once the comments are addressed

smitebot/README.md
```diff
@@ -53,6 +53,16 @@ After spawning, startup is verified by polling for `fuzzer_stats` files. Because
...
 
+### smitebot stop
+
+Stops a running campaign: reaps every runner's process group — afl-fuzz and its Nyx QEMU child, which shares the group — tears down the tmux session, and records the stop time in `state.json`.
+
+```bash
+smitebot stop <campaign-id>
+```
+
+`<campaign-id>` is the directory name under `~/.smitebot/runs` (printed by `smitebot start`). Each runner group is sent SIGTERM, then SIGKILL after a grace period for anything still alive (Nyx QEMU in fast-reload mode did not exit on SIGTERM in testing). Runner PIDs are read from the live tmux session rather than `state.json`: the recorded PIDs can be stale (e.g. after a reboot) and may have been recycled by the OS, so group-killing them could hit an unrelated process. Reading from tmux also means stop works even when startup verification left the PIDs unrecorded in `state.json`.
```
morehouse: I think everything after the first sentence are implementation details that don't belong in the README.


```suggestion
`<campaign-id>` is the directory name under `~/.smitebot/runs` (printed by `smitebot start`).
```

morehouse (approved):

smitebot/README.md
Ashish-Kumar-Dash: Added this in the PR description, good to write the workflow somewhere for history

morehouse (approved):
LGTM

PR #146, opened 2026-07-13
smite-scenarios: skip gossip messages in recv_non_ping

NishantBansal2003:
ref: https://github.com/morehouse/smite/pull/141#pullrequestreview-4674442898

Gossip messages are not consumed by any scenario currently, so skip them like other unknown message types instead of returning them to callers. Revisit once we need to extract their fields.

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -1118,6 +1118,16 @@ fn recv_non_ping(conn: &mut impl Connection, timeout: Duration) -> Result<Messag
             Message::Unknown { msg_type, .. } => {
                 log::debug!("skipping unknown message type {msg_type}");
             }
+            // TODO: Gossip messages are not currently consumed by any scenario,
+            // so skip them for now. Revisit this once we want to extract their
+            // fields.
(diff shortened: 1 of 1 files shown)


smite-scenarios/src/executor.rs
```diff
@@ -2308,6 +2318,43 @@ mod tests {
         assert_eq!(pong.ignored.len(), 4);
     }
 
+    #[test]
+    fn execute_recv_skips_gossip() {
```
erickcestari, line 2322: nit: we are only testing if `GossipTimestampFilter` is being skipped, since this is temporary I think it's fine to not test others gossip messages.

erickcestari (approved):
ACK 6235b0e8cbe6b76c1d8c150f0be9c992fa7b761d

ekzyis (approved):
ACK 6235b0e

Only asked myself if `message.is_gossip()`  would make sense, but we only need to distinguish gossip messages in one place, and it's intended to be temporary, so I think it's not needed. Would help with testing if all gossip messages are skipped, if we want that, though.

morehouse (approved):
LGTM

PR #148, opened 2026-07-14
smite: return early when no UTXOs are available for funding

NishantBansal2003:
If no UTXOs are available, we cannot fund the transaction, so return early instead of relying on the normal funds availability check, which would otherwise be bypassed when both the funding amount and feerate are zero. This previously allowed a transaction with no inputs to be created, which in turn caused smite to panic when `bitcoin-cli signrawtransactionwithwallet` was called.

The following crash was observed while fuzzing the funding flow:
```
thread 'main' (1) panicked at smite/src/bitcoin.rs:295:9:
bitcoin-cli signrawtransactionwithwallet failed: error code: -22
error message:
TX decode failed. Make sure the tx has at least one input.
```

### DIFF
--- a/smite/src/channel_tx/funding.rs
+++ b/smite/src/channel_tx/funding.rs
@@ -64,6 +64,15 @@ pub fn build_funding_transaction(
         });
     }
 
+    // Return early if no UTXOs are available, since the funding transaction
+    // cannot be funded.
+    if utxos.is_empty() {
+        return Err(InsufficientFunds {
+            required: funding_amt,
+            available: Amount::ZERO,
+        });
+    }
+
     let funding_spk =
         build_funding_witness_script(opener_funding_pubkey, acceptor_funding_pubkey).to_p2wsh();
 
@@ -536,7 +545,30 @@ mod tests {
         )
         .unwrap_err();
         assert!(matches!(err, InsufficientFunds { .. }));
-        assert_eq!(err.required, Amount::from_sat(10_003_180));
+        assert_eq!(err.required, Amount::from_sat(10_000_000));
+        assert_eq!(err.available, Amount::from_sat(0));
+    }
+
+    /// Not from BOLT 3 test vectors.
+    /// Tests that building a funding transaction fails when the funding amount,
+    /// feerate, and available UTXOs are all zero.
+    #[test]
+    fn funding_tx_with_zero_funding_amt_feerate_and_utxos() {
+        let change_spk = ScriptBuf::from(
+            hex::decode("00143ca33c2e4446f4a305f23c80df8ad1afdcf652f9")
+                .expect("valid P2WPKH scriptpubkey hex"),
+        );
+        let err = build_funding_transaction(
+            &pubkey("023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb"),
+            &pubkey("030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c1"),
+            0,
+            0,
+            vec![],
+            change_spk,
+        )
+        .unwrap_err();
+        assert!(matches!(err, InsufficientFunds { .. }));
+        assert_eq!(err.required, Amount::from_sat(0));
         assert_eq!(err.available, Amount::from_sat(0));
     }
 


smite/src/channel_tx/funding.rs
```diff
@@ -536,7 +545,30 @@ mod tests {
...
+            &pubkey("023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb"),
+            &pubkey("030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c1"),
+            0,
+            0,
+            vec![],
+            change_spk,
+        )
+        .unwrap_err();
+        assert!(matches!(err, InsufficientFunds { .. }));
+        assert_eq!(err.required, Amount::from_sat(0));
```
ekzyis, line 571: Should we add `expected_fee_no_change` to `required`? I think an error when `required` is the same amount as `available` is confusing. Also, then:

* you don't need to change the test above
* the `is_empty()` check is closer to when we need `utxos` for the first time

```diff
diff --git a/smite/src/channel_tx/funding.rs b/smite/src/channel_tx/funding.rs
index 323e3ba..de16946 100644
--- a/smite/src/channel_tx/funding.rs
+++ b/smite/src/channel_tx/funding.rs
@@ -64,15 +64,6 @@ pub fn build_funding_transaction(
         });
     }
 
-    // Return early if no UTXOs are available, since the funding transaction
-    // cannot be funded.
-    if utxos.is_empty() {
-        return Err(InsufficientFunds {
-            required: funding_amt,
-            available: Amount::ZERO,
-        });
-    }
-
     let funding_spk =
         build_funding_witness_script(opener_funding_pubkey, acceptor_funding_pubkey).to_p2wsh();
 
@@ -86,6 +77,15 @@ pub fn build_funding_transaction(
     let mut expected_fee_no_change =
         predict_tx_fee(feerate_per_kw, &input_weights, &[funding_spk.len()]);
 
+    // Return early if no UTXOs are available, since the funding transaction
+    // cannot be funded.
+    if utxos.is_empty() {
+        return Err(InsufficientFunds {
+            required: funding_amt + expected_fee_no_change,
+            available: Amount::ZERO,
+        });
+    }
+
     for utxo in utxos {
         total += utxo.amount;
 
@@ -545,7 +545,7 @@ mod tests {
         )
         .unwrap_err();
         assert!(matches!(err, InsufficientFunds { .. }));
-        assert_eq!(err.required, Amount::from_sat(10_000_000));
+        assert_eq!(err.required, Amount::from_sat(10_003_180));
         assert_eq!(err.available, Amount::from_sat(0));
     }
 
```

ekzyis (commented):
Concept ACK

Left a question regarding error semantics

smite/src/channel_tx/funding.rs
NishantBansal2003: Yup, can do that, but my thinking was -- why calculate a value just to error out later (we already followed a similar approach above)? As for required == available, that can still happen if the feerate, funding amount, and UTXO are all 0 (as in the test I added). So I was thinking a new error type just for this case wouldn’t be necessary.

But I’m fine with adding the fee to required as well if you want to go that way. Let me know

erickcestari (approved):
ACK 4314d51103e4a531c9f19cc7c3c22e45d88035c7

I'm curious how did you get in the state of not having any utxo to spend.  Did it locked all the utxo by having many `CreateFundingTransaction` operations?

NishantBansal2003:
> I'm curious how did you get in the state of not having any utxo to spend. Did it locked all the utxo by having many `CreateFundingTransaction` operations?

So, at the start we mine 101 blocks to mature the coinbase from block 1. The 1st `CreateFundingTransaction` locks that UTXO, and the 2nd `CreateFundingTransaction` is left with an empty UTXO

smite/src/channel_tx/funding.rs
morehouse: Yeah even if we move the check to include the fees, we can still get `required = 0`.  It's a small wart that surfaces in an edge case, and we don't actually rely on the error fields for anything important -- it's more for informational/debugging purposes.

Would be nice to fix, but I don't think it's worth holding up this PR for it.

morehouse (approved):

PR #150, opened 2026-07-15
smitebot: add status command

Ashish-Kumar-Dash:
# What

Adds `smitebot status <campaign-id> [--summary]`, which reports how a campaign is progressing. It detects whether the campaign is live (its tmux session exists), then either drops into a live dashboard or prints a one-shot snapshot.

- **Live (default):** adds a self-reloading `status` window to the campaign's tmux session and attaches, so you can flip between the `runner-*` windows and the aggregated view. Re-running on the same live campaign reuses the existing `status` window instead of stacking a second.
- **Live + `--summary`:** prints one snapshot and exits, no attach.
- **Dead:** prints a static snapshot of the last-written stats.

## Sample output

```
campaign
  id         lnd-encrypted_bytes-1784060278
  target     lnd
  scenario   encrypted_bytes
  status     running
  uptime     14m 31s

runners
┌────────┬───────┬───────────┬─────────────┬─────────┬──────────┬────────┬─────────┬──────────────┐
│ runner │ state │   execs/s │       execs │  corpus │  crashes │  hangs │     cov │    last find │
├────────┼───────┼───────────┼─────────────┼─────────┼──────────┼────────┼─────────┼──────────────┤
│      0 │ alive │    328.47 │      256677 │     471 │        0 │      0 │   6.05% │      45s ago │
│      1 │ alive │    339.01 │      264909 │     382 │        0 │      0 │   6.00% │      46s ago │
├────────┼───────┼───────────┼─────────────┼─────────┼──────────┼────────┼─────────┼──────────────┤
│ totals │       │    667.48 │      521586 │         │        0 │      0 │         │              │
└────────┴───────┴───────────┴─────────────┴─────────┴──────────┴────────┴─────────┴──────────────┘
```

The totals row sums only the additive columns (execs/s, execs, crashes, hangs); corpus, cov, and last find are left blank because they don't sum meaningfully (a synced corpus overlaps across runners; coverage is a percentage; last find is a timestamp). After `stop`, the header reads `stopped` and each runner reads `dead`.

## Design notes
  - **Own `fuzzer_stats` parser.** One `render_summary` produces the text for all three paths (dead dump, `--summary`, and the live window, which just re-runs `status --summary`). `afl-whatsup` was not wrapped because its output isn't stably machine-readable across versions and would fork the code paths.
  - **Per-runner `state`** comes from tmux's `pane_dead` (via `dead_windows`), a dead `remain-on-exit` pane can report a stale pid. One runner can die while the campaign runs on.
  - **Status reconciliation.** The persisted `state.status` drives the reported status; tmux liveness only refines the `running` case, since a session outlives its runners  (`remain-on-exit`). So a `running` campaign whose runners have all exited reads `dead`, not `running`. `starting` is shown throughout bring-up (the session may not exist yet, or  runners may still be calibrating — minutes under Nyx). A session-gone campaign that was never cleanly stopped reads `dead` (it crashed or the host rebooted), as opposed to  `stopped`/`failed`. Full mapping:

    | `state.status` | session | runner alive | reported |
    | :--- | :--- | :--- | :--- |
    | starting | any | any | `starting` |
    | running | yes | yes | `running` |
    | running | yes | no | `dead` |
    | running | no | — | `dead` |
    | failed | any | any | `failed` |
    | stopped | no | — | `stopped` |

  - **Read-only:** no `state.json` schema change.
  - The summary aggregates each runner's `fuzzer_stats`: a live/dead `state`, executions per second, total executions, corpus count, saved crashes and hangs, bitmap coverage, and time since the last new find, plus campaign-wide totals. A runner that has not yet written `fuzzer_stats` , AFL++ writes it only after calibrating every seed — shows `-`  placeholders in place of numbers.

  Ref. #70

### DIFF
--- a/smitebot/src/commands/status.rs
+++ b/smitebot/src/commands/status.rs
@@ -1,0 +1,665 @@
+//! Campaign status reporting: a one-shot text summary and a live tmux dashboard.
+//!
+//! `status` detects whether the campaign is still running by checking for its
+//! tmux session. A dead campaign (or `--summary`) prints a static snapshot of
+//! each runner's `fuzzer_stats`. A live campaign gets an extra tmux window that
+//! re-runs the snapshot on an interval, then attaches so the user can flip
+//! between the runner windows and the summary.
+
+use std::collections::HashMap;
+use std::fs;
+use std::io;
+use std::path::Path;
+
+use clap::Args;
+
+use crate::state::{CampaignState, RunnerState, Status};
+use crate::tmux;
+use crate::utils::{self, shell_quote};
+
+/// Name of the tmux window hosting the reloading status view.
+const STATUS_WINDOW: &str = "status";
+
+/// Reload interval (seconds) for the live status window.
+///
+/// AFL++ rewrites `fuzzer_stats` at most every `STATS_UPDATE_SEC` (60, see
+/// AFL++ `include/config.h`), so refreshing faster only redraws identical
+/// numbers.
+const RELOAD_SECS: u32 = 60;
+
+/// Command handler for `smitebot status`.
+pub struct StatusCommand;
+
+/// CLI arguments for `smitebot status`.
+#[derive(Debug, Args)]
+pub struct StatusArgs {
+    /// Campaign ID to inspect (a directory name under `~/.smitebot/runs`).
+    campaign_id: String,
+    /// Print a one-shot text summary instead of attaching to the live dashboard.
+    #[arg(long)]
+    summary: bool,
+}
+
+impl StatusCommand {
+    /// Reports the status of a campaign, either as a one-shot summary or by
+    /// attaching to its live tmux dashboard.
+    pub fn execute(args: &StatusArgs) -> bool {
+        let Some(runs_dir) = CampaignState::runs_dir() else {
+            log::error!("unable to determine home directory");
+            return false;
+        };
+        let state_path = runs_dir.join(&args.campaign_id).join("state.json");
+
+        let state = match CampaignState::load(&state_path) {
+            Ok(state) => state,
+            Err(e) => {
+                log::error!("{e}");
+                log::error!(
+                    "campaign '{}' not found; list campaigns with: ls {}",
+                    args.campaign_id,
+                    runs_dir.display()
+                );
+                return false;
+            }
+        };
+
+        // state.status is written only at start and by `stop`, so it can be
+        // stale (e.g. still "running" for a campaign that has since crashed).
+        // The live tmux session reconciles it against reality — both for the
+        // snapshot-vs-dashboard choice below and for the reported status.
+        let session_alive = tmux::session_exists(&state.tmux_session);
+
+        if !session_alive || args.summary {
+            let alive_runner_ids = match alive_runners(&state, session_alive) {
+                Ok(ids) => ids,
+                Err(e) => {
+                    // A live session we can't query is anomalous (both this and
+                    // the session check use tmux). Fail loudly rather than
+                    // silently render every runner as dead.
+                    log::error!(
+                        "could not determine per-runner liveness in session '{}': {e}",
+                        state.tmux_session
+                    );
+                    return false;
+                }
+            };
+            print!(
+                "{}",
+                render_summary(&state, session_alive, &alive_runner_ids)
+            );
+            return true;
+        }
+
+        attach_dashboard(&state)
+    }
+}
+
+/// Ensures a single reloading status window exists in the campaign session,
+/// then attaches to it.
+fn attach_dashboard(state: &CampaignState) -> bool {
+    let session = &state.tmux_session;
+
+    // Re-run guard: reuse the window if `status` was already run on this live
+    // campaign, rather than stacking a second status window.
+    match tmux::window_exists(session, STATUS_WINDOW) {
+        Ok(true) => log::info!("status window already exists in session '{session}', attaching"),
+        Ok(false) => {
+            let cmd = dashboard_command(&state.id);
+            if let Err(e) = tmux::add_window(session, STATUS_WINDOW, &cmd) {
+                log::error!("failed to create status window in session '{session}': {e}");
+                return false;
+            }
+        }
+        Err(e) => {
+            log::error!("failed to query tmux session '{session}': {e}");
+            return false;
+        }
+    }
+
+    // `attach`/`switch-client` land on whatever window is currently active. The
+    // status window is active right after we create it, but if the user switched
+    // to a runner window and detached, a later `status` run would otherwise
+    // attach to that runner window — so select the status window explicitly.
+    if let Err(e) = tmux::select_window(session, STATUS_WINDOW) {
+        log::error!("failed to select status window in session '{session}': {e}");
+        return false;
+    }
+
+    if let Err(e) = tmux::attach(session) {
+        log::error!("failed to attach to tmux session '{session}': {e}");
+        return false;
+    }
+    true
+}
+
+/// Builds the shell command for the reloading status window.
+///
+/// A bash loop re-runs `smitebot status <id> --summary` every [`RELOAD_SECS`]
+/// seconds, clearing the terminal between frames with an ANSI home+erase escape
+/// (`\033[H\033[2J`) so no `clear`/`watch` binary is required.
+fn dashboard_command(campaign_id: &str) -> String {
+    // Explicit path to this binary rather than a bare `smitebot`, so the window
+    // runs the same binary the user invoked, not whatever is on PATH.
+    let exe = std::env::current_exe()
+        .map_or_else(|_| "smitebot".to_string(), |p| p.display().to_string());
+    let exe = shell_quote(&exe);
+    let id = shell_quote(campaign_id);
+    format!(
+        "while true; do printf '\\033[H\\033[2J'; {exe} status {id} --summary; sleep {RELOAD_SECS}; done"
+    )
+}
+
+/// Returns the ids of runners whose tmux window is still alive.
+///
+/// A dead campaign (no session) has no live runners. For a live campaign, a
+/// runner is alive only if its window still exists with a running pane (via
+/// [`tmux::alive_windows`]). Keying off live panes rather than dead-but-open
+/// ones means a runner whose pane was closed entirely — its window gone from
+/// both lists — is correctly reported dead, not frozen-but-alive. A tmux query
+/// failure propagates rather than being reported as "all runners dead".
+fn alive_runners(state: &CampaignState, session_alive: bool) -> io::Result<Vec<u16>> {
+    if !session_alive {
+        return Ok(Vec::new());
+    }
+    let alive = tmux::alive_windows(&state.tmux_session)?;
+    Ok(alive_ids(&state.runners, &alive))
+}
+
+/// Returns the ids of runners whose window is among `alive_windows`.
+fn alive_ids(runners: &[RunnerState], alive_windows: &[String]) -> Vec<u16> {
+    runners
+        .iter()
+        .filter(|runner| alive_windows.contains(&tmux::runner_window_name(runner.id)))
+        .map(|runner| runner.id)
+        .collect()
+}
+
+/// Renders a one-shot text summary of the campaign.
+///
+/// `session_alive` is the tmux-session liveness signal, used to reconcile the
+/// displayed status (see [`display_status`]). `alive_runner_ids` holds the ids of
+/// runners whose window is still alive, driving the per-runner `state` column.
+fn render_summary(state: &CampaignState, session_alive: bool, alive_runner_ids: &[u16]) -> String {
+    use std::fmt::Write;
+
+    let now = utils::epoch_secs();
+    let elapsed = state
+        .stop_time
+        .unwrap_or(now)
+        .saturating_sub(state.start_time);
+
+    let status = display_status(state, session_alive, !alive_runner_ids.is_empty());
+    let mut out = render_header(state, status, &fmt_duration(elapsed));
+    let _ = writeln!(out);
+    out.push_str(&render_runner_table(state, alive_runner_ids, now));
+    out
+}
+
+/// Renders the per-runner stats table: one row per runner plus a totals row,
+/// each column right-aligned and boxed with border glyphs.
+///
+/// Column widths are derived from the actual cell contents (see
+/// [`collect_rows`]) so values never overflow their column.
+fn render_runner_table(state: &CampaignState, alive_runner_ids: &[u16], now: u64) -> String {
+    use std::fmt::Write;
+
+    let header = [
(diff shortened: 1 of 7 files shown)


smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,457 @@
...
+
+/// Reconciles the displayed campaign status against tmux liveness.
+///
+/// A campaign whose session is gone but that was never `stop`ped (state still
+/// reads `starting`/`running`) is reported as `dead`: it crashed or the host
+/// rebooted.
+fn display_status(state: &CampaignState, live: bool) -> &'static str {
+    if live {
+        return "running";
+    }
```
NishantBansal2003: So this logic is based solely on `session_exists`, which means that if the tmux session exists, it reports the campaign as running. But, the campaign session is created with `remain-on-exit` enabled, so dead runner windows keep the session alive. Means, a campaign where every runner has crashed (and was never stopped) still has an active tmux session and is incorrectly reported as running

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,457 @@
...
+        .map(|runner| runner.id)
+        .collect()
+}
+
+/// Renders a one-shot text summary of the campaign.
+///
+/// `live` is the tmux-session liveness signal, used to reconcile the displayed
+/// status (see [`display_status`]). `alive` holds the ids of runners whose
+/// window is still running, driving the per-runner `state` column.
+fn render_summary(state: &CampaignState, live: bool, alive: &HashSet<u16>) -> String {
```
NishantBansal2003: These live and alive names are a little confusing. Can we make them more explicit?

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,457 @@
...
+            .get("last_find")
+            .and_then(|v| v.parse::<u64>().ok())
+            .map_or_else(|| "-".to_string(), |t| fmt_ago(now, t));
+
+        let _ = writeln!(
+            out,
+            "{:<7} {:<6} {:>9} {:>12} {:>8} {:>8} {:>6} {:>8} {:>10}",
+            runner.id,
+            liveness,
+            field("execs_per_sec"),
```
NishantBansal2003: We can cache `execs_per_sec`, since it is fetched and parsed twice per runner

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,457 @@
...
+        "while true; do printf '\\033[H\\033[2J'; {exe} status {id} --summary; sleep {RELOAD_SECS}; done"
+    )
+}
+
+/// Returns the ids of runners whose tmux window is still alive.
+///
+/// A dead campaign (no session) has no live runners. For a live campaign, a
+/// runner is dead if its window's pane has exited — `remain-on-exit` keeps the
+/// dead window queryable via [`tmux::dead_windows`].
+fn alive_runners(state: &CampaignState, live: bool) -> HashSet<u16> {
```
NishantBansal2003: Do we really need a `HashSet` here? I think the runner count is small enough that we don’t need to optimize for lookup time, and I’d rather keep the code simpler. I also don’t think we’re using any `HashSet` specific functionality

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,457 @@
...
+    #[test]
+    fn display_status_reports_dead_when_session_gone_without_stop() {
+        let dir = tempfile::tempdir().unwrap();
+        let state = sample_state(dir.path().to_path_buf(), 1);
+        // state.status is Running but the tmux session is gone (live = false).
+        assert_eq!(display_status(&state, false), "dead");
+    }
+
+    #[test]
+    fn display_status_reports_stopped_after_clean_stop() {
```
NishantBansal2003, line 597: Can we add more tests for `display_status`? It looks like this is where some edge cases may have been missed, a few ideas:
- reports failed when session gone
- reports running when session live with alive runner
- reports when session live but all runners dead

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,457 @@
...
+
+    let now = utils::epoch_secs();
+    let elapsed = state
+        .stop_time
+        .unwrap_or(now)
+        .saturating_sub(state.start_time);
+
+    let mut out = String::new();
+    let _ = writeln!(out, "campaign: {}", state.id);
+    let _ = writeln!(out, "target:   {} / {}", state.target, state.scenario);
```
NishantBansal2003: targets are lightning implementations, and then we have scenarios. Can we separate them?

NishantBansal2003 (changes requested):
Currently, the status looks something like this:
```sh
campaign: lnd-encrypted_bytes-1784401575
target:   lnd / encrypted_bytes
status:   running
uptime:   1m 12s

runner  state    execs/s        execs   corpus  crashes  hangs      cov  last find
0       alive   12000.00           12        1        0      0    5.60%      never
1       alive   12000.00           12        1        0      0    5.60%      never
totals           24000.0                              0      0
```

Can we display this as a table (with lines) instead to reduce clutter? I think it would be much easier to read and visualize. Also, could we use section titles similar to AFL++’s status screen? I find those titles more explicit and easier to understand at a glance

smitebot/src/commands/status.rs
Ashish-Kumar-Dash: Yes its a real bug, fixed in latest commit

smitebot/src/commands/status.rs
Ashish-Kumar-Dash: switched to Vec

smitebot/src/commands/status.rs
Ashish-Kumar-Dash: fixed

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,602 @@
...
+        };
+
+        // state.status is written only at start and by `stop`, so it can be
+        // stale (e.g. still "running" for a campaign that has since crashed).
+        // The live tmux session reconciles it against reality — both for the
+        // snapshot-vs-dashboard choice below and for the reported status.
+        let session_alive = tmux::session_exists(&state.tmux_session);
+
+        if !session_alive || args.summary {
+            let running_runners = alive_runners(&state, session_alive);
```
NishantBansal2003: nit:
```suggestion
            let alive_runner_ids = alive_runners(&state, session_alive);
```

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,602 @@
...
+        return Vec::new();
+    }
+    match tmux::dead_windows(&state.tmux_session) {
+        Ok(dead) => alive_ids(&state.runners, &dead),
+        Err(e) => {
+            log::warn!(
+                "could not determine per-runner liveness in session '{}': {e}",
+                state.tmux_session
+            );
+            Vec::new()
```
NishantBansal2003: I think if this errors out, we should surface the error instead of falsely reporting the runner as dead. But, I don’t think this should fail under normal circumstances, so if it does, it’s likely due to a bug in smitebot or an external system. Should we panic in that case to make it easier to debug?

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,602 @@
...
+        .unwrap_or(now)
+        .saturating_sub(state.start_time);
+
+    let status = display_status(state, session_alive, !running_runners.is_empty());
+    let mut out = render_header(state, status, &fmt_duration(elapsed));
+    let _ = writeln!(out);
+
+    let _ = writeln!(out, "runners");
+    let _ = writeln!(out, "{}", table_rule('┌', '┬', '┐'));
+    let _ = writeln!(
```
NishantBansal2003: Can we refactor this and the code below into `render_runner_table` so it’s consistent with the code above?

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,602 @@
...
+
+/// Reload interval (seconds) for the live status window.
+///
+/// AFL++ rewrites `fuzzer_stats` at most every `STATS_UPDATE_SEC` (60, see
+/// AFL++ `include/config.h`), so refreshing faster only redraws identical
+/// numbers.
+const RELOAD_SECS: u32 = 60;
+
+/// Column widths for the runner table, indexed like the header cells.
+const COL_W: [usize; 9] = [6, 5, 9, 11, 7, 8, 6, 7, 12];
```
NishantBansal2003: I wasn’t expecting this to be implemented manually. The fixed widths could misalign if the values exceed them (unless they were chosen with the worst case in mind). Also, there’s quite a bit of logic just for formatting, whereas the previous approach was much simpler. If there’s an easier way to implement this, we can do that here, otherwise, I think it’s fine to leave it as is and open a follow-up PR, since this adds a fair amount of code

smitebot/src/commands/status.rs
NishantBansal2003: I think the same applies to `saved_crashes` and `saved_hangs` as well

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,602 @@
...
+            continue;
+        };
+        let field = |key: &str| runner_stats.get(key).map_or("-", String::as_str);
+        let execs_per_sec = field("execs_per_sec");
+        let execs_done = field("execs_done");
+
+        total_eps += execs_per_sec.parse::<f64>().unwrap_or(0.0);
+        total_execs += execs_done.parse::<u64>().unwrap_or(0);
+        total_crashes += field("saved_crashes").parse::<u64>().unwrap_or(0);
+        total_hangs += field("saved_hangs").parse::<u64>().unwrap_or(0);
```
NishantBansal2003: I think if parsing fails, we’ll end up with incorrect totals while still showing values in the runner table. I don’t think parsing should ever fail, so let’s use expect with an appropriate message to make it clear that this isn’t expected

NishantBansal2003 (changes requested):

smitebot/src/commands/status.rs
Ashish-Kumar-Dash: Agreed the old path was wrong, returning Vec::new() reported every runner dead on a tmux hiccup. Fixed, live_runners now returns io::Result, and execute logs the failure (naming the session) and returns false. That matches the command boundary here, commands return bool and log at the failure site, expect/panic is reserved for invariants.

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,653 @@
...
+                return false;
+            }
+        }
+        Err(e) => {
+            log::error!("failed to query tmux session '{session}': {e}");
+            return false;
+        }
+    }
+
+    if let Err(e) = tmux::attach(session) {
```
NishantBansal2003, line 128: This only attaches to the currently active window. On the first run, that will always be the status window. However, if the user later switches to another window, detaches, and then runs the status command, it will attach to the runner window instead of the status window

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,653 @@
...
+        // place (src/afl-fuzz-stats.c, v4.40c), so `status` never reads a partial
+        // file — a present field is a complete number. `get` is None only when the
+        // field is absent (file not written yet), which renders "-" and adds 0. A
+        // parse failure would mean a smitebot bug (wrong key/corrupt file), not
+        // runtime variability, so it asserts. One value feeds both cell and total,
+        // so they can't disagree.
+        let eps = runner_stats.get("execs_per_sec").map(|v| {
+            v.parse::<f64>()
+                .expect("execs_per_sec in fuzzer_stats is a float")
+        });
```
NishantBansal2003: This (and below) repeated pattern can be extracted into a helper and reused here, something like: `stat_num::<f64>(&runner_stats, "execs_per_sec")`

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,653 @@
...
+    alive_runner_ids: &[u16],
+    now: u64,
+) -> (Vec<[String; 9]>, [String; 9]) {
+    let mut rows: Vec<[String; 9]> = Vec::new();
+    let mut total_eps = 0.0f64;
+    let mut total_execs = 0u64;
+    let mut total_crashes = 0u64;
+    let mut total_hangs = 0u64;
+
+    let dash_u64 = |v: Option<u64>| v.map_or_else(|| "-".to_string(), |n| n.to_string());
```
NishantBansal2003: nit: This name isn't clear, Consider a better name or add a comment

NishantBansal2003 (changes requested):

smitebot/src/commands/status.rs
Ashish-Kumar-Dash: yes, fixed the edge case

NishantBansal2003 (approved):

smitebot/src/state.rs
```diff
@@ -77,6 +77,14 @@ impl RunnerState {
     }
 }
 
+/// Returns the tmux window name that hosts the runner with the given id.
+///
+/// Shared by `start` (which creates the window) and `status` (which checks
+/// whether it is still alive via that window).
+pub fn runner_window_name(id: u16) -> String {
```
morehouse: Does this belong in `tmux.rs` instead?

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,661 @@
...
+
+/// Name of the tmux window hosting the reloading status view.
+const STATUS_WINDOW: &str = "status";
+
+/// Reload interval (seconds) for the live status window.
+///
+/// AFL++ rewrites `fuzzer_stats` at most every `STATS_UPDATE_SEC` (60, see
+/// AFL++ `include/config.h`), so refreshing faster only redraws identical
+/// numbers.
+const RELOAD_SECS: u32 = 60;
```
morehouse, line 28: Ah, yeah this is a mild annoyance, especially at the start of a campaign where things are changing fast and the status isn't updating.

But I don't think we want to change AFL's `fuzzer_stats` refresh rate either (extra disk syncing is overhead).

smitebot/src/commands/status.rs
```diff
@@ -0,0 +1,661 @@
...
+}
+
+/// Returns the ids of runners whose window is not among `dead_windows`.
+fn alive_ids(runners: &[RunnerState], dead_windows: &[String]) -> Vec<u16> {
+    runners
+        .iter()
+        .filter(|runner| !dead_windows.contains(&runner_window_name(runner.id)))
+        .map(|runner| runner.id)
+        .collect()
+}
```
morehouse, line 175: While testing I found an edge case here.  If I kill a runner and then close its tmux pane, it doesn't show up in `dead_windows`, so the status table still shows it as running and alive, though the stats are frozen forever.

Ideally we would determine the live IDs from the live panes rather than the dead-but-open panes.

morehouse (approved):

morehouse (approved):
LGTM

PR #153, opened 2026-07-16
scripts: Add automated fuzzing trial orchestrator

Chand-ra:
Introduce `smite-orchestrator.py` to handle the automated execution of parallel fuzzing trials across multiple targets for rigorous A/B coverage evaluation.

The script automatically queues jobs and maps CPU cores using `taskset` for strict isolation. It manages the full lifecycle of standalone Nyx mode (`-X`) fuzzers, including the generation and cleanup of isolated sharedirs per trial.

It features a live Rich TUI dashboard to monitor real-time coverage metrics and track failed trials, and ensures graceful teardown of active processes upon completion or interrupt.

### DIFF
--- a/scripts/smite-orchestrator.py
+++ b/scripts/smite-orchestrator.py
@@ -1,0 +1,1012 @@
+#!/usr/bin/env python3
+"""
+Smite Fuzzing Campaign Orchestrator
+
+This script automates the execution of parallel fuzzing trials across multiple targets
+for rigorous A/B coverage evaluation. It queues jobs, maps cores to trials for strict
+CPU isolation via `taskset`, and launches `afl-fuzz` against Nyx sharedirs.
+
+`smitebot` is used for one-time setup steps (`smitebot build` to build Docker workload
+images, and `smitebot doctor` to validate the host). Similar to `smitebot start`, this
+script handles building `libsmite_ir_mutator.so`, preparing the Nyx sharedir, and
+launching `afl-fuzz` with the right strategy flags and environment variables directly.
+
+A live Rich TUI dashboard monitors coverage (Edges and Execs/s) of all active cores in
+real-time. Press Ctrl+C to stop scheduling new jobs and safely kill active fuzzers.
+
+Requirements:
+    pip install rich
+    smitebot (must be available in your PATH: `cargo install --path smitebot`)
+
+Generated Directory Structure:
+    The script creates an output directory populated with isolated trial runs based on
+    your defined configuration labels:
+
+    <out_dir>/
+    ├── .default-seeds/                  # Fallback '\x00' seed (if --seed-dir omitted)
+    ├── <label_a>/                       # e.g., 'baseline'
+    │   ├── <target_1>/                  # e.g., 'cln'
+    │   │   ├── trial-01/
+    │   │   │   ├── afl-fuzz.log         # Raw afl-fuzz stdout/stderr for this trial
+    │   │   │   ├── sharedir/            # Nyx sharedir (deleted on cleanup)
+    │   │   │   └── afl-out/default/     # Fuzzer output (stats, plot_data, bitmap)
+    │   │   ├── trial-02/
+    │   │   └── ...
+    │   └── <target_2>/                  # e.g., 'lnd'
+    └── <label_b>/                       # e.g., 'experimental'
+        ├── <target_1>/
+        └── <target_2>/
+
+Usage:
+    python smite-orchestrator.py \
+    --out-dir OUT_DIR \
+    --configs LABEL:SMITE_DIR[,LABEL:SMITE_DIR...] \
+    --scenario SCENARIO \
+    --targets TARGET[,TARGET...] \
+    --cores CORE[,CORE...] \
+    --afl-dir AFL_DIR \
+    [--trials N | --trial-ids ID[,ID...]] \
+    [--timeout SECONDS] \
+    [--seed-dir SEED_DIR]
+
+Examples:
+    # Standard 24-hour evaluation (4 isolated cores, 30 trials per target/config)
+    python smite-orchestrator.py \
+        --out-dir ./eval-results \
+        --configs baseline:~/smite,experimental:~/smite-new-mutator \
+        --scenario ir \
+        --targets cln,lnd,ldk,eclair \
+        --cores 0,1,2,3 \
+        --afl-dir ~/AFLplusplus
+
+    # Fast exploratory test run (1-hour timeout, 5 trials, with seed corpus)
+    python smite-orchestrator.py \
+        --out-dir ./eval-results \
+        --configs control:~/smite,test:~/smite-new-mutator \
+        --scenario ir \
+        --targets cln \
+        --cores 4,5,6,7,8 \
+        --trials 5 \
+        --timeout 3600 \
+        --afl-dir ~/AFLplusplus \
+        --seed-dir ./my_seeds
+
+    # Targeted re-run of specific failed trials (preserves all other data)
+    python smite-orchestrator.py \
+        --out-dir ./eval-results \
+        --configs baseline:~/smite \
+        --scenario ir \
+        --targets lnd \
+        --cores 0,1 \
+        --trial-ids 1,15,20 \
+        --afl-dir ~/AFLplusplus
+"""
+
+import argparse
+import collections
+import json
+import os
+import shutil
+import signal
+import subprocess
+import sys
+import threading
+import time
+from dataclasses import dataclass
+from pathlib import Path
+from queue import Empty, Queue
+from typing import Optional
+
+from rich.console import Console, Group
+from rich.live import Live
+from rich.panel import Panel
+from rich.table import Table
+
+# ────────────────────────────  CONFIGURATION  ────────────────────────────
+
+
+def testcache_size_mb() -> Optional[int]:
+    """Suggested AFL_TESTCACHE_SIZE (MB) from available RAM.
+
+    Mirrors smitebot's conservative thresholds (50/250/500 MB), since
+    machines here are shared across several concurrently-fuzzing cores.
+    """
+    try:
+        meminfo = Path("/proc/meminfo").read_text()
+    except OSError:
+        return None
+
+    for line in meminfo.splitlines():
+        if line.startswith("MemAvailable:"):
+            try:
+                free_mb = int(line.split()[1]) // 1024
+            except (IndexError, ValueError):
+                return None
+            if free_mb > 32_000:
+                return 500
+            elif free_mb > 8_000:
+                return 250
+            else:
+                return 50
+    return None
+
+
+@dataclass(frozen=True)
+class TrialConfig:
+    """Immutable configuration and derived path/command resolution for a single
+    fuzzing trial."""
+
+    core: int
+    label: str
+    target: str
+    trial_num: int
+    scenario: str
+    out_dir: Path
+    smite_dir: Path
+    afl_dir: Path
+    timeout: int
+    seed_dir: Path
+
+    @property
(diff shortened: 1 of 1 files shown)


Chand-ra:
Here's what the dashboard looks like:
```
python scripts/smite-orchestrator.py --out-dir ~/smite-results --configs baseline:encrypted_bytes,experimental:ir --targets lnd,ldk,cln,eclair --cores 0,1 --smite-dir ~/smite --afl-dir ~/AFLplusplus/ --trials 5 --timeout 120
```
Screenshot 2026-07-16 125255

CLN and Eclair ran pretty flaky: sometimes all of their trials would succeed and other times none would, but those two targets are flaky on my machine even when ran _normally_ (without this orchestrator). This is probably because I use a VM, but would love if someone else could confirm it.

Edit: The top of the dashboard should read **Smite Orchestrator (Parallel Fuzzing Trials)** instead of **Smite Orchestrator (Coverage Campaign)**. The screenshot was taken before this particular change was made, sorry.

erickcestari:
It's looking really nice!

Q: Couldn't this live in the smitebot cli?

Chand-ra:
> It's looking really nice!

Thanks, Eric!

> Q: Couldn't this live in the smitebot cli?

Right, the more I was working on the script the more I realized that this is a better fit as a feature in `smitebot`. We could reuse a lot of functionality from `smitebot start` (like automated seed creation, sharedir creation) that we need to re-implement here.

However, I decided to not go down that path because:

i) I already had a similar script in Python, and creating this script just meant making a couple of changes in that script (although it _did_ end up being more than a couple 😞). This is probably the biggest reason.

ii) This script uses `Rich` to build the TUI and that's something that I was familiar with, I'm completely blank on TUI libraries in Rust though. Although now that I look around, [`ratatui`](https://github.com/ratatui/ratatui) would have probably worked.

iii) I had a discussion with Matt and Nishant on this some time ago and we agreed that it was fine to implement the script in Python.

It might be worthwhile to note this (porting this script to `smitebot` as a separate command) as something we may want to do in the future though.

Ashish-Kumar-Dash:
Yes, I'm on board with this. Looking at it, the overlap with smitebot goes a bit further than `start`, the process-group teardown to reap orphaned QEMU is basically what `stop` does too, minus the graceful TERM→KILL escalation. Even the monitoring side wouldn't really need ratatui, #150 does its live view as a self-reloading tmux window over a single `fuzzer_stats` parser. Though the Rich dashboard aggregates many trials at once, which the single-campaign status view doesn't, so they might serve different purposes.

From my understanding, the one thing that'd make a port non-trivial is that smitebot is built entirely around `-Y` parallel campaigns, tmux, shared sharedir, runners cooperating on one coverage map, and this wants the opposite- isolated `-X` single-core trials. So `start` would need a single-trial `-X` mode before any of that reuse actually kicks in. Definitely worth writing down as future work, as `start.rs` is due for edits after `status` lands.

Chand-ra:
> From my understanding, the one thing that'd make a port non-trivial is that smitebot is built entirely around `-Y` parallel campaigns, tmux, shared sharedir, runners cooperating on one coverage map, and this wants the opposite- isolated `-X` single-core trials. So `start` would need a single-trial `-X` mode before any of that reuse actually kicks in. Definitely worth writing down as future work, as `start.rs` is due for edits after `status` lands.

Right, I first implemented this using `smitebot start` to run a single trial, and that worked for a while, but a couple of things proved to be confounding:
1. `smitebot start` uses UNIX timestamps to write `state.json` files to a globally shared tracking folder. Parallel trials that get executed at the same instance cause race conditions.

    I tried side-stepping this using time-based locks, but that turned out to be a flaky workaround. The best solution is probably to never write these files in the first place since we're directly polling `fuzzer_stats` to display the relevant data and `afl-fuzz.log` provides a window to each fuzzer's live state.

2. I wanted more granular control over when we create the sharedirs and how we use them. Right now we create them at the start of each trial (similar to `smitebot start`), but I can envision a future where we might want to create those before we run any trial (similar to what we're doing with the Docker images) and possibly re-use them across trials.

3. `tmux` proved to be the biggest obstacle. Running parallel `tmux` instances completely fudges the TUI dashboard. Need to entirely disable it for the parallel fuzzing trials.

morehouse:
> CLN and Eclair ran pretty flaky: sometimes all of their trials would succeed and other times none would, but those two targets are flaky on my machine even when ran normally (without this orchestrator). This is probably because I use a VM, but would love if someone else could confirm it.

Doesn't happen for me on bare metal.  Could you file a bug with the startup errors you get?

> Q: Couldn't this live in the smitebot cli?

I do eventually want this ported over to smitebot.  But a Python script works for now.

scripts/smite-orchestrator.py
```diff
@@ -0,0 +1,996 @@
+#!/usr/bin/env python
```
morehouse:
```suggestion
#!/usr/bin/env python3
```

scripts/smite-orchestrator.py
```diff
@@ -0,0 +1,996 @@
...
+    │   │   └── ...
+    │   └── <target_2>/                  # e.g., 'lnd'
+    └── <label_b>/                       # e.g., 'experimental'
+        ├── <target_1>/
+        └── <target_2>/
+
+Usage:
+    python smite-orchestrator.py \
+    --out-dir OUT_DIR \
+    --configs LABEL:SCENARIO[,LABEL:SCENARIO...] \
```
morehouse: I expect we will ~never compare different scenarios to each other.  Instead we'll probably be comparing the same scenario with different Smite branches or worktrees.

I think it would be worthwhile to optimize this script for the different-worktree case -- ideally we'd be able to specify a different smite directory for each configuration and the script would ensure that none of the build artifacts contaminate each other (e.g., Docker images are named according to the config label).

scripts/smite-orchestrator.py
```diff
@@ -0,0 +1,996 @@
...
+        return self.afl_out_dir / "default" / "fuzzer_stats"
+
+    @property
+    def ir_mutator_path(self) -> Path:
+        return self.smite_dir / "target" / "release" / "libsmite_ir_mutator.so"
+
+    def build_afl_cmd(self) -> list[str]:
+        """The exact afl-fuzz invocation for a standalone runner."""
+        # Fixed power schedule for every standalone trial.
+        POWER_SCHEDULE = "fast"
```
morehouse: Why "fast" instead of the default "explore"?

scripts/smite-orchestrator.py
```diff
@@ -0,0 +1,996 @@
...
+        ]
+
+    def build_afl_env(self) -> dict:
+        env = os.environ.copy()
+        env.update(
+            {
+                "AFL_NO_AFFINITY": "1",
+                "AFL_NO_UI": "1",
+                "AFL_NO_COLOR": "1",
+                "AFL_FORKSRV_INIT_TMOUT": "1800000",
```
morehouse, line 217: I haven't had to touch the forkserver timeout; I assume it's needed for your nested virtualization setup?

scripts/smite-orchestrator.py
```diff
@@ -0,0 +1,996 @@
...
+            self.cleanup()
+
+    def prepare_fs(self):
+        """Reset this trial's output directory and ensure its Nyx sharedir exists."""
+        shutil.rmtree(self.cfg.trial_dir, ignore_errors=True)
+        self.cfg.afl_out_dir.mkdir(parents=True)
+        self.ensure_sharedir()
+
+    def ensure_sharedir(self):
+        """Run `setup-nyx.sh` to create this trial's fresh Nyx sharedir."""
```
morehouse, line 608: I think we actually may want to serialize the `./setup-nyx.sh` runs since that script can cause a build to happen in the `AFLplusplus/nyx_mode` directory and concurrent builds there would corrupt stuff.

Chand-ra:
> > CLN and Eclair ran pretty flaky: sometimes all of their trials would succeed and other times none would, but those two targets are flaky on my machine even when ran normally (without this orchestrator). This is probably because I use a VM, but would love if someone else could confirm it.
> 
> Doesn't happen for me on bare metal. Could you file a bug with the startup errors you get?

Sure, I'll do that if I encounter those again.

scripts/smite-orchestrator.py
Chand-ra: Yes. Spawning parallel Nyx-mode AFL++ instances on my potato laptop takes some time, the default timeout frequently produced false positive crashes😅. This is supposed to be a generous upper bound to prevent that.

scripts/smite-orchestrator.py
Chand-ra: It is supposed to match what [`smitebot start` uses](https://github.com/morehouse/smite/blob/fe76ac48d467b3afb89de5ad23a1d3605c133dc8/smitebot/src/commands/start.rs#L410) for the primary node (runner 0). Worth changing to explore?

scripts/smite-orchestrator.py
Chand-ra: I think upfront serialization (similar to the docker images) would take too much time before any trial is run. Let us just use a thread lock around the `setup-nyx.sh` execution instead.

scripts/smite-orchestrator.py
Chand-ra: I personally have a use-case for comparing different scenarios, but I agree, the biggest use-case for our evaluation framework is most likely comparing different IR scenario configurations.

I'll make the script accept a `--scenario` flag and modify the `--smite-dir` flag to `--configs label:smite_dir`.

scripts/smite-orchestrator.py
Ashish-Kumar-Dash: The `fast` on start's runner 0 is just index 0 of the round-robin. `start` spreads schedules across cooperating runners (`POWER_SCHEDULES = [fast, explore, coe, ...]`, indexed by `runner_id % len`), so each `-Y` runner explores differently. Runner 0 lands on `fast` only because it's first in the array.

A standalone `-X` trial has no round-robin to diversify, so ig there's nothing to match. So it should default to `explore` imo, I will change this in `start.rs` to avoid confusion in the future.

Chand-ra:
> > > CLN and Eclair ran pretty flaky: sometimes all of their trials would succeed and other times none would, but those two targets are flaky on my machine even when ran normally (without this orchestrator). This is probably because I use a VM, but would love if someone else could confirm it.
> > 
> > 
> > Doesn't happen for me on bare metal. Could you file a bug with the startup errors you get?
> 
> Sure, I'll do that if I encounter those again.

I think now I have a good idea of why this happens.

To make the required changes to this script, I had to revert a change I had made to `setup-nyx.sh` which decreased the [memory allocation](https://github.com/morehouse/smite/blob/2eefc44fb3624c811ff4a166751864d04bd9c585/scripts/setup-nyx.sh#L75) for a new VM instance from 4 GB to 2 GB. I forgot to re-apply the said change after modifying the script. So now when I invoked the orchestrator, all the targets (including previously non-failing targets LND and LDK) failed with the same errors that CLN and Eclair did earlier.

I believe this is because a single target consumed (4 GB + some overhead) of memory, and because my VM has only 10 GB of memory, spawning two such instances forced the OOM killer to terminate one of these instances. Changing the allocated memory to 2 GB fixed the issue for LND, LDK, and Eclair.

As for why the issue only surfaced in CLN and Eclair earlier, I believe it is because these two targets are the most memory-hungry out of the set (CLN spawns a bunch of sub-daemons and Eclair needs JVM).

morehouse (approved):
Used this to evaluate https://github.com/morehouse/smite/pull/154#issuecomment-5071449199, and everything was seamless.

The ability to choose the CPU cores was especially helpful so that I could make sure P cores were used for all trials.

PR #155, opened 2026-07-17
smite-ir: add LookupShortChannelId operation

devvaansh:
Adds `LookupShortChannelId`, which takes a `FundingTransaction` and produces the BOLT 7 `short_channel_id` derived from its confirmed block position.

This is the missing piece needed to build validly-signed gossip messages that reference a real on-chain UTXO and can pass the UTXO validation checks in CLN and LND.

If the transaction is not yet confirmed (e.g. a mutator dropped `MineBlocks`), the sentinel `ShortChannelId::new(0, 0, 0)` is returned rather than panicking. A valid but unconfirmed program should not be a fatal error.

Ref: #71

### DIFF
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -226,6 +226,32 @@ pub enum Operation {
     /// Sign wallet inputs of the transaction and broadcast it via `bitcoin-cli`.
     /// Input: `FundingTransaction`.
     BroadcastTransaction,
+    /// Look up the confirmed block position of a broadcast funding transaction
+    /// and produce the corresponding BOLT 7 `short_channel_id`.
+    ///
+    /// This is the bridge between an on-chain funding output and the gossip
+    /// layer: the resulting `ShortChannelId` can be fed into
+    /// `BuildChannelAnnouncement`, `BuildChannelUpdate`, and
+    /// `BuildAnnouncementSignatures` so that gossip messages reference a real
+    /// UTXO and can pass the on-chain validation performed by CLN and LND.
+    ///
+    /// The typical program shape is:
+    ///
+    /// ```text
+    /// ft   = CreateFundingTransaction(...)
+    /// _    = BroadcastTransaction(ft)
+    /// _    = MineBlocks(k)          // k >= 1 for the tx to be confirmed
+    /// scid = LookupShortChannelId(ft)
+    /// ```
+    ///
+    /// If the funding transaction is unknown to the node or still in the
+    /// mempool (e.g. `MineBlocks` was dropped by a mutator), the sentinel
+    /// `ShortChannelId::new(0, 0, 0)` is produced. This keeps downstream
+    /// consumers well-typed without introducing a target- or program-side
+    /// error: any resulting gossip message simply fails on-chain validation.
+    ///
+    /// Input: `FundingTransaction`.
+    LookupShortChannelId,
 }
 
 /// A BOLT 2 compliant `upfront_shutdown_script` template.
@@ -680,6 +706,7 @@ impl fmt::Display for Operation {
             Self::RecvFundingSigned => write!(f, "RecvFundingSigned"),
             Self::RecvChannelReady => write!(f, "RecvChannelReady()"),
             Self::BroadcastTransaction => write!(f, "BroadcastTransaction"),
+            Self::LookupShortChannelId => write!(f, "LookupShortChannelId"),
         }
     }
 }
@@ -691,7 +718,9 @@ impl Operation {
     pub fn output_type(&self) -> Option<VariableType> {
         match self {
             Self::LoadAmount(_) => Some(VariableType::Amount),
-            Self::LoadShortChannelId(_) => Some(VariableType::ShortChannelId),
+            Self::LoadShortChannelId(_) | Self::LookupShortChannelId => {
+                Some(VariableType::ShortChannelId)
+            }
             Self::LoadFeeratePerKw(_) => Some(VariableType::FeeratePerKw),
             Self::LoadBlockHeight(_) => Some(VariableType::BlockHeight),
             Self::LoadTimestamp(_) => Some(VariableType::Timestamp),
@@ -768,7 +797,9 @@ impl Operation {
             ],
             Self::RecvAcceptChannel => vec![VariableType::SentOpenChannel],
             Self::RecvFundingSigned => vec![VariableType::SentFundingCreated],
-            Self::BroadcastTransaction => vec![VariableType::FundingTransaction],
+            Self::BroadcastTransaction | Self::LookupShortChannelId => {
+                vec![VariableType::FundingTransaction]
+            }
 
             Self::BuildOpenChannel => vec![
                 VariableType::ChainHash,    // chain_hash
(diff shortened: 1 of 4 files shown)


smite-scenarios/src/executor.rs
```diff
@@ -1423,6 +1459,13 @@ mod tests {
         fn get_transaction_confirmations(&mut self, _txid: Txid) -> u32 {
             self.confirmations
         }
+
+        fn get_transaction_block_position(&mut self, _txid: Txid) -> Option<TxBlockPosition> {
+            (self.confirmations > 0).then_some(TxBlockPosition {
+                block_height: 0,
+                tx_index: 0,
+            })
+        }
```
morehouse: This is currently unused.  We should add a test that uses it via a `LookupShortChannelId` op.

smite-ir/src/operation.rs
```diff
@@ -899,7 +931,8 @@ impl Operation {
             | Self::RecvChannelReady
             | Self::MineBlocks(_)
             | Self::CreateFundingTransaction
-            | Self::BroadcastTransaction => true,
+            | Self::BroadcastTransaction
+            | Self::LookupShortChannelId => true,
```
morehouse, line 935: I would argue that `LookupShortChannelId` does *not* have side effects, since it is a read-only operation.  But I see that we need it here to prevent CSE from merging it with other lookups, since the result is nondeterministic (depends on the relative position of a `MineBlocks` operation).

I think it's fine to put it here to keep this PR simple.  But this inconsistency hints that we may want a separate `is_deterministic` function for use by CSE, which could be a follow-up PR.

morehouse (commented):
Just missing an executor test.

smite-scenarios/src/executor.rs
devvaansh: added `execute_lookup_short_channel_id_confirmed` and `execute_lookup_short_channel_id_unconfirmed_returns_sentinel`. The mock now records `block_position_lookups` so the tests can assert the executor queried it with the correct txid, covering both the confirmed and sentinel paths.

smite-ir/src/operation.rs
devvaansh: Agreed. It's purely a read- no I/O, no state mutation. The only reason it's in `has_side_effects` is to keep CSE from collapsing two lookups that straddle a `MineBlocks`. Will track `is_deterministic` as a follow-up.

smite-scenarios/src/executor.rs
```diff
@@ -2882,21 +2895,78 @@ mod tests {
...
+        // The executor must have queried the mock with the broadcast
+        // transaction's txid.
+        assert_eq!(executor.bitcoin_cli.block_position_lookups.len(), 1);
+        let broadcast_txid = executor.bitcoin_cli.broadcast_calls[0].compute_txid();
         assert_eq!(
-            executor.bitcoin_cli.broadcast_calls[0].compute_txid(),
-            executor.bitcoin_cli.broadcast_calls[1].compute_txid(),
+            executor.bitcoin_cli.block_position_lookups[0],
+            broadcast_txid,
         );
```
morehouse, line 2982: We should also send a `channel_announcement` using the looked-up SCID, so we can assert that it matches what we expect here.

smite-scenarios/src/executor.rs
```diff
@@ -2882,21 +2895,78 @@ mod tests {
...
+                &Program {
+                    instructions: instrs,
+                },
+                std::time::Instant::now(),
+            )
+            .expect("lookup on unconfirmed tx should not fail");
+        // The mock was queried but returned None (zero confirmations), so the
+        // executor took the sentinel path without panicking.
+        assert!(executor.bitcoin_cli.mine_blocks_calls.is_empty());
+        assert_eq!(executor.bitcoin_cli.block_position_lookups.len(), 1);
```
morehouse, line 3052: We should also send the SCID in `channel_announcement` for this test, so we can verify it.

smite-scenarios/src/executor.rs
devvaansh: Done, both tests now feed the looked-up SCID into `BuildChannelAnnouncement → SendMessage`, decode the sent bytes, and assert `ca.short_channel_id`. Confirmed path asserts `ShortChannelId::new(800_042, 7, 0)`, sentinel path asserts `ShortChannelId::new(0, 0, 0)`.

smite-scenarios/src/executor.rs
devvaansh: Done - same as the confirmed test, the sentinel test now sends a `channel_announcement` with the looked-up SCID and asserts `ca.short_channel_id == ShortChannelId::new(0, 0, 0)`

smite-scenarios/src/executor.rs
```diff
@@ -467,36 +484,49 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
...
+                    // the intended fuzzing behaviour for a valid but
+                    // unconfirmed program.
+                    let scid = match self.bitcoin_cli.get_transaction_block_position(txid) {
+                        // Truncating cast: BOLT 7 `short_channel_id` allocates
+                        // 16 bits to the output index. In practice the
+                        // 2-of-2 funding output is at vout 0 or 1.
+                        #[allow(clippy::cast_possible_truncation)]
+                        Some(pos) => {
+                            ShortChannelId::new(pos.block_height, pos.tx_index, ft.vout as u16)
+                        }
```
morehouse: The comment is wrong -- I'm pretty sure the vout is always 0.  Also we should probably just panic if the vout is invalid as we currently do in `build_funding_created`.


```suggestion
                        Some(pos) => {
                            let funding_output_index = u16:try_from(ft.vout)
                                .expect("funding output index fits in u16");
                            ShortChannelId::new(pos.block_height, pos.tx_index, funding_output_index)
                        }
```

smite-scenarios/src/executor.rs
```diff
@@ -2857,30 +2934,30 @@ mod tests {
...
-            operation: Operation::MineBlocks(1),
-            inputs: vec![],
+            // Feed the FundingTransaction produced by
+            // CreateFundingTransaction (instruction 6) into the lookup. The
+            // resulting ShortChannelId is variable 9.
+            operation: Operation::LookupShortChannelId,
+            inputs: vec![6],
         });
+        // Build and send a channel_announcement carrying the looked-up SCID.
+        instrs.extend(channel_announcement_from_scid_instructions(10, 9));
```
```suggestion
        // Build and send a channel_announcement carrying the looked-up SCID.
        instrs.extend(channel_announcement_from_scid_instructions(instrs.len(), 9));
```

smite-scenarios/src/executor.rs
```diff
@@ -2890,21 +2967,90 @@ mod tests {
...
+                operation: Operation::CreateFundingTransaction,
+                inputs: vec![1, 3, 4, 5],
+            },
+            // The looked-up SCID is variable 7.
+            Instruction {
+                operation: Operation::LookupShortChannelId,
+                inputs: vec![6],
+            },
+        ];
+        instrs.extend(channel_announcement_from_scid_instructions(8, 7));
```
```suggestion
        instrs.extend(channel_announcement_from_scid_instructions(instrs.len(), 7));
```

morehouse (approved):

smite-scenarios/src/executor.rs
devvaansh: I added the comment since BOLT 7 technically reserves 16 bits for vout and I wasn't sure the funding output would always land at index 0 -- wanted to be explicit about the cast. But you're right, in practice it's always 0, and panicking on an invalid vout is the right approach. Updated to `u16::try_from(...).expect(...)`.

morehouse (approved):
LGTM

PR #157, opened 2026-07-20
smite-scenarios: expect channel_ready only for a valid funding outpoint

NishantBansal2003:
When the advertised funding outpoint does not pay the negotiated 2-of-2 funding script and amount, the target is unlikely to complete the funding flow or send an error, so we should not expect a `channel_ready` back. Track whether the advertised funding outpoint is valid and require it before expecting `channel_ready`, keeping `RecvChannelReady` a no-op otherwise


morehouse (approved):
LGTM

PR #158, opened 2026-07-20
smite-scenarios: expect channel_ready at negotiated minimum_depth

NishantBansal2003:
Previously, `channel_ready` was only expected after 8 confirmations, the maximum default `minimum_depth` advertised by Lightning implementations. This was overly conservative, since a smaller `minimum_depth` may be negotiated in `accept_channel`. Instead, expect `channel_ready` once the negotiated `minimum_depth` is reached.

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -56,11 +56,6 @@ pub const RECV_IDLE_TIMEOUT: Duration = Duration::from_secs(1);
 /// reconfiguring or patching CLN to poll more frequently.
 pub const RECV_CHANNEL_READY_TIMEOUT: Duration = Duration::from_secs(5);
 
-/// Default maximum `minimum_depth` that Lightning implementations advertise in
-/// `accept_channel`. Once the funding transaction reaches this many
-/// confirmations the target is guaranteed to have sent its `channel_ready`.
-const MAX_MINIMUM_DEPTH: u32 = 8;
-
 /// Abstraction over bitcoin-cli operations, allowing mock implementations in tests.
 pub trait BitcoinRpc {
     /// Mines the given number of blocks, including any transactions in the
@@ -868,6 +863,7 @@ fn build_funding_created(
         channel_type: open_channel.tlvs.channel_type.clone().unwrap_or_default(),
         opener,
         acceptor,
+        minimum_depth: accept_channel.minimum_depth,
     };
 
     let state = config.new_initial_commitment(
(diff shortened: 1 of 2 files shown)


smite/src/channel_tx/commitment.rs
```diff
@@ -130,6 +130,9 @@ pub struct ChannelState {
     /// revealed by `channel_ready` and then each `revoke_and_ack`. `None` until
     /// known.
     pub acceptor_next_per_commitment_point: Option<PublicKey>,
+    /// Minimum funding confirmations required before `channel_ready` is sent,
+    /// advertised in `accept_channel` during channel creation.
+    pub minimum_depth: u32,
```
ekzyis: nit: according to the documentation of `config`, `minimum_depth` could also go in there, because it will never change?

smite-scenarios/src/executor.rs
```diff
@@ -3398,13 +3394,13 @@ mod tests {
...
     #[test]
-    fn execute_recv_channel_ready_below_eight_confirmations_is_noop() {
+    fn execute_recv_channel_ready_below_minimum_depth_is_noop() {
         let (mut executor, channel_id, _) = recv_channel_ready_executor();
 
         let mut instrs = send_funding_created_and_recv_funding_signed_instructions();
         instrs.extend([
             Instruction {
-                operation: Operation::MineBlocks(6),
+                operation: Operation::MineBlocks(5),
```
ekzyis, line 3543: nit: could somehow make clear this is `accept_channel.minimum_depth` in `sample_funding_negotiation()` minus 1, but might be easy enough to understand

smite-scenarios/src/executor.rs
```diff
@@ -3432,13 +3429,13 @@ mod tests {
...
     #[test]
-    fn execute_recv_channel_ready_at_eight_confirmations_records_point() {
+    fn execute_recv_channel_ready_at_minimum_depth_records_point() {
         let (mut executor, channel_id, target_pcp) = recv_channel_ready_executor();
 
         let mut instrs = send_funding_created_and_recv_funding_signed_instructions();
         instrs.extend([
             Instruction {
-                operation: Operation::MineBlocks(8),
+                operation: Operation::MineBlocks(6),
```
ekzyis, line 3581: similar nit here

ekzyis (approved):
ACK fa089ea

smite-scenarios/src/executor.rs
```diff
@@ -1195,7 +1190,8 @@ fn recv_channel_ready(
 ///
 /// A `channel_ready` is expected when a tracked channel is still at commitment
 /// number 0, the counterparty's next per-commitment point is unknown, and the
-/// funding transaction has at least [`MAX_MINIMUM_DEPTH`] confirmations.
+/// funding transaction has at least the `minimum_depth` specified in
+/// `accept_channel` confirmations.
```
morehouse: Nit: I think this is slightly easier to understand

```suggestion
/// funding transaction has at least `minimum_depth` confirmations (as
/// specified in the received `accept_channel`).
```

smite/src/channel_tx/commitment.rs
morehouse: Right, `ChannelConfig` has the funding info, so I think `minimum_depth` fits well there.  Also then we don't have to add *another* parameter to `ChannelState::new`.

morehouse (approved):
LGTM

PR #160, opened 2026-07-21
smitebot: add bench command to measure Nyx execution speed

erickcestari:
The goal is to tell whether a change concretely improved the target's throughput or not: run bench before and after and compare execs/sec.

It runs a single input through the target's Nyx VM many times, timing the snapshot-restore-plus-target-run loop. Each exec is one snapshot restore plus one target run, so the numbers isolate VM/snapshot and target speed from AFL++'s mutation and scheduling overhead.

Like start, it builds the image and sets up the sharedir by default (pass --no-build to reuse an existing one), and loads libnyx.so at runtime the same way afl-fuzz does. --repeat averages over several fresh VM boots so a real change stands out from boot/snapshot variance, and --timeout bounds a single execution (default 2s).

### DIFF
--- a/smitebot/Cargo.toml
+++ b/smitebot/Cargo.toml
@@ -9,6 +9,7 @@ workspace = true
 
 [dependencies]
 clap = { version = "4.6", features = ["derive"] }
+libc.workspace = true
 log.workspace = true
 postcard.workspace = true
 serde.workspace = true
--- a/smitebot/README.md
+++ b/smitebot/README.md
@@ -82,6 +82,59 @@ smitebot status <campaign-id> --summary
 - `--summary`: Print a one-shot text summary to the terminal instead of attaching to the live dashboard.
 - `<campaign-id>` is the directory name under `~/.smitebot/runs` (printed by `smitebot start`).
 
+### smitebot bench-exec
+
+Benchmarks Nyx execution speed by running a single input through the target's VM many times. Each execution is a snapshot restore plus one target run, so the results isolate VM/snapshot and target speed from AFL++'s mutation and scheduling overhead.
+
+```bash
+smitebot bench-exec campaign.toml
+smitebot bench-exec campaign.toml --input testcase.bin --iterations 5000
+```
+
+- `--input`: Input file to execute repeatedly. Defaults to a single `0x00` byte. The default is handy for quickly measuring performance gains that come from the target itself rather than a specific test case: every exec still does the live-target ping-pong the executor uses to confirm the VM is up, so it exercises that path. It's a fast way to check improvements like JVM warmup/optimization.
+- `-n`, `--iterations`: Number of timed executions per run (default `1000`).
+- `-r`, `--repeat`: Number of full boot+measure runs to average over (default `1`). Each run boots a fresh VM, so repeats capture boot- and snapshot-level variance a single run cannot.
+- `--worker-id`: Nyx worker id, which namespaces the VM's workdir files (default `0`).
+- `--cpu`: CPU to pin the benchmark and its VM to (defaults to `--worker-id`). libnyx does no CPU binding of its own, so `bench-exec` sets its own affinity before booting and QEMU inherits it, the same way `afl-fuzz` does. On a hybrid CPU, pin to a core of the type you mean to measure: a P-core and an E-core give materially different numbers.
+- `--max-input-size`: Input buffer size in bytes for the VM, which must be a multiple of 4096 (QEMU-Nyx asserts the payload buffer is page-aligned). Defaults to `1048576`, AFL++'s `MAX_FILE` and therefore the buffer a real campaign runs with.
+- `-t`, `--timeout`: Per-execution timeout in seconds; an exec that runs longer is counted as timed out (default `2`).
+- `--no-build`: Skip building the image and setting up the sharedir; reuse an existing one.
+
+Like `start`, `bench-exec` builds the Docker image and sets up the Nyx sharedir before running, so it works from a clean checkout. Pass `--no-build` to skip that and reuse an already-prepared sharedir (`sharedir/config.ron`) from a prior `start`, `bench-exec`, or `scripts/setup-nyx.sh`. Either way it needs `libnyx.so` under `aflpp_path`, which it loads at runtime the same way `afl-fuzz` does.
+
+Booting the VM is what pays for target init and root snapshot creation: libnyx does not return from boot until the guest has reached its ready-to-fuzz state, so that cost is timed on its own. Every execution after that is a snapshot restore plus one target run; they are all timed together and reported as steady-state throughput:
+
+```
+Smite Nyx benchmark
+  target:      ldk/ir
+  sharedir:    /tmp/smite-nyx
+  input size:  406 bytes
+  input buffer:1048576 bytes
+  iterations:  100
+  repeats:     1
+
+  boot + snapshot creation: 4.056 s
+  steady-state (snapshot restore + target run):
+    execs/sec: 8.8
+    exec time: 11.324 s
+    latency:   min 106.16 ms  mean 113.24 ms  median 113.84 ms  p99 116.02 ms  max 116.06 ms
+    input execution: mean 109.35 ms  median 110.31 ms   (guest runtime)
+    nyx overhead:    mean 3.89 ms  median 4.00 ms   (restore + reset + ipc; 4045 dirty pages/exec)
+    failed iterations: 0 / 100
+    coverage determinism: 84.8% stable (2808 / 18456 edges fluctuated)
+```
+
+Each per-exec latency is split into two parts read from libnyx's auxiliary buffer:
+
+- **input execution** is the guest runtime the target actually spent processing the input, measured inside the VM. This is the "true" cost of the input; use it to tell whether a change made the target's own work faster.
+- **nyx overhead** is the wall time minus that guest runtime: snapshot restore/reset plus hypercall round-trips. `dirty pages/exec` is the number of pages restored each execution, the main driver of that overhead, so a change that dirties more memory (e.g. spawning a process per block) shows up as both higher overhead and more dirty pages.
+
+`coverage determinism` reports how reproducible the target's edge coverage is across the identical executions. Each execution's AFL coverage bitmap is captured and its per-edge hit counts are bucketed with AFL's count classes; an edge whose bucketed count is not identical across all executions is counted as fluctuated, and the denominator is the edges hit at least once. High determinism (close to 100%) means the same input reliably reaches the same code; fluctuation points to nondeterminism (uninitialized memory, timing-dependent branches, ASLR, background threads) that adds noise to the fuzzer's coverage signal. Bucketing means benign loop-trip jitter is not counted.
+
+With `--repeat` greater than 1 each run is reported separately, followed by an aggregate (mean, stddev, and range of execs/sec across runs; coverage determinism is pooled over all runs).
+
+`failed iterations` counts executions that ended in a crash, timeout, or error rather than a clean run; a non-zero count means the input is not a clean steady-state case and the numbers are skewed.
+
 ### smitebot config
 
 Validates a campaign configuration file, reports the resolved settings, and checks that referenced paths exist on disk.
--- a/smitebot/src/commands/bench_exec.rs
+++ b/smitebot/src/commands/bench_exec.rs
@@ -1,0 +1,971 @@
+//! Single-input Nyx benchmark.
+//!
+//! Boots the target's Nyx VM and runs one fixed input through it a fixed number
+//! of times, measuring steady-state throughput and per-execution latency.
+//! Because each execution is a snapshot restore plus one target run, the numbers
+//! isolate VM/snapshot and target speed from AFL++'s mutation and scheduling
+//! overhead.
+//!
+//! Like `start`, `bench-exec` builds the Docker image and sets up the Nyx
+//! sharedir before running. Pass `--no-build` to skip that and reuse an
+//! already-prepared sharedir (e.g. from a prior `start`, `bench-exec`, or
+//! `scripts/setup-nyx.sh`).
+
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::time::{Duration, Instant};
+
+use clap::Args;
+
+use crate::commands::build::{BuildInputs, run_build};
+use crate::config::CampaignConfig;
+use crate::latency_stats::{LatencyStats, avg_duration, mean_stddev};
+use crate::libnyx::{Libnyx, PAYLOAD_HEADER_SIZE};
+use crate::utils::{pin_to_cpu, setup_nyx};
+
+/// Default number of timed executions when `--iterations` is not given.
+const DEFAULT_ITERATIONS: u64 = 1000;
+
+/// Upper bound on `--iterations`. The run preallocates and retains timing
+/// samples per iteration, so an unbounded count would reserve gigabytes.
+const MAX_ITERATIONS: u64 = 1_000_000;
+
+/// Upper bound on `--repeat`, which likewise sizes a preallocation.
+const MAX_REPEAT: u32 = 1_000;
+
+/// Guest page size (x86-64 base page, `x86_64_PAGE_SIZE` in QEMU-Nyx) that the
+/// input buffer must be a whole multiple of.
+///
+/// QEMU-Nyx maps the payload buffer into the guest one page at a time and
+/// asserts `shared_payload_buffer_size % x86_64_PAGE_SIZE == 0`
+/// (`nyx_mode/QEMU-Nyx/nyx/memory_access.c`), so any buffer size we pass must be
+/// page-rounded or the VM aborts. AFL++ never trips this because its buffer is a
+/// page-multiple 1 MiB.
+const NYX_PAGE_SIZE: u32 = 4096;
+
+/// Input buffer size the VM is booted with unless `--max-input-size` says
+/// otherwise: AFL++'s `MAX_FILE`, which is what afl-fuzz hands to
+/// `nyx_config_set_input_buffer_size`.
+///
+/// Matching AFL++ by default means the benchmark measures the buffer a real
+/// campaign runs with. Page-aligned, so it satisfies the QEMU-Nyx assertion
+/// above.
+const DEFAULT_MAX_INPUT_SIZE: u32 = 1 << 20;
+
+/// Default per-execution timeout in seconds. An input that runs longer is
+/// reported as a failed (timed-out) iteration rather than hanging the run.
+const BENCH_TIMEOUT_SECS: u8 = 2;
+
+/// Command handler for `smitebot bench-exec`.
+pub struct BenchExecCommand;
+
+/// CLI arguments for `smitebot bench-exec`.
+#[derive(Debug, Args)]
+pub struct BenchExecArgs {
+    /// Path to the campaign configuration TOML file.
+    path: PathBuf,
+    /// Input file to execute repeatedly; defaults to a single 0x00 byte.
+    #[arg(long, short)]
+    input: Option<PathBuf>,
+    /// Number of timed executions to run (at most 1000000, since every
+    /// execution's timing samples are held until the run ends).
+    #[arg(long, short = 'n', default_value_t = DEFAULT_ITERATIONS)]
+    iterations: u64,
+    /// Number of full boot+measure runs to repeat and average over (at most 1000).
+    ///
+    /// Each run boots a fresh VM, so repeats capture boot- and snapshot-level
+    /// variance that a single run cannot. Results are reported per run plus an
+    /// aggregate.
+    #[arg(long, short = 'r', default_value_t = 1)]
+    repeat: u32,
+    /// Nyx worker id for the VM, which namespaces its workdir files so several
+    /// VMs can share a workdir.
+    #[arg(long, default_value_t = 0)]
+    worker_id: u32,
+    /// CPU to pin the benchmark and its VM to; defaults to `--worker-id`.
+    ///
+    /// Pinning keeps the scheduler from migrating the VM mid-run. On a hybrid
+    /// CPU, pick a core of the type you mean to measure: a P-core and an E-core
+    /// give materially different numbers.
+    #[arg(long)]
+    cpu: Option<usize>,
+    /// Input buffer size in bytes for the Nyx VM; must be a multiple of 4096.
+    ///
+    /// Defaults to AFL++'s 1 MiB, the buffer a real campaign runs with.
+    #[arg(long, default_value_t = DEFAULT_MAX_INPUT_SIZE)]
+    max_input_size: u32,
+    /// Per-execution timeout in seconds before an exec is counted as timed out.
+    /// Must be at least 1; Nyx reads zero as "no timeout".
+    #[arg(long = "timeout", short = 't', default_value_t = BENCH_TIMEOUT_SECS)]
+    timeout_secs: u8,
+    /// Skip building the image and setting up the sharedir; reuse an existing one.
+    #[arg(long)]
+    no_build: bool,
+}
+
+impl BenchExecCommand {
+    /// Runs the benchmark and returns whether it completed successfully.
+    pub fn execute(args: &BenchExecArgs) -> bool {
+        let iterations = match validate_iterations(args.iterations) {
+            Ok(n) => n,
+            Err(e) => {
+                log::error!("{e}");
+                return false;
+            }
+        };
+        let repeat = match validate_repeat(args.repeat) {
+            Ok(n) => n,
+            Err(e) => {
+                log::error!("{e}");
+                return false;
+            }
+        };
+        // A zero timeout causes Nyx to silently turn off the timing the
+        // benchmark exists to measure.
+        if args.timeout_secs == 0 {
+            log::error!("Nyx requires --timeout to be at least 1 second");
+            return false;
+        }
+
+        let config = match CampaignConfig::load(&args.path) {
+            Ok(c) => c,
+            Err(e) => {
+                log::error!("{e}");
+                return false;
+            }
+        };
+
+        let input = match load_input(args) {
+            Ok(input) => input,
+            Err(e) => {
+                log::error!("{e}");
+                return false;
+            }
+        };
+        let max_input_size = match validate_max_input_size(args.max_input_size, input.len()) {
+            Ok(size) => size,
+            Err(e) => {
+                log::error!("{e}");
+                return false;
+            }
+        };
+
+        let Some(libnyx_path) = locate_libnyx(&config) else {
+            return false;
+        };
+
+        if !ensure_sharedir(&config, args.no_build) {
+            return false;
+        }
+
+        let libnyx = match Libnyx::load(&libnyx_path) {
+            Ok(lib) => lib,
+            Err(e) => {
+                log::error!("failed to load {}: {e}", libnyx_path.display());
+                return false;
+            }
+        };
+
+        // Pin before booting: libnyx spawns QEMU as a child, so the VM inherits
+        // this affinity mask. Doing it here also pins the timing loop itself.
+        let cpu = args.cpu.unwrap_or(args.worker_id as usize);
+        if let Err(e) = pin_to_cpu(cpu) {
+            log::error!("{e}");
+            return false;
+        }
+        log::info!("pinned to CPU {cpu}");
+
+        // Each repeat is a fresh VM boot so the aggregate captures boot- and
+        // snapshot-level variance, which dominates run-to-run spread.
+        let mut runs = Vec::with_capacity(repeat);
+        for run_index in 0..repeat {
+            if repeat > 1 {
+                log::info!("benchmark run {}/{repeat}", run_index + 1);
+            }
+            match run_once(
+                &libnyx,
+                &config,
+                &input,
+                args,
+                max_input_size,
+                iterations,
+                run_index,
+            ) {
+                Ok(result) => runs.push(result),
+                Err(e) => {
+                    log::error!("{e}");
+                    return false;
+                }
+            }
+        }
+
+        let report = BenchReport {
+            target: format!("{}/{}", config.target, config.scenario),
+            sharedir: config.sharedir.clone(),
+            input_len: input.len(),
+            max_input_size,
+            iterations,
+            runs,
+        };
+        report.print();
+
+        true
+    }
+}
+
+/// Boots a fresh VM, times one benchmark run, and tears the VM down.
+fn run_once(
+    libnyx: &Libnyx,
+    config: &CampaignConfig,
+    input: &[u8],
+    args: &BenchExecArgs,
+    max_input_size: u32,
+    iterations: usize,
+    run_index: usize,
+) -> Result<RunResult, String> {
+    let workdir = BenchWorkdir::create(&config.output_dir, run_index)
+        .map_err(|e| format!("failed to create benchmark workdir: {e}"))?;
+
+    // libnyx does not return until the guest reaches the ready-to-fuzz state, so
+    // this call is where VM boot, target init, and root snapshot creation are
+    // paid for (`QemuProcess::new` in libnyx spins until `result.state == 3`).
+    log::info!("booting Nyx VM from {}", config.sharedir.display());
+    let boot_start = Instant::now();
+    let mut vm = libnyx.boot(
+        &config.sharedir,
+        workdir.path(),
+        max_input_size,
+        args.worker_id,
+        args.timeout_secs,
+    )?;
+    let boot_time = boot_start.elapsed();
+
+    // The coverage bitmap lets us measure how deterministic the target's edge
+    // coverage is across identical executions. Skip it if libnyx reports no map.
+    let bitmap_size = vm.bitmap_size();
+    let mut coverage = (bitmap_size > 0).then(|| CoverageTracker::new(bitmap_size));
+    let mut bitmap_buf = vec![0u8; bitmap_size];
+
+    log::info!("running {iterations} timed executions");
+    let mut latencies = Vec::with_capacity(iterations);
+    let mut target_runtimes = Vec::with_capacity(iterations);
+    let mut overheads = Vec::with_capacity(iterations);
+    let mut failed_iterations = 0u64;
+    let mut dirty_pages_total = 0u64;
+    // Throughput is the sum of the timed exec latencies, not the loop's wall
+    // clock.
+    let mut total = Duration::ZERO;
+    for _ in 0..iterations {
+        let exec_start = Instant::now();
+        let stats = vm.exec_with_stats(input);
+        let latency = exec_start.elapsed();
+        total += latency;
+
+        latencies.push(latency);
+        target_runtimes.push(stats.target_runtime);
+        // Overhead is the wall time not spent running the target: snapshot
+        // restore/reset plus hypercall round-trips. Saturating guards the rare
+        // case where the guest-measured runtime rounds above the host latency.
+        overheads.push(latency.saturating_sub(stats.target_runtime));
+        dirty_pages_total += u64::from(stats.dirty_pages);
+        if !stats.result.is_normal() {
+            failed_iterations += 1;
+        }
+
+        if let Some(tracker) = coverage.as_mut() {
+            vm.copy_bitmap_into(&mut bitmap_buf);
+            tracker.record(&bitmap_buf);
+        }
+    }
+
+    // `vm` (QEMU shutdown) then `workdir` (rm -rf) drop in reverse order here.
+    Ok(RunResult {
+        boot_time,
+        total,
+        failed_iterations,
+        mean_dirty_pages: mean_u64(dirty_pages_total, iterations),
+        stats: LatencyStats::summarize(&mut latencies),
+        target: LatencyStats::summarize(&mut target_runtimes),
+        overhead: LatencyStats::summarize(&mut overheads),
+        coverage: coverage.map(|c| c.summary()),
+    })
+}
+
+/// Tracks per-edge coverage stability across the executions of one run.
+///
+/// AFL++'s bitmap holds one byte per edge, a raw hit count. The same input on a
+/// deterministic target should light the same edges with the same *bucketed*
+/// count every execution; an edge whose bucketed count varies across executions
+/// is unstable (nondeterminism from uninitialized memory, timing-dependent
+/// branches, ASLR, background threads, etc.). Bucketing with AFL++'s live count
+/// classes means benign within-bucket jitter (e.g. 100 vs 101 iterations, both
+/// in `[64..127]`) is not flagged, matching what actually perturbs AFL's
+/// coverage.
+struct CoverageTracker {
+    /// Bucketed hit count from the first recorded execution, per edge.
+    reference: Vec<u8>,
+    /// Whether each edge was hit (nonzero) in any recorded execution.
+    hit: Vec<bool>,
+    /// Whether each edge's bucketed count ever differed from `reference`.
+    fluctuated: Vec<bool>,
+    /// Number of executions recorded so far.
+    samples: u64,
+}
+
+impl CoverageTracker {
+    fn new(bitmap_size: usize) -> Self {
+        Self {
+            reference: vec![0; bitmap_size],
+            hit: vec![false; bitmap_size],
+            fluctuated: vec![false; bitmap_size],
+            samples: 0,
+        }
+    }
+
+    /// Folds one execution's raw bitmap into the running tally.
+    fn record(&mut self, bitmap: &[u8]) {
+        for (i, &raw) in bitmap.iter().enumerate() {
+            let bucket = classify_count(raw);
+            if bucket != 0 {
+                self.hit[i] = true;
+            }
+            if self.samples == 0 {
+                self.reference[i] = bucket;
+            } else if bucket != self.reference[i] {
+                self.fluctuated[i] = true;
+            }
+        }
+        self.samples += 1;
+    }
+
+    /// Collapses the tally into executed- and fluctuated-edge counts.
+    fn summary(&self) -> CoverageSummary {
+        let executed_edges = self.hit.iter().filter(|&&h| h).count() as u64;
+        let fluctuated_edges = self.fluctuated.iter().filter(|&&f| f).count() as u64;
+        CoverageSummary {
+            executed_edges,
+            fluctuated_edges,
+        }
+    }
+}
+
+/// AFL's count-class bucketing: raw hit counts collapse to fixed buckets so that
+/// small variations within a bucket are treated as identical coverage.
+///
+/// Returns the bucket's index rather than AFL's bucket value. Only equality
+/// between two executions matters here, so the index is equivalent and keeps the
+/// ranges and their labels lined up.
+fn classify_count(count: u8) -> u8 {
+    match count {
+        0 => 0,
+        1 => 1,
+        2..=3 => 2,
+        4..=7 => 3,
+        8..=15 => 4,
+        16..=31 => 5,
+        32..=63 => 6,
+        64..=127 => 7,
+        128..=255 => 8,
+    }
+}
+
+/// Executed- and fluctuated-edge counts for one run's coverage stability.
+#[derive(Clone, Copy)]
+struct CoverageSummary {
+    /// Edges hit (nonzero) in at least one execution.
+    executed_edges: u64,
+    /// Executed edges whose bucketed count was not identical across executions.
+    fluctuated_edges: u64,
+}
+
+impl CoverageSummary {
+    /// Percentage of executed edges that were stable across all executions.
+    #[allow(clippy::cast_precision_loss)] // display figure; edge counts are small
+    fn stable_pct(&self) -> f64 {
+        if self.executed_edges == 0 {
+            100.0
+        } else {
+            let stable = self.executed_edges - self.fluctuated_edges;
+            stable as f64 / self.executed_edges as f64 * 100.0
+        }
+    }
+}
+
+/// Mean of a summed total over `count` samples, as a float (0.0 if no samples).
+#[allow(clippy::cast_precision_loss)] // display figure; exact precision unneeded
+fn mean_u64(total: u64, count: usize) -> f64 {
+    if count == 0 {
+        0.0
+    } else {
+        total as f64 / count as f64
+    }
+}
+
+/// Validates `--iterations` and returns it as a `usize`, the length the run's
+/// sample vectors are allocated from.
+fn validate_iterations(iterations: u64) -> Result<usize, String> {
+    if iterations == 0 {
+        return Err("--iterations must be at least 1".to_string());
+    }
+    if iterations > MAX_ITERATIONS {
+        return Err(format!(
+            "--iterations must be at most {MAX_ITERATIONS}, got {iterations}"
+        ));
+    }
+    usize::try_from(iterations)
+        .map_err(|_| format!("--iterations is {iterations}, more than this platform can address"))
+}
+
+/// Validates `--repeat` and returns it as the capacity for the results vector.
+fn validate_repeat(repeat: u32) -> Result<usize, String> {
+    if repeat == 0 {
+        return Err("--repeat must be at least 1".to_string());
+    }
+    if repeat > MAX_REPEAT {
+        return Err(format!(
+            "--repeat must be at most {MAX_REPEAT}, got {repeat}"
+        ));
+    }
+    usize::try_from(repeat)
+        .map_err(|_| format!("--repeat is {repeat}, more than this platform can address"))
+}
+
+/// Validates `--max-input-size` against the input it has to carry.
+///
+/// The size reaches QEMU-Nyx verbatim, so it is rejected rather than adjusted:
+/// silently resizing the buffer would change the very thing the benchmark
+/// measures.
+fn validate_max_input_size(size: u32, input_len: usize) -> Result<u32, String> {
+    if size == 0 || !size.is_multiple_of(NYX_PAGE_SIZE) {
+        return Err(format!(
+            "--max-input-size must be a nonzero multiple of {NYX_PAGE_SIZE} (QEMU-Nyx asserts \
+             the payload buffer is page-aligned), got {size}"
+        ));
+    }
+    // Saturating: the sum only has to be compared against a `u32`, so clamping an
+    // absurd length at `u64::MAX` still rejects it, where adding would overflow.
+    if (input_len as u64).saturating_add(u64::from(PAYLOAD_HEADER_SIZE)) > u64::from(size) {
+        return Err(format!(
+            "input is {input_len} bytes and its {PAYLOAD_HEADER_SIZE}-byte length prefix shares \
+             the buffer, but --max-input-size is {size}; raise it (the guest allocates the whole \
+             buffer every execution, so it must fit the VM's RAM)"
+        ));
+    }
+    Ok(size)
+}
+
+/// Loads the input to benchmark, defaulting to a single zero byte.
+fn load_input(args: &BenchExecArgs) -> Result<Vec<u8>, String> {
+    match &args.input {
+        Some(path) => {
+            fs::read(path).map_err(|e| format!("failed to read input {}: {e}", path.display()))
+        }
+        None => Ok(vec![0u8]),
+    }
+}
+
+/// Validates `aflpp_path` and locates `libnyx.so`, returning its path.
+///
+/// AFL++'s Nyx build is required both to prepare the sharedir (setup-nyx.sh)
+/// and to drive the VM at runtime, so it is checked before either.
+fn locate_libnyx(config: &CampaignConfig) -> Option<PathBuf> {
+    if !config.aflpp_path.exists() {
+        log::error!("aflpp_path does not exist: {}", config.aflpp_path.display());
+        return None;
+    }
+
+    let libnyx_path = config.aflpp_path.join("libnyx.so");
+    if !libnyx_path.exists() {
+        log::error!(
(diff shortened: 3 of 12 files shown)


smitebot/README.md
```diff
@@ -53,6 +53,55 @@ After spawning, startup is verified by polling for `fuzzer_stats` files. Because
...
+### smitebot bench
+
+Benchmarks Nyx execution speed by running a single input through the target's VM many times. Each execution is a snapshot restore plus one target run, so the results isolate VM/snapshot and target speed from AFL++'s mutation and scheduling overhead.
+
+```bash
+smitebot bench campaign.toml
+smitebot bench campaign.toml --input testcase.bin --iterations 5000
+```
+
+- `--input`: Input file to execute repeatedly. Defaults to a single `0x00` byte.
```
morehouse: I think `--input` should be required -- it should be rare that anyone wants to benchmark an empty program.

smitebot/README.md
```diff
@@ -53,6 +53,55 @@ After spawning, startup is verified by polling for `fuzzer_stats` files. Because
...
+- `-n`, `--iterations`: Number of timed executions per run (default `1000`).
+- `-r`, `--repeat`: Number of full boot+measure runs to average over (default `1`). Each run boots a fresh VM, so repeats capture boot- and snapshot-level variance a single run cannot.
+- `--worker-id`: Nyx worker id, which also selects the pinned CPU (default `0`).
+- `--max-input-size`: Input buffer size in bytes for the VM (default `1048576`).
```
morehouse: Can we automatically determine the max input size using the actual `--input` size?

smitebot/src/commands/start.rs
```diff
@@ -166,7 +166,7 @@ impl StartCommand {
 }
 
 /// Runs `scripts/setup-nyx.sh` to prepare the Nyx sharedir.
-fn setup_nyx(config: &CampaignConfig, image: &str) -> bool {
+pub(crate) fn setup_nyx(config: &CampaignConfig, image: &str) -> bool {
```
Chand-ra: At this point, I think it is better to move this is a separate file where it can be accessed by both these commands. Maybe `commands/utils/build.rs` or `build_utils.rs`?

The new orchestrator script also needs this functionality and we may want to port it as a `smitebot` command in the future, so there's at least 3 independent use-cases for this functionality. I believe that is enough to warrant a separate, easily reusable module.

smitebot/src/commands/bench.rs
```diff
@@ -0,0 +1,682 @@
...
+    let mean = values.iter().sum::<f64>() / n as f64;
+    if n < 2 {
+        return (mean, 0.0);
+    }
+    let variance = values.iter().map(|v| (v - mean).powi(2)).sum::<f64>() / (n - 1) as f64;
+    (mean, variance.sqrt())
+}
+
+/// Averages an iterator of durations (nanosecond precision).
+fn avg_duration(durations: impl Iterator<Item = Duration>) -> Duration {
```
Chand-ra: This single file implements three separate functionalities: calculating statistics, running the actual command, and formatting the results.

I think it is fine that we run the command and format the results in the same file (since they're connected operations), but scattering statistical calculations between them doesn't look right to me. They should also be moved to a separate utility module.

This should also make the reviewing process easier, right now it takes quite some mental overhead to code-switch between these parts.

smitebot/README.md
```diff
@@ -53,6 +53,55 @@ After spawning, startup is verified by polling for `fuzzer_stats` files. Because
 
 Campaign state is saved to `~/.smitebot/runs/<campaign-id>/state.json` for use by future `stop` and `status` commands.
 
+### smitebot bench
```
Chand-ra: `bench` is too generic, imo. We might want to benchmark other metrics (like coverage) using `smitebot` in the future. Something like `bench-exec` would be better.

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,682 @@
+//! Single-input Nyx benchmark.
+//!
+//! Boots the target's Nyx VM and runs one fixed input through it a fixed number
+//! of times, measuring steady-state throughput and per-execution latency.
```
Chand-ra, line 4: I think it would be nice if we could also measure **edge instability**: the proportion of executed code edges that fluctuate across the N iterations. Perhaps using the bitmap that AFL++ outputs on each run? Should look something like this in the output:
```
coverage determinism: 98.4% stable (12 / 750 edges fluctuated)
```

Chand-ra (changes requested):
Looks like a great addition overall. Following are some improvement suggestions.

smitebot/src/commands/bench_exec.rs
erickcestari: Good idea! I've added this feature to it.

erickcestari:
Thanks for the review @Chand-ra ! I've added the coverage metric, which is a nice feature to have. Specially when we start using the `bedrock` as one of the hypervisor.

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,377 @@
...
+
+/// Loads a required symbol out of a [`Library`], copying the function pointer.
+macro_rules! sym {
+    ($lib:expr, $name:literal, $ty:ty) => {{
+        // SAFETY: the caller vouches that `$lib` is a genuine libnyx.so whose
+        // exported symbol `$name` has the C signature `$ty`.
+        let symbol: Symbol<$ty> = unsafe { $lib.get(concat!($name, "\0").as_bytes())? };
+        *symbol
+    }};
+}
```
morehouse: Nit: IMO it's cleaner to inline this everywhere, with one unsafe block around it all.  The code becomes clearer and function types should resolve automatically.

Also it looks like if we use libloading 0.9 we don't need add a null terminator or cast to bytes.

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,377 @@
...
+    pub fn copy_bitmap_into(&self, dst: &mut [u8]) {
+        // SAFETY: `get_bitmap_buffer` returns the live bitmap mapping whose
+        // length is `get_bitmap_buffer_size`; copying the min of that and
+        // `dst.len()` stays in bounds of both, and the regions do not overlap.
+        unsafe {
+            let src = (self.lib.get_bitmap_buffer)(self.process);
+            let size = self.bitmap_size();
+            let n = size.min(dst.len());
+            std::ptr::copy_nonoverlapping(src, dst.as_mut_ptr(), n);
+        }
```
morehouse, line 459: Nit: the `n` calculation can go outside the unsafe block


```suggestion
        let n = self.bitmap_size().min(dst.len());
        // SAFETY: `get_bitmap_buffer` returns the live bitmap mapping whose
        // length is `get_bitmap_buffer_size`; copying the min of that and
        // `dst.len()` stays in bounds of both, and the regions do not overlap.
        unsafe {
            let src = (self.lib.get_bitmap_buffer)(self.process);
            std::ptr::copy_nonoverlapping(src, dst.as_mut_ptr(), n);
        }
```

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,377 @@
...
+    pub const VERSION: u16 = 3;
+
+    /// Start of the `result` section within the buffer.
+    const RESULT: usize = 896;
+    /// `result.dirty_pages` (`u32`): pages restored for the last exec.
+    pub const DIRTY_PAGES: usize = RESULT + 16;
+    /// `result.runtime_usec` (`u32`): microsecond part of guest payload runtime.
+    pub const RUNTIME_USEC: usize = RESULT + 28;
+    /// `result.runtime_sec` (`u32`): second part of guest payload runtime.
+    pub const RUNTIME_SEC: usize = RESULT + 32;
```
morehouse, line 37: These offsets are all correct AFAICT.  But it seems fragile -- it would be nice if Nyx provided a clearer interface.

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,768 @@
...
+        CoverageSummary {
+            executed_edges,
+            fluctuated_edges,
+        }
+    }
+}
+
+/// AFL's count-class bucketing: raw hit counts collapse to fixed buckets so that
+/// small variations within a bucket are treated as identical coverage.
+fn classify_count(count: u8) -> u8 {
```
Ashish-Kumar-Dash, line 358: This is AFL's old count-class table, the one AFLplusplus keeps only as a /* OLD */ comment. The live table `(src/afl-fuzz-bitmap.c:51-59)` groups **[2..3]=2** and splits **[32..63]=32**/ **[64..127]=64**. This buckets 3 with 4 and merges **32..127**.

  Two real consequences for the determinism metric-
  - edge toggling **2↔3 hits**: AFL treats it as stable, this flags it fluctuated.
  - edge toggling **40↔100 hits**: AFL treats it as changed, this reports it stable.

So the comment above ("matching what actually perturbs AFL's coverage") doesn't quite hold with this table. I'd suggest we can switch to the live buckets and cite `afl-fuzz-bitmap.c.` Note the current tests
  (**[5]→[7]**, **[100]→[101]**) pass under both tables, a 2 vs 3 case is needed to pin it.

smitebot/README.md
```diff
@@ -53,6 +53,58 @@ After spawning, startup is verified by polling for `fuzzer_stats` files. Because
...
+
+- `--input`: Input file to execute repeatedly. Defaults to a single `0x00` byte.
+- `-n`, `--iterations`: Number of timed executions per run (default `1000`).
+- `-r`, `--repeat`: Number of full boot+measure runs to average over (default `1`). Each run boots a fresh VM, so repeats capture boot- and snapshot-level variance a single run cannot.
+- `--worker-id`: Nyx worker id, which also selects the pinned CPU (default `0`).
+- `--max-input-size`: Input buffer size in bytes for the VM (default `1048576`).
+- `-t`, `--timeout`: Per-execution timeout in seconds; an exec that runs longer is counted as timed out (default `2`).
+- `--no-build`: Skip building the image and setting up the sharedir; reuse an existing one.
+
+Like `start`, `bench` builds the Docker image and sets up the Nyx sharedir before running, so it works from a clean checkout. Pass `--no-build` to skip that and reuse an already-prepared sharedir (`sharedir/config.ron`) from a prior `start`, `bench`, or `scripts/setup-nyx.sh`. Either way it needs `libnyx.so` under `aflpp_path`, which it loads at runtime the same way `afl-fuzz` does.
```
Ashish-Kumar-Dash: Nit- The command is bench-exec, this line says bench twice ("Like start, bench builds…", "from a prior start, bench").

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,377 @@
...
+    /// to compare coverage across executions it must be copied out before the
+    /// next `exec`. At most `min(bitmap_size(), dst.len())` bytes are copied;
+    /// pass a `dst` of `bitmap_size()` bytes to capture the whole map.
+    pub fn copy_bitmap_into(&self, dst: &mut [u8]) {
+        // SAFETY: `get_bitmap_buffer` returns the live bitmap mapping whose
+        // length is `get_bitmap_buffer_size`; copying the min of that and
+        // `dst.len()` stays in bounds of both, and the regions do not overlap.
+        unsafe {
+            let src = (self.lib.get_bitmap_buffer)(self.process);
+            let size = self.bitmap_size();
```
Ashish-Kumar-Dash: Nit- `self.bitmap_size()` is an FFI call, and it runs once per timed iteration when coverage is on. The size is fixed after boot. We can cache it at boot (or pass it in) to drop the call from the loop.

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,768 @@
...
+    /// Edge-coverage stability across the run, if a bitmap was available.
+    coverage: Option<CoverageSummary>,
+}
+
+impl RunResult {
+    /// Steady-state throughput for this run.
+    // Execs/sec is a display figure; f64 precision loss on the exec count is
+    // immaterial at any realistic iteration count.
+    #[allow(clippy::cast_precision_loss)]
+    fn execs_per_sec(&self, iterations: u64) -> f64 {
```
Ashish-Kumar-Dash: Nit - The timed total still includes the per-iteration timer reads and vector pushes, so at sub-µs execs it reads a bit under the latency line and the two won't match. We can either subtract it like `coverage_overhead` already does, or add a small note in the report that execs/sec includes a little harness overhead.

smitebot/Cargo.toml
```diff
@@ -9,6 +9,7 @@ workspace = true
 
 [dependencies]
 clap = { version = "4.6", features = ["derive"] }
+libloading = "0.8"
```
Ashish-Kumar-Dash: The code already copies the fn pointers out and keeps the lib open for the whole process, so libloading's lifetime-safety isn't in play, and it's a cross-platform loader dragging windows feature into a primarily Linux project. libc is already in our workspace, so `libc::dlopen` can swap it out (or bare dl externs drop the dep entirely, but i'm not sure). Every symbol's a function, so a null check is all we need. WDYT?

Ashish-Kumar-Dash (changes requested):
I had a few thoughts/nits, otherwise its a solid PR! The only real concern I have is the order in which we merge status, bench and print-ir since one of us will have to rebase, but I think it'd be much better if bench,print-ir could get merged and then we land status. I don't think it will be an issue on your part, since I'd have to do the rebase then :)

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,768 @@
...
+        0 => 0,
+        1 => 1,
+        2 => 2,
+        3 => 4,
+        4..=7 => 8,
+        8..=15 => 16,
+        16..=31 => 32,
+        32..=127 => 64,
+        128..=255 => 128,
+    }
```
morehouse, line 369: What is the purpose of these mappings?  It seems like the sequence is chosen to be ascending powers of 2, but wouldn't it be simpler to map to sequential integers (0, 1, 2, 3, ..., 7, 8)?

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,768 @@
...
+        let executed_edges = self.hit.iter().filter(|&&h| h).count() as u64;
+        // Only hit edges can be "executed code edges"; an edge that was 0 in
+        // every execution never fluctuated, so `hit` gating is redundant but
+        // makes the invariant explicit.
+        let fluctuated_edges = self
+            .fluctuated
+            .iter()
+            .zip(&self.hit)
+            .filter(|&(&f, &h)| f && h)
+            .count() as u64;
```
morehouse: Let's just count the fluctuated edges instead of doing the extra no-op check against hit edges?

smitebot/README.md
erickcestari: It's useful for quickly checking performance improvements. I've used it to quickly test JVM optimization because the `executor` checks if the target is live, which does a ping-pong.

smitebot/src/commands/bench_exec.rs
erickcestari: The mapping copies AFL++'s count-class buckets that group raw hit counts. We keep AFL's exact values so our determinism metric matches AFL's stability.

smitebot/src/libnyx.rs
erickcestari: I've dropped the `libloading` since we can use `libc` directly and suggested by @Ashish-Kumar-Dash . I think now the macro makes more sense to exist, since inline it would look worse.

smitebot/src/commands/bench_exec.rs
erickcestari: Good catch!! Fixed now!

erickcestari:
Thanks @morehouse and @Ashish-Kumar-Dash for the review.

smitebot/src/stats.rs
```diff
@@ -0,0 +1,127 @@
...
+
+impl LatencyStats {
+    /// Computes summary statistics, sorting `latencies` in place.
+    pub fn summarize(latencies: &mut [Duration]) -> Self {
+        assert!(!latencies.is_empty(), "latencies must be non-empty");
+        latencies.sort_unstable();
+
+        let sum: Duration = latencies.iter().sum();
+        let count = u32::try_from(latencies.len()).unwrap_or(u32::MAX);
+        let mean = sum / count;
```
Chand-ra: why not use `avg_duration()` here instead of manually calculating the average?

smitebot/src/latency_stats.rs
```diff
@@ -0,0 +1,127 @@
+//! Statistical summaries for benchmark timings.
```
Chand-ra, line 1: Nit: "stats" feels too broad for the filename, why not just use "latency_stats" instead, since that's what we're actually dealing with?

smitebot/src/stats.rs
```diff
@@ -0,0 +1,127 @@
...
+            median: percentile(latencies, 50),
+            p99: percentile(latencies, 99),
+            max: latencies[latencies.len() - 1],
+        }
+    }
+}
+
+/// Returns the value at the given percentile from a sorted slice.
+pub fn percentile(sorted: &[Duration], pct: usize) -> Duration {
+    debug_assert!(!sorted.is_empty());
```
Chand-ra: Since this is an internal benchmarking tool, I think it is better that we fail fast with `assert!` everywhere. Also, let us add:
```
assert!(pct <= 100, "percentile must be <= 100");
```
to prevent logical bugs from callers asking for the 999th percentile.

smitebot/src/latency_stats.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn percentile_nearest_rank() {
+        let mut v: Vec<Duration> = (1..=100).map(ms).collect();
+        v.sort_unstable();
+        assert_eq!(percentile(&v, 50), ms(50));
+        assert_eq!(percentile(&v, 99), ms(99));
+        assert_eq!(percentile(&v, 100), ms(100));
+    }
+
+    #[test]
+    fn percentile_single_element() {
```
Chand-ra, line 88: It should be worthwhile adding a `latency_stats_single_element()` test as well.

smitebot/src/latency_stats.rs
```diff
@@ -0,0 +1,127 @@
...
+        Duration::from_millis(n)
+    }
+
+    #[test]
+    fn percentile_nearest_rank() {
+        let mut v: Vec<Duration> = (1..=100).map(ms).collect();
+        v.sort_unstable();
+        assert_eq!(percentile(&v, 50), ms(50));
+        assert_eq!(percentile(&v, 99), ms(99));
+        assert_eq!(percentile(&v, 100), ms(100));
```
Chand-ra, line 84: \+ `assert_eq!(percentile(&v, 0), ms(1));`

smitebot/src/latency_stats.rs
```diff
@@ -0,0 +1,127 @@
...
+    #[test]
+    fn mean_stddev_computes_sample_stddev() {
+        // For [2, 4, 4, 4, 5, 5, 7, 9]: mean 5, sample stddev ~2.138.
+        let (mean, stddev) = mean_stddev(&[2.0, 4.0, 4.0, 4.0, 5.0, 5.0, 7.0, 9.0]);
+        assert!((mean - 5.0).abs() < 1e-9, "mean was {mean}");
+        assert!((stddev - 2.1380).abs() < 1e-3, "stddev was {stddev}");
+    }
+
+    #[test]
+    fn avg_duration_averages_nanoseconds() {
```
Chand-ra, line 132: Let us add a `avg_duration_prevents_overflow()` test as well.

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,827 @@
+//! Single-input Nyx benchmark.
```
Chand-ra, line 1: The second commit message's header should read `bench-exec` instead of `bench`.

smitebot/README.md
Chand-ra: I think it is worth mentioning in the documentation then: the default input can be used to quickly check performance improvements due to factors like target optimization.

smitebot/README.md
```diff
@@ -53,6 +53,59 @@ After spawning, startup is verified by polling for `fuzzer_stats` files. Because
...
+
+`coverage determinism` reports how reproducible the target's edge coverage is across the identical executions. Each execution's AFL coverage bitmap is captured and its per-edge hit counts are bucketed with AFL's count classes; an edge whose bucketed count is not identical across all executions is counted as fluctuated, and the denominator is the edges hit at least once. High determinism (close to 100%) means the same input reliably reaches the same code; fluctuation points to nondeterminism (uninitialized memory, timing-dependent branches, ASLR, background threads) that adds noise to the fuzzer's coverage signal. Bucketing means benign loop-trip jitter is not counted. Capturing the bitmap is timed separately and excluded from `execs/sec`, so the throughput numbers are unaffected.
+
+With `--repeat` greater than 1 each run is reported separately, followed by an aggregate (mean, stddev, and range of execs/sec across runs; coverage determinism is pooled over all runs).
+
+`non-normal results` counts executions that ended in a crash, timeout, or error rather than a clean run; a non-zero count means the input is not a clean steady-state case and the numbers are skewed.
```
Chand-ra: nit: `non-normal results` doesn't really convey this meaning. What about something like `failed iterations` instead?

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,450 @@
...
+            .into_owned()
+    }
+}
+
+/// Converts a path to a C string, rejecting non-UTF-8 or NUL-containing paths.
+fn path_to_cstring(path: &Path) -> Result<CString, String> {
+    let s = path
+        .to_str()
+        .ok_or_else(|| format!("path is not valid UTF-8: {}", path.display()))?;
+    CString::new(s).map_err(|_| format!("path contains an interior NUL byte: {}", path.display()))
```
Chand-ra: I don't think Linux file paths are required to be UTF-8. What about this instead?
```
use std::os::unix::ffi::OsStrExt;
fn path_to_cstring(path: &Path) -> Result<CString, String> {
    CString::new(path.as_os_str().as_bytes())
        .map_err(|_| format!("path contains an interior NUL byte: {}", path.display()))
}
```

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,827 @@
...
+/// keeps the snapshot's reset region as small as the input allows rather than
+/// AFL++'s 1 MiB default. The size is rounded up to a whole [`INPUT_BUFFER_PAGE`]
+/// (never below one page) because QEMU-Nyx asserts the payload buffer is a page
+/// multiple; the raw input length would trip that assertion. Clamps to
+/// `u32::MAX` for inputs too large to express as a `u32` byte count (well beyond
+/// any realistic seed).
+fn auto_max_input_size(input_len: usize) -> u32 {
+    let pages = input_len.max(1).div_ceil(INPUT_BUFFER_PAGE);
+    let bytes = pages.saturating_mul(INPUT_BUFFER_PAGE);
+    u32::try_from(bytes).unwrap_or(u32::MAX)
```
Chand-ra: Since `u32::MAX` isn't necessarily a multiple of `INPUT_BUFFER_PAGE`, wouldn't the `unwrap_or()` case lead to the same error that we're trying to avoid here? Let us clamp to the highest `u32` value that is a multiple of `INPUT_BUFFER_PAGE`:
```
const MAX_PAGE_ALIGNED_U32 = (u32::MAX / INPUT_BUFFER_PAGE as u32) * INPUT_BUFFER_PAGE as u32
```

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,827 @@
...
+
+    #[test]
+    fn auto_max_input_size_rounds_up_to_a_page() {
+        // Small inputs (including the empty case) get one full page; a page-sized
+        // input stays one page; a byte over rolls to the next page.
+        assert_eq!(auto_max_input_size(0), 4096);
+        assert_eq!(auto_max_input_size(1), 4096);
+        assert_eq!(auto_max_input_size(4096), 4096);
+        assert_eq!(auto_max_input_size(4097), 8192);
+        assert_eq!(auto_max_input_size(9000), 12288);
```
Chand-ra: Let us add an input for the `unwrap_or` case as well.

Chand-ra (changes requested):
GitHub started acting up so I couldn't use the "suggest change" functionality🤦. Looks mostly good to me, following are a couple of suggestions.

smitebot/src/libnyx.rs
erickcestari: Right.  Probably no one will have a non UTF-8 path. But I've applied the change.

Chand-ra (approved):
Took a quick look and verified that all the pointers from the previous feedback were addressed, looks good to me now!

Ashish-Kumar-Dash:
@erickcestari  I pushed some recent commits, you might have to rebase

erickcestari:
> @erickcestari I pushed some recent commits, you might have to rebase

Thanks! I've rebased it now.

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,850 @@
...
+
+/// Guest page size (x86-64 base page, `x86_64_PAGE_SIZE` in QEMU-Nyx) that the
+/// input buffer must be a whole multiple of.
+///
+/// QEMU-Nyx maps the payload buffer into the guest one page at a time and
+/// asserts `shared_payload_buffer_size % x86_64_PAGE_SIZE == 0`
+/// (`nyx_mode/QEMU-Nyx/nyx/memory_access.c`), so any buffer size we pass must be
+/// page-rounded or the VM aborts. AFL++ never trips this because its buffer is a
+/// page-multiple 1 MiB; our auto-sizing from the raw input length must round up.
+const INPUT_BUFFER_PAGE: u32 = 4096;
```
morehouse: Nit:


```suggestion
const NYX_PAGE_SIZE: u32 = 4096;
```

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,850 @@
...
+
+/// Largest `u32` byte count that is still a whole [`INPUT_BUFFER_PAGE`]. Used as
+/// the overflow clamp so the buffer stays page-aligned; `u32::MAX` itself is not
+/// a page multiple and would trip the same QEMU-Nyx assertion we round to avoid.
+const MAX_PAGE_ALIGNED_U32: u32 = (u32::MAX / INPUT_BUFFER_PAGE) * INPUT_BUFFER_PAGE;
```
morehouse: 4GB seems a bit excessive for the input buffer.  Now that we allocate 2GB memory by default to Nyx VMs, we don't even have enough memory to load inputs that large.

Maybe we should just cap the buffer size at 1 MB like AFL++ does?

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,850 @@
...
+    /// aggregate.
+    #[arg(long, short = 'r', default_value_t = 1)]
+    repeat: u32,
+    /// Nyx worker id (also selects the pinned CPU) for the VM.
+    #[arg(long, default_value_t = 0)]
+    worker_id: u32,
+    /// Maximum input buffer size in bytes for the Nyx VM. Defaults to the input
+    /// size rounded up to a 4 KiB page; pass 1048576 to match AFL++'s 1 MiB buffer.
+    #[arg(long)]
+    max_input_size: Option<u32>,
```
morehouse: Is there ever a reason to use this flag now that we automatically calculate the buffer size?

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,850 @@
...
+/// Input-buffer size for an `input_len`-byte benchmark input when
+/// `--max-input-size` is not given.
+///
+/// The benchmark feeds one fixed input, so the buffer only has to hold it, which
+/// keeps the snapshot's reset region as small as the input allows rather than
+/// AFL++'s 1 MiB default. The size is rounded up to a whole [`INPUT_BUFFER_PAGE`]
+/// (never below one page) because QEMU-Nyx asserts the payload buffer is a page
+/// multiple; the raw input length would trip that assertion. Clamps to
+/// [`MAX_PAGE_ALIGNED_U32`] for inputs too large to express as a `u32` byte count
+/// (well beyond any realistic seed), keeping the result page-aligned.
```
morehouse: Rather than clamping, I think we want the user to see an error if the input size is too large.

Probably we should just cap the input length and report an error earlier if it's too large, then we can use `expect` in this function.

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,850 @@
...
+                log::error!("{e}");
+                return false;
+            }
+        };
+        // With no explicit `--max-input-size`, size the VM buffer to the input
+        // itself: a fixed input is the only thing ever fed, so the buffer (and
+        // the pages the snapshot resets) need not be AFL++'s 1 MiB default.
+        let max_input_size = args
+            .max_input_size
+            .unwrap_or_else(|| auto_max_input_size(input.len()));
```
morehouse: A user-provided `max_input_size` isn't validated to be a multiple of the page size.

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,850 @@
...
+    /// Number of full boot+measure runs to repeat and average over.
+    ///
+    /// Each run boots a fresh VM, so repeats capture boot- and snapshot-level
+    /// variance that a single run cannot. Results are reported per run plus an
+    /// aggregate.
+    #[arg(long, short = 'r', default_value_t = 1)]
+    repeat: u32,
+    /// Nyx worker id (also selects the pinned CPU) for the VM.
+    #[arg(long, default_value_t = 0)]
+    worker_id: u32,
```
morehouse: Where is the logic to pin the CPU?  I couldn't find anything in libnyx that does it, so I think we need to do it ourselves like AFL++ does.

This is important for getting consistent benchmark results when there's both P and E cores.

smitebot/Cargo.toml
```diff
@@ -9,6 +9,7 @@ workspace = true
 
 [dependencies]
 clap = { version = "4.6", features = ["derive"] }
+libc = "0.2"
```
morehouse: There's now two crates that depend on libc, so we should make it a workspace dependency.

smitebot/src/commands/bench_exec.rs
morehouse: I understand the bucketing, but the values that the buckets map to seem arbitrary.  We could just as easily map to (0, 1, 2, 3, ..., 8) instead for simplicity.

I don't think it's as confusing anymore since at least with the new bucket bounds the mapped values correspond to the starting count for each bucket.  But with the previous bucket bounds the values were entirely unrelated.

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,850 @@
...
+    }
+
+    // The coverage bitmap lets us measure how deterministic the target's edge
+    // coverage is across identical executions. Skip it if libnyx reports no map.
+    let bitmap_size = vm.bitmap_size();
+    let mut coverage = (bitmap_size > 0).then(|| CoverageTracker::new(bitmap_size));
+    let mut bitmap_buf = vec![0u8; bitmap_size];
+
+    log::info!("running {} timed executions", args.iterations);
+    let cap = usize::try_from(args.iterations).unwrap_or(0);
```
morehouse: An iterations value that doesn't fit in `usize` seems extreme.  Perhaps we should validate `iterations` and then `expect()` here.

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,449 @@
...
+        }
+    }
+
+    /// Runs `input` once through the VM and returns the outcome.
+    ///
+    /// `input` is copied into libnyx's input buffer, so the same slice can be
+    /// executed repeatedly. libnyx treats the buffer as read-only (write
+    /// protection is enabled at boot), but its C signature takes `*mut u8`.
+    pub fn exec(&self, input: &[u8]) -> NyxReturn {
+        let len = u32::try_from(input.len()).unwrap_or(u32::MAX);
```
morehouse: If we validate the input len earlier, we can use `expect()` here.

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,850 @@
...
+    let mut latencies = Vec::with_capacity(cap);
+    let mut target_runtimes = Vec::with_capacity(cap);
+    let mut overheads = Vec::with_capacity(cap);
```
morehouse: This is another reason to bound `iterations`.  We preallocate these vectors.  If the user chooses 100M iterations or more, we end up with several GB of allocated memory here.

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,449 @@
...
+    /// only the target/overhead split is unreliable.
+    #[must_use]
+    pub fn aux_buffer_layout_matches(&self) -> bool {
+        // SAFETY: as in `exec_with_stats`; the header sits at offset 0.
+        unsafe {
+            let aux = (self.lib.get_aux_buffer)(self.process).cast_const();
+            let magic = aux.cast::<u64>().read_unaligned();
+            let version = aux.add(8).cast::<u16>().read_unaligned();
+            magic == aux::MAGIC && version == aux::VERSION
+        }
```
morehouse: Should we also check the `hash` field from the `auxiliary_buffer_header_s` struct?

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,850 @@
...
+    /// Nyx worker id (also selects the pinned CPU) for the VM.
+    #[arg(long, default_value_t = 0)]
+    worker_id: u32,
+    /// Maximum input buffer size in bytes for the Nyx VM. Defaults to the input
+    /// size rounded up to a 4 KiB page; pass 1048576 to match AFL++'s 1 MiB buffer.
+    #[arg(long)]
+    max_input_size: Option<u32>,
+    /// Per-execution timeout in seconds before an exec is counted as timed out.
+    #[arg(long = "timeout", short = 't', default_value_t = BENCH_TIMEOUT_SECS)]
+    timeout_secs: u8,
```
morehouse, line 100: We should validate `timeout` to make sure it's not 0.  Apparently giving a 0 value to Nyx  disables the timing measurements.

smitebot/README.md
```diff
@@ -60,6 +60,59 @@ After spawning, startup is verified by polling for `fuzzer_stats` files. Because
 
 Campaign state is saved to `~/.smitebot/runs/<campaign-id>/state.json` for use by future `stop` and `status` commands.
 
+### smitebot bench-exec
```
morehouse, line 85: Nit: putting it here splits up the `start`, `stop`, and `status` commands.  Please put it after.

smitebot/README.md
```diff
@@ -60,6 +60,59 @@ After spawning, startup is verified by polling for `fuzzer_stats` files. Because
...
+- **nyx overhead** is the wall time minus that guest runtime: snapshot restore/reset plus hypercall round-trips. `dirty pages/exec` is the number of pages restored each execution, the main driver of that overhead, so a change that dirties more memory (e.g. spawning a process per block) shows up as both higher overhead and more dirty pages.
+
+`coverage determinism` reports how reproducible the target's edge coverage is across the identical executions. Each execution's AFL coverage bitmap is captured and its per-edge hit counts are bucketed with AFL's count classes; an edge whose bucketed count is not identical across all executions is counted as fluctuated, and the denominator is the edges hit at least once. High determinism (close to 100%) means the same input reliably reaches the same code; fluctuation points to nondeterminism (uninitialized memory, timing-dependent branches, ASLR, background threads) that adds noise to the fuzzer's coverage signal. Bucketing means benign loop-trip jitter is not counted. Capturing the bitmap is timed separately and excluded from `execs/sec`, so the throughput numbers are unaffected.
```
morehouse: Nit: the bitmap capture isn't timed at all anymore.  I think we can drop the sentence entirely since it isn't related to coverage determinism.


```suggestion
`coverage determinism` reports how reproducible the target's edge coverage is across the identical executions. Each execution's AFL coverage bitmap is captured and its per-edge hit counts are bucketed with AFL's count classes; an edge whose bucketed count is not identical across all executions is counted as fluctuated, and the denominator is the edges hit at least once. High determinism (close to 100%) means the same input reliably reaches the same code; fluctuation points to nondeterminism (uninitialized memory, timing-dependent branches, ASLR, background threads) that adds noise to the fuzzer's coverage signal. Bucketing means benign loop-trip jitter is not counted.
```

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,850 @@
...
+/// (never below one page) because QEMU-Nyx asserts the payload buffer is a page
+/// multiple; the raw input length would trip that assertion. Clamps to
+/// [`MAX_PAGE_ALIGNED_U32`] for inputs too large to express as a `u32` byte count
+/// (well beyond any realistic seed), keeping the result page-aligned.
+fn auto_max_input_size(input_len: usize) -> u32 {
+    let page = INPUT_BUFFER_PAGE as usize;
+    let pages = input_len.max(1).div_ceil(page);
+    let bytes = pages.saturating_mul(page);
+    u32::try_from(bytes).unwrap_or(MAX_PAGE_ALIGNED_U32)
+}
```
morehouse: We actually need to add 4 bytes to `input_len` before rounding up to the next page size, since the payload has its size prepended.

See https://github.com/morehouse/smite/blob/085ed95411329bddd6d21c66574c939caec8117c/smite-nyx-sys/src/nyx.h#L119-L122 and https://github.com/nyx-fuzz/libnyx/blob/f07a41fcf0a01be5f8b8397b26cb914a149b9198/libnyx/src/lib.rs#L368-L373.

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,449 @@
...
+        let ptr = unsafe { libc::dlsym($handle, concat!($name, "\0").as_ptr().cast::<c_char>()) };
+        if ptr.is_null() {
+            return Err(format!("libnyx.so is missing required symbol `{}`", $name));
+        }
+        // SAFETY: `$name` names a function whose C signature the caller vouches
+        // is `$ty`; transmuting its address to that `extern "C"` fn pointer is
+        // the standard `dlsym` idiom (both are a single pointer wide).
+        unsafe { std::mem::transmute::<*mut c_void, $ty>(ptr) }
+    }};
+}
```
morehouse: This would be simpler and more readable as a function.


```suggestion
unsafe fn sym<T: Copy>(handle: *mut c_void, name: &CStr) -> Result<T, String> {
    const { assert!(size_of::<T>() == size_of::<*mut c_void>()) } 
    // SAFETY: `$handle` is a live `dlopen` handle for a genuine libnyx.so, and `name`
    // is NUL-terminated by construction.
    let ptr = unsafe { libc::dlsym(handle, name.as_ptr()) };
    if ptr.is_null() {
        return Err(format!("libnyx.so is missing required symbol `{}`", name.to_string_lossy()));
    }
    // SAFETY: `T` is pointer-sized, and the caller vouches for the signature; transmuting its
    // address is the standard `dlsym` idiom.
    Ok(unsafe { std::mem::transmute_copy(ptr) })
    }};
}
```

Then each call site benefits from type inference as well: `config_load: sym(handle, c"nyx_config_load")?,`

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,850 @@
...
+        args.timeout_secs,
+    )?;
+
+    // The smite Nyx agent takes the snapshot on the first execution, so the
+    // first exec pays for VM boot, target init, and snapshot creation. Time it
+    // separately; the timed loop then measures steady-state restores.
+    log::info!("creating snapshot (first execution)");
+    let snapshot_start = Instant::now();
+    let first = vm.exec(input);
+    let snapshot_time = snapshot_start.elapsed();
```
morehouse: AFAICT this is false.  VM boot, target init, and snapshot creation all already happened in `libnyx.boot` above.  Maybe the first execution is slower due to a cold cache, but it's not due to Nyx boot, JVM warmup, snapshot etc.

If we actually want to know the boot + snapshot time, we need to time `libnyx.boot`.

morehouse (changes requested):

smitebot/src/commands/bench_exec.rs
erickcestari: Actually I'll remove the adaptive buffer size, since when running with AFL++ the value used is always the 1MiB `MAX_FILE`. So I believe it makes more sense to bench in the same way AFL++ runs. Also I'll keep an option to change the max buffer size to test if changing it makes any difference on performance.

smitebot/src/commands/bench_exec.rs
erickcestari: Good catch. I completely forgot about that. I've added by using the libc

smitebot/src/libnyx.rs
erickcestari: Sure. I've added the new check.

smitebot/src/commands/bench_exec.rs
erickcestari: True, I've interpreted wrongly what `nyx_get_fuzz_input` actually does.

erickcestari:
Thanks @morehouse. It was an insightful review. It's ready for another one.

smitebot/src/utils.rs
```diff
@@ -41,11 +44,108 @@ fn find_in_path_with_path(tool: &str, path_var: &OsStr) -> Option<PathBuf> {
...
+    log::info!("Nyx sharedir ready at {}", config.sharedir.display());
+    true
+}
+
+/// Pins the calling process, and every process it later spawns, to `cpu`.
+///
+/// # Errors
+/// Returns an error if `cpu` is out of range or is not an online CPU this
+/// process is allowed to run on.
+pub fn pin_to_cpu(cpu: usize) -> Result<(), String> {
```
morehouse, line 82: This should not be added in the commit labeled "pure refactor".  It also looks like it's introduced before we even add the needed libc dependency.

smitebot/src/utils.rs
```diff
@@ -41,11 +44,108 @@ fn find_in_path_with_path(tool: &str, path_var: &OsStr) -> Option<PathBuf> {
...
+    if cpu >= max {
+        return Err(format!("CPU {cpu} is out of range (0..{max})"));
+    }
+
+    // SAFETY: `set` is zero-initialized before use, `CPU_SET` writes within it
+    // given the bound check above, and `sched_setaffinity` reads exactly
+    // `size_of::<cpu_set_t>()` bytes from the pointer. A pid of 0 means the
+    // calling thread.
+    let failed = unsafe {
+        let mut set: libc::cpu_set_t = std::mem::zeroed();
```
morehouse, line 95: I think we should probably use `CPU_ZERO` rather than assuming a manual zero-init does the same.

smitebot/src/utils.rs
```diff
@@ -41,11 +44,108 @@ fn find_in_path_with_path(tool: &str, path_var: &OsStr) -> Option<PathBuf> {
...
+    if !status.success() {
+        log::error!("setup-nyx.sh failed with {status}");
+        return false;
+    }
+
+    log::info!("Nyx sharedir ready at {}", config.sharedir.display());
+    true
+}
+
+/// Pins the calling process, and every process it later spawns, to `cpu`.
```
morehouse:
```suggestion
/// Pins the calling thread, and every process it later spawns, to `cpu`.
```

smitebot/src/utils.rs
```diff
@@ -41,11 +44,108 @@ fn find_in_path_with_path(tool: &str, path_var: &OsStr) -> Option<PathBuf> {
...
+        assert!(pin_to_cpu(libc::CPU_SETSIZE as usize).is_err());
+        assert!(pin_to_cpu(usize::MAX).is_err());
+    }
+
+    #[test]
+    fn pin_to_cpu_binds_the_calling_thread() {
+        // Affinity is per-thread, and this runs on the test's own thread, so
+        // pinning here leaves the rest of the suite alone. Pin to a CPU the
+        // process is already allowed on so a restricted cpuset does not fail it.
+        let mut set: libc::cpu_set_t = unsafe { std::mem::zeroed() };
```
morehouse, line 128: Should also be using `CPU_ZERO` in this test.

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,985 @@
...
+const MAX_REPEAT: u32 = 1_000;
+
+/// Guest page size (x86-64 base page, `x86_64_PAGE_SIZE` in QEMU-Nyx) that the
+/// input buffer must be a whole multiple of.
+///
+/// QEMU-Nyx maps the payload buffer into the guest one page at a time and
+/// asserts `shared_payload_buffer_size % x86_64_PAGE_SIZE == 0`
+/// (`nyx_mode/QEMU-Nyx/nyx/memory_access.c`), so any buffer size we pass must be
+/// page-rounded or the VM aborts. AFL++ never trips this because its buffer is a
+/// page-multiple 1 MiB; a `--max-input-size` that is not is rejected up front.
```
morehouse: Not sure what the last part is about.


```suggestion
/// page-multiple 1 MiB.
```

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,486 @@
...
+            result,
+            target_runtime,
+            dirty_pages,
+        }
+    }
+
+    /// Checks that the auxiliary buffer has the magic and version this binding's
+    /// field offsets were derived from, so [`ExecStats`] readings are trustworthy.
+    /// Returns `false` (rather than erroring) if not: benchmarking can still run,
+    /// only the target/overhead split is unreliable.
```
morehouse: This seems like it would be confusing to the user.  If we're reading garbage data, we should fail loudly rather than expecting the user to figure out that the data is garbage.

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,985 @@
...
+    /// give materially different numbers.
+    #[arg(long)]
+    cpu: Option<usize>,
+    /// Input buffer size in bytes for the Nyx VM; must be a multiple of 4096.
+    ///
+    /// Defaults to AFL++'s 1 MiB, the buffer a real campaign runs with. Lower it
+    /// to measure what a smaller buffer buys: the guest allocates the whole
+    /// buffer on every execution and the snapshot resets every page of it
+    /// dirtied, so buffer size shows up directly in the Nyx overhead and dirty
+    /// pages per exec.
```
morehouse: I don't think this logic checks out.  Assume the input size is `X` pages and the buffer size is `Y > X` pages.  Then regardless of what `Y` is, only `X` pages ever get dirtied and need to be reset.

Have you actually observed differences in performance with varying buffer sizes?

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,985 @@
...
+                return false;
+            }
+        };
+        // A zero timeout reaches Nyx as "no timeout" rather than "expire
+        // immediately", which silently turns off the timing the benchmark exists
+        // to measure.
+        if args.timeout_secs == 0 {
+            log::error!("--timeout must be at least 1 second; Nyx reads zero as no timeout");
+            return false;
+        }
```
morehouse: This is confusing to the user, who may very well want "no timeout" (a.k.a. super high timeout).  The actual issue is that we lose benchmark timings .


```suggestion
        // A zero timeout causes Nyx to silently turn off the timing the
        // benchmark exists to measure.
        if args.timeout_secs == 0 {
            log::error!("Nyx requires --timeout to be at least 1 second");
            return false;
        }
```

smitebot/src/commands/bench_exec.rs
```diff
@@ -0,0 +1,985 @@
...
+
+        // Pin before booting: libnyx spawns QEMU as a child, so the VM inherits
+        // this affinity mask. Doing it here also pins the timing loop itself.
+        let cpu = args.cpu.unwrap_or(args.worker_id as usize);
+        if let Err(e) = pin_to_cpu(cpu) {
+            log::error!("{e}");
+            return false;
+        }
+        log::info!("pinned to CPU {cpu}");
```
morehouse: Perhaps we should pin closer to where we actually start benchmarking.  Currently we still load and validate a whole bunch of stuff after pinning.

smitebot/README.md
```diff
@@ -82,6 +82,60 @@ smitebot status <campaign-id> --summary
...
+  boot + snapshot creation: 20.822 s
+
+  steady-state (snapshot restore + target run):
+    execs/sec: 1234.5
+    exec time: 810.20 ms
+    latency:   min 0.7 µs  mean 0.8 µs  median 0.8 µs  p99 1.2 µs  max 3.4 ms
+    input execution: mean 0.5 µs  median 0.5 µs   (guest runtime)
+    nyx overhead:    mean 0.3 µs  median 0.3 µs   (restore + reset + ipc; 42 dirty pages/exec)
+    failed iterations: 0 / 1000
+    coverage determinism: 98.4% stable (12 / 750 edges fluctuated)
```
morehouse: These stats are inconsistent.  exec time 810 ms implies 810 us latency per iteration.

Let's regenerate with data from a real run.

smitebot/src/commands/bench_exec.rs
erickcestari: I'll provide a benchmark later. Currently I'm in a laptop without the charger which would result in unreliable benchmark.

smitebot/src/libnyx.rs
erickcestari: I agree. I've added assertions instead of logging it once.

smitebot/src/utils.rs
erickcestari: Fixed

smitebot/src/commands/bench_exec.rs
erickcestari: Actually, that's not true at all. When I ran it, it ended up using an E-core instead of a P-core because I hadn't pinned the CPU. It produced a different result, which led me to think the difference was caused by `max-input-size`.

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,510 @@
...
+    /// Coverage bitmap size, fixed once the VM is booted. Cached so hot loops
+    /// don't re-cross the FFI boundary for a constant.
+    bitmap_size: usize,
+    /// Input buffer size the VM was booted with; the largest input it can run.
+    input_buffer_size: u32,
+}
+
+impl NyxVm<'_> {
+    /// Enables or disables snapshot reload before each execution.
+    pub fn set_reload_mode(&self, enable: bool) {
```
morehouse: I think this technically mutates `self` through the raw `process` pointer


```suggestion
    pub fn set_reload_mode(&mut self, enable: bool) {
```

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,510 @@
...
+        // SAFETY: `process` is a live handle; option changes require a following
+        // `option_apply` to take effect.
+        unsafe {
+            (self.lib.option_set_reload_mode)(self.process, enable);
+            (self.lib.option_apply)(self.process);
+        }
+    }
+
+    /// Sets the per-execution timeout.
+    pub fn set_timeout(&self, seconds: u8, micros: u32) {
```
morehouse:
```suggestion
    pub fn set_timeout(&mut self, seconds: u8, micros: u32) {
```

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,510 @@
...
+    /// executed repeatedly. libnyx treats the buffer as read-only (write
+    /// protection is enabled at boot), but its C signature takes `*mut u8`.
+    ///
+    /// # Panics
+    ///
+    /// If `input` does not fit the buffer the VM was booted with, alongside the
+    /// length prefix libnyx stores in front of it. Callers size that buffer from
+    /// the input, so an input that does not fit is a bug in that validation
+    /// rather than a runtime condition.
+    pub fn exec(&self, input: &[u8]) -> NyxReturn {
```
morehouse:
```suggestion
    pub fn exec(&mut self, input: &[u8]) -> NyxReturn {
```

smitebot/src/libnyx.rs
morehouse:
```suggestion
    pub fn set_timeout(&mut self, seconds: u8, micros: u32) {
```

smitebot/src/commands/start.rs
```diff
@@ -16,6 +16,7 @@ use crate::commands::build::{BuildInputs, run_build};
 use crate::config::CampaignConfig;
 use crate::state::{CampaignState, RunnerState, Status};
 use crate::tmux;
+use crate::utils::setup_nyx;
 use crate::utils::shell_quote;
```
morehouse: Nit:


```suggestion
use crate::utils::{setup_nyx, shell_quote};
```

smitebot/src/libnyx.rs
```diff
@@ -0,0 +1,510 @@
...
+/// section (`auxilary_buffer_result_s`) starts at byte 896. Within it, the
+/// `#[repr(C, packed)]` fields we care about sit at fixed offsets. QEMU-Nyx
+/// rewrites these on every execution, so they describe the exec that just ran.
+/// See `nyx_mode/libnyx/fuzz_runner/src/nyx/aux_buffer.rs` in AFL++.
+mod aux {
+    /// `0x54502d554d4551` ("QEMU-PT" little-endian), at buffer offset 0.
+    pub const MAGIC: u64 = 0x0054_502d_554d_4551;
+    /// Aux-buffer format version this binding's offsets were derived from.
+    pub const VERSION: u16 = 3;
+    // Hash defined on libnyx
```
morehouse:
```suggestion
    /// Hash defined on libnyx
```

morehouse (approved):

erickcestari:
@morehouse this should be ready for merge! Thanks for the review!

PR #162, opened 2026-07-23
smite-ir: add SendShutdown operation

ekzyis:
This implements the `SendShutdown` operation for #98.

### DIFF
--- a/smite-ir/src/mutators/operation_param.rs
+++ b/smite-ir/src/mutators/operation_param.rs
@@ -113,6 +113,7 @@ fn mutate_operation(op: &mut Operation, rng: &mut impl Rng) -> bool {
         | Operation::SendMessage
         | Operation::SendOpenChannel
         | Operation::SendFundingCreated
+        | Operation::SendShutdown
         | Operation::RecvAcceptChannel
         | Operation::RecvFundingSigned
         | Operation::RecvChannelReady
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -207,6 +207,13 @@ pub enum Operation {
         /// If `false`, the TLV is omitted and input 2 is ignored.
         include_alias: bool,
     },
+    /// Build and send a `shutdown` message (BOLT 2, type 38).
+    /// Produces a `SentShutdown` variable.
+    ///
+    /// Inputs (2):
+    ///   0: `channel_id`   (`ChannelId`)
+    ///   1: `scriptpubkey` (`Bytes`)
+    SendShutdown,
     /// Receive and parse an `accept_channel` response.
     /// Produces an `AcceptChannel` compound variable.
     RecvAcceptChannel,
@@ -702,6 +709,7 @@ impl fmt::Display for Operation {
             Self::SendChannelReady { include_alias } => {
                 write!(f, "SendChannelReady{{include_alias={include_alias}}}")
             }
+            Self::SendShutdown => write!(f, "SendShutdown"),
             Self::RecvAcceptChannel => write!(f, "RecvAcceptChannel"),
             Self::RecvFundingSigned => write!(f, "RecvFundingSigned"),
             Self::RecvChannelReady => write!(f, "RecvChannelReady()"),
@@ -747,6 +755,7 @@ impl Operation {
             | Self::BroadcastTransaction => None,
             Self::SendOpenChannel => Some(VariableType::SentOpenChannel),
             Self::SendFundingCreated => Some(VariableType::SentFundingCreated),
+            Self::SendShutdown => Some(VariableType::SentShutdown),
             Self::RecvAcceptChannel => Some(VariableType::AcceptChannel),
         }
     }
@@ -794,6 +803,10 @@ impl Operation {
                 VariableType::ChannelId,      // channel_id
                 VariableType::Point,          // second_per_commitment_point
                 VariableType::ShortChannelId, // short_channel_id (alias)
+            ],
+            Self::SendShutdown => vec![
+                VariableType::ChannelId, // channel_id
+                VariableType::Bytes,     // scriptpubkey
             ],
             Self::RecvAcceptChannel => vec![VariableType::SentOpenChannel],
             Self::RecvFundingSigned => vec![VariableType::SentFundingCreated],
@@ -904,6 +917,7 @@ impl Operation {
             | Self::SendOpenChannel
             | Self::SendFundingCreated
             | Self::SendChannelReady { .. }
+            | Self::SendShutdown
             | Self::RecvFundingSigned
             | Self::RecvChannelReady
             | Self::MineBlocks(_)
@@ -926,6 +940,7 @@ impl Operation {
             | Self::SendOpenChannel
             | Self::SendFundingCreated
             | Self::SendChannelReady { .. }
+            | Self::SendShutdown
             | Self::RecvAcceptChannel
             | Self::RecvFundingSigned
             | Self::RecvChannelReady
@@ -994,6 +1009,7 @@ impl Operation {
             | Self::SendMessage
             | Self::SendOpenChannel
             | Self::SendFundingCreated
+            | Self::SendShutdown
             | Self::RecvAcceptChannel
             | Self::RecvFundingSigned
             | Self::RecvChannelReady
--- a/smite-ir/src/tests.rs
+++ b/smite-ir/src/tests.rs
@@ -553,6 +553,40 @@ fn display_send_and_recv_channel_ready_program() {
 }
 
 #[test]
+fn display_send_shutdown_program() {
+    let instructions = vec![
+        Instruction {
+            operation: Operation::LoadChannelId([0xcd; 32]),
+            inputs: vec![],
+        },
+        Instruction {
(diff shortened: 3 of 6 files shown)


smite-scenarios/src/executor.rs
```diff
@@ -3434,6 +3453,90 @@ mod tests {
...
+        let sd = match Message::decode(&executor.conn.sent[0]).expect("valid message") {
+            Message::Shutdown(sd) => sd,
+            other => panic!("expected Shutdown, got type {}", other.msg_type()),
+        };
+        assert_eq!(sd.channel_id, channel_id);
+        assert_eq!(sd.scriptpubkey, scriptpubkey);
+    }
+
+    #[test]
+    fn execute_send_shutdown_empty_scriptpubkey() {
```
ekzyis, line 3740: I used `SendChannelReady` as a reference, but diverged from it when writing this test for an empty, protocol-invalid scriptpubkey in the `shutdown` message. In `execute_send_channel_ready`, we test two variations (with and without alias). I preferred to write two separate tests so it's more clear just by looking at the names of our test cases what we've covered. The downside is that this duplicates most of the code from the other test.

smite-scenarios/src/executor.rs
```diff
@@ -939,6 +951,13 @@ fn build_channel_ready(
     }
 }
 
+/// Builds a `Shutdown` from 2 input variables (wire order).
```
ekzyis: nit: I feel like it should say "`Shutdown` message", but then the comment would be inconsistent with the other ones.

erickcestari (approved):
ACK 187f6aa2c8842abd1e62e53df70b7c59d709aae0

smite-ir/src/mutators/operation_param.rs
```diff
@@ -116,6 +116,7 @@ fn mutate_operation(op: &mut Operation, rng: &mut impl Rng) -> bool {
         | Operation::RecvAcceptChannel
         | Operation::RecvFundingSigned
         | Operation::RecvChannelReady
+        | Operation::SendShutdown
```
NishantBansal2003: I think we bundled all the Send and Recv operations separately, so I think this should go after `SendFundingCreated / SendChannelReady`

Same applies to all the other places as well

smite-ir/src/operation.rs
```diff
@@ -221,6 +221,8 @@ pub enum Operation {
     /// point unknown) and its funding transaction has enough confirmations for
     /// the target to have sent `channel_ready`.
     RecvChannelReady,
+    /// Build and send a `shutdown` message (BOLT 2, type 38).
+    SendShutdown,
```
NishantBansal2003: Needs an `Inputs` comment, just like the other `Send*` operations, It's pretty handy when selecting inputs in tests or generators

smite-ir/src/operation.rs
```diff
@@ -714,6 +717,7 @@ impl Operation {
             Self::SendMessage
             | Self::SendChannelReady { .. }
             | Self::RecvChannelReady
+            | Self::SendShutdown
```
NishantBansal2003: I think Erick already flagged this, but documenting it here as well: `SendShutdown` should return an affine type, `SentShutdown`, which is then consumed by subsequent `RecvShutdown` operation to avoid impossible Recv cases

smite-ir/src/operation.rs
```diff
@@ -766,6 +770,7 @@ impl Operation {
                 VariableType::Point,          // second_per_commitment_point
                 VariableType::ShortChannelId, // short_channel_id (alias)
             ],
+            Self::SendShutdown => vec![VariableType::ChannelId, VariableType::Bytes],
```
NishantBansal2003:
```suggestion
            Self::SendShutdown => vec![
                VariableType::ChannelId, // channel_id
                VariableType::Bytes,     // scriptpubkey
            ],
```

smite-ir/src/tests.rs
```diff
@@ -552,6 +552,39 @@ fn display_send_and_recv_channel_ready_program() {
...
+fn display_send_shutdown_program() {
+    let instructions = vec![
+        Instruction {
+            operation: Operation::LoadChannelId([0xcd; 32]),
+            inputs: vec![],
+        },
+        Instruction {
+            operation: Operation::LoadBytes(vec![0x00, 0x14, 0xab, 0xcd]),
+            inputs: vec![],
+        },
```
NishantBansal2003, line 565: I think we will use a valid spk in the generator, just like we do for `open_channel`, so we should follow the same pattern here as well, see: `display_open_channel_program`

smite-scenarios/src/executor.rs
NishantBansal2003: I think we can add `message` since it's used in many places (though not all), e.g: `build_announcement_signatures`, `build_funding_created`

smite-scenarios/src/executor.rs
```diff
@@ -3434,6 +3453,90 @@ mod tests {
         );
     }
 
+    #[test]
+    fn execute_send_shutdown() {
+        let channel_id = ChannelId::new([0x7a; 32]);
+
+        // random p2wpkh
+        let mut scriptpubkey = vec![0x00, 0x14];
+        scriptpubkey.extend_from_slice(&[0xab; 20]);
```
NishantBansal2003: Could use: `ShutdownScriptVariant / LoadShutdownScript` here

smite-scenarios/src/executor.rs
```diff
@@ -3434,6 +3453,90 @@ mod tests {
...
+        let program = Program {
+            instructions: vec![
+                Instruction {
+                    operation: Operation::LoadChannelId([0x7a; 32]),
+                    inputs: vec![],
+                },
+                Instruction {
+                    // the fuzzer should allow empty scriptpubkey in the shutdown message to test
+                    // target's behavior even though it's protocol-invalid
+                    operation: Operation::LoadBytes(vec![]),
```
NishantBansal2003: Could use: `Operation::LoadShutdownScript(ShutdownScriptVariant::Empty)` here

smite-scenarios/src/executor.rs
NishantBansal2003: I think we can refactor the common setup into a helper function to reduce redundancy? similar to what we did in `recv_channel_ready_executor`

NishantBansal2003 (changes requested):

smite-scenarios/src/executor.rs
ekzyis: Good call! Done in 82978d6

smite-scenarios/src/executor.rs
ekzyis: Done in 82978d6

smite-ir/src/mutators/operation_param.rs
ekzyis: Done in 82978d6

smite-ir/src/operation.rs
ekzyis: Done in 82978d6

smite-ir/src/operation.rs
ekzyis: Done in 82978d6

smite-ir/src/operation.rs
ekzyis: Done in 82978d6

smite-ir/src/tests.rs
ekzyis: Done in 82978d6, it now uses `LoadShutdownScript`

smite-scenarios/src/executor.rs
ekzyis: Done in 82978d6

smite-scenarios/src/executor.rs
ekzyis: Done in 82978d6

smite-scenarios/src/executor.rs
```diff
@@ -466,6 +471,18 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                     None
                 }
 
+                Operation::SendShutdown => {
```
morehouse: Nit: this block should come after the `SendChannelReady` block.

smite-scenarios/src/executor.rs
```diff
@@ -3442,6 +3454,59 @@ mod tests {
         );
     }
 
+    fn send_shutdown_executor(channel_id: ChannelId, script: ShutdownScriptVariant) -> Shutdown {
```
morehouse: The name of this helper implies that it returns an executor similar to `recv_channel_ready_executor`.  I think we should rename this to something clearer, or just get rid of it entirely and move the instructions and executor runs inline like we do for most other tests.  For these two tests it would be a small duplication but would make it easier to read the tests.

smite-scenarios/src/executor.rs
ekzyis: whoops, done in 4345df1

smite-scenarios/src/executor.rs
ekzyis: I split it into two separate tests again in 4345df1. While writing tests for #163, I might find a better pattern to share between the tests.

morehouse (approved):
LGTM

PR #163, opened 2026-07-24
smite-ir: add RecvShutdown operation

ekzyis:
This implements the `RecvShutdown` operation for https://github.com/morehouse/smite/issues/98.

`RecvShutdown` consumes the `SentShutdown` of a previous `SendShutdown`, which now carries the `shutdown` we sent, and waits for the target's `shutdown` in reply. It returns the target's scriptpubkey, or empty bytes if no `shutdown` was received.

The reply is checked by a new `ShutdownOracle`. It flags the target if:

* the reply is for a channel we never established
* we sent an invalid signature, so the target should have failed the channel instead of replying
* our scriptpubkey doesn't match the `upfront_shutdown_script` we committed to and `option_upfront_shutdown_script` was negotiated, so the target must fail the connection instead of replying
* its scriptpubkey doesn't match the `upfront_shutdown_script` it committed to, or isn't a standard form for the negotiated features

If our own scriptpubkey isn't standard, BOLT 2 says the target should send a warning instead of replying. If it breaks our `upfront_shutdown_script`, the target may send a warning before failing the connection. In both cases we accept a warning for the channel as a reply.

A `shutdown` for another tracked channel may answer one we sent there earlier, which we can't check against the `shutdown` we sent here, so it ends the program as an unexpected message.

`RecvShutdown` is a no-op if we don't track the channel or the target already replied. Before the target sent `channel_ready`, it may choose not to reply, but we still expect one: LDK always replies, and a target doesn't only costs us a receive timeout, which isn't reported as a violation.

### DIFF
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -221,6 +221,8 @@ pub enum Operation {
     /// point unknown) and its funding transaction has enough confirmations for
     /// the target to have sent `channel_ready`.
     RecvChannelReady,
+    /// Receive and parse a `shutdown` message.
+    RecvShutdown,
     /// Mines the given number of blocks on the Bitcoin network.
     MineBlocks(u8),
     /// Sign wallet inputs of the transaction and broadcast it via `bitcoin-cli`.
@@ -679,6 +681,7 @@ impl fmt::Display for Operation {
             Self::RecvAcceptChannel => write!(f, "RecvAcceptChannel"),
             Self::RecvFundingSigned => write!(f, "RecvFundingSigned"),
             Self::RecvChannelReady => write!(f, "RecvChannelReady()"),
+            Self::RecvShutdown => write!(f, "RecvShutdown()"),
             Self::BroadcastTransaction => write!(f, "BroadcastTransaction"),
         }
     }
@@ -714,6 +717,7 @@ impl Operation {
             Self::SendMessage
             | Self::SendChannelReady { .. }
             | Self::RecvChannelReady
+            | Self::RecvShutdown
             | Self::MineBlocks(_)
             | Self::BroadcastTransaction => None,
             Self::SendOpenChannel => Some(VariableType::SentOpenChannel),
@@ -744,6 +748,7 @@ impl Operation {
             | Self::LoadTargetPubkeyFromContext
             | Self::LoadChainHashFromContext
             | Self::RecvChannelReady
+            | Self::RecvShutdown
             | Self::MineBlocks(_) => vec![],
 
             Self::DerivePoint => vec![VariableType::PrivateKey],
@@ -875,6 +880,7 @@ impl Operation {
             | Self::SendChannelReady { .. }
             | Self::RecvFundingSigned
             | Self::RecvChannelReady
+            | Self::RecvShutdown
             | Self::MineBlocks(_)
             | Self::BroadcastTransaction => vec![],
 
@@ -897,6 +903,7 @@ impl Operation {
             | Self::RecvAcceptChannel
             | Self::RecvFundingSigned
             | Self::RecvChannelReady
+            | Self::RecvShutdown
             | Self::MineBlocks(_)
             | Self::CreateFundingTransaction
             | Self::BroadcastTransaction => true,
@@ -964,6 +971,7 @@ impl Operation {
             | Self::RecvAcceptChannel
             | Self::RecvFundingSigned
             | Self::RecvChannelReady
+            | Self::RecvShutdown
             | Self::BroadcastTransaction => false,
         }
     }
--- a/smite-ir/src/tests.rs
+++ b/smite-ir/src/tests.rs
@@ -553,6 +553,21 @@ fn display_send_and_recv_channel_ready_program() {
 }
 
 #[test]
+fn display_recv_shutdown_program() {
+    let instructions = vec![Instruction {
+        operation: Operation::RecvShutdown,
+        inputs: vec![],
+    }];
+
+    let program = Program { instructions };
+    let text = program.to_string();
+    let lines: Vec<&str> = text.lines().collect();
+
+    let expected = vec!["RecvShutdown()"];
+    assert_eq!(lines, expected);
+}
+
+#[test]
 fn postcard_roundtrip() {
     let program = Program {
         instructions: vec![
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -10,7 +10,7 @@ use smite::bitcoin::{BitcoinCli, Utxo};
 use smite::bolt::{
     AcceptChannel, AnnouncementSignatures, ChannelAnnouncement, ChannelId, ChannelReady,
     ChannelReadyTlvs, ChannelUpdate, FundingCreated, FundingSigned, Message, NodeAnnouncement,
-    OpenChannel, OpenChannelTlvs, Pong, ShortChannelId, msg_type,
+    OpenChannel, OpenChannelTlvs, Pong, ShortChannelId, Shutdown, msg_type,
 };
 use smite::channel_tx::{
     ChannelConfig, ChannelPartyConfig, ChannelState, FundingTransaction, HolderIdentity, Side,
@@ -468,6 +468,13 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                         recv_channel_ready(&mut self.conn, &mut self.channel_states)?;
                         log::debug!("[{:?}] RecvChannelReady: received", start.elapsed());
                     }
+                    None
+                }
+
+                Operation::RecvShutdown => {
+                    log::debug!("[{:?}] RecvShutdown: waiting", start.elapsed());
+                    recv_shutdown(&mut self.conn)?;
+                    log::debug!("[{:?}] RecvShutdown: received", start.elapsed());
                     None
                 }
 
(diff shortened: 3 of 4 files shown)


smite-scenarios/src/executor.rs
```diff
@@ -1170,6 +1177,17 @@ fn recv_funding_signed(conn: &mut impl Connection) -> Result<FundingSigned, Exec
...
+fn recv_shutdown(conn: &mut impl Connection) -> Result<Shutdown, ExecuteError> {
+    match recv_non_ping(conn, RECV_IDLE_TIMEOUT)? {
+        Message::Shutdown(sd) => Ok(sd),
+        other => Err(ExecuteError::UnexpectedMessage {
+            expected: msg_type::SHUTDOWN,
+            got: other.msg_type(),
+        }),
+    }
+}
+
```
erickcestari, line 1297: Should `RecvShutdown` take an affine `SentShutdown` input?

Right now it has no inputs, so nothing stops a mutator from orphaning it (e.g. `InstructionDeleteMutator` dropping the paired send), leaving a blocking recv for a message the target never sends that results on `RECV_IDLE_TIMEOUT` on every run.

If we model self-initiated cooperative close (`SendShutdown` -> `RecvShutdown`), an affine `SentShutdown` couples the pair through the existing delete/reorder guards and makes orphan waits unrepresentable, same as `SentOpenChannel`.

erickcestari (changes requested):

smite-scenarios/src/executor.rs
ekzyis: I had the same question. *In theory*, the peer could send `shutdown` at any time, so it's not necessarily a response (which is why I also use "message" instead of "response" at operation.rs:224).

However, I agree that, for a self-initiated cooperative close, it makes sense to use affine inputs for the reasons you mentioned. I just wasn't sure whether that would limit us too much in the future, so I thought it would be better to discuss the more general problem of messages that can be either responses or unsolicited messages now. But since I can't think of a scenario where we would expect the peer to send an unsolicited `shutdown` and would therefore want to use `RecvShutdown` without an input, I'm going to accept your suggestion (will push the code later).

erickcestari: We could add an `Operation` that triggers the target to initiate a cooperative close with us. I don't think any Lightning implementation will initiate a cooperative close on its own without an external request (e.g., an RPC call), so this operation could be paired with an affine type to model that behavior.

The same idea applies to channel opening. Instead of always acting as the `funder`, we could add an operation that triggers the target to open a channel with us, allowing us to act as the `fundee`. This would exercise additional protocol paths and cover a broader set of channel states.

I think this is not at the current IR design #5, but it could be a great addition to it.

cc @morehouse

ekzyis: I also think this could be a great addition!

smite-ir/src/tests.rs
```diff
@@ -552,6 +552,21 @@ fn display_send_and_recv_channel_ready_program() {
     }
 }
 
+#[test]
+fn display_recv_shutdown_program() {
```
NishantBansal2003: Once rebased, I think we can merge this with `display_send_shutdown_program`, similar to `displays_send_funding_created_recv_funding_signed_program`

smite-scenarios/src/executor.rs
```diff
@@ -3434,6 +3452,36 @@ mod tests {
         );
     }
 
+    #[test]
+    fn execute_recv_shutdown() {
+        let channel_id = ChannelId::new([0x7a; 32]);
+
+        let mut scriptpubkey = vec![0x00, 0x14];
+        scriptpubkey.extend_from_slice(&[0xab; 20]);
```
NishantBansal2003: Could use: `ShutdownScriptVariant` here

smite-ir/src/operation.rs
```diff
@@ -714,6 +717,7 @@ impl Operation {
             Self::SendMessage
             | Self::SendChannelReady { .. }
             | Self::RecvChannelReady
+            | Self::RecvShutdown
```
NishantBansal2003, line 794: I think this should output the spk so it can be used in subsequent channel closing messages, otherwise, we will always be rejected in those cases later

smite-scenarios/src/executor.rs
```diff
@@ -1170,6 +1177,17 @@ fn recv_funding_signed(conn: &mut impl Connection) -> Result<FundingSigned, Exec
     }
 }
 
+/// Receives and decodes a `shutdown` message.
+fn recv_shutdown(conn: &mut impl Connection) -> Result<Shutdown, ExecuteError> {
```
NishantBansal2003, line 1288: Nit: Could move this below `recv_channel_ready` for consistency

NishantBansal2003 (changes requested):
I think this needs to be rebased on top of https://github.com/morehouse/smite/pull/162 to consume the affine type `SentShutdown`

smite-scenarios/src/executor.rs
morehouse: I agree we should consume an affine variable to ensure we never execute a `RecvShutdown` before a corresponding `SendShutdown` has been sent.

In the short-medium term I think we should prioritize flows that are initiated by us (e.g., we send the first `shutdown`), since bugs found that way are generally more severe (an attacker can more easily trigger them).

Target-initiated flows can still have bad bugs, but since they're less exploitable I think that work should be lower-priority.

morehouse (changes requested):
I think we want the affine type wired in for sure.

But that probably isn't enough -- we can't expect the target to send us a `shutdown` message until all HTLCs have been resolved as well.  There's also probably a case where we could send two `shutdown` messages and the target may ignore any of them after the first one.

I think for the HTLCs we can't really implement that part until #111 is implemented, so we can just add a TODO for that.

For the duplicate `shutdown` case we could probably add a flag to the channel state that indicates whether the peer has already responded to the first `shutdown`, and if they have then any subsequent `RecvShutdown` message becomes a no-op (similar to `RecvChannelReady`).

smite-ir/src/operation.rs
ekzyis: done in 77cf4f8

smite-scenarios/src/executor.rs
ekzyis: done in 77cf4f8

smite-scenarios/src/executor.rs
ekzyis: done in 77cf4f8

smite-ir/src/tests.rs
ekzyis: done in 77cf4f8

ekzyis:
> I think for the HTLCs we can't really implement that part until https://github.com/lnfuzz/smite/issues/111 is implemented, so we can just add a TODO for that.

Added TODOs in 77cf4f8. I thought it would be useful to group all HTLC TODOs together with `TODO(htlc)` as the prefix, so I updated the existing HTLC TODOs in c9a2c60:

```
$ git grep "TODO(htlc)"
smite-scenarios/src/executor.rs:                    // TODO(htlc): we only expect `shutdown` when all HTLCs are resolved, else this
smite/src/channel_tx/commitment.rs:    // TODO(htlc): When adding HTLC support, store pending HTLCs (offered/received) for both sides
smite/src/channel_tx/commitment.rs:    // TODO(htlc): When adding HTLC support, add `get_next_commitment_state` to build the next
smite/src/oracles/shutdown.rs:        // TODO(htlc): BOLT 2 forbids sending `shutdown` while HTLCs are still pending
```

---

> For the duplicate shutdown case we could probably add a flag to the channel state that indicates whether the peer has already responded to the first shutdown

We now check in 77cf4f8 if we expect a shutdown on _any_ channel. Only then we wait for a shutdown response. We don't know on which channel we will receive a shutdown, but I could change `SentShutdown` to `SentShutdown(ChannelId)`, then we would know (just realized while typing this).

TODO:

- [x] use `SentShutdown(ChannelId)` and check `peer_shutdown_received` for that channel in `is_shutdown_expected`

---

Mhh, I wonder if there's a conflict between this (https://github.com/lnfuzz/smite/pull/163#discussion_r3652211529):

> I think this should output the spk so it can be used in subsequent channel closing messages, otherwise, we will always be rejected in those cases later

and this (https://github.com/lnfuzz/smite/pull/163#pullrequestreview-4789346842):

> if they have then any subsequent RecvShutdown message becomes a no-op (similar to RecvChannelReady).

In 77cf4f8, `RecvShutdown` returns empty bytes as a sentinel similar to `LookupShortChannelId` when no shutdown is expected on any channel. It's clearly an invalid shutdown script. Could this be a problem for downstream consumers of its return value, who might expect a valid shutdown script? Should they then also be no-ops, at least in the case of `RecvShutdown`?

ekzyis:
c9a2c60...d083969:

* `SentShutdown(ChannelId)` instead of `SentShutdown` so we can check if we expect a shutdown for the specific channel.
* Consider `upfront_shutdown_script` in `ShutdownOracle`. It doesn't check if it's standard as it should be according to BOLT 2, since I think this check should be added to `AcceptChannelOracle`.
* Updated commit message.

I forgot tests for `ShutdownOracle` like for `AcceptChannelOracle`. Will add them.

ekzyis:
This is ready for review now. I'm working on a new PR with a `ShutdownGenerator` so we can test `RecvShutdown` against the targets.

morehouse:
This PR has grown a bit.

@ekzyis would it be possible to split this into smaller PRs?  Perhaps something like this:

- PR 1: `RecvShutdown` operation
- PR 2: initial shutdown oracle
- PR 3: check `upfront_shutdown_script` on shutdown

ekzyis:
fd0bbbb: rebased on master (dc2ef28) and only includes the new `RecvShutdown` operation now. Will push PRs for the two other commits.

FYI, this question in https://github.com/lnfuzz/smite/pull/163#issuecomment-5541217268 is still open for me:

> `RecvShutdown` returns empty bytes as a sentinel similar to `LookupShortChannelId` when no shutdown is expected on any channel. It's clearly an invalid shutdown script. Could this be a problem for downstream consumers of its return value, who might expect a valid shutdown script? Should they then also be no-ops, at least in the case of `RecvShutdown`?

But I think it's more relevant when we start consuming the output of `RecvShutdown`.

ekzyis:
8d65b49: add `execute_recv_shutdown_unknown_channel` test in this commit and update `UnexpectedMessage` with `InvalidShutdown` in the next PR, instead of adding the test in the next PR

PR #165, opened 2026-07-25
smitebot: hash-distribute runner strategy modifiers

Ashish-Kumar-Dash:
Addresses the strategy-distribution feedback from PR #139 .

  ## What changed

Reworks `runner_strategy` in `start.rs` so mutation modifiers are spread by a deterministic per-modifier hash instead of index thresholds, and adds the remaining issue #138 modifiers. 

  - **`-p` for secondaries only.** The primary (runner 0) runs AFL++'s default  `explore` schedule with no `-p` flag; secondaries round-robin the schedule  list (reordered so `explore` is index 0).
  - **New modifiers on secondaries:** `-a binary`/`text`, `-P explore`/  `exploit`, `-L 0`, `-Z`. 
  - **Hash-based assignment.** For each modifier, `X = fnv1a(runner_id ‖ name) % 100`; applied when `X < probability`. Mutually-exclusive flags (`-a`, `-P`) share one hash mapped to disjoint ranges so a runner never draws both values. `AFL_DISABLE_TRIM` moves onto this scheme too, so no two modifiers correlate by runner index. 
  - **`-L` skipped when a custom mutator is set** (IR scenario or user `afl_env`  supplying `AFL_CUSTOM_MUTATOR_LIBRARY`) . MOpt is incompatible with custom  mutators (`afl-fuzz.c:2417` FATALs).
    
  Distribution per secondary: `-a` binary ~30% / text ~30%; `-P` explore ~40% / exploit ~20%; `-L 0` ~10%; `-Z` ~10%; `AFL_DISABLE_TRIM` ~60%. Probabilities are asymptotic, as each runner is an independent deterministic draw, so small runner counts won't quite hit the nominal ratios .

### DIFF
--- a/smitebot/README.md
+++ b/smitebot/README.md
@@ -37,9 +37,16 @@ Launches a fuzzing campaign. Builds the Docker image, sets up the Nyx sharedir, 
 smitebot start campaign.toml
 ```
 
-Each runner gets a deterministic strategy distribution:
-- Power schedule (`-p`): round-robin across `fast`, `explore`, `coe`, `lin`, `quad`, `exploit`, `rare`
-- `-a binary`: ~70% of secondary runners
+Each runner gets a deterministic strategy distribution. The primary runner (0)
+runs AFL++'s default schedule with no modifiers; secondaries carry the schedule
+and mutation modifiers below. The power schedule is round-robin by index; the
+remaining modifiers are spread across secondaries by a fixed hash of the runner
+index, so a given runner always draws the same flags:
+- Power schedule (`-p`, secondaries only): round-robin across `explore`, `fast`, `coe`, `lin`, `quad`, `exploit`, `rare`
+- `-a binary`: ~70% of secondaries (wire messages are binary; hints AFL++ to use binary mutation strategies)
+- `-P` (fixed mutation strategy): `explore` ~40% / `exploit` ~20% of secondaries
+- `-L 0` (MOpt mode): ~10% of secondaries; skipped when a custom mutator is set (incompatible with MOpt)
+- `-Z` (sequential queue selection): ~10% of secondaries
 - `AFL_DISABLE_TRIM`: ~60% of secondary runners
 - `AFL_FINAL_SYNC`: primary runner only
 - `AFL_IMPORT_FIRST`: enabled when runner count < 16
--- a/smitebot/src/commands/start.rs
+++ b/smitebot/src/commands/start.rs
@@ -32,7 +32,9 @@ const VERIFY_TIMEOUT: Duration = Duration::from_mins(10);
 const VERIFY_POLL_INTERVAL: Duration = Duration::from_secs(5);
 
 /// AFL++ power schedules for round-robin distribution across runners.
-const POWER_SCHEDULES: &[&str] = &["fast", "explore", "coe", "lin", "quad", "exploit", "rare"];
+// Index 0 (explore) is AFL++'s default schedule; the primary runs it implicitly
+// with no -p flag, so secondaries cycling past the end wrap onto the safe default.
+const POWER_SCHEDULES: &[&str] = &["explore", "fast", "coe", "lin", "quad", "exploit", "rare"];
 
 /// Command handler for `smitebot start`.
 pub struct StartCommand;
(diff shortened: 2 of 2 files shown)


smitebot/src/commands/start.rs
```diff
@@ -392,39 +394,81 @@ fn read_fuzzer_pid(path: &Path) -> Option<u32> {
         .and_then(|v| v.trim().parse().ok())
 }
 
+/// FNV-1a hash of `runner_id` followed by `name`, reduced to `0..100`.
+///
+/// Used to spread mutation modifiers across runners deterministically, so a
+/// given runner always draws the same flags. std's `DefaultHasher` is
+/// explicitly not stable across Rust releases (see `std::hash` docs), which
+/// would break reproducibility, hence this fixed inline hash.
+fn modifier_hash(runner_id: u16, name: &str) -> u64 {
```
NishantBansal2003, line 403: Can we use sha256 instead of implementing fnv-1a hash algorithm? It's already well established across the codebase

smitebot/src/commands/start.rs
```diff
@@ -392,39 +394,81 @@ fn read_fuzzer_pid(path: &Path) -> Option<u32> {
...
             envs.push(("AFL_DISABLE_TRIM".to_string(), "1".to_string()));
         }
+
+        // -a: target input format (afl-fuzz.c:288). binary 30% / text 30% / none.
+        let a = modifier_hash(runner_id, "-a");
+        if a < 30 {
+            flags.extend(["-a".to_string(), "binary".to_string()]);
+        } else if a < 60 {
+            flags.extend(["-a".to_string(), "text".to_string()]);
+        }
```
NishantBansal2003: Why was `-a binary` changed from 70% for this strategy? I think we should stick with it, since the target input format is binary only

NishantBansal2003 (changes requested):

smitebot/src/commands/start.rs
Ashish-Kumar-Dash: sha256 is used in the smite crate via `bitcoin::hashes`, but smitebot doesn't depend on bitcoin. Using it here means adding a new dep. FNV-1a is 6 lines, no new dep. I didn't think that the tradeoff was worth it.

smitebot/README.md
```diff
@@ -37,9 +37,16 @@ Launches a fuzzing campaign. Builds the Docker image, sets up the Nyx sharedir,
...
-Each runner gets a deterministic strategy distribution:
-- Power schedule (`-p`): round-robin across `fast`, `explore`, `coe`, `lin`, `quad`, `exploit`, `rare`
-- `-a binary`: ~70% of secondary runners
+Each runner gets a deterministic strategy distribution. The primary runner (0)
+runs AFL++'s default schedule with no modifiers; secondaries carry the schedule
+and mutation modifiers below. The power schedule is round-robin by index; the
+remaining modifiers are spread across secondaries by a fixed hash of the runner
+index, so a given runner always draws the same flags:
+- Power schedule (`-p`, secondaries only): round-robin across `explore`, `fast`, `coe`, `lin`, `quad`, `exploit`, `rare`
+- `-a binary`: ~70% of secondaries (target input is binary network protocol)
```
NishantBansal2003: -a specifies the target input format for mutation purposes, not the corpus type, which affects the mutation strategies used by Afl++

Since the messages fed to the target are wire messages, -a binary fits our case. It's not like the target is getting input in a binary network protocol, please fix that

smitebot/src/commands/start.rs
```diff
@@ -672,81 +719,154 @@ mod tests {
...
+    fn runner_strategy_deterministic() {
+        // Same inputs always yield the same flags — reproducibility guarantee.
+        assert_eq!(
+            runner_strategy(5, 16, None, false),
+            runner_strategy(5, 16, None, false)
+        );
+    }
+
+    #[test]
+    fn runner_strategy_disable_trim_hash_distributed() {
```
NishantBansal2003: Can we add tests that actually assert `modifier_flag` < `modifier_probability` with some spread, instead of using `saw` and `saw_not`, which don't give much insight?

same applies to the other modifier tests as well

smitebot/src/commands/start.rs
```diff
@@ -672,81 +719,154 @@ mod tests {
...
+        // 70% of 63 secondaries = ~44. Allow ±10 for hash spread.
+        assert!(
+            (35..=55).contains(&binary_count),
+            "binary_count={binary_count}"
+        );
+        // -a text is never set.
+        for id in 1..64u16 {
+            let (flags, _) = runner_strategy(id, 64, None, false);
+            assert!(!flags.contains(&"text".to_string()));
+        }
```
NishantBansal2003: nit: I don't think this test assert is required, since we never use this at all

smitebot/src/commands/start.rs
```diff
@@ -392,39 +394,79 @@ fn read_fuzzer_pid(path: &Path) -> Option<u32> {
...
 
         // AFL_DISABLE_TRIM on ~60% of secondary runners.
-        let trim_count = (secondary_count * 6).div_ceil(10);
-        if ((runner_id - 1) as usize) < trim_count {
+        if modifier_hash(runner_id, "AFL_DISABLE_TRIM") < 60 {
             envs.push(("AFL_DISABLE_TRIM".to_string(), "1".to_string()));
         }
+
+        // -a binary on ~70% of secondaries per AFL++ docs and issue #138;
+        // target input is binary (network protocol messages). afl-fuzz.c:288.
```
NishantBansal2003: In general, it would be better to cite the actual links rather than just code lines and issue numbers, as those don't provide much context

smitebot/src/commands/start.rs
NishantBansal2003: Not sure whether fnv-1a is on par with sha256 for our use case or whether it has any input bias, but I'm fine with it if others think fnv-1a is sufficient and simpler for our use case

smitebot/src/commands/start.rs
```diff
@@ -392,39 +394,80 @@ fn read_fuzzer_pid(path: &Path) -> Option<u32> {
...
         }
+
+        // -a binary on ~70% of secondaries: wire messages are binary, so hinting
+        // AFL++ to use binary mutation strategies (not text tokenization) fits the
+        // target. See https://github.com/AFLplusplus/AFLplusplus/blob/e0b7cae5/src/afl-fuzz.c#L288
+        if modifier_hash(runner_id, "-a") < 70 {
+            flags.extend(["-a".to_string(), "binary".to_string()]);
+        }
+
+        // -P: fixed mutation strategy (afl-fuzz.c:250). explore 40% / exploit 20% / none.
```
NishantBansal2003: This applies here as well: https://github.com/morehouse/smite/pull/165#discussion_r3655296590 and to all other places that use only the code file and line number instead of actual links. Or, remove this altogether and use it only where citing the AFL++ doc alone is not sufficient (e.g. for `MOpt`)

smitebot/src/commands/start.rs
```diff
@@ -672,81 +720,158 @@ mod tests {
...
+    #[test]
+    fn runner_strategy_primary_gets_no_modifiers() {
+        // Modifiers apply to secondaries only, never the primary.
+        let (flags, _) = runner_strategy(0, 10, None, false);
+        for modifier in ["-p", "-a", "-P", "-L", "-Z"] {
+            assert!(
+                !flags.contains(&modifier.to_string()),
+                "primary got {modifier}"
+            );
+        }
```
NishantBansal2003:
```suggestion
        let (flags, envs) = runner_strategy(0, 10, None, false);
        for modifier in ["-p", "-a", "-P", "-L", "-Z"] {
            assert!(
                !flags.contains(&modifier.to_string()),
                "primary got {modifier}"
            );
        }
        assert!(!envs.iter().any(|(k, _)| k == "AFL_DISABLE_TRIM"));
```

smitebot/src/commands/start.rs
```diff
@@ -672,81 +720,158 @@ mod tests {
...
     fn runner_strategy_wraps_schedules_past_seven() {
-        let (flags, _) = runner_strategy(7, 8, None);
-        assert_eq!(flags[1], "fast");
+        // id 7 wraps to index 0 = explore.
+        let (flags, _) = runner_strategy(7, 16, None, false);
+        assert_eq!(flags[1], "explore");
+    }
+
+    #[test]
+    fn runner_strategy_input_format_binary_70_percent() {
```
NishantBansal2003, line 771: Could add a test for `-Z` flag as well

NishantBansal2003 (approved):

morehouse (approved):
LGTM

PR #166, opened 2026-07-27
smite: add tu32/tu64 truncated integer wire types

erickcestari:
This will be needed for BOLT 4 hop payloads, which use `tu32` and `tu64` to encode fields.

### DIFF
--- a/smite/src/bolt/wire.rs
+++ b/smite/src/bolt/wire.rs
@@ -3,7 +3,7 @@
 use crate::bolt::BoltError;
 use crate::bolt::types::{
     BigSize, CHANNEL_ID_SIZE, COMPACT_SIGNATURE_SIZE, ChannelId, PUBLIC_KEY_SIZE, SHA256_HASH_SIZE,
-    ShortChannelId, TXID_SIZE,
+    ShortChannelId, TXID_SIZE, Tu32, Tu64,
 };
 use bitcoin::Txid;
 use bitcoin::hashes::{Hash, sha256};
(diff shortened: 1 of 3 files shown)


smite/src/bolt/wire.rs
```diff
@@ -57,6 +57,64 @@ impl<const N: usize> WireFormat for [u8; N] {
...
+            /// # Errors
+            ///
+            /// Returns `TruncatedIntTooLong` if more bytes remain than the
+            /// integer's width, or `TruncatedIntNotMinimal` if the first byte
+            /// is zero.
+            fn read(data: &mut &[u8]) -> Result<Self, BoltError> {
+                const WIDTH: usize = std::mem::size_of::<$inner>();
+
+                let bytes = *data;
+                *data = &[];
```
NishantBansal2003: nit: Could move this down after all error paths, We currently advance the cursor only on success, so let's keep that here as well

smite/src/bolt/wire.rs
```diff
@@ -57,6 +57,64 @@ impl<const N: usize> WireFormat for [u8; N] {
...
+                }
+                if bytes.first() == Some(&0) {
+                    return Err(BoltError::TruncatedIntNotMinimal);
+                }
+
+                let mut value: $inner = 0;
+                for byte in bytes {
+                    value = (value << 8) | <$inner>::from(*byte);
+                }
+                Ok(Self(value))
```
NishantBansal2003: nit: This is very optional, so no changes are really required -- just a thought. The code below is the exact inverse of the write logic and is currently used by both LDK and LND

```rust
  let mut buf = [0u8; WIDTH];
  buf[WIDTH - bytes.len()..].copy_from_slice(bytes);
  Ok(Self(<$inner>::from_be_bytes(buf)))
```

smite/src/bolt/wire.rs
```diff
@@ -1007,4 +1065,59 @@ mod tests {
...
+        let mut data: &[u8] = &[0x01, 0x00];
+        assert_eq!(Tu64::read(&mut data).unwrap(), Tu64(256));
+        assert!(data.is_empty());
+
+        let mut empty: &[u8] = &[];
+        assert_eq!(Tu32::read(&mut empty).unwrap(), Tu32(0));
+    }
+
+    #[test]
+    fn truncated_int_roundtrips() {
```
NishantBansal2003: nit: Could add tests for tu64 as well, covering only the cases whose width exceeds tu32

NishantBansal2003 (approved):
Looks good!

smite/src/bolt/wire.rs
erickcestari: Indeed it looks better

smite/src/bolt/wire.rs
erickcestari: Updated!

smite/src/bolt/wire.rs
erickcestari: Added

erickcestari:
Thanks for the review @NishantBansal2003 . I've addressed your nits!

smite/src/bolt/wire.rs
```diff
@@ -1007,4 +1064,80 @@ mod tests {
...
+        // Five bytes still fit in a tu64.
+        let mut data: &[u8] = &[0x01; 5];
+        assert_eq!(Tu64::read(&mut data).unwrap(), Tu64(0x01_0101_0101));
+
+        let mut data: &[u8] = &[0x01; 9];
+        assert_eq!(
+            Tu64::read(&mut data),
+            Err(BoltError::TruncatedIntTooLong { max: 8, actual: 9 })
+        );
+    }
```
morehouse, line 1169: I think we should also implement all the relevant test vectors from [BOLT 1 Appendix B](https://github.com/lightning/bolts/blob/master/01-messaging.md#appendix-b-type-length-value-test-vectors), potentially replacing any similar tests here.

smite/src/bolt/wire.rs
```diff
@@ -57,6 +57,63 @@ impl<const N: usize> WireFormat for [u8; N] {
...
+macro_rules! impl_truncated_int {
+    ($name:ident, $inner:ty, $doc:literal) => {
+        #[doc = $doc]
+        ///
+        /// Its width is only recoverable from the enclosing TLV record, so
+        /// `read` consumes every remaining byte. Decode one only from a slice
+        /// already narrowed to the TLV value it ends, never from a cursor
+        /// running over a message body.
+        #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Default, Hash)]
+        pub struct $name(pub $inner);
```
morehouse: The current convention is to put type definitions like this one in `types.rs` and then the `impl WireFormat` goes here.

smite/src/bolt/wire.rs
erickcestari: Added!

smite/src/bolt/wire.rs
erickcestari: Makes sense! Updated!

morehouse (approved):
LGTM

PR #169, opened 2026-07-28
scripts: make Nyx VM memory size configurable

Chand-ra:
The memory allocation for Nyx VM instances is hardcoded  to 4096 MB during the packer configuration generation step.

Introduce the `VM_SIZE` environment variable, allowing users to dynamically scale the VM memory down for resource-constrained deployments such as parallel fuzzing trials.

Retain the default 4096 MB fallback if the variable is unset and `echo` the chosen size to stdout for better setup observability.

### DIFF
--- a/scripts/setup-nyx.sh
+++ b/scripts/setup-nyx.sh
@@ -4,8 +4,11 @@
 #
 # Usage: ./scripts/setup-nyx.sh <sharedir> <docker-image> <aflplusplus-path>
 #
+# Environment Variables:
+#   NYX_MEM_MB: Sets the memory allocation for the Nyx VM in MB (default: 2048).
+#
 # Example:
-#   ./scripts/setup-nyx.sh /tmp/smite-lnd-nyx smite-lnd ~/AFLplusplus
+#   NYX_MEM_MB=4096 ./scripts/setup-nyx.sh /tmp/smite-lnd-nyx smite-lnd ~/AFLplusplus
 #
 # Creates the necessary files in sharedir for snapshot fuzzing using
 # docker-image.  Requires that docker-image is already built and that AFL++ has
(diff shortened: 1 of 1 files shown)


erickcestari (approved):
LGTM! Maybe we could also reduce the default value. I haven't tested other values, but a lower default should make it easier to run more fuzzing jobs in parallel.

I have a dedicated server with 62 GiB of RAM, and it can run at most 15 fuzzing instances simultaneously.

Chand-ra:
> LGTM! Maybe we could also reduce the default value. I haven't tested other values, but a lower default should make it easier to run more fuzzing jobs in parallel.
> 
> I have a dedicated server with 62 GiB of RAM, and it can run at most 15 fuzzing instances simultaneously.

Agreed. I've ran a lot of trials with 2048 as the image size and haven't ever had any problems with it, but I remember @NishantBansal2003 telling me once that it caused some issues with his setup for Eclair or CLN.

I personally prefer that as the default value and can vouch for it working fine, but let us see what others have to say.

NishantBansal2003 (approved):
Looks good!

> Agreed. I've ran a lot of trials with 2048 as the image size and haven't ever had any problems with it, but I remember @NishantBansal2003 telling me once that it caused some issues with his setup for Eclair or CLN.

I had issues with the nested virtualization setup, though the machine specs were not quite good, so I didn’t expect good results. On my dedicated machine, I lowered the VM size to 2048 and tried running the CLN campaign, and I haven’t faced any issues though

scripts/setup-nyx.sh
```diff
@@ -72,7 +72,12 @@ cat > "$FORCE_FORK_DIR/sitecustomize.py" <<'EOF'
 import multiprocessing
 multiprocessing.set_start_method("fork", force=True)
 EOF
-(cd "$PACKER_PATH" && PYTHONPATH="$FORCE_FORK_DIR${PYTHONPATH:+:$PYTHONPATH}" ./nyx_config_gen.py "$SHAREDIR" Kernel -m 4096)
+
+# The size of a single QEMU VM instance. Defaults to 4096 MB.
+VM_SIZE="${VM_SIZE:-4096}"
```
morehouse: Nit:  this isn't the size of the VM, it's the amount of memory allocated to the VM.  Also we can make the name self-documenting.


```suggestion
NYX_MEM_MB="${SMITE_NYX_MEM_MB:-4096}"
```

scripts/setup-nyx.sh
morehouse, line 15: We should document the new environment variable here. in the header comment.

morehouse (approved):

Chand-ra:
Since no one seems to have a problem with it, I have changed the default to 2048 MB.

morehouse (approved):
Tested the lower 2GB memory limit by fuzzing the IR scenario against all targets for 1h.  No issues encountered.

PR #171, opened 2026-07-28
smite-scenarios: skip RecvChannelReady on premature funding tx

morehouse:
If the funding transaction was mined before sending it to the target via `funding_created`, the target may never detect the funding transaction on chain and will therefore fail to send us `channel_ready`.  Make `RecvChannelReady` a no-op in this case.

This was found to occur while fuzzing both LDK and LND.

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -21,7 +21,7 @@ use smite::pending_channel::PendingChannel;
 use smite::violation::Violation;
 use smite_ir::operation::AcceptChannelField;
 use smite_ir::{Operation, Program, Variable};
-use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


NishantBansal2003 (approved):
LGTM!

This approach gave me an idea that could be implemented in the future: maintain a `HashMap<txid, confirmations>`. This would replace both the `unmined_txids` and `mined_txids` `HashSet` with a single map, while avoiding repeated `get_transaction_confirmations` RPC calls during sequences like `MineBlock` -> `RecvChannelReady` until `state.config.minimum_depth` is reached

morehouse:
Cool idea.  Anything to reduce unnecessary RPC calls would be good for improving speed.

PR #172, opened 2026-07-29
smite-nyx-sys+workloads/ldk: include panic messages in crash reports

NishantBansal2003:
LDK workload builds with `panic = "abort"`, so panic messages are only written to stderr, which the scenario/Nyx discards. As a result, crash reports contained only an `abort` with no indication of what actually failed.

see:
https://github.com/morehouse/smite/blob/06a5edfacc9a0656cca9dd6b3abc92e0ac0c5993/workloads/ldk/Cargo.toml#L7-L9

Install a panic hook that writes the panic message and its source location to `/tmp/smite-panic.log.<pid>`, and have the crash handler append target-generated logs to the crash report.

ref (fixed in https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/pulls/4833):

```sh
abort
panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lightning-0.2.1/src/sign/mod.rs:2440:9:
assertion failed: expected_max_weight <=
    spend_tx.weight().to_wu() + descriptors.len() as u64 * 3
```

### DIFF
--- a/smite-nyx-sys/src/nyx-crash-handler.c
+++ b/smite-nyx-sys/src/nyx-crash-handler.c
@@ -34,6 +34,8 @@
 #include "nyx.h"
 #endif
 
+// Must match PANIC_LOG_PATH in workloads/ldk/src/main.rs.
+#define PANIC_LOG_PATH "/tmp/smite-panic.log"
 #define ASAN_LOG_PATH "/tmp/asan.log"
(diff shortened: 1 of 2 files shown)


morehouse (approved):
Tested this locally.  An injected panic was found and reported successfully.  The new overhead was also negligible (~10 us).

PR #173, opened 2026-07-29
smitebot: document tmux_session in sample config

Ashish-Kumar-Dash:
`tmux_session` is a supported optional config key (added in #139 ) but was missing from the sample file. Added it as an example.


morehouse (approved):
LGTM

PR #181, opened 2026-07-30
Add HTLC support to BOLT 3 commitment transaction

NishantBansal2003:
ref: #111 

Bolts ref:
- https://github.com/lightning/bolts/blob/master/03-transactions.md
- https://github.com/lightning/bolts/blob/master/05-onchain.md

#### Verification

Anchor test vectors (local signatures): https://github.com/ACINQ/eclair/blob/master/eclair-core/src/test/resources/bolt3-tx-test-vectors-anchor-outputs-zero-fee-htlc-tx-format.txt

Custom tests on top of commit https://github.com/ACINQ/eclair/commit/26d035070 (after this commit, eclair removed support for non-anchor channels), with the diff below:

<details>

<summary>Eclair diff</summary>

```diff
diff --git a/eclair-core/src/test/scala/fr/acinq/eclair/transactions/TestVectorsSpec.scala b/eclair-core/src/test/scala/fr/acinq/eclair/transactions/TestVectorsSpec.scala
index 1931aa8b1..1623fb59e 100644
--- a/eclair-core/src/test/scala/fr/acinq/eclair/transactions/TestVectorsSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/transactions/TestVectorsSpec.scala
@@ -456,6 +456,199 @@ class StaticRemoteKeyTestVectorSpec extends TestVectorsSpec {
   override def filename: String = "/bolt3-tx-test-vectors-static-remotekey-format.txt"
   override def channelFeatures: Set[ChannelTypeFeature] = Set(Features.StaticRemoteKey)
   // @formatter:on
+
+  test("commitment_tx_with_htlc_amount_msat_not_multiple_of_1000_legacy") {
+    val name = "commitment_tx_with_htlc_amount_msat_not_multiple_of_1000_legacy"
+    val htlc = OutgoingHtlc(UpdateAddHtlc(ByteVector32.Zeroes, 0, 2000999 msat, ByteVector32.Zeroes, CltvExpiry(500), TestConstants.emptyOnionPacket, None, Reputation.maxEndorsement, None))
+    val spec = CommitmentSpec(Set(htlc), FeeratePerKw(647 sat), toLocal = (7000000000L - 2000999L) msat, toRemote = 3000000000L msat)
+    val dustLimit = 546.sat
+
+    println()
+    println(s"name: $name")
+
+    val outputs = Transactions.makeCommitTxOutputs(
+      Local.funding_pubkey,
+      Remote.funding_pubkey,
+      localCommitmentKeys.publicKeys,
+      payCommitTxFees = true,
+      dustLimit,
+      Local.toSelfDelay,
+      spec,
+      commitmentFormat
+    )
+    val unsignedCommitTx = Transactions.makeCommitTx(
+      commitTxInput = commitmentInput,
+      commitTxNumber = Local.commitTxNumber,
+      localPaymentBasePoint = Local.payment_basepoint,
+      remotePaymentBasePoint = Remote.payment_basepoint,
+      localIsChannelOpener = true,
+      outputs = outputs)
+    val local_sig = unsignedCommitTx.sign(Local.funding_privkey, Remote.funding_pubkey)
+    val remote_sig = unsignedCommitTx.sign(Remote.funding_privkey, Local.funding_pubkey)
+    println(s"local_signature: ${Scripts.der(local_sig.sig).dropRight(1).toHex}")
+    println(s"remote_signature: ${Scripts.der(remote_sig.sig).dropRight(1).toHex}")
+    val commitTx = unsignedCommitTx.aggregateSigs(Local.funding_pubkey, Remote.funding_pubkey, local_sig, remote_sig)
+    Transaction.correctlySpends(commitTx, Seq(fundingTx), ScriptFlags.STANDARD_SCRIPT_VERIFY_FLAGS)
+
+    val htlcTxs = Transactions.makeHtlcTxs(commitTx, outputs, commitmentFormat).sortBy(_.input.outPoint.index)
+    println(s"num_htlcs: ${htlcTxs.length}")
+    htlcTxs.foreach {
+      case tx: UnsignedHtlcTimeoutTx =>
+        val remoteHtlcSig = tx.localSig(remoteCommitmentKeys)
+        val withRemoteSig = tx.addRemoteSig(localCommitmentKeys, remoteHtlcSig)
+        val localHtlcSig = withRemoteSig.localSig(WalletInputs(Nil, None))
+        println(s"# signature for output #${tx.input.outPoint.index} (htlc-timeout for htlc #${tx.htlcId})")
+        println(s"local_htlc_signature: ${Scripts.der(localHtlcSig).dropRight(1).toHex}")
+        println(s"remote_htlc_signature: ${Scripts.der(remoteHtlcSig).dropRight(1).toHex}")
+      case tx: UnsignedHtlcSuccessTx =>
+        fail(s"unexpected htlc-success tx for htlc #${tx.htlcId}")
+    }
+  }
+
+  test("commitment_tx_with_htlc_value_equals_non_htlc_value_legacy") {
+    val name = "commitment_tx_with_htlc_value_equals_non_htlc_value_legacy"
+    val htlc = OutgoingHtlc(UpdateAddHtlc(ByteVector32.Zeroes, 0, 2000000000 msat, ByteVector32.Zeroes, CltvExpiry(500), TestConstants.emptyOnionPacket, None, Reputation.maxEndorsement, None))
+    val spec = CommitmentSpec(Set(htlc), FeeratePerKw(15_000 sat), toLocal = (8000000000L - 2000000000L) msat, toRemote = 2000000000L msat)
+    val dustLimit = 546.sat
+
+    println()
+    println(s"name: $name")
+
+    val outputs = Transactions.makeCommitTxOutputs(
+      Local.funding_pubkey,
+      Remote.funding_pubkey,
+      localCommitmentKeys.publicKeys,
+      payCommitTxFees = true,
+      dustLimit,
+      Local.toSelfDelay,
+      spec,
+      commitmentFormat
+    )
+    val unsignedCommitTx = Transactions.makeCommitTx(
+      commitTxInput = commitmentInput,
+      commitTxNumber = Local.commitTxNumber,
+      localPaymentBasePoint = Local.payment_basepoint,
+      remotePaymentBasePoint = Remote.payment_basepoint,
+      localIsChannelOpener = true,
+      outputs = outputs)
+    val local_sig = unsignedCommitTx.sign(Local.funding_privkey, Remote.funding_pubkey)
+    val remote_sig = unsignedCommitTx.sign(Remote.funding_privkey, Local.funding_pubkey)
+    println(s"local_signature: ${Scripts.der(local_sig.sig).dropRight(1).toHex}")
+    println(s"remote_signature: ${Scripts.der(remote_sig.sig).dropRight(1).toHex}")
+    val commitTx = unsignedCommitTx.aggregateSigs(Local.funding_pubkey, Remote.funding_pubkey, local_sig, remote_sig)
+    Transaction.correctlySpends(commitTx, Seq(fundingTx), ScriptFlags.STANDARD_SCRIPT_VERIFY_FLAGS)
+
+    val htlcTxs = Transactions.makeHtlcTxs(commitTx, outputs, commitmentFormat).sortBy(_.input.outPoint.index)
+    println(s"num_htlcs: ${htlcTxs.length}")
+    htlcTxs.foreach {
+      case tx: UnsignedHtlcTimeoutTx =>
+        val remoteHtlcSig = tx.localSig(remoteCommitmentKeys)
+        val withRemoteSig = tx.addRemoteSig(localCommitmentKeys, remoteHtlcSig)
+        val localHtlcSig = withRemoteSig.localSig(WalletInputs(Nil, None))
+        println(s"# signature for output #${tx.input.outPoint.index} (htlc-timeout for htlc #${tx.htlcId})")
+        println(s"local_htlc_signature: ${Scripts.der(localHtlcSig).dropRight(1).toHex}")
+        println(s"remote_htlc_signature: ${Scripts.der(remoteHtlcSig).dropRight(1).toHex}")
+      case tx: UnsignedHtlcSuccessTx =>
+        fail(s"unexpected htlc-success tx for htlc #${tx.htlcId}")
+    }
+  }
+
+  test("commitment_tx_with_non_htlc_outputs_between_htlc_outputs_legacy") {
+    val name = "commitment_tx_with_non_htlc_outputs_between_htlc_outputs_legacy"
+    val htlc1 = OutgoingHtlc(UpdateAddHtlc(ByteVector32.Zeroes, 0, 2000000000L msat, ByteVector32.Zeroes, CltvExpiry(500), TestConstants.emptyOnionPacket, None, Reputation.maxEndorsement, None))
+    val htlc2 = OutgoingHtlc(UpdateAddHtlc(ByteVector32.Zeroes, 1, 4000000000L msat, ByteVector32.Zeroes, CltvExpiry(501), TestConstants.emptyOnionPacket, None, Reputation.maxEndorsement, None))
+    val spec = CommitmentSpec(Set[DirectedHtlc](htlc1, htlc2), FeeratePerKw(15000 sat), toLocal = (7000000000L - 2000000000L - 4000000000L) msat, toRemote = 3000000000L msat)
+    val dustLimit = 546.sat
+
+    println()
+    println(s"name: $name")
+
+    val outputs = Transactions.makeCommitTxOutputs(
+      Local.funding_pubkey,
+      Remote.funding_pubkey,
+      localCommitmentKeys.publicKeys,
+      payCommitTxFees = true,
+      dustLimit,
+      Local.toSelfDelay,
+      spec,
+      commitmentFormat
+    )
+    val unsignedCommitTx = Transactions.makeCommitTx(
+      commitTxInput = commitmentInput,
+      commitTxNumber = Local.commitTxNumber,
+      localPaymentBasePoint = Local.payment_basepoint,
+      remotePaymentBasePoint = Remote.payment_basepoint,
+      localIsChannelOpener = true,
+      outputs = outputs)
+    val local_sig = unsignedCommitTx.sign(Local.funding_privkey, Remote.funding_pubkey)
+    val remote_sig = unsignedCommitTx.sign(Remote.funding_privkey, Local.funding_pubkey)
+    println(s"local_signature: ${Scripts.der(local_sig.sig).dropRight(1).toHex}")
+    println(s"remote_signature: ${Scripts.der(remote_sig.sig).dropRight(1).toHex}")
+    val commitTx = unsignedCommitTx.aggregateSigs(Local.funding_pubkey, Remote.funding_pubkey, local_sig, remote_sig)
+    Transaction.correctlySpends(commitTx, Seq(fundingTx), ScriptFlags.STANDARD_SCRIPT_VERIFY_FLAGS)
+
+    val htlcTxs = Transactions.makeHtlcTxs(commitTx, outputs, commitmentFormat).sortBy(_.input.outPoint.index)
+    println(s"num_htlcs: ${htlcTxs.length}")
+    htlcTxs.foreach {
+      case tx: UnsignedHtlcTimeoutTx =>
+        val remoteHtlcSig = tx.localSig(remoteCommitmentKeys)
+        val withRemoteSig = tx.addRemoteSig(localCommitmentKeys, remoteHtlcSig)
+        val localHtlcSig = withRemoteSig.localSig(WalletInputs(Nil, None))
+        println(s"# signature for output #${tx.input.outPoint.index} (htlc-timeout for htlc #${tx.htlcId})")
+        println(s"local_htlc_signature: ${Scripts.der(localHtlcSig).dropRight(1).toHex}")
+        println(s"remote_htlc_signature: ${Scripts.der(remoteHtlcSig).dropRight(1).toHex}")
+      case tx: UnsignedHtlcSuccessTx =>
+        fail(s"unexpected htlc-success tx for htlc #${tx.htlcId}")
+    }
+  }
+
+  test("commitment_tx_with_asymmetric_dust_limits_legacy") {
+    val name = "commitment_tx_with_asymmetric_dust_limits_legacy"
+    val htlc = IncomingHtlc(UpdateAddHtlc(ByteVector32.Zeroes, 0, 20000000 msat, ByteVector32.Zeroes, CltvExpiry(500), TestConstants.emptyOnionPacket, None, Reputation.maxEndorsement, None))
+    val spec = CommitmentSpec(Set[DirectedHtlc](htlc), FeeratePerKw(15000 sat), toLocal = 7000000000L msat, toRemote = (3000000000L - 20000000L) msat)
+    val dustLimit = 546.sat
+
+    println()
+    println(s"name: $name")
+
+    val outputs = Transactions.makeCommitTxOutputs(
+      Local.funding_pubkey,
+      Remote.funding_pubkey,
+      localCommitmentKeys.publicKeys,
+      payCommitTxFees = true,
+      dustLimit,
+      Local.toSelfDelay,
+      spec,
+      commitmentFormat
+    )
+    val unsignedCommitTx = Transactions.makeCommitTx(
+      commitTxInput = commitmentInput,
+      commitTxNumber = Local.commitTxNumber,
+      localPaymentBasePoint = Local.payment_basepoint,
+      remotePaymentBasePoint = Remote.payment_basepoint,
+      localIsChannelOpener = true,
+      outputs = outputs)
+    val local_sig = unsignedCommitTx.sign(Local.funding_privkey, Remote.funding_pubkey)
+    val remote_sig = unsignedCommitTx.sign(Remote.funding_privkey, Local.funding_pubkey)
+    println(s"local_signature: ${Scripts.der(local_sig.sig).dropRight(1).toHex}")
+    println(s"remote_signature: ${Scripts.der(remote_sig.sig).dropRight(1).toHex}")
+    val commitTx = unsignedCommitTx.aggregateSigs(Local.funding_pubkey, Remote.funding_pubkey, local_sig, remote_sig)
+    Transaction.correctlySpends(commitTx, Seq(fundingTx), ScriptFlags.STANDARD_SCRIPT_VERIFY_FLAGS)
+
+    val htlcTxs = Transactions.makeHtlcTxs(commitTx, outputs, commitmentFormat).sortBy(_.input.outPoint.index)
+    println(s"num_htlcs: ${htlcTxs.length}")
+    htlcTxs.foreach {
+      case tx: UnsignedHtlcSuccessTx =>
+        val remoteHtlcSig = tx.localSig(remoteCommitmentKeys)
+        val redeemScript = Scripts.htlcReceived(localCommitmentKeys.publicKeys, tx.paymentHash, tx.htlcExpiry, commitmentFormat)
+        val localHtlcSig = Transaction.signInput(tx.tx, 0, redeemScript, fr.acinq.bitcoin.SigHash.SIGHASH_ALL, tx.input.txOut.amount, fr.acinq.bitcoin.SigVersion.SIGVERSION_WITNESS_V0, Local.htlc_privkey)
+        println(s"# signature for output #${tx.input.outPoint.index} (htlc-success for htlc #${tx.htlcId})")
+        println(s"local_htlc_signature: ${localHtlcSig.dropRight(1).toHex}")
+        println(s"remote_htlc_signature: ${Scripts.der(remoteHtlcSig).dropRight(1).toHex}")
+      case tx: UnsignedHtlcTimeoutTx =>
+        fail(s"unexpected htlc-timeout tx for htlc #${tx.htlcId}")
+    }
+  }
 }
 
 class AnchorOutputsTestVectorSpec extends TestVectorsSpec {
@@ -470,4 +663,245 @@ class AnchorOutputsZeroFeeHtlcTxTestVectorSpec extends TestVectorsSpec {
   override def filename: String = "/bolt3-tx-test-vectors-anchor-outputs-zero-fee-htlc-tx-format.txt"
   override def channelFeatures: Set[ChannelTypeFeature] = Set(Features.StaticRemoteKey, Features.AnchorOutputsZeroFeeHtlcTx)
   // @formatter:on
+
+  test("commitment_tx_with_htlc_amount_msat_not_multiple_of_1000_anchor") {
+    val name = "commitment_tx_with_htlc_amount_msat_not_multiple_of_1000_anchor"
+    val htlc = OutgoingHtlc(UpdateAddHtlc(ByteVector32.Zeroes, 0, 2000999 msat, ByteVector32.Zeroes, CltvExpiry(500), TestConstants.emptyOnionPacket, None, Reputation.maxEndorsement, None))
+    val spec = CommitmentSpec(Set(htlc), FeeratePerKw(647 sat), toLocal = (7000000000L - 2000999L) msat, toRemote = 3000000000L msat)
+    val dustLimit = 546.sat
+
+    println()
+    println(s"name: $name")
+
+    val outputs = Transactions.makeCommitTxOutputs(
+      Local.funding_pubkey,
+      Remote.funding_pubkey,
+      localCommitmentKeys.publicKeys,
+      payCommitTxFees = true,
+      dustLimit,
+      Local.toSelfDelay,
+      spec,
+      commitmentFormat
+    )
+    val unsignedCommitTx = Transactions.makeCommitTx(
+      commitTxInput = commitmentInput,
+      commitTxNumber = Local.commitTxNumber,
+      localPaymentBasePoint = Local.payment_basepoint,
+      remotePaymentBasePoint = Remote.payment_basepoint,
+      localIsChannelOpener = true,
+      outputs = outputs)
+    val local_sig = unsignedCommitTx.sign(Local.funding_privkey, Remote.funding_pubkey)
+    val remote_sig = unsignedCommitTx.sign(Remote.funding_privkey, Local.funding_pubkey)
+    println(s"local_signature: ${Scripts.der(local_sig.sig).dropRight(1).toHex}")
+    println(s"remote_signature: ${Scripts.der(remote_sig.sig).dropRight(1).toHex}")
+    val commitTx = unsignedCommitTx.aggregateSigs(Local.funding_pubkey, Remote.funding_pubkey, local_sig, remote_sig)
+    Transaction.correctlySpends(commitTx, Seq(fundingTx), ScriptFlags.STANDARD_SCRIPT_VERIFY_FLAGS)
+
+    val htlcTxs = Transactions.makeHtlcTxs(commitTx, outputs, commitmentFormat).sortBy(_.input.outPoint.index)
+    println(s"num_htlcs: ${htlcTxs.length}")
+    htlcTxs.foreach {
+      case tx: UnsignedHtlcTimeoutTx =>
+        val remoteHtlcSig = tx.localSig(remoteCommitmentKeys)
+        val withRemoteSig = tx.addRemoteSig(localCommitmentKeys, remoteHtlcSig)
+        val localHtlcSig = withRemoteSig.localSig(WalletInputs(Nil, None))
+        println(s"# signature for output #${tx.input.outPoint.index} (htlc-timeout for htlc #${tx.htlcId})")
+        println(s"local_htlc_signature: ${Scripts.der(localHtlcSig).dropRight(1).toHex}")
+        println(s"remote_htlc_signature: ${Scripts.der(remoteHtlcSig).dropRight(1).toHex}")
+      case tx: UnsignedHtlcSuccessTx =>
+        fail(s"unexpected htlc-success tx for htlc #${tx.htlcId}")
+    }
+  }
+
+  test("commitment_tx_with_htlc_value_equals_non_htlc_value_anchor") {
+    val name = "commitment_tx_with_htlc_value_equals_non_htlc_value_anchor"
+    val htlc = OutgoingHtlc(UpdateAddHtlc(ByteVector32.Zeroes, 0, 2000000000L msat, ByteVector32.Zeroes, CltvExpiry(500), TestConstants.emptyOnionPacket, None, Reputation.maxEndorsement, None))
+    val spec = CommitmentSpec(Set(htlc), FeeratePerKw(15000 sat), toLocal = (8000000000L - 2000000000L) msat, toRemote = 2000000000L msat)
+    val dustLimit = 546.sat
+
+    println()
+    println(s"name: $name")
+
+    val outputs = Transactions.makeCommitTxOutputs(
+      Local.funding_pubkey,
+      Remote.funding_pubkey,
+      localCommitmentKeys.publicKeys,
+      payCommitTxFees = true,
+      dustLimit,
+      Local.toSelfDelay,
+      spec,
+      commitmentFormat
+    )
+    val unsignedCommitTx = Transactions.makeCommitTx(
+      commitTxInput = commitmentInput,
+      commitTxNumber = Local.commitTxNumber,
+      localPaymentBasePoint = Local.payment_basepoint,
+      remotePaymentBasePoint = Remote.payment_basepoint,
+      localIsChannelOpener = true,
+      outputs = outputs)
+    val local_sig = unsignedCommitTx.sign(Local.funding_privkey, Remote.funding_pubkey)
+    val remote_sig = unsignedCommitTx.sign(Remote.funding_privkey, Local.funding_pubkey)
+    println(s"local_signature: ${Scripts.der(local_sig.sig).dropRight(1).toHex}")
+    println(s"remote_signature: ${Scripts.der(remote_sig.sig).dropRight(1).toHex}")
+    val commitTx = unsignedCommitTx.aggregateSigs(Local.funding_pubkey, Remote.funding_pubkey, local_sig, remote_sig)
+    Transaction.correctlySpends(commitTx, Seq(fundingTx), ScriptFlags.STANDARD_SCRIPT_VERIFY_FLAGS)
+
+    val htlcTxs = Transactions.makeHtlcTxs(commitTx, outputs, commitmentFormat).sortBy(_.input.outPoint.index)
+    println(s"num_htlcs: ${htlcTxs.length}")
+    htlcTxs.foreach {
+      case tx: UnsignedHtlcTimeoutTx =>
+        val remoteHtlcSig = tx.localSig(remoteCommitmentKeys)
+        val withRemoteSig = tx.addRemoteSig(localCommitmentKeys, remoteHtlcSig)
+        val localHtlcSig = withRemoteSig.localSig(WalletInputs(Nil, None))
+        println(s"# signature for output #${tx.input.outPoint.index} (htlc-timeout for htlc #${tx.htlcId})")
+        println(s"local_htlc_signature: ${Scripts.der(localHtlcSig).dropRight(1).toHex}")
+        println(s"remote_htlc_signature: ${Scripts.der(remoteHtlcSig).dropRight(1).toHex}")
+      case tx: UnsignedHtlcSuccessTx =>
+        fail(s"unexpected htlc-success tx for htlc #${tx.htlcId}")
+    }
+  }
+
+  test("commitment_tx_with_non_htlc_outputs_between_htlc_outputs_anchor") {
+    val name = "commitment_tx_with_non_htlc_outputs_between_htlc_outputs_anchor"
+    val htlc1 = OutgoingHtlc(UpdateAddHtlc(ByteVector32.Zeroes, 0, 2000000000L msat, ByteVector32.Zeroes, CltvExpiry(500), TestConstants.emptyOnionPacket, None, Reputation.maxEndorsement, None))
+    val htlc2 = OutgoingHtlc(UpdateAddHtlc(ByteVector32.Zeroes, 1, 4000000000L msat, ByteVector32.Zeroes, CltvExpiry(501), TestConstants.emptyOnionPacket, None, Reputation.maxEndorsement, None))
+    val spec = CommitmentSpec(Set[DirectedHtlc](htlc1, htlc2), FeeratePerKw(15000 sat), toLocal = (7000000000L - 2000000000L - 4000000000L) msat, toRemote = 3000000000L msat)
+    val dustLimit = 546.sat
+
+    println()
+    println(s"name: $name")
+
+    val outputs = Transactions.makeCommitTxOutputs(
+      Local.funding_pubkey,
+      Remote.funding_pubkey,
+      localCommitmentKeys.publicKeys,
+      payCommitTxFees = true,
+      dustLimit,
+      Local.toSelfDelay,
+      spec,
+      commitmentFormat
+    )
+    val unsignedCommitTx = Transactions.makeCommitTx(
+      commitTxInput = commitmentInput,
+      commitTxNumber = Local.commitTxNumber,
+      localPaymentBasePoint = Local.payment_basepoint,
+      remotePaymentBasePoint = Remote.payment_basepoint,
+      localIsChannelOpener = true,
+      outputs = outputs)
+    val local_sig = unsignedCommitTx.sign(Local.funding_privkey, Remote.funding_pubkey)
+    val remote_sig = unsignedCommitTx.sign(Remote.funding_privkey, Local.funding_pubkey)
+    println(s"local_signature: ${Scripts.der(local_sig.sig).dropRight(1).toHex}")
+    println(s"remote_signature: ${Scripts.der(remote_sig.sig).dropRight(1).toHex}")
+    val commitTx = unsignedCommitTx.aggregateSigs(Local.funding_pubkey, Remote.funding_pubkey, local_sig, remote_sig)
+    Transaction.correctlySpends(commitTx, Seq(fundingTx), ScriptFlags.STANDARD_SCRIPT_VERIFY_FLAGS)
+
+    val htlcTxs = Transactions.makeHtlcTxs(commitTx, outputs, commitmentFormat).sortBy(_.input.outPoint.index)
+    println(s"num_htlcs: ${htlcTxs.length}")
+    htlcTxs.foreach {
+      case tx: UnsignedHtlcTimeoutTx =>
+        val remoteHtlcSig = tx.localSig(remoteCommitmentKeys)
+        val withRemoteSig = tx.addRemoteSig(localCommitmentKeys, remoteHtlcSig)
+        val localHtlcSig = withRemoteSig.localSig(WalletInputs(Nil, None))
+        println(s"# signature for output #${tx.input.outPoint.index} (htlc-timeout for htlc #${tx.htlcId})")
+        println(s"local_htlc_signature: ${Scripts.der(localHtlcSig).dropRight(1).toHex}")
+        println(s"remote_htlc_signature: ${Scripts.der(remoteHtlcSig).dropRight(1).toHex}")
+      case tx: UnsignedHtlcSuccessTx =>
+        fail(s"unexpected htlc-success tx for htlc #${tx.htlcId}")
+    }
+  }
+
+  test("commitment_tx_with_asymmetric_dust_limits_anchor") {
+    val name = "commitment_tx_with_asymmetric_dust_limits_anchor"
+    val htlc = IncomingHtlc(UpdateAddHtlc(ByteVector32.Zeroes, 0, 15000000 msat, ByteVector32.Zeroes, CltvExpiry(500), TestConstants.emptyOnionPacket, None, Reputation.maxEndorsement, None))
+    val spec = CommitmentSpec(Set[DirectedHtlc](htlc), FeeratePerKw(15000 sat), toLocal = 7000000000L msat, toRemote = (3000000000L - 15000000L) msat)
+    val dustLimit = 546.sat
+
+    println()
+    println(s"name: $name")
+
+    val outputs = Transactions.makeCommitTxOutputs(
+      Local.funding_pubkey,
+      Remote.funding_pubkey,
+      localCommitmentKeys.publicKeys,
+      payCommitTxFees = true,
+      dustLimit,
+      Local.toSelfDelay,
+      spec,
+      commitmentFormat
+    )
+    val unsignedCommitTx = Transactions.makeCommitTx(
+      commitTxInput = commitmentInput,
+      commitTxNumber = Local.commitTxNumber,
+      localPaymentBasePoint = Local.payment_basepoint,
+      remotePaymentBasePoint = Remote.payment_basepoint,
+      localIsChannelOpener = true,
+      outputs = outputs)
+    val local_sig = unsignedCommitTx.sign(Local.funding_privkey, Remote.funding_pubkey)
+    val remote_sig = unsignedCommitTx.sign(Remote.funding_privkey, Local.funding_pubkey)
+    println(s"local_signature: ${Scripts.der(local_sig.sig).dropRight(1).toHex}")
+    println(s"remote_signature: ${Scripts.der(remote_sig.sig).dropRight(1).toHex}")
+    val commitTx = unsignedCommitTx.aggregateSigs(Local.funding_pubkey, Remote.funding_pubkey, local_sig, remote_sig)
+    Transaction.correctlySpends(commitTx, Seq(fundingTx), ScriptFlags.STANDARD_SCRIPT_VERIFY_FLAGS)
+
+    val htlcTxs = Transactions.makeHtlcTxs(commitTx, outputs, commitmentFormat).sortBy(_.input.outPoint.index)
+    println(s"num_htlcs: ${htlcTxs.length}")
+    htlcTxs.foreach {
+      case tx: UnsignedHtlcSuccessTx =>
+        val remoteHtlcSig = tx.localSig(remoteCommitmentKeys)
+        val redeemScript = Scripts.htlcReceived(localCommitmentKeys.publicKeys, tx.paymentHash, tx.htlcExpiry, commitmentFormat)
+        val localHtlcSig = Transaction.signInput(tx.tx, 0, redeemScript, fr.acinq.bitcoin.SigHash.SIGHASH_ALL, tx.input.txOut.amount, fr.acinq.bitcoin.SigVersion.SIGVERSION_WITNESS_V0, Local.htlc_privkey)
+        println(s"# signature for output #${tx.input.outPoint.index} (htlc-success for htlc #${tx.htlcId})")
+        println(s"local_htlc_signature: ${localHtlcSig.dropRight(1).toHex}")
+        println(s"remote_htlc_signature: ${Scripts.der(remoteHtlcSig).dropRight(1).toHex}")
+      case tx: UnsignedHtlcTimeoutTx =>
+        fail(s"unexpected htlc-timeout tx for htlc #${tx.htlcId}")
+    }
+  }
+
+  test("commitment_tx_with_main_output_trimmed_but_htlcs_present_anchor") {
+    val name = "commitment_tx_with_main_output_trimmed_but_htlcs_present_anchor"
+    val htlc = OutgoingHtlc(UpdateAddHtlc(ByteVector32.Zeroes, 0, 5000000 msat, ByteVector32.Zeroes, CltvExpiry(500), TestConstants.emptyOnionPacket, None, Reputation.maxEndorsement, None))
+    val spec = CommitmentSpec(Set(htlc), FeeratePerKw(15000 sat), toLocal = (10000000000L - 5000000L) msat, toRemote = 0 msat)
+    val dustLimit = 546.sat
+
+    println()
+    println(s"name: $name")
+
+    val outputs = Transactions.makeCommitTxOutputs(
+      Local.funding_pubkey,
+      Remote.funding_pubkey,
+      localCommitmentKeys.publicKeys,
+      payCommitTxFees = true,
+      dustLimit,
+      Local.toSelfDelay,
+      spec,
+      commitmentFormat
+    )
+    val unsignedCommitTx = Transactions.makeCommitTx(
+      commitTxInput = commitmentInput,
+      commitTxNumber = Local.commitTxNumber,
+      localPaymentBasePoint = Local.payment_basepoint,
+      remotePaymentBasePoint = Remote.payment_basepoint,
+      localIsChannelOpener = true,
+      outputs = outputs)
+    val local_sig = unsignedCommitTx.sign(Local.funding_privkey, Remote.funding_pubkey)
+    val remote_sig = unsignedCommitTx.sign(Remote.funding_privkey, Local.funding_pubkey)
+    println(s"local_signature: ${Scripts.der(local_sig.sig).dropRight(1).toHex}")
+    println(s"remote_signature: ${Scripts.der(remote_sig.sig).dropRight(1).toHex}")
+    val commitTx = unsignedCommitTx.aggregateSigs(Local.funding_pubkey, Remote.funding_pubkey, local_sig, remote_sig)
+    Transaction.correctlySpends(commitTx, Seq(fundingTx), ScriptFlags.STANDARD_SCRIPT_VERIFY_FLAGS)
+
+    val htlcTxs = Transactions.makeHtlcTxs(commitTx, outputs, commitmentFormat).sortBy(_.input.outPoint.index)
+    println(s"num_htlcs: ${htlcTxs.length}")
+    htlcTxs.foreach {
+      case tx: UnsignedHtlcTimeoutTx =>
+        val remoteHtlcSig = tx.localSig(remoteCommitmentKeys)
+        val withRemoteSig = tx.addRemoteSig(localCommitmentKeys, remoteHtlcSig)
+        val localHtlcSig = withRemoteSig.localSig(WalletInputs(Nil, None))
+        println(s"# signature for output #${tx.input.outPoint.index} (htlc-timeout for htlc #${tx.htlcId})")
+        println(s"local_htlc_signature: ${Scripts.der(localHtlcSig).dropRight(1).toHex}")
+        println(s"remote_htlc_signature: ${Scripts.der(remoteHtlcSig).dropRight(1).toHex}")
+      case tx: UnsignedHtlcSuccessTx =>
+        fail(s"unexpected htlc-success tx for htlc #${tx.htlcId}")
+    }
+  }
 }
```

</details>

Note: I split this into smaller commits, so some functions and fields temporarily use `#[allow(dead_code)]` so that each commit is clippy clean. These annotations are removed in later commits

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -858,7 +858,7 @@ fn build_open_channel(variables: &[Option<Variable>], inputs: &[usize]) -> OpenC
     }
 }
 
-/// Builds a `funding_created` message from 3 input variables.
+/// Builds a `funding_created` message from 4 input variables.
 ///
 /// Channel parameters are read from the negotiated `open_channel` and
 /// `accept_channel` messages recorded in `negotiations`, ensuring the
@@ -876,7 +876,8 @@ fn build_funding_created(
 ) -> Result<FundingCreated, ExecuteError> {
     let funding_tx = resolve_funding_transaction(variables, inputs[0]);
     let opener_funding_privkey_bytes = resolve_private_key(variables, inputs[1]);
-    let temporary_channel_id = resolve_channel_id(variables, inputs[2]);
+    let opener_htlc_basepoint_privkey_bytes = resolve_private_key(variables, inputs[2]);
+    let temporary_channel_id = resolve_channel_id(variables, inputs[3]);
 
     let funding_outpoint = OutPoint {
         txid: funding_tx.tx.compute_txid(),
@@ -913,11 +914,16 @@ fn build_funding_created(
     let secp = Secp256k1::new();
     let opener_funding_pubkey = PublicKey::from_secret_key(&secp, &opener_funding_privkey);
 
+    let opener_htlc_basepoint_privkey =
+        SecretKey::from_slice(&opener_htlc_basepoint_privkey_bytes).expect("valid private key");
+    let opener_htlc_basepoint = PublicKey::from_secret_key(&secp, &opener_htlc_basepoint_privkey);
+
     let opener = ChannelPartyConfig {
         funding_pubkey: opener_funding_pubkey,
         payment_basepoint: open_channel.payment_basepoint,
         revocation_basepoint: open_channel.revocation_basepoint,
         delayed_payment_basepoint: open_channel.delayed_payment_basepoint,
+        htlc_basepoint: opener_htlc_basepoint,
         dust_limit_satoshis: open_channel.dust_limit_satoshis,
         to_self_delay: open_channel.to_self_delay,
     };
@@ -926,6 +932,7 @@ fn build_funding_created(
         payment_basepoint: accept_channel.payment_basepoint,
         revocation_basepoint: accept_channel.revocation_basepoint,
         delayed_payment_basepoint: accept_channel.delayed_payment_basepoint,
+        htlc_basepoint: accept_channel.htlc_basepoint,
         dust_limit_satoshis: accept_channel.dust_limit_satoshis,
         to_self_delay: accept_channel.to_self_delay,
     };
@@ -947,6 +954,7 @@ fn build_funding_created(
     let holder = HolderIdentity {
         side: Side::Opener,
         funding_privkey: opener_funding_privkey,
+        htlc_basepoint_privkey: opener_htlc_basepoint_privkey,
     };
     let signature = config.sign_counterparty_commitment(&state, &holder);
 
@@ -3336,19 +3344,28 @@ mod tests {
     }
 
     fn send_funding_created_and_recv_funding_signed_instructions() -> Vec<Instruction> {
+        let opener_htlc_basepoint_privkey =
+            SecretKey::from_str("1111111111111111111111111111111111111111111111111111111111111111")
+                .unwrap()
+                .secret_bytes();
+
         let mut instrs = create_and_broadcast_tx_instructions();
         instrs.extend(vec![
+            Instruction {
+                operation: Operation::LoadPrivateKey(opener_htlc_basepoint_privkey),
+                inputs: vec![],
+            },
             Instruction {
                 operation: Operation::LoadChannelId([0xbb; 32]),
                 inputs: vec![],
             },
             Instruction {
                 operation: Operation::SendFundingCreated,
-                inputs: vec![6, 0, 8],
+                inputs: vec![6, 0, 8, 9],
             },
             Instruction {
                 operation: Operation::RecvFundingSigned,
-                inputs: vec![9],
+                inputs: vec![10],
             },
         ]);
         instrs
@@ -3412,6 +3429,10 @@ mod tests {
             side: Side::Acceptor,
             funding_privkey: SecretKey::from_str(
                 "1552dfba4f6cf29a62a0af13c8d6981d36d0ef8d61ba10fb0fe90da7634d7e13",
+            )
+            .unwrap(),
+            htlc_basepoint_privkey: SecretKey::from_str(
+                "4444444444444444444444444444444444444444444444444444444444444444",
             )
             .unwrap(),
         };
@@ -3668,13 +3689,13 @@ mod tests {
                 operation: Operation::SendChannelReady {
                     include_alias: false,
                 },
-                inputs: vec![10, 1, 11],
+                inputs: vec![11, 1, 12],
             },
             Instruction {
                 operation: Operation::SendChannelReady {
                     include_alias: true,
                 },
-                inputs: vec![10, 3, 11],
+                inputs: vec![11, 3, 12],
             },
         ]);
 
@@ -4004,7 +4025,7 @@ mod tests {
             },
             Instruction {
                 operation: Operation::SendFundingCreated,
-                inputs: vec![6, 0, 9],
+                inputs: vec![6, 0, 2, 9],
             },
             Instruction {
                 operation: Operation::RecvFundingSigned,
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -956,7 +956,8 @@ fn build_funding_created(
         funding_privkey: opener_funding_privkey,
         htlc_basepoint_privkey: opener_htlc_basepoint_privkey,
     };
-    let signature = config.sign_counterparty_commitment(&state, &holder);
+    let (signature, htlc_signature) = config.sign_counterparty_commitment(&state, &holder);
+    assert!(htlc_signature.is_empty()); // There are no HTLCs in the initial commitment transaction.
 
     let channel_id = ChannelId::v1_from_funding_outpoint(config.funding_outpoint);
 
@@ -1342,7 +1343,7 @@ fn verify_funding_signed(
 
     state
         .config
-        .verify_counterparty_signature(&state.commitment, &state.holder, &fs.signature)
+        .verify_counterparty_signature(&state.commitment, &state.holder, &fs.signature, &[])
         .then_some(())
         .ok_or(Violation::InvalidCounterpartySignature(fs.channel_id))
 }
@@ -3440,7 +3441,8 @@ mod tests {
         assert!(state.config.verify_counterparty_signature(
             &state.commitment,
             &holder,
-            &fc.signature
+            &fc.signature,
+            &[]
         ));
 
         let pending = executor
--- a/smite/src/channel_tx/commitment.rs
+++ b/smite/src/channel_tx/commitment.rs
@@ -165,6 +165,13 @@ pub struct CommitmentCost {
     pub anchor_cost_sat: u64,
 }
 
+/// A fully built commitment transaction with the metadata needed to construct
+/// and sign its second-stage HTLC transactions.
+struct BuiltCommitmentTx {
+    /// The assembled commitment transaction.
+    tx: Transaction,
+}
+
 /// State of a single channel, including its static configuration, holder
 /// identity, and current commitment state.
 pub struct ChannelState {
@@ -343,22 +350,21 @@ impl ChannelConfig {
         state: &CommitmentState,
         holder: &HolderIdentity,
     ) -> Signature {
-        let sighash = self.build_commitment_sighash(state, holder.counterparty_side());
-        sign(&sighash, &holder.funding_privkey)
+        let commitment = self.build_commitment_tx(state, holder.counterparty_side());
+        self.sign_commitment_tx(&commitment, &holder.funding_privkey)
     }
 
-    /// Verifies a signature received from the counterparty for the holder's
-    /// commitment transaction. Returns `true` if the signature is valid.
+    /// Verifies the counterparty's signatures on the holder's commitment
+    /// transaction.
     #[must_use]
     pub fn verify_counterparty_signature(
         &self,
         state: &CommitmentState,
         holder: &HolderIdentity,
-        signature: &Signature,
+        commit_sig: &Signature,
     ) -> bool {
-        let sighash = self.build_commitment_sighash(state, holder.side);
-        let counterparty = self.party(holder.counterparty_side());
-        verify(&sighash, signature, &counterparty.funding_pubkey)
+        let commitment = self.build_commitment_tx(state, holder.side);
+        self.verify_commit_sig(&commitment, holder, commit_sig)
     }
 
     /// Builds the signature for the holder's commitment transaction.
@@ -369,16 +375,16 @@ impl ChannelConfig {
         state: &CommitmentState,
         holder: &HolderIdentity,
     ) -> Signature {
-        let sighash = self.build_commitment_sighash(state, holder.side);
-        sign(&sighash, &holder.funding_privkey)
+        let commitment = self.build_commitment_tx(state, holder.side);
+        self.sign_commitment_tx(&commitment, &holder.funding_privkey)
     }
 
-    /// Builds the sighash for the commitment transaction. The commitment
-    /// format (legacy or anchor) is determined by the `channel_type`.
+    /// Builds the commitment transaction. The commitment format (legacy or
+    /// anchor) is determined by the `channel_type`.
     ///
     /// `local_side` selects whose commitment is built: the opener's or
     /// the acceptor's.
-    fn build_commitment_sighash(&self, state: &CommitmentState, local_side: Side) -> [u8; 32] {
+    fn build_commitment_tx(&self, state: &CommitmentState, local_side: Side) -> BuiltCommitmentTx {
         // Obscured commitment number.
         let obscuring_factor = compute_obscuring_factor(
             &self.opener.payment_basepoint,
@@ -416,6 +422,11 @@ impl ChannelConfig {
             output: outputs,
         };
 
+        BuiltCommitmentTx { tx }
+    }
+
+    /// Builds the sighash for the given commitment transaction.
+    fn build_commitment_sighash(&self, tx: &Transaction) -> [u8; 32] {
         // Funding output witness script.
         let funding_witness_script = build_funding_witness_script(
             &self.opener.funding_pubkey,
@@ -423,7 +434,7 @@ impl ChannelConfig {
         );
 
         // Compute the BIP143 sighash
-        let sighash = SighashCache::new(&tx)
+        let sighash = SighashCache::new(tx)
             .p2wsh_signature_hash(
                 0,
                 &funding_witness_script,
@@ -477,13 +488,6 @@ impl ChannelConfig {
                 value: Amount::from_sat(to_local_value),
                 script_pubkey: to_local_spk,
             });
-
-            if anchor {
-                outputs.push(TxOut {
-                    value: Amount::from_sat(ANCHOR_OUTPUT_VALUE),
-                    script_pubkey: build_anchor_scriptpubkey(&local.funding_pubkey),
-                });
-            }
         }
         if to_remote_value >= local.dust_limit_satoshis {
             let to_remote_spk = build_to_remote_scriptpubkey(&remote.payment_basepoint, anchor);
@@ -492,8 +496,17 @@ impl ChannelConfig {
                 value: Amount::from_sat(to_remote_value),
                 script_pubkey: to_remote_spk,
             });
+        }
 
-            if anchor {
+        if anchor {
+            if to_local_value >= local.dust_limit_satoshis {
+                outputs.push(TxOut {
+                    value: Amount::from_sat(ANCHOR_OUTPUT_VALUE),
+                    script_pubkey: build_anchor_scriptpubkey(&local.funding_pubkey),
+                });
+            }
+
+            if to_remote_value >= local.dust_limit_satoshis {
                 outputs.push(TxOut {
                     value: Amount::from_sat(ANCHOR_OUTPUT_VALUE),
                     script_pubkey: build_anchor_scriptpubkey(&remote.funding_pubkey),
@@ -509,6 +522,30 @@ impl ChannelConfig {
         });
 
         outputs
+    }
+
+    /// Signs the commitment transaction using the local party's funding private
+    /// key.
+    fn sign_commitment_tx(
+        &self,
+        commitment: &BuiltCommitmentTx,
+        funding_privkey: &SecretKey,
+    ) -> Signature {
+        let sighash = self.build_commitment_sighash(&commitment.tx);
+        sign(&sighash, funding_privkey)
+    }
+
+    /// Verifies the commitment signature against the counterparty's funding
+    /// public key.
+    fn verify_commit_sig(
+        &self,
+        commitment: &BuiltCommitmentTx,
+        holder: &HolderIdentity,
+        commit_sig: &Signature,
+    ) -> bool {
+        let sighash = self.build_commitment_sighash(&commitment.tx);
+        let counterparty = self.party(holder.counterparty_side());
+        verify(&sighash, commit_sig, &counterparty.funding_pubkey)
     }
 }
 
@@ -602,14 +639,12 @@ impl CommitmentState {
 impl Htlc {
     /// Returns whether this HTLC is offered on the commitment owned by
     /// `local_side` (otherwise it is received).
-    #[allow(dead_code)]
     fn is_offered(&self, local_side: Side) -> bool {
         self.offerer == local_side
     }
 
     /// Returns whether this HTLC would be trimmed from the commitment
     /// transaction due to dust limits.
-    #[allow(dead_code)]
     fn is_dust(
         &self,
         dust_limit_satoshis: u64,
@@ -675,7 +710,6 @@ fn total_anchors_sat(channel_type: &[u8]) -> u64 {
 
 /// Get the fee cost of a second-stage HTLC transaction in satoshis.
 /// `is_offered` selects between the HTLC-timeout and HTLC-success weights.
-#[allow(dead_code)]
 fn htlc_tx_fee_sat(channel_type: &[u8], feerate_per_kw: u32, is_offered: bool) -> u64 {
     if supports_option_anchors(channel_type) {
         return 0;
--- a/smite/src/channel_tx/commitment.rs
+++ b/smite/src/channel_tx/commitment.rs
@@ -3,6 +3,7 @@
 use super::funding::build_funding_witness_script;
 
 use bitcoin::absolute::LockTime;
+use bitcoin::hashes::ripemd160::Hash as Ripemd160;
 use bitcoin::hashes::sha256::Hash as Sha256;
 use bitcoin::hashes::{Hash, HashEngine};
 use bitcoin::opcodes::all as opcodes;
@@ -12,7 +13,8 @@ use bitcoin::secp256k1::{Message, PublicKey, Scalar, Secp256k1, SecretKey};
 use bitcoin::sighash::{EcdsaSighashType, SighashCache};
 use bitcoin::transaction::Version;
 use bitcoin::{
-    Amount, CompressedPublicKey, OutPoint, ScriptBuf, Sequence, Transaction, TxIn, TxOut, Witness,
+    Amount, CompressedPublicKey, OutPoint, PubkeyHash, ScriptBuf, Sequence, Transaction, TxIn,
+    TxOut, Witness,
 };
 
 /// Anchor output value in satoshis.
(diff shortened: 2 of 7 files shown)


smite/src/channel_tx/commitment.rs
```diff
@@ -398,82 +587,401 @@ impl ChannelConfig {
...
+
+    /// Settles the in-flight HTLC that `offerer` added with the given `id`,
+    /// removing it from the in-flight set and crediting its amount to the
+    /// receiver's balance.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`CommitmentError::HtlcNotFound`] if no in-flight HTLC matches
+    /// `id` and `offerer`.
+    pub fn fulfill_htlc(&mut self, id: u64, offerer: Side) -> Result<(), CommitmentError> {
```
erickcestari, line 902: `fulfill_htlc`, `fail_htlc`, `update_fee`, `update_per_commitment_point` and `advance_commitment_number` are missing unit tests.

smite/src/channel_tx/commitment.rs
```diff
@@ -87,6 +111,25 @@ pub struct ChannelConfig {
     pub minimum_depth: u32,
 }
 
+/// An in-flight HTLC that appears (subject to dust trimming) as an output in
+/// the commitment transaction.
+#[derive(Clone, Copy)]
+pub struct Htlc {
+    /// HTLC ID, unique per channel and offering direction.
+    id: u64,
```
erickcestari, line 119: The `Htlc` cannot be built outside the `commitment` module. 

```suggestion
    pub id: u64,
```

smite/src/channel_tx/commitment.rs
```diff
@@ -814,249 +1524,1723 @@ mod tests {
...
+    // name: commitment tx with six outputs untrimmed (minimum feerate) (BOLT 3 Appendix C)
+    #[test]
+    fn commitment_tx_with_six_outputs_untrimmed_minimum_feerate_legacy() {
+        let (chan_config, mut commitment_params, opener_holder, acceptor_holder) =
+            bolt3_commitment_params(648, 6_993_000_000, 3_006_000_000, 546, vec![]);
+        let htlcs = bolt3_htlc_list();
+        commitment_params.add_htlc(htlcs[1]).unwrap();
+        commitment_params.add_htlc(htlcs[2]).unwrap();
+        commitment_params.add_htlc(htlcs[3]).unwrap();
+        commitment_params.add_htlc(htlcs[4]).unwrap();
```
erickcestari, line 1708: Shouldn't all of these `minimum_feerate` tests add the 5 HTLCs to actually verify that the HTLCs are being trimmed correctly?

```suggestion
        let htlcs = bolt3_htlc_list();
        commitment_params.add_htlc(htlcs[0]).unwrap();
        commitment_params.add_htlc(htlcs[1]).unwrap();
        commitment_params.add_htlc(htlcs[2]).unwrap();
        commitment_params.add_htlc(htlcs[3]).unwrap();
        commitment_params.add_htlc(htlcs[4]).unwrap();
```

smite-scenarios/src/executor.rs
```diff
@@ -913,11 +914,16 @@ fn build_funding_created(
     let secp = Secp256k1::new();
     let opener_funding_pubkey = PublicKey::from_secret_key(&secp, &opener_funding_privkey);
 
+    let opener_htlc_basepoint_privkey =
+        SecretKey::from_slice(&opener_htlc_basepoint_privkey_bytes).expect("valid private key");
+    let opener_htlc_basepoint = PublicKey::from_secret_key(&secp, &opener_htlc_basepoint_privkey);
```
erickcestari, line 919: Shouldn't we use the same `htlc_basepoint` that we sent to the target in the `open_channel` message?

smite/src/channel_tx/commitment.rs
```diff
@@ -398,82 +587,401 @@ impl ChannelConfig {
...
+    }
+
+    /// Adds `htlc` to the in-flight set, debiting its amount from the offerer's
+    /// balance.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`CommitmentError::HtlcExceedsBalance`] if the HTLC amount
+    /// would underflow the offerer's balance.
+    pub fn add_htlc(&mut self, htlc: Htlc) -> Result<(), CommitmentError> {
```
erickcestari, line 885: We could either return an error or assert when an attempt is made to add a duplicate `Htlc`.

smite-scenarios/src/executor.rs
```diff
@@ -947,8 +954,10 @@ fn build_funding_created(
     let holder = HolderIdentity {
         side: Side::Opener,
         funding_privkey: opener_funding_privkey,
+        htlc_basepoint_privkey: opener_htlc_basepoint_privkey,
     };
-    let signature = config.sign_counterparty_commitment(&state, &holder);
+    let (signature, htlc_signature) = config.sign_counterparty_commitment(&state, &holder);
+    assert!(htlc_signature.is_empty()); // There are no HTLCs in the initial commitment transaction.
```
erickcestari, line 960: nit:

```suggestion
    let (signature, htlc_signatures) = config.sign_counterparty_commitment(&state, &holder);
    assert!(htlc_signatures.is_empty()); // There are no HTLCs in the initial commitment transaction.
```

smite/src/channel_tx/commitment.rs
```diff
@@ -814,249 +1524,1723 @@ mod tests {
...
+            bolt3_commitment_params(0, 6_993_000_000, 3_007_000_000, 546, vec![]);
+        let htlcs = bolt3_htlc_list();
+        commitment_params.add_htlc(htlcs[0]).unwrap();
+        commitment_params.add_htlc(htlcs[1]).unwrap();
+        commitment_params.add_htlc(htlcs[2]).unwrap();
+        commitment_params.add_htlc(htlcs[3]).unwrap();
+        commitment_params.add_htlc(htlcs[4]).unwrap();
 
         // Opener signs own commitment.
+        let (local_signature, local_htlc_signsignature) =
```
erickcestari, line 1554: nit:

```suggestion
        let (local_signature, local_htlc_signatures) =
```

erickcestari (changes requested):
Nice PR! I've added some notes:

smite-scenarios/src/executor.rs
NishantBansal2003: Yes, but with this, we can also execute invalid signature cases via `opener_htlc_basepoint_privkey`, and invalid script cases via `opener_htlc_basepoint`, since the latter was used in `build_htlc_witness_script`

But I don't think it helps us much, as `opener_funding_pubkey` used to, since the signature is already invalid due to the invalid private key. So I think we can use `open_channel.htlc_basepoint` instead

smite/src/channel_tx/commitment.rs
NishantBansal2003: Currently, we are the initiator here (In the future, when the target might be the initiator, we should definitely sanitize the result in the oracle itself before coming here), If we prohibit the duplicate HTLC case, we won't be able to test that case with the target. So I think we should just add it (it's not an issue in the commitment tx anyway), and if the peer's response is not an error/warning, that should be handled in the oracle.

I was also thinking of removing the `HtlcExceedsBalance` case as well, similar to how we handled the commitment fee case in the commitment tx with `saturating_sub` (Will also add a comment discussing the edge cases that are not handled here and should be handled only by the oracle)

smite/src/channel_tx/commitment.rs
```diff
@@ -398,82 +587,401 @@ impl ChannelConfig {
...
+    /// # Errors
+    ///
+    /// Returns [`CommitmentError::HtlcNotFound`] if no in-flight HTLC matches
+    /// `id` and `offerer`.
+    pub fn fulfill_htlc(&mut self, id: u64, offerer: Side) -> Result<(), CommitmentError> {
+        let pos = self
+            .htlcs
+            .iter()
+            .position(|h| h.id == id && h.offerer == offerer)
+            .ok_or(CommitmentError::HtlcNotFound)?;
```
NishantBansal2003, line 907: I think, based on the discussion in https://github.com/lnfuzz/smite/pull/181#discussion_r3825882289, there might be a comment suggesting that we should remove the `HtlcNotFound` error and make it a hard assert, since this will also be checked by the oracle before getting here. But I think we should keep this until we have those oracles

smite-scenarios/src/executor.rs
erickcestari: Yes. I'm also not sure. It feels interesting that we have the possibility to test those edge cases. But at the same time the fuzzer might expend more time on it then it should.

smite/src/channel_tx/commitment.rs
erickcestari: Makes sense to me!

PR #182, opened 2026-07-30
smitebot: treat alive runners as started when verify_startup times out

### LINKED ISSUE #176: smitebot: `start` times out on large seed corpus
Attempting to launch a campaign with a large seed corpus causes the startup timeout to be exceeded, and `smitebot start` then reports a failure.  In reality, the tmux session is still live and can be attached to without any issues, but the failure message is still confusing for the user.

We either need a way to detect that the runner is in "dry-run" mode on the seed corpus, or perhaps we can just remove the timeout logic and attach immediately to the tmux session (as `aflr` does).

Ashish-Kumar-Dash:
## Problem

`fuzzer_stats` is only written after AFL++ finishes calibrating all seeds. Under Nyx, each seed exec restores a VM snapshot, so a large corpus can blow past the 10-minute deadline before the file appears. At deadline the old code returned `false` unconditionally, triggering `fail_campaign` even though every runner window was still alive.

Fixes #176 .

  ## Fix

On deadline, call `alive_windows` for an authoritative liveness check instead of failing immediately:
  - Windows still alive, campaign marked `running`, per-runner `warn` logged with the session name to attach to. PIDs stay `None`; `smitebot stop` reads them from tmux pane PIDs at teardown (`list_pane_pids`).
  - Session gone or a window missing is marked `false` as before.

Genuine failures (window dies before `fuzzer_stats` appears) are still caught immediately mid-loop via `dead_windows`, unchanged.

### DIFF
--- a/smitebot/src/commands/start.rs
+++ b/smitebot/src/commands/start.rs
@@ -18,15 +18,15 @@ use crate::state::{CampaignState, RunnerState, Status};
 use crate::tmux;
 use crate::utils::shell_quote;
 
-/// Safety ceiling for startup verification.
+/// How long to wait for `fuzzer_stats` before treating alive runners as started.
 ///
-/// Genuine launch failures are caught immediately when a runner's tmux window
-/// dies (see `verify_startup`), so this only bounds an alive-but-hung runner.
-/// Nyx seed calibration can take several minutes because each exec restores a
-/// VM snapshot, so the ceiling is deliberately generous; its exact value is not
-/// load-bearing — it only prevents an indefinite hang. Kept comfortably above a
-/// measured fresh start (~45s) so a large seed corpus is not falsely failed.
-const VERIFY_TIMEOUT: Duration = Duration::from_mins(10);
+/// This is not a correctness gate: a runner whose window dies is caught within
+/// one `VERIFY_POLL_INTERVAL` (see `verify_startup`), and a large seed corpus can
+/// take minutes to calibrate before `fuzzer_stats` appears. So on deadline, live
+/// windows are handed off as running rather than failed. The value only bounds
+/// how long we block the terminal before that hand-off; kept short so a slow
+/// calibration is not held hostage to a long wait.
+const VERIFY_TIMEOUT: Duration = Duration::from_mins(1);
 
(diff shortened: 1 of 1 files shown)


NishantBansal2003:
WDYT about this?

> we can just remove the timeout logic and attach immediately to the tmux session (as aflr does).

I think we originally added the 10min timeout to account for seed calibration, waiting up to 10min before assuming the seeds are still calibrating seems too long now. I think we should either lower the timeout or attach immediately, WDYT?

Ashish-Kumar-Dash:
Lowered VERIFY_TIMEOUT to 60s

smitebot/src/commands/start.rs
```diff
@@ -372,16 +371,37 @@ fn verify_startup(
         thread::sleep(VERIFY_POLL_INTERVAL);
     }
 
+    // Deadline reached. Check liveness one final time: alive_windows errors when
+    // the session is gone, so an Ok result is authoritative. Runners still alive
+    // are calibrating a large seed corpus; treat the campaign as running. PIDs
+    // remain None and `smitebot stop` reads them from tmux pane PIDs at teardown.
+    let alive = match tmux::alive_windows(session) {
```
NishantBansal2003, line 378: This will be executed after the deadline, and just before the deadline, we already do `dead_windows` check. So, if no runner had exited just before the deadline (in `dead_windows`), I don’t think anyone will exit just after it. So, the `alive_windows` check seems redundant, but I think there is a small race that could happen, so it's fine to keep this check here

NishantBansal2003 (approved):
Verified with 2 runners, and it works fine:

```sh
INFO  [smitebot::commands::start] starting 2 runners in tmux session 'cln-ir-1785826380'
INFO  [smitebot::commands::start] verifying runners started
WARN  [smitebot::commands::start] runner 0 is still calibrating seeds; attach to session 'cln-ir-1785826380' to monitor
WARN  [smitebot::commands::start] runner 1 is still calibrating seeds; attach to session 'cln-ir-1785826380' to monitor
INFO  [smitebot::commands::start] campaign cln-ir-1785826380 is running
```

morehouse (approved):

PR #183, opened 2026-07-30
workloads: bump target versions

morehouse:
```
CLN v26.06.2 -> v26.06.6
Eclair v0.14.0 -> v0.14.1
LDK v0.2.1 -> v0.2.4
```

Smoke tested with ~1h IR fuzzing campaigns against all three targets.  No issues encountered.


NishantBansal2003 (approved):
LGTM!

Issue #174, opened 2026-07-29
smitebot corpus merge + smitebot corpus minimize


Ashish-Kumar-Dash:
## Problem

After a multi-runner campaign, each AFL++ instance writes its finds to its own `output_dir/<n>/queue/`. Before reusing the corpus for a new campaign, cross-campaign analysis, or hand-off, two manual steps are needed:

  1. Collect unique inputs across all runner queues (currently: `cp output_dir/*/queue/* merged/`)
  2. Prune inputs that don't add new coverage (currently: manual `afl-cmin` invocation)

  This issue presents the workflow and use-cases for adding both as smitebot commands.

  ---

  ## CLI

  smitebot corpus merge  [...] [-o ]
  smitebot corpus minimize  [-i ] [-o ]

  `corpus` would be a new top-level clap variant wrapping a `CorpusCommand` subcommand enum. 

  ---

  ## `corpus merge`

For each `<id>`, reads `output_dir` and runner list from `~/.smitebot/runs/<id>/state.json`. Walks `output_dir/<n>/queue/` across every runner. SHA-256 deduplicates by content (same bytes, different filename = one copy). Copies unique files to output.

  - Default output: `~/.smitebot/runs/<first-id>/corpus/`
  - Accepts multiple campaign IDs and replaces the manual `output_dir/*/queue/` glob across campaigns

  ---

  ## `corpus minimize`

  Reads `output_dir`, `sharedir`, and `aflpp_path` from `~/.smitebot/runs/<id>/state.json`. Invokes:

  afl-cmin -X -i  -o  --

 `afl-cmin -X` calls `afl-showmap` internally, restores the Nyx VM snapshot per input, scores coverage, drops inputs made redundant by later finds. The sharedir is self-contained and survives `smitebot stop`, so no live campaign is needed.

  - Default input: `~/.smitebot/runs/<id>/corpus/` (merge output)
  - Default output: `~/.smitebot/runs/<id>/corpus-min/`
  - Overridable via `-i` / `-o` for externally merged corpora
  - Streams `afl-cmin` output to terminal (`.status()`, not captured)

  ---

  ## Possible Workflows

  ```bash
  # single campaign — merge only
  smitebot corpus merge lnd-enc-123

  # single campaign — merge then minimize
  smitebot corpus merge lnd-enc-123
  smitebot corpus minimize lnd-enc-123

  # cross-campaign merge (replaces output_dir/*/queue/ glob)
[...]

PR #184, opened 2026-08-01
smitebot: implement corpus merge and minimize commands

Ashish-Kumar-Dash:
## Problem

After a multi-runner campaign, collecting unique inputs across runner queues and pruning redundant ones requires manual shell scripting (`cp output_dir/*/queue/* ...`, `afl-cmin ...`). These commands automate both steps.

Ref. #174 #70 

  ## Changes

  **`smitebot corpus merge <id> [<id>...] [-o <dir>]`**
  Walks `output_dir/<n>/queue/` for every runner across all specified campaigns, deduplicates by content, and copies unique files to the output directory. Accepts multiple campaign IDs to merge across independent runs.
  Default output: `~/.smitebot/runs/<first-id>/corpus/`. No target binary required.

  **`smitebot corpus minimize <id> [-i <input>] [-o <dir>]`**
  Runs `afl-cmin -X -i <input> -o <output> -- <sharedir>`. Reads `sharedir` and `aflpp_path` from `state.json`; no live campaign required. Default input: `<output_dir>/*/queue/`. Default output: `~/.smitebot/runs/<id>/corpus-min/`.

  **`state.json`**: adds `aflpp_path: Option<PathBuf>` (`#[serde(default)]` for backward compat with existing state files).

### DIFF
--- a/smite-nyx-sys/src/nyx-crash-handler.c
+++ b/smite-nyx-sys/src/nyx-crash-handler.c
@@ -34,6 +34,8 @@
 #include "nyx.h"
 #endif
 
+// Must match PANIC_LOG_PATH in workloads/ldk/src/main.rs.
+#define PANIC_LOG_PATH "/tmp/smite-panic.log"
 #define ASAN_LOG_PATH "/tmp/asan.log"
 #define MAX_CUSTOM_BACKTRACE_SIZE 50
 
@@ -58,11 +60,11 @@ void append_log(const char *msg) {
   strcat(log, msg);
 }
 
-// Fetch the ASan log from the log file and append it to the global log
-void append_asan_log() {
+// Fetch a report the target left behind at "<path>.<pid>" and append it to the
+// global log
+void append_target_log(const char *path) {
   char log_file_path[256];
-  snprintf(log_file_path, sizeof(log_file_path), "%s.%d", ASAN_LOG_PATH,
-           getpid());
+  snprintf(log_file_path, sizeof(log_file_path), "%s.%d", path, getpid());
 
   FILE *file = fopen(log_file_path, "r");
 
@@ -121,7 +123,8 @@ void panic_with_backtrace(const char *extra_msg) {
     append_log("\n");
   }
 
-  append_asan_log();
+  append_target_log(PANIC_LOG_PATH);
+  append_target_log(ASAN_LOG_PATH);
 
 #ifdef CUSTOM_BACKTRACE
   char custom_backtrace[0x10000];
--- a/workloads/ldk/src/main.rs
+++ b/workloads/ldk/src/main.rs
@@ -27,7 +27,27 @@ extern "C" fn handle_block(_: libc::c_int) {
     NEW_BLOCK.store(true, Ordering::SeqCst);
 }
 
+/// Path prefix the crash handler reads panic reports from.
+const PANIC_LOG_PATH: &str = "/tmp/smite-panic.log";
+
+/// Stores the panic message where the LD_PRELOADed crash handler can retrieve
+/// it. With `panic = "abort"`, panics only reach stderr, which scenario/Nyx
+/// discards,so this allows crash reports to include the panic message and its
+/// source location.
+fn install_panic_hook() {
+    let path = format!("{PANIC_LOG_PATH}.{}", std::process::id());
+    let default_hook = std::panic::take_hook();
+
+    std::panic::set_hook(Box::new(move |info| {
+        // Renders as "panicked at <file>:<line>:<col>:\n<message>".
+        let _ = std::fs::write(&path, format!("{info}\n"));
+        default_hook(info);
+    }));
+}
+
 fn main() {
+    install_panic_hook();
+
     // Set up signal handlers for graceful shutdown
     unsafe {
         // Cast through pointer to satisfy Rust's function-to-integer cast rules


smitebot/src/state.rs
```diff
@@ -41,6 +41,9 @@ pub struct CampaignState {
     pub tmux_session: String,
     /// State of each AFL++ runner process.
     pub runners: Vec<RunnerState>,
+    /// Path to the AFL++ source tree; used by `corpus minimize` to locate `afl-cmin`.
+    #[serde(default)]
```
NishantBansal2003: not needed, since the field type is `Option`

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,308 @@
...
+    Minimize(MinimizeArgs),
+}
+
+/// CLI arguments for `smitebot corpus merge`.
+#[derive(Debug, Args)]
+pub struct MergeArgs {
+    /// Campaign IDs whose runner queues to merge (directories under `~/.smitebot/runs`).
+    #[arg(required = true)]
+    campaign_ids: Vec<String>,
+    /// Output directory (default: `~/.smitebot/runs/<first-id>/corpus/`).
```
NishantBansal2003: I think it would be better to make it required, since using `<first-id>` as the default doesn't seem to make much sense

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,308 @@
...
+    #[arg(short = 'o', long)]
+    output: Option<PathBuf>,
+}
+
+/// CLI arguments for `smitebot corpus minimize`.
+#[derive(Debug, Args)]
+pub struct MinimizeArgs {
+    /// Campaign ID whose sharedir and `aflpp_path` to use.
```
NishantBansal2003: I think we should have an optional `aflpp_path` flag here as well for both cmds, It would be useful if the path has changed from what's stored in `states.json`, as the flag could override the stored value

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,308 @@
...
+        let state_path = runs_dir.join(campaign_id).join("state.json");
+        let state = match CampaignState::load(&state_path) {
+            Ok(s) => s,
+            Err(e) => {
+                log::error!("{e}");
+                log::error!(
+                    "campaign '{campaign_id}' not found; list campaigns with: ls {}",
+                    runs_dir.display()
+                );
+                return false;
```
NishantBansal2003: If the user passes a wrong id, we'll be left with partial merge files. I think we should validate the input before starting the merge

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,308 @@
...
+
+            for entry in entries.flatten() {
+                let path = entry.path();
+                if !path.is_file() {
+                    continue;
+                }
+                let contents = match fs::read(&path) {
+                    Ok(c) => c,
+                    Err(e) => {
+                        log::error!("failed to read {}: {e}", path.display());
```
NishantBansal2003: In any error or failure case, I don't think we should be left with partial merge files, we should either remove them on error or log a message asking the user to remove them

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,308 @@
...
+                        return false;
+                    }
+                };
+                total_in += 1;
+                if seen.insert(contents) {
```
NishantBansal2003: I think this dedupe strategy could use a lot of memory if the seed is very large. We probably shouldn't store the entire file contents, instead, we could deduplicate based on a content hash, that way, we only need to store 32 bytes per entry rather than the entire file

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,308 @@
...
+    log::info!("running afl-cmin on {input}");
+    log::info!("output: {}", output.display());
+
+    let status = match Command::new(&afl_cmin)
+        .arg("-X")
+        .arg("-i")
+        .arg(&input)
+        .arg("-o")
+        .arg(&output)
+        .arg(&state.sharedir)
```
NishantBansal2003:
```suggestion
        .arg("-i")
        .arg(&input)
        .arg("-o")
        .arg(&output)
        .arg("-X")
        .arg(&state.sharedir)
```

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,308 @@
...
+        // Either found on PATH or not — just confirm no panic.
+        let _ = result;
+    }
+
+    #[test]
+    fn minimize_default_input_uses_output_dir_glob() {
+        let output_dir = PathBuf::from("/tmp/smite-out");
+        let resolved = format!("{}/*/queue/", output_dir.display());
+        assert_eq!(resolved, "/tmp/smite-out/*/queue/");
+    }
```
NishantBansal2003: test name doesn't seem to match what it’s testing?

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,308 @@
...
+#[cfg(test)]
+mod tests {
+    use std::fs;
+
+    use tempfile::tempdir;
+
+    use super::*;
+
+    #[test]
+    fn merge_copies_unique_files_from_queue() {
```
NishantBansal2003: This test is reimplementing the logic of `execute_merge` instead of actually calling `execute_merge`, so it could pass even if `execute_merge` itself is broken

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,308 @@
...
+        fs::write(&cmin, "#!/bin/sh").unwrap();
+
+        // File exists at aflpp_path — should be returned even if not executable.
+        // (afl-cmin.py is a Python script, not necessarily +x in all installs.)
+        let result = find_afl_cmin(Some(dir.path()));
+        assert_eq!(result, Some(cmin));
+    }
+
+    #[test]
+    fn find_afl_cmin_falls_back_to_path_when_aflpp_missing() {
```
NishantBansal2003: This test doesn't really test the fallback to `PATH`, so it doesn’t seem very useful to me

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    #[test]
+    fn minimize_default_input_uses_output_dir_glob() {
+        let output_dir = PathBuf::from("/tmp/smite-out");
+        let resolved = format!("{}/*/queue/", output_dir.display());
+        assert_eq!(resolved, "/tmp/smite-out/*/queue/");
+    }
+
+    #[test]
+    fn minimize_default_output_is_corpus_min() {
```
NishantBansal2003: test name doesn't seem to match what it’s testing?

NishantBansal2003 (changes requested):

smitebot/src/commands/corpus.rs
Ashish-Kumar-Dash: Left it off merge since merge never calls afl-cmin (it's a pure file dedup/copy), so there's nothing for the path to point at. Do let me know your thoughts on this

smitebot/src/commands/corpus.rs
```diff
@@ -178,8 +210,13 @@ fn execute_minimize(args: &MinimizeArgs) -> bool {
...
     };
 
-    let Some(afl_cmin) = find_afl_cmin(state.aflpp_path.as_deref()) else {
-        log::error!("afl-cmin not found; set aflpp_path in campaign.toml or add AFL++ to PATH");
+    // The --aflpp flag overrides the path recorded at campaign start, for when
+    // the AFL++ checkout has since moved.
+    let aflpp = args.aflpp.as_deref().or(state.aflpp_path.as_deref());
+    let Some(afl_cmin) = find_afl_cmin(aflpp) else {
+        log::error!(
+            "afl-cmin not found; pass --aflpp <path>, set aflpp_path in campaign.toml, or add AFL++ to PATH"
```
NishantBansal2003: I'm sure we're not passing or reading the `campaign.toml` path here

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+
+    if let Err(e) = fs::create_dir_all(&args.output) {
+        log::error!(
+            "failed to create output directory {}: {e}",
+            args.output.display()
+        );
+        return false;
+    }
```
NishantBansal2003: I think we should error out and refuse to merge into a directory that already contains corpus files to prevent accidental overwrites, the user must either choose a new directory or remove the existing corpus

NishantBansal2003 (changes requested):

smitebot/src/commands/corpus.rs
NishantBansal2003: Same applies to the minimize as well

NishantBansal2003:
I'm not sure if this is just an issue on my end, but I get the following error both when running `afl-cmin` manually and through `smitebot`:

```sh
[-] PROGRAM ABORT : Library 'libnyx.so' not found
         Location : find_afl_binary(), src/afl-common.c:574
```
I need to set `AFL_PATH=~/AFLplusplus` for it to work, **if others** are seeing the same issue?, I think we should set `AFL_PATH` as well in cmd

Ashish-Kumar-Dash:
> I'm not sure if this is just an issue on my end, but I get the following error both when running `afl-cmin` manually and through `smitebot`:
> 
> ```shell
> [-] PROGRAM ABORT : Library 'libnyx.so' not found
>          Location : find_afl_binary(), src/afl-common.c:574
> ```
> 
> I need to set `AFL_PATH=~/AFLplusplus` for it to work, **if others** are seeing the same issue?, I think we should set `AFL_PATH` as well in cmd

Yes, it was a genuine error, fixed in recent commit

NishantBansal2003 (approved):

smitebot/README.md
```diff
@@ -165,3 +165,29 @@ smitebot print-ir output/default/crashes/id:000000,sig:06,...
...
+
+Collects all queue files from one or more campaign runner directories, deduplicates by content, and writes unique files to an output directory. Accepts multiple campaign IDs to merge corpora across independent runs.
+
+```bash
+smitebot corpus merge <campaign-id> -o <output-dir>
+smitebot corpus merge <campaign-id-1> <campaign-id-2> ... -o <output-dir>
+```
+
+- `<campaign-id>` — directory name(s) under `~/.smitebot/runs`
+- `-o, --output <output-dir>` — output directory for the merged corpus (required)
```
morehouse: Nit: current convention in the README is to use colons, not emdashes


```suggestion
- `<campaign-id>`: directory name(s) under `~/.smitebot/runs`
- `-o, --output <output-dir>`: output directory for the merged corpus (required)
```

Also applies below.

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+                    Ok(c) => c,
+                    Err(e) => {
+                        log::error!("failed to read {}: {e}", path.display());
+                        return None;
+                    }
+                };
+                total_in += 1;
+                if seen.insert(content_hash(&contents)) {
+                    let dest = output.join(format!("{total_out:06}"));
+                    if let Err(e) = fs::copy(&path, &dest) {
```
morehouse: Nit: we already have the bytes in `contents`.  We could just directly write them to `dest` rather than rereading via `copy`.

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+    cmd.arg("-i")
+        .arg(&input)
+        .arg("-o")
+        .arg(&output)
+        .arg("-X")
+        .arg(&state.sharedir);
+
+    if let Some(afl_root) = afl_cmin.parent() {
+        cmd.env("AFL_PATH", afl_root);
+    }
```
morehouse: Does this actually provide benefit?  `afl-cmin` should automatically find `libnyx.so` when it is in the same directory as `afl-cmin`.  And if `afl-cmin` is installed on PATH, then `afl_root` probably points to the `bin/` directory, not the `lib/` directory where libnyx would be.

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+
+/// Reports (with an error log) whether `output` already contains files.
+///
+/// Both subcommands refuse a populated output directory so an existing corpus
+/// is never mixed into or overwritten. A `read_dir` error (e.g. the directory
+/// does not exist yet) counts as unoccupied.
+fn output_dir_occupied(output: &Path) -> bool {
+    if output
+        .read_dir()
+        .is_ok_and(|mut entries| entries.next().is_some())
```
morehouse: This counts any `read_dir` error as "unoccupied".  There's edge cases where this is incorrect (e.g., permissions issues).

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    #[test]
+    fn merge_states_deduplicates_across_campaigns() {
+        let dir = tempdir().unwrap();
+        let out_a = dir.path().join("a");
+        let out_b = dir.path().join("b");
+        // Same input present in both campaigns collapses to one.
+        write_queue(&out_a, 0, &[("id:0", "shared")]);
+        write_queue(&out_b, 0, &[("id:0", "shared"), ("id:1", "unique")]);
```
morehouse: Nit: we could also name the inputs differently to make sure we're deduping on content, not name.


```suggestion
        write_queue(&out_a, 0, &[("id:2", "shared")]);
        write_queue(&out_b, 0, &[("id:0", "shared"), ("id:1", "unique")]);
```

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    #[test]
+    fn find_afl_cmin_prefers_aflpp_path_over_system() {
+        let dir = tempdir().unwrap();
+        let cmin = dir.path().join("afl-cmin");
+        fs::write(&cmin, "#!/bin/sh").unwrap();
+
+        // File exists at aflpp_path — should be returned even if not executable.
+        // (afl-cmin.py is a Python script, not necessarily +x in all installs.)
```
morehouse: Sure, but the actual `afl-cmin` script should always be executable, right?

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+        write_queue(&output_dir, 0, &[("id:0", "aaa"), ("id:1", "bbb")]);
+        write_queue(&output_dir, 1, &[("id:0", "aaa"), ("id:1", "ccc")]);
+        let out = dir.path().join("corpus");
+        fs::create_dir_all(&out).unwrap();
+
+        let (total_in, total_out) = merge_states(&[sample_state(output_dir, 2)], &out).unwrap();
+
+        assert_eq!(total_in, 4);
+        assert_eq!(total_out, 3);
+        assert_eq!(fs::read_dir(&out).unwrap().count(), 3);
```
morehouse: Let's also verify the file contents in these tests.

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+
+/// Locates `afl-cmin`: checks `aflpp_path` first, falls back to searching `$PATH`.
+fn find_afl_cmin(aflpp_path: Option<&Path>) -> Option<PathBuf> {
+    if let Some(path) = aflpp_path {
+        let candidate = path.join("afl-cmin");
+        if candidate.exists() {
+            return Some(candidate);
+        }
+    }
+    find_in_path("afl-cmin")
```
morehouse: Perhaps we should check that `afl-cmin` is executable too, and if not then fall back to PATH


```suggestion
    if let Some(path) = aflpp_path {
        let candidate = path.join("afl-cmin");
        if is_executable(&candidate) {
            return Some(candidate);
        }
    }
    find_in_path("afl-cmin")
```

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+    log::info!("running afl-cmin on {input}");
+    log::info!("output: {}", output.display());
+
+    let mut cmd = Command::new(&afl_cmin);
+    cmd.arg("-i")
+        .arg(&input)
+        .arg("-o")
+        .arg(&output)
+        .arg("-X")
+        .arg(&state.sharedir);
```
morehouse: When I try to directly use `smitebot corpus minimize` without a merge first, I get errors due to the pathnames being too long.  Perhaps we could handle that better in a future PR (e.g., rename the corpus files before minimizing).

```
OSError: [Errno 36] File name too long: '/home/m/Code/smite/ldk-test-output/0/queue/id:000121,src:000002,time:49073,execs:6254,smite-ir:instr-reorder,instr-reorder,input-swap,input-swap,instr-delete,gen-insert,op-param,op-param,instr-delete,instr-delete,instr-reorder,instr-reorder,instr-delete,input-swap,instr-reorder,instr-reorde,+cov' -> 'ldk-minimized/.filelist.1086221.d/06abd306_id:000121,src:000002,time:49073,execs:6254,smite-ir:instr-reorder,instr-reorder,input-swap,input-swap,instr-delete,gen-insert,op-param,op-param,instr-delete,instr-delete,instr-reorder,instr-reorder,instr-delete,input-swap,instr-reorder,instr-reorde,+cov'
```

smitebot/src/commands/corpus.rs
Ashish-Kumar-Dash: fixed in latest commit

smitebot/src/commands/corpus.rs
Ashish-Kumar-Dash: yes, fixed in latest commit to enforce it to be an executable

smitebot/src/commands/corpus.rs
Ashish-Kumar-Dash: merge to minimize avoids the blow-up;, it's only direct minimize on raw runner queues that overflows. I'll fix this in a follow-up

smitebot/src/commands/corpus.rs
Ashish-Kumar-Dash: Yes, AFAIK `afl-cmin` resolves `afl-showmap` via $PATH before the checkout (search_binary, afl-cmin.py), so the showmap that runs can sit in a dir without libnyx.so and abort, then `find_afl_binary` checks $AFL_PATH first, which is why exporting it fixed the libnyx.so not found abort here.
Now I've set AFL_PATH only when libnyx.so actually sits next to afl-cmin (source checkout), packaged installs fall through as usual.

smitebot/src/commands/corpus.rs
```diff
@@ -0,0 +1,476 @@
...
+    // before the checkout (afl-cmin.py search order), so the afl-showmap that
+    // runs may sit in a directory without libnyx.so and abort. Point AFL_PATH at
+    // the AFL++ root when libnyx.so lives beside afl-cmin (a source checkout);
+    // skip it for a bin/-installed afl-cmin, where libnyx.so is elsewhere and
+    // find_afl_binary's built-in path handles it.
+    if let Some(afl_root) = afl_cmin.parent()
+        && afl_root.join("libnyx.so").exists()
+    {
+        cmd.env("AFL_PATH", afl_root);
+    }
```
morehouse: Again, maybe I'm missing something, but this seems useless.  When I run `afl-cmin` locally, it automatically finds the `libnyx.so` in the same directory without me needing to set `AFL_PATH`.  And this code *only* sets `AFL_PATH` in exactly that scenario where it isn't needed.

smitebot/README.md
```diff
@@ -165,3 +165,29 @@ smitebot print-ir output/default/crashes/id:000000,sig:06,...
...
+- `<campaign-id>`: directory name(s) under `~/.smitebot/runs`
+- `-o, --output <output-dir>`: output directory for the merged corpus (required)
+
+### smitebot corpus minimize
+
+Removes corpus inputs that do not contribute new coverage, using `afl-cmin` in Nyx mode (`-X`). Reads `sharedir` and `aflpp_path` from the campaign's `state.json`; no live campaign required.
+
+```bash
+smitebot corpus minimize <campaign-id>
+smitebot corpus minimize <campaign-id> [-i <input>] [-o <output-dir>] [--aflpp <path>]
```
morehouse: Other commands use `--aflpp-path`.  We should be consistent.

smitebot/src/commands/corpus.rs
NishantBansal2003: Previously, I built afl++ with all modes enabled, which I think was causing the `libnyx.so not found` issue. I rebuilt afl++ with only nyx mode enabled, and the issue is now fixed for me without adding any extra flags

@Ashish-Kumar-Dash , I think you can try this as well if you're getting the same issue. We shouldn't need to set the `AFL_PATH` env variable in this case

Ashish-Kumar-Dash: Yes, I've deleted it, verified in source too, I will document this

PR #185, opened 2026-08-03
smite: add accept_channel oracle

NishantBansal2003:
Add `AcceptChannelOracle` and consolidate all the previously scattered checks relevant to the initial funding flow negotiation into it, including unknown `temporary_channel_id`, `temporary_channel_id` reuse before `funding_created`, cases where the target accepted an invalid `open_channel` we sent, and cases where the target sent an `accept_channel` that does not follow the BOLT 2 `open_channel <-> accept_channel` requirements

Note: Some checks that depend on negotiated features or `channel_type` are still missing and will be added in a follow-up PR to keep this one focused
Also, I only added the checks explicitly mentioned in the BOLT 2 and haven't added some obvious checks, such as rejecting `max_accepted_htlcs < 1` in `open_channel`. I do think those checks are valuable, so I was thinking of adding them in a follow up PR to avoid adding too much here. But let me know if you think all of them (excluding feature-related checks) should be included here.

I'm also planning to add BOLT 9 feature flag primitives so feature related logic can be consolidated in one place instead of being scattered across `operation.rs`, `setup.rs`, and `AcceptChannelOracle`. The remaining feature dependent checks will be added to this oracle once that is in place

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -13,10 +13,11 @@ use smite::bolt::{
     OpenChannel, OpenChannelTlvs, Pong, ShortChannelId, Shutdown, msg_type,
 };
 use smite::channel_tx::{
-    ChannelConfig, ChannelPartyConfig, ChannelState, CommitmentCost, FundingTransaction,
-    HolderIdentity, Side, build_funding_transaction,
+    ChannelConfig, ChannelPartyConfig, ChannelState, FundingTransaction, HolderIdentity, Side,
+    build_funding_transaction,
 };
 use smite::noise::{ConnectionError, NoiseConnection};
+use smite::oracles::{AcceptChannelContext, AcceptChannelOracle, Oracle};
 use smite::pending_channel::PendingChannel;
 use smite::violation::Violation;
 use smite_ir::operation::AcceptChannelField;
@@ -477,7 +478,11 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                     log::debug!("[{:?}] RecvAcceptChannel: waiting", start.elapsed());
                     let ac = recv_accept_channel(&mut self.conn)?;
                     log::debug!("[{:?}] RecvAcceptChannel: received", start.elapsed());
-                    record_recv_accept_channel(&mut self.negotiations, &ac)?;
+                    AcceptChannelOracle.evaluate(&AcceptChannelContext {
+                        accept_channel: &ac,
+                        negotiation: self.negotiations.get(&ac.temporary_channel_id),
+                    })?;
+                    record_recv_accept_channel(&mut self.negotiations, &ac);
                     Some(Variable::AcceptChannel(ac))
                 }
 
@@ -1317,9 +1322,8 @@ fn is_channel_ready_expected(
 /// # Errors
 ///
 /// Returns [`Violation::UnknownChannel`] if no channel state exists for the
-/// given `channel_id`, [`Violation::OpenerCannotAffordFee`] if the opener
-/// cannot afford the commitment feerate, or [`Violation::InvalidCounterpartySignature`]
-/// if the signature is invalid for the holder's initial commitment transaction.
+/// given `channel_id`, or [`Violation::InvalidCounterpartySignature`] if the
+/// signature is invalid for the holder's initial commitment transaction.
 fn verify_funding_signed(
     fs: &FundingSigned,
     channel_states: &HashMap<ChannelId, ChannelState>,
@@ -1327,17 +1331,6 @@ fn verify_funding_signed(
     let state = channel_states
         .get(&fs.channel_id)
         .ok_or(Violation::UnknownChannel(fs.channel_id))?;
-
-    // The opener cannot afford the fee, so the acceptor must not send
-    // `funding_signed`. Receiving one is a protocol violation.
-    let commitment_cost =
-        CommitmentCost::new(state.commitment.feerate_per_kw, &state.config.channel_type);
-    if (state.commitment.opener.balance_msat / 1000)
-        .checked_sub(commitment_cost.total_sat())
-        .is_none()
-    {
-        return Err(Violation::OpenerCannotAffordFee(fs.channel_id));
-    }
 
     state
         .config
@@ -1377,29 +1370,18 @@ fn record_send_open_channel(
 /// Pairs a received `accept_channel` with the recorded `open_channel` of the
 /// same `temporary_channel_id`.
 ///
-/// # Errors
+/// # Panics
 ///
-/// Returns [`Violation::UnknownChannel`] if no `open_channel` was recorded
-/// for the message's `temporary_channel_id`.
-///
-/// Returns [`Violation::TempChannelIdReuse`] if the negotiation already has an
-/// `accept_channel` but has not yet reached `funding_created`.
+/// Panics if no matching `open_channel` exists. This should be unreachable, as
+/// `AcceptChannelOracle` reports such messages as a [`Violation`].
 fn record_recv_accept_channel(
     negotiations: &mut HashMap<ChannelId, PendingChannel>,
     accept_channel: &AcceptChannel,
-) -> Result<(), Violation> {
-    let pending = negotiations
+) {
+    negotiations
         .get_mut(&accept_channel.temporary_channel_id)
-        .ok_or(Violation::UnknownChannel(
-            accept_channel.temporary_channel_id,
-        ))?;
-    if pending.accept_channel.is_some() && !pending.funding_built {
-        return Err(Violation::TempChannelIdReuse(
-            accept_channel.temporary_channel_id,
-        ));
-    }
-    pending.accept_channel = Some(accept_channel.clone());
-    Ok(())
+        .expect("AcceptChannelOracle guaranteed this temporary_channel_id exists")
+        .accept_channel = Some(accept_channel.clone());
 }
 
 /// Extracts a field from a parsed `accept_channel` message.
@@ -1620,7 +1602,7 @@ mod tests {
             first_per_commitment_point: sample_pubkey(6),
             tlvs: AcceptChannelTlvs {
                 upfront_shutdown_script: Some(vec![0xde, 0xad]),
-                channel_type: Some(vec![0x01]),
+                channel_type: Some(vec![0x40, 0x10, 0x00]),
             },
         }
     }
@@ -1705,7 +1687,7 @@ mod tests {
                 inputs: vec![],
             },
             Instruction {
-                operation: Operation::LoadFeatures(vec![]),
+                operation: Operation::LoadFeatures(vec![0x40, 0x10, 0x00]),
                 inputs: vec![],
             },
         ]
@@ -1894,7 +1876,7 @@ mod tests {
         assert_eq!(oc.first_per_commitment_point, pk);
         assert_eq!(oc.channel_flags, 1);
         assert_eq!(oc.tlvs.upfront_shutdown_script, Some(vec![]));
-        assert!(oc.tlvs.channel_type.is_none());
+        assert_eq!(oc.tlvs.channel_type, Some(vec![0x40, 0x10, 0x00]));
     }
 
     #[test]
@@ -2650,9 +2632,55 @@ mod tests {
             )
             .unwrap_err();
 
-        assert!(
-            matches!(err, ExecuteError::Violation(Violation::UnknownChannel(id)) if id == unknown_id)
+        let ExecuteError::Violation(Violation::InvalidAcceptChannel(id, reason)) = &err else {
+            panic!("unexpected error: {err:?}");
+        };
+        assert_eq!(*id, unknown_id);
+        assert!(reason.contains(
+            "unknown temporary_channel_id: no open_channel was sent for this negotiation"
+        ));
+    }
+
+    #[test]
+    fn execute_recv_accept_channel_opener_cannot_afford_fee() {
+        let temporary_channel_id = ChannelId::new([0xbb; 32]);
+        let ac_bytes = Message::AcceptChannel(sample_accept_channel()).encode();
+
+        // Set `push_msat` so the opener cannot afford the commitment fee
+        // requiring the peer to reject the `open_channel` per BOLT 2.
+        let mut instrs = send_open_channel_instructions();
+        instrs[3] = Instruction {
+            operation: Operation::LoadAmount(99_900_000),
+            inputs: vec![],
+        };
+        let sent_open_channel = instrs.len() - 1;
+        instrs.push(Instruction {
+            operation: Operation::RecvAcceptChannel,
+            inputs: vec![sent_open_channel],
+        });
+
+        let mut executor = Executor::new(
+            MockConnection::new(),
+            MockBitcoinCli::default(),
+            sample_context(),
         );
+        executor.conn.queue_recv(ac_bytes);
+        let err = executor
+            .execute(
+                &Program {
+                    instructions: instrs,
+                },
+                std::time::Instant::now(),
+            )
+            .unwrap_err();
+
+        let ExecuteError::Violation(Violation::InvalidAcceptChannel(id, reason)) = &err else {
+            panic!("unexpected error: {err:?}");
+        };
+        assert_eq!(*id, temporary_channel_id);
+        assert!(reason.contains(
+            "invalid open_channel: opener balance 100 sat cannot cover the commitment fee"
+        ));
     }
 
     #[test]
@@ -2692,9 +2720,13 @@ mod tests {
                 std::time::Instant::now(),
             )
             .unwrap_err();
-        assert!(matches!(
-            err,
-            ExecuteError::Violation(Violation::TempChannelIdReuse(id)) if id == temporary_channel_id
+
+        let ExecuteError::Violation(Violation::InvalidAcceptChannel(id, reason)) = &err else {
+            panic!("unexpected error: {err:?}");
+        };
+        assert_eq!(*id, temporary_channel_id);
+        assert!(reason.contains(
+            "temporary_channel_id reuse: previous negotiation has not reached funding_created"
         ));
     }
 
@@ -3572,54 +3604,6 @@ mod tests {
     }
 
     #[test]
-    fn execute_recv_funding_signed_opener_cannot_afford_fee() {
-        let mock_cli = MockBitcoinCli {
-            utxos: vec![sample_utxo()],
-            change_spk: sample_change_spk(),
-            ..Default::default()
-        };
-
-        let channel_id = ChannelId::v1_from_funding_outpoint(OutPoint {
-            txid: "09b0549b35f14ee862f63bd75811c6c27963c4dea6766ec6836952ec78df1e7e"
-                .parse()
-                .unwrap(),
-            vout: 0,
-        });
-
-        // The expected signature here was computed using LDK as the source of
-        // truth.
-        let fs_bytes = Message::FundingSigned(FundingSigned {
-            channel_id,
-            signature: "304502210096c5e8ad834af46b42a4301828852205655d16dc8d55333831de49642d70c60a02205466283b9557447dd4c5374b90eda80f023017164dd04deb7c45cfc472e03023".parse().unwrap(),
-        })
-        .encode();
-
-        let mut executor = Executor::new(MockConnection::new(), mock_cli, sample_context());
-        executor.conn.queue_recv(fs_bytes);
-
-        // Increase the pushed amount so the opener cannot afford the required
-        // fee when the commitment is built and funding_signed is received.
-        let mut negotiation = sample_funding_negotiation();
-        negotiation.open_channel.push_msat = 10_000_000_000;
-        executor
-            .negotiations
-            .insert(ChannelId::new([0xbb; 32]), negotiation);
-
-        let err = executor
-            .execute(
-                &Program {
-                    instructions: send_funding_created_and_recv_funding_signed_instructions(),
-                },
-                std::time::Instant::now(),
-            )
-            .unwrap_err();
-        assert!(matches!(
-            err,
-            ExecuteError::Violation(Violation::OpenerCannotAffordFee(id)) if id == channel_id
-        ));
-    }
-
-    #[test]
     fn execute_recv_funding_signed_invalid_signature() {
         let mock_cli = MockBitcoinCli {
             utxos: vec![sample_utxo()],
@@ -4141,7 +4125,7 @@ mod tests {
         );
         assert_eq!(
             extract_field(&ac, AcceptChannelField::ChannelType),
-            Variable::Features(vec![0x01])
+            Variable::Features(vec![0x40, 0x10, 0x00])
         );
     }
 
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -13,8 +13,8 @@ use smite::bolt::{
     OpenChannel, OpenChannelTlvs, Pong, ShortChannelId, Shutdown, msg_type,
 };
 use smite::channel_tx::{
-    ChannelConfig, ChannelPartyConfig, ChannelState, FundingTransaction, HolderIdentity, Side,
-    build_funding_transaction,
+    ChannelConfig, ChannelPartyConfig, ChannelState, CommitmentCost, FundingTransaction,
+    HolderIdentity, Side, build_funding_transaction,
 };
 use smite::noise::{ConnectionError, NoiseConnection};
 use smite::pending_channel::PendingChannel;
@@ -1330,7 +1330,12 @@ fn verify_funding_signed(
 
     // The opener cannot afford the fee, so the acceptor must not send
     // `funding_signed`. Receiving one is a protocol violation.
-    if !state.config.can_opener_afford_feerate(&state.commitment) {
+    let commitment_cost =
+        CommitmentCost::new(state.commitment.feerate_per_kw, &state.config.channel_type);
+    if (state.commitment.opener.balance_msat / 1000)
+        .checked_sub(commitment_cost.total_sat())
+        .is_none()
+    {
         return Err(Violation::OpenerCannotAffordFee(fs.channel_id));
     }
 
--- a/smite/src/channel_tx/commitment.rs
+++ b/smite/src/channel_tx/commitment.rs
@@ -113,6 +113,15 @@ pub struct CommitmentState {
     // to correctly compute balances and construct HTLC outputs in the commitment transaction.
 }
 
+/// Costs associated with a commitment transaction, including transaction fee
+/// and anchor outputs.
+pub struct CommitmentCost {
+    /// Commitment transaction fee in satoshis.
+    pub fee_sat: u64,
+    /// Total cost of anchor outputs in satoshis.
+    pub anchor_cost_sat: u64,
+}
+
 /// State of a single channel, including its static configuration, holder
 /// identity, and current commitment state.
 pub struct ChannelState {
@@ -266,19 +275,6 @@ impl ChannelConfig {
         })
     }
 
-    /// Checks whether the opener can afford the commitment fee at the given
-    /// feerate, after accounting for the anchor outputs.
-    #[must_use]
-    pub fn can_opener_afford_feerate(&self, state: &CommitmentState) -> bool {
-        let fee = commit_tx_fee_sat(state.feerate_per_kw, &self.channel_type);
-        let anchor_cost = total_anchors_sat(&self.channel_type);
-
-        (state.opener.balance_msat / 1000)
-            .checked_sub(fee)
-            .and_then(|balance| balance.checked_sub(anchor_cost))
-            .is_some()
-    }
-
     /// Builds the signature for the counterparty's commitment transaction.
     #[must_use]
     pub fn sign_counterparty_commitment(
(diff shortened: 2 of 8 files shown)


smite/src/channel_tx/commitment.rs
```diff
@@ -475,6 +462,26 @@ impl CommitmentState {
...
+    channel_type: &[u8],
+) -> Option<u64> {
+    let fee = commit_tx_fee_sat(feerate_per_kw, channel_type);
+    let anchor_cost = total_anchors_sat(channel_type);
+
+    opener_balance_sat
+        .checked_sub(fee)?
+        .checked_sub(anchor_cost)
+}
+
```
ekzyis: Mhh, I think returning `None` can be confused with 0. To me, this code:

```rust
if opener_balance_after_fee_sat(
        state.commitment.opener.balance_msat,
        state.commitment.feerate_per_kw,
        &state.config.channel_type,
    )
    .is_none()
```

reads like it will be true if the balance is 0 instead of negative. Did you consider returning a negative number instead? I think `balance < 0` is easier to understand than `balance.is_none()`.

smite/src/oracles.rs
ekzyis, line 1: I see that this file was already defined, but is there a reason why not update the return type of `evaluate` from `OracleResult` to `Result<(), Violation>`?

The documentation of `Violation` mentions this:

```
//! Each [`Violation`] variant names a buggy target behavior (e.g., crashing,
//! hanging, breaking a protocol invariant). This is how the fuzzer reports bugs
//! in the target.
```

The documentation of oracles.rs also mentions protocol invariants, so it seems like the perfect match to me. I think this would make the `evaluate_accept_channel_oracle` function obsolete.

Not sure how useful this would be, but we could then even consider oracles to return all violations instead of only the first one.

smite-scenarios/src/executor.rs
```diff
@@ -1322,12 +1323,6 @@ fn verify_funding_signed(
         .get(&fs.channel_id)
         .ok_or(Violation::UnknownChannel(fs.channel_id))?;
 
-    // The opener cannot afford the fee, so the acceptor must not send
-    // `funding_signed`. Receiving one is a protocol violation.
-    if !state.config.can_opener_afford_feerate(&state.commitment) {
-        return Err(Violation::OpenerCannotAffordFee(fs.channel_id));
-    }
-
```
ekzyis, line 1336: `verify_funding_signed` can still throw a `Violation`. Is the plan to replace `verify_funding_signed` with an oracle? I expected all `Violation`s to be thrown in oracles.

smite-scenarios/src/executor.rs
```diff
@@ -1432,6 +1416,35 @@ fn nonempty_or_none(bytes: &[u8]) -> Option<Vec<u8>> {
...
 
+// -- Oracle evaluation --
+
+/// Evaluates the [`AcceptChannelOracle`] against the negotiation a received
+/// `accept_channel` answers, if any.
+///
+/// # Errors
+///
+/// Returns [`Violation::InvalidAcceptChannel`] if the oracle fails.
+fn evaluate_accept_channel_oracle(
```
ekzyis: I think this function isn't needed, see other comment

smite-scenarios/src/executor.rs
```diff
@@ -1366,29 +1361,18 @@ fn record_send_open_channel(
...
-/// # Errors
-///
-/// Returns [`Violation::UnknownChannel`] if no `open_channel` was recorded
-/// for the message's `temporary_channel_id`.
+/// # Panics
 ///
-/// Returns [`Violation::TempChannelIdReuse`] if the negotiation already has an
-/// `accept_channel` but has not yet reached `funding_created`.
+/// Panics if no matching `open_channel` exists. This should be unreachable, as
+/// `evaluate_accept_channel_oracle` reports such messages as a [`Violation`].
```
ekzyis: Mhh, would be nice if we could couple oracle evaluation ("verification") with receiving or recording.

I think conceptually, verification is closer to receiving. If we would move it there, we would still need to document a possible panic here, but if the only function that receives also verifies, and we use naming conventions like `recv_<message>` and `record_<message>` for the functions we call in the matching arm of `Executor::execute()`, it would be easier to parse the code and have confidence in no panics.

(Or we move verification to recording and then we don't need to panic.)

ekzyis (commented):
I was working on using #186 to verify the `upfront_shutdown_script` TLV and I saw this PR is related. I like the idea of oracles to localize code to detect protocol invariants! I only took a quick look for now. Left some comments regarding the design.

smite/src/channel_tx/commitment.rs
NishantBansal2003: > Did you consider returning a negative number instead?

We could use that, but then we need to handle type conversions of balances, I think we could use `Result<u64, String>` for that case, that would be clearer

smite-scenarios/src/executor.rs
NishantBansal2003: Yup, that is originally my next plan to migrate those to oracles as well

smite-scenarios/src/executor.rs
NishantBansal2003: I do prefer all three to be separate, we already follow the same idea currently with `funding_signed` -- receive and then verify instead of clubbing them together (for logging purposes as well)

I can certainly club verify with record, but I actually wanted to keep the oracle evaluation blocks separate as `// -- Oracle evaluation --`, so we can put all the oracle evaluation there, which is easier to understand. Then we can follow a pattern like recv -> verify oracle -> record

smite/src/oracles.rs
NishantBansal2003: I don't have a strong opinion here and can certainly change `OracleResult to Result<(), Violation>`, though I do prefer a separate block for oracle evaluation and hence would like to keep `evaluate_accept_channel_oracle`

> Not sure how useful this would be, but we could then even consider oracles to return all violations instead of only the first one.

How will this work? Do you mean each oracle failure is only logged, or we store each oracle fail in a Vec and then output it?

smite/src/channel_tx/commitment.rs
morehouse: Maybe we want to make this function more general so that the caller has more data to work with.  For example:

```rust
pub struct CommitmentCost {
  pub fee_sat: u64,
  pub anchor_cost_sat: u64,
}

impl CommitmentCost {
  pub fn total_sat(&self) -> u64 {
    self.fee_sat + self.anchor_cost_sat
  }
}

pub fn commitment_cost_sat(feerate_per_kw: u32, channel_type: &[u8]) -> CommitmentCost {
  CommitmentCost {
    fee_sat: commit_tx_fee_sat(feerate_per_kw, channel_type),
    anchor_cost_sat: total_anchors_sat(channel_type),
  }
}
```

This would let the oracle do the `checked_sub` and report a more detailed error on failure since it would have the actual commitment cost breakdown.  It would also allow this to be reused with `saturating_sub` in `build_commitment_transaction`.

smite-scenarios/src/executor.rs
```diff
@@ -1432,6 +1416,35 @@ fn nonempty_or_none(bytes: &[u8]) -> Option<Vec<u8>> {
...
+    if let OracleResult::Fail(reason) = accept_channel_oracle.evaluate(&AcceptChannelContext {
+        accept_channel,
+        negotiation: negotiations.get(&temporary_channel_id),
+    }) {
+        log::debug!("{} failed: {reason}", accept_channel_oracle.name());
+        return Err(Violation::InvalidAcceptChannel(
+            temporary_channel_id,
+            reason,
+        ));
+    }
```
morehouse: Perhaps `Oracle::evaluate` should return a `Result<(), Violation>` to simplify call sites?


```suggestion
    accept_channel_oracle.evaluate(&AcceptChannelContext {
        accept_channel,
        negotiation: negotiations.get(&temporary_channel_id),
    })?;
```

We could probably even inline the oracle checks in the match arms at that point.

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+        return Some(format!(
+            "dust_limit_satoshis {} is below the minimum of {MIN_DUST_LIMIT_SATOSHIS} sat",
+            open_channel.dust_limit_satoshis,
+        ));
+    }
+
+    // Check that the channel type was included.
+    let Some(channel_type) = open_channel.tlvs.channel_type.as_deref() else {
+        return Some("open_channel does not include a channel_type".to_string());
+    };
```
morehouse: I know the spec says the receiver must fail in this case, but I wouldn't be surprised if there's still an implementation that allows the old "implicit" negotiation...

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+            &accept_channel(),
+            Some(&pending_negotiation(oc)),
+            "invalid open_channel: funding_satoshis 2100000000000001 exceeds maximum funding",
+        );
+    }
+
+    #[test]
+    fn push_msat_above_the_funding_amount() {
+        let mut oc = open_channel();
+        oc.push_msat = 10_000_000_001;
```
morehouse: Nit


```suggestion
        oc.push_msat = oc.funding_satoshis * 1000 + 1;
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+            &accept_channel(),
+            Some(&pending_negotiation(oc)),
+            "invalid open_channel: open_channel does not include a channel_type",
+        );
+    }
+
+    #[test]
+    fn opener_cannot_afford_commitment_fee() {
+        let mut oc = open_channel();
+        oc.push_msat = 9_990_000_000;
```
morehouse: Nit


```suggestion
        oc.push_msat = oc.funding_satoshis *1000 - 10_000_000;
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+            &accept_channel(),
+            Some(&pending_negotiation(oc)),
+            "invalid open_channel: opener balance 10000 sat cannot cover the commitment fee",
+        );
+    }
+
+    #[test]
+    fn opener_cannot_cover_anchor_outputs() {
+        let mut oc = open_channel();
+        oc.push_msat = 9_983_000_000;
```
morehouse: Nit


```suggestion
        oc.push_msat = oc.funding_satoshis * 1000 - 17_000_000;
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+            &accept_channel(),
+            Some(&pending_negotiation(oc)),
+            "invalid open_channel: push_msat 10000000001 exceeds funding amount",
+        );
+    }
+
+    #[test]
+    fn open_channel_dust_limit_above_its_channel_reserve() {
+        let mut oc = open_channel();
+        oc.dust_limit_satoshis = 20_000;
```
morehouse: Nit


```suggestion
        oc.dust_limit_satoshis = oc.channel_reserve_satoshis + 1;
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+            &accept_channel(),
+            Some(&pending_negotiation(oc)),
+            "invalid open_channel: dust_limit_satoshis 20000 exceeds channel_reserve_satoshis",
+        );
+    }
+
+    #[test]
+    fn open_channel_max_accepted_htlcs_above_the_limit() {
+        let mut oc = open_channel();
+        oc.max_accepted_htlcs = 484;
```
morehouse: Nit


```suggestion
        oc.max_accepted_htlcs = MAX_ACCEPTED_HTLCS_LIMIT + 1;
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+            &accept_channel(),
+            Some(&pending_negotiation(oc)),
+            "invalid open_channel: max_accepted_htlcs 484 exceeds the limit of 483",
+        );
+    }
+
+    #[test]
+    fn open_channel_dust_limit_below_the_minimum() {
+        let mut oc = open_channel();
+        oc.dust_limit_satoshis = 353;
```
morehouse: Nit


```suggestion
        oc.dust_limit_satoshis = MIN_DUST_LIMIT_SATOSHIS - 1;
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+        return Some(format!(
+            "dust_limit_satoshis {} exceeds the open_channel channel_reserve_satoshis {}",
+            accept_channel.dust_limit_satoshis, open_channel.channel_reserve_satoshis,
+        ));
+    }
+
+    // Check that the channel type was included.
+    let Some(channel_type) = accept_channel.tlvs.channel_type.as_deref() else {
+        return Some("accept_channel does not include a channel_type".to_string());
+    };
```
morehouse: `channel_type` also must match the type from `open_channel`.

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+
+    // Check the channel reserve covers the dust limit.
+    if accept_channel.dust_limit_satoshis > accept_channel.channel_reserve_satoshis {
+        return Some(format!(
+            "dust_limit_satoshis {} exceeds channel_reserve_satoshis {}",
+            accept_channel.dust_limit_satoshis, accept_channel.channel_reserve_satoshis,
+        ));
+    }
```
morehouse, line 176: I think this isn't strictly required by the spec (though maybe it should be?).  Still good to have this, though it might be worth adding a comment explaining that it's not in the spec.


```suggestion
    // Check the channel reserve covers the dust limit. Not strictly required by the spec.
    if accept_channel.dust_limit_satoshis > accept_channel.channel_reserve_satoshis {
        return Some(format!(
            "dust_limit_satoshis {} exceeds channel_reserve_satoshis {} (not strictly required by spec)",
            accept_channel.dust_limit_satoshis, accept_channel.channel_reserve_satoshis,
        ));
    }
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+            &accept_channel(),
+            Some(&pending_negotiation(oc)),
+            "invalid open_channel: neither side exceeds channel reserve",
+        );
+    }
+
+    #[test]
+    fn accept_channel_reserve_below_the_open_channel_dust_limit() {
+        let mut ac = accept_channel();
+        ac.channel_reserve_satoshis = 545;
```
morehouse: Nit

```suggestion
        ac.channel_reserve_satoshis = oc.dust_limit_satoshis - 1;
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+            &ac,
+            Some(&pending_negotiation(open_channel())),
+            "invalid accept_channel: channel_reserve_satoshis 545 is below the open_channel dust_limit_satoshis 546",
+        );
+    }
+
+    #[test]
+    fn accept_channel_dust_limit_above_the_open_channel_reserve() {
+        let mut ac = accept_channel();
+        ac.dust_limit_satoshis = 10_001;
```
morehouse: Nit

```suggestion
        ac.dust_limit_satoshis = oc.channel_reserve_satoshis + 1;
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+            &ac,
+            Some(&pending_negotiation(open_channel())),
+            "invalid accept_channel: dust_limit_satoshis 5000 exceeds channel_reserve_satoshis 4000",
+        );
+    }
+
+    #[test]
+    fn accept_channel_max_accepted_htlcs_above_the_limit() {
+        let mut ac = accept_channel();
+        ac.max_accepted_htlcs = 484;
```
morehouse: Nit

```suggestion
        ac.max_accepted_htlcs = MAX_ACCEPTED_HTLCS_LIMIT + 1;
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+            &ac,
+            Some(&pending_negotiation(open_channel())),
+            "invalid accept_channel: max_accepted_htlcs 484 exceeds the limit of 483",
+        );
+    }
+
+    #[test]
+    fn accept_channel_dust_limit_below_the_minimum() {
+        let mut ac = accept_channel();
+        ac.dust_limit_satoshis = 353;
```
morehouse: NIt

```suggestion
        ac.dust_limit_satoshis = MIN_DUST_LIMIT_SATOSHIS - 1;
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+        accept_channel: &AcceptChannel,
+        negotiation: Option<&PendingChannel>,
+        expected: &str,
+    ) {
+        match AcceptChannelOracle.evaluate(&AcceptChannelContext {
+            accept_channel,
+            negotiation,
+        }) {
+            OracleResult::Pass => panic!("expected failure: {expected}"),
+            OracleResult::Fail(reason) => assert!(reason.contains(expected)),
```
morehouse: Nit: easier to debug if we print the reason


```suggestion
            OracleResult::Fail(reason) => assert!(reason.contains(expected), "unexpected failure reason: {reason}"),
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+        }) = context.negotiation
+        else {
+            return OracleResult::Fail(
+                "unknown temporary_channel_id: no open_channel was sent for this negotiation"
+                    .to_string(),
+            );
+        };
+
+        // Check that the `open_channel` was valid to accept.
+        if let Some(reason) = must_reject_open_channel(open_channel) {
```
morehouse: Should we call this `verify_open_channel` or `verify_accepted_open_channel` instead?

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+
+    fn name(&self) -> &'static str {
+        "AcceptChannelOracle"
+    }
+}
+
+/// Returns the BOLT 2 requirement our `open_channel` breaches, i.e. the reason
+/// its receiver had to fail the channel instead of accepting it, or `None` if
+/// it breaches none.
+fn must_reject_open_channel(open_channel: &OpenChannel) -> Option<String> {
```
morehouse: I think the usual pattern for these verify-type functions is to return a Result


```suggestion
fn must_reject_open_channel(open_channel: &OpenChannel) -> Result<(), String> {
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,553 @@
...
+}
+
+/// Verifies that the initial commitment can cover its fee and satisfies the
+/// channel reserve requirement, returning the first requirement it breaches,
+/// or None if both are met.
+fn verify_initial_commitment(
+    open_channel: &OpenChannel,
+    channel_type: &[u8],
+    channel_reserve_satoshis: u64,
+) -> Option<String> {
```
morehouse:
```suggestion
fn verify_initial_commitment(
    open_channel: &OpenChannel,
    channel_type: &[u8],
    channel_reserve_satoshis: u64,
) -> Result<(), String> {
```

smite-scenarios/src/executor.rs
```diff
@@ -1694,7 +1707,7 @@ mod tests {
                 inputs: vec![],
             },
             Instruction {
-                operation: Operation::LoadFeatures(vec![]),
+                operation: Operation::LoadFeatures(vec![0x40, 0x00, 0x00]),
```
morehouse: `option_anchors` requires `option_static_remotekey` too
```suggestion
                operation: Operation::LoadFeatures(vec![0x40, 0x10, 0x00]),
```

morehouse (commented):
My main concern is that we might get false positives for channel types we haven't implemented yet (i.e. taproot, zero-fee commits).  Please double check that and add comments explaining why we're safe (if we are).

Otherwise this is great -- I have only minor comments inline.

Also things that we should consider adding in follow-up PRs:

##### `open_channel` expected failures

- `zero_fee_commitments && max_accepted_htlcs > 114`
- `zero_fee_commitments && feerate_per_kw != 0`
- wumbo limit exceeded
- maybe `feerate_per_kw == 0`?
- maybe `channel_reserve_satoshis >= funding_satoshis`
- maybe `chain_hash != regtest` if we add non-regtest chain hashes to the IR
- maybe `announce_channel && option_scid_alias`
- maybe absurd `dust_limit_satoshis` values (e.g., more than 10,000 sat?)

##### `accept_channel` violations

- `channel_type != open_channel.channel_type`
- `option_zeroconf && minimum_depth > 0`
- invalid `upfront_shutdown_script`
- `zero_fee_commitments && max_accepted_htlcs > 114`
- maybe when any pubkeys match a pubkey from `open_channel`
- maybe `per_commitment_point` reuse -- requires maintaining a global PCP set.  But we'll want to maintain that set anyway for commitment fuzzing
- maybe `to_self_delay == 0`
- maybe `htlc_minimum_msat > max_htlc_value_in_flight_msat`
- maybe `htlc_minimum_msat > open_channel.funding_satoshis`
- maybe `max_accepted_channels == 0`

smite/src/oracles.rs
morehouse: I also think `Result<(), Violation>` would be simpler.

I would rather not have oracles return multiple violations though -- it's simpler to just return the first one.  Realistically I would be surprised (in a bad way) if multiple violations are common, and even then we should be able to find the next violation after the first one is fixed.

smite/src/oracles/accept_channel.rs
NishantBansal2003: I was thinking of adding this in follow-up of `channel_type` as well, but yeah, let's just add it here since it's only a small check

smite/src/oracles/accept_channel.rs
NishantBansal2003: It's mentioned for `open_channel`, and I assume that mentioning `Other fields have the same requirements as their counterparts in open_channel.` for `accept_channel` means that all relevant checks for `open_channel` also apply to `accept_channel`. So, this means it is required by the spec, right?

smite/src/oracles/accept_channel.rs
NishantBansal2003: I think then that's a non-spec-compliant bug, I will add a TODO to check `option_channel_type` in the negotiated features since it is assumed to be supported

NishantBansal2003:
Had to force-push since there were some design changes

> My main concern is that we might get false positives for channel types we haven't implemented yet (i.e. taproot, zero-fee commits). Please double check that and add comments explaining why we're safe (if we are).

I verified this and added comments inline. For the fee and anchor cost calculations, we might be undercalculating the fee for 0FC or taproot channels, but definitely not overcalculating it. So we will not get false positives, though we might miss some invalid cases. In some cases, we might also hit a `Violation` where the error string is misleading, but it will still be an actual error.

I think this will also be resolved once I add support for verifying oracles based on the negotiated features in a follow-up PR.

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,611 @@
...
+    verify_initial_commitment(
+        open_channel,
+        channel_type,
+        accept_channel.channel_reserve_satoshis,
+    )
+}
+
+/// Verifies that the initial commitment can cover its fee and satisfies the
+/// channel reserve requirement, returning the first requirement it breaches,
+/// or None if both are met.
```
morehouse: Doesn't return `None`.

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,611 @@
...
+    // Check the initial commitment satisfies the channel reserve.
+    verify_initial_commitment(
+        open_channel,
+        channel_type,
+        open_channel.channel_reserve_satoshis,
+    )
+}
+
+/// Verifies the `accept_channel` against the BOLT 2 requirements it must meet,
+/// returning the one it breaches, or `None` if it meets them all.
```
morehouse: Doesn't return `None`.

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,611 @@
...
+        Ok(())
+    }
+
+    fn name(&self) -> &'static str {
+        "AcceptChannelOracle"
+    }
+}
+
+/// Returns the BOLT 2 requirement our `open_channel` breaches, i.e. the reason
+/// its receiver had to fail the channel instead of accepting it, or `None` if
```
morehouse: Doesn't return `None`.

smite/src/oracles.rs
```diff
@@ -1,19 +1,20 @@
...
 pub trait Oracle<C> {
     /// Evaluate the oracle against the given context
-    fn evaluate(&self, context: &C) -> OracleResult;
+    ///
+    /// # Errors
+    ///
+    /// Returns `Violation` if an oracle invariant is violated.
+    fn evaluate(&self, context: &C) -> Result<(), Violation>;
     /// Return the name of this oracle for logging
     fn name(&self) -> &str;
```
morehouse: Looks like we can delete `name` entirely.

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,611 @@
...
+use crate::violation::Violation;
+
+use bitcoin::Amount;
+
+// Constants from the BOLT 2 `open_channel` and `accept_channel` requirements:
+// https://github.com/lightning/bolts/blob/master/02-peer-protocol.md#requirements-8
+const MAX_ACCEPTED_HTLCS_LIMIT: u16 = 483;
+const MIN_DUST_LIMIT_SATOSHIS: u64 = 354;
+
+/// `AcceptChannelContext` is the context for the `AcceptChannelOracle`
```
morehouse: Nit

```suggestion
/// Context for `AcceptChannelOracle`
```

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,611 @@
...
+pub struct AcceptChannelContext<'a> {
+    /// The `accept_channel` received from the peer.
+    pub accept_channel: &'a AcceptChannel,
+    /// The negotiation the `accept_channel` answers, identified by its
+    /// `temporary_channel_id`, or `None` if no matching `open_channel` was sent.
+    pub negotiation: Option<&'a PendingChannel>,
+}
+
+/// `AcceptChannelOracle` checks whether the `open_channel` answered by an
```
morehouse: Nit


```suggestion
/// Checks whether the `open_channel` answered by an
```

smite/src/channel_tx/commitment.rs
```diff
@@ -1365,39 +1375,39 @@ mod tests {
     }
 
     #[test]
-    fn can_opener_afford_feerate_checks() {
+    fn opener_balance_after_fee_sat_checks() {
```
morehouse: This test name no longer matches the API.

morehouse (approved):
Ready to squash the fixup.

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,611 @@
...
+            open_channel.dust_limit_satoshis, open_channel.channel_reserve_satoshis,
+        ));
+    }
+
+    // Check that the channel type was included.
+    // TODO: Check option_channel_type in negotiated features since it is
+    // assumed to be supported.
+    let Some(channel_type) = open_channel.tlvs.channel_type.as_deref() else {
+        return Err("open_channel does not include a channel_type".to_string());
+    };
```
morehouse, line 110: As I suspected, LND still allows implicit negotiation:

```
Test case failed: invalid accept_channel for temporary_channel_id ChannelId([112, 255, 15, 48, 188, 74, 47, 111, 121, 162, 25, 100, 124, 255, 90, 141, 148, 4, 178, 182, 147, 115, 26, 82, 167, 128, 214, 55, 227, 109, 97, 90]): accepted invalid open_channel: open_channel does not include a channel_type
```

Also, it would be good to print the channel ID in hex.

NishantBansal2003: Then isn't this a bug in LND? Since LND always sets `option_channel_type` in its init (https://github.com/lightningnetwork/lnd/pull/9637), and we're echoing the same feature, shouldn't LND require `channel_type` to be set when both peers have negotiated it?

I think LND is being lenient here for backward compatibility, but if both peers explicitly negotiated `option_channel_type`, it should reject an `open_channel` with a missing `channel_type` instead of falling back to implicit negotiation. WDYT?

smite/src/oracles/accept_channel.rs
```diff
@@ -0,0 +1,607 @@
...
+        ));
+    }
+
+    // Check the channel reserve covers the dust limit.
+    if open_channel.dust_limit_satoshis > open_channel.channel_reserve_satoshis {
+        return Err(format!(
+            "dust_limit_satoshis {} exceeds channel_reserve_satoshis {}",
+            open_channel.dust_limit_satoshis, open_channel.channel_reserve_satoshis,
+        ));
+    }
```
morehouse: LDK deliberately allows violations of this requirement (and it also violates the matching `open_channel.channel_reserve_satoshis >= accept_channel.dust_limit_satoshis` requirement) to enable zero (or below dust) reserve.  The relevant commit is https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/commit/25f4a54a2bdda91232a62034befc3c6d20611058.

I think the logic checks out, and probably we should just not enforce these two requirements in the oracle.  But we should probably keep a comment explaining why the checks are omitted.

smite/src/oracles/accept_channel.rs
morehouse: It's a spec violation, and we should report it (or submit a fix ourselves).  But we also probably want a workaround for smite until it is fixed in LND.

NishantBansal2003: Okay, I'm working on the LND fix. For the workaround, I'll handle it in a follow-up PR, since CLN also needs a workaround for the `funding_satoshis > 21M` failure violation.

morehouse (approved):
LGTM

PR #186, opened 2026-08-03
smite: add is_standard_shutdown_script helper

ekzyis:
Depends on #192.

From the commit message:

> BOLT-02 specifies sender requirements for shutdown scripts. They must be
> witness v0 (P2WPKH, P2WSH) or following features must be negotiated:
>
> * `option_shutdown_anysegwit`: witness v1-v16 with a 2..=40 byte program
> * `option_simple_close`: `OP_RETURN` with a single data push of 6..=80
>   bytes
> 
> Legacy scripts (P2PKH, P2SH) may be accepted for backward compatibility.
>
> This applies to the `shutdown` and `closing_complete` messages, and the
> `upfront_shutdown_script` TLV in the `open_channel`, `open_channel2`,
> `accept_channel` and `accept_channel2` messages.
>
> This commit adds a helper to catch targets that don't comply with the
> spec.

~As per the note I added to the code, I'm not sure if the fuzzer should also reject legacy scripts, since a target must not send them.~ _update: decided to reject them, see discussion_

I haven't wired this into existing code or #163 yet, but I thought the introduction of the helper might be worthwile to review itself, especially considering the question wrt legacy scripts.

### DIFF
--- a/smite/src/bolt/shutdown.rs
+++ b/smite/src/bolt/shutdown.rs
@@ -1,6 +1,11 @@
 //! BOLT 2 shutdown message.
 
+use bitcoin::opcodes::all::OP_RETURN;
+use bitcoin::script::Instruction;
+use bitcoin::{Script, WitnessVersion};
+
 use super::BoltError;
+use super::Features;
 use super::types::ChannelId;
 use super::wire::WireFormat;
 
@@ -51,10 +56,115 @@ impl Shutdown {
     }
 }
 
+/// Returns `true` if `spk` is a standard shutdown scriptpubkey per BOLT 2: P2WPKH or P2WSH.
+/// Negotiated `features` widen the accepted set.
+///
+/// Legacy P2PKH/P2SH are rejected. A receiver may accept them for backward compatibility, but this
+/// oracle judges the sender's output.
+#[must_use]
+pub fn is_standard_shutdown_script(spk: &[u8], features: &Features) -> bool {
+    let script = Script::from_bytes(spk);
+    let witness_v0 = script.is_p2wpkh() || script.is_p2wsh();
+    let anysegwit = features.supports_feature(Features::OPTION_SHUTDOWN_ANYSEGWIT)
+        && matches!(script.witness_version(), Some(v) if v != WitnessVersion::V0);
+    let simple_close = features.supports_feature(Features::OPTION_SIMPLE_CLOSE)
+        && is_simple_close_op_return(script);
+    witness_v0 || anysegwit || simple_close
+}
+
+/// Returns `true` if `spk` is a shutdown scriptpubkey a receiver may accept per
+/// BOLT 2.
+///
+/// This includes the standard shutdown scriptpubkey forms, as well as legacy
+/// P2PKH/P2SH outputs accepted for backward compatibility.
+#[must_use]
+pub fn is_acceptable_shutdown_script(spk: &[u8], features: &Features) -> bool {
+    let script = Script::from_bytes(spk);
+    is_standard_shutdown_script(spk, features) || script.is_p2pkh() || script.is_p2sh()
+}
+
+/// Returns `true` if `script` is a BOLT 2 `option_simple_close` `OP_RETURN` script: `OP_RETURN`
+/// followed by a single minimal push of 6..=80 bytes.
+///
+/// A non-minimal push here would be `OP_PUSHDATA1` used for a payload of fewer than 76 bytes.
+fn is_simple_close_op_return(script: &Script) -> bool {
+    let mut instrs = script.instructions_minimal();
+    if !matches!(instrs.next(), Some(Ok(Instruction::Op(op))) if op == OP_RETURN) {
+        return false;
+    }
+    match instrs.next() {
+        // matches any minimal push
+        Some(Ok(Instruction::PushBytes(bytes))) => {
+            (6..=80).contains(&bytes.len()) && instrs.next().is_none()
+        }
+        _ => false,
+    }
+}
+
 #[cfg(test)]
 mod tests {
+    use bitcoin::opcodes::all::{
+        OP_CHECKSIG, OP_DUP, OP_EQUAL, OP_EQUALVERIFY, OP_HASH160, OP_PUSHBYTES_0, OP_PUSHBYTES_1,
+        OP_PUSHBYTES_20, OP_PUSHBYTES_21, OP_PUSHBYTES_32, OP_PUSHDATA1, OP_PUSHNUM_1,
+        OP_PUSHNUM_16, OP_RETURN,
+    };
+
     use super::super::CHANNEL_ID_SIZE;
     use super::*;
+
+    fn no_features() -> Features {
+        Features::new()
+    }
+    fn with_anysegwit() -> Features {
+        Features::from_bits(&[Features::OPTION_SHUTDOWN_ANYSEGWIT])
+    }
+    fn with_simple_close() -> Features {
+        Features::from_bits(&[Features::OPTION_SIMPLE_CLOSE])
+    }
+    fn all_shutdown_features() -> Features {
+        Features::from_bits(&[
+            Features::OPTION_SHUTDOWN_ANYSEGWIT,
+            Features::OPTION_SIMPLE_CLOSE,
+        ])
+    }
+
+    fn p2pkh() -> Vec<u8> {
+        let mut spk = vec![OP_DUP.to_u8(), OP_HASH160.to_u8(), OP_PUSHBYTES_20.to_u8()];
+        spk.extend_from_slice(&[0x11; 20]);
+        spk.extend_from_slice(&[OP_EQUALVERIFY.to_u8(), OP_CHECKSIG.to_u8()]);
+        spk
+    }
+
+    fn p2sh() -> Vec<u8> {
+        let mut spk = vec![OP_HASH160.to_u8(), OP_PUSHBYTES_20.to_u8()];
+        spk.extend_from_slice(&[0x22; 20]);
+        spk.push(OP_EQUAL.to_u8());
+        spk
+    }
+
+    fn p2wpkh() -> Vec<u8> {
+        let mut spk = vec![OP_PUSHBYTES_0.to_u8(), OP_PUSHBYTES_20.to_u8()];
+        spk.extend_from_slice(&[0x33; 20]);
+        spk
+    }
+
+    fn p2wsh() -> Vec<u8> {
+        let mut spk = vec![OP_PUSHBYTES_0.to_u8(), OP_PUSHBYTES_32.to_u8()];
+        spk.extend_from_slice(&[0x44; 32]);
+        spk
+    }
+
+    fn anysegwit() -> Vec<u8> {
+        let mut spk = vec![OP_PUSHNUM_1.to_u8(), 32];
+        spk.extend_from_slice(&[0x00; 32]);
+        spk
+    }
+
+    fn simple_close() -> Vec<u8> {
+        let mut spk = vec![OP_RETURN.to_u8(), 6];
+        spk.extend_from_slice(&[0xab; 6]);
+        spk
+    }
 
     #[test]
     fn shutdown_for_channel() {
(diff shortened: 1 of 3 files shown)


smite/src/bolt/shutdown.rs
```diff
@@ -51,11 +54,72 @@ impl Shutdown {
...
+    pub option_simple_close: bool,
+}
+
+/// Returns `true` if `spk` is a standard `shutdown` scriptpubkey per BOLT 2:
+/// P2PKH, P2SH, P2WPKH, or P2WSH. Negotiated `features` widen the accepted set.
+#[must_use]
+pub fn is_standard_shutdown_script(spk: &[u8], features: ShutdownScriptFeatures) -> bool {
+    let script = Script::from_bytes(spk);
+    // NOTE(ek): senders MUST NOT send legacy scripts, but receivers MAY accept them for backward
+    // compatibility. It's not clear to me if the fuzzer should therefore accept or reject them.
```
NishantBansal2003: I think the fuzzer should reject it if the target sends a legacy script. The reasons are: Our underlying `bitcoind` uses `SegWit` by default, so if the target picks a default address, it will definitely be `SegWit`. But, if the target explicitly asks for a legacy script, I think that should be considered incorrect based on the BOLT specification. WDYT?

smite/src/bolt/shutdown.rs
```diff
@@ -51,11 +54,72 @@ impl Shutdown {
...
 
+/// Feature bits that widen the set of standard `shutdown` scriptpubkeys.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct ShutdownScriptFeatures {
+    /// Additionally permits witness program versions 1..=16 with a 2..=40 byte
+    /// program.
+    pub option_shutdown_anysegwit: bool,
+    /// Additionally permits a single-push `OP_RETURN` script.
+    pub option_simple_close: bool,
+}
```
NishantBansal2003: I think this can be removed/simplified once https://github.com/lnfuzz/smite/pull/192 gets merged, so we can just use just use: `negotiated_features.supports_feature(Features::OPTION_SHUTDOWN_ANYSEGWIT)` or `negotiated_features.supports_feature(Features::OPTION_SIMPLE_CLOSE)`

smite/src/bolt/shutdown.rs
```diff
@@ -51,11 +54,72 @@ impl Shutdown {
...
+    let witness_v0 = script.is_p2wpkh() || script.is_p2wsh();
+    let anysegwit = features.option_shutdown_anysegwit
+        && matches!(script.witness_version(), Some(v) if v != WitnessVersion::V0);
+    let simple_close = features.option_simple_close && is_simple_close_op_return(spk);
+    legacy || witness_v0 || anysegwit || simple_close
+}
+
+/// Returns `true` if `spk` is a BOLT 2 `option_simple_close` `OP_RETURN`
+/// script: `OP_RETURN` followed by a single data push of 6 to 80 bytes.
+fn is_simple_close_op_return(spk: &[u8]) -> bool {
```
NishantBansal2003: nit: using the `bitcoin` crate would make this more understandable/consistent. See: https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/src/commit/384e0d61305b24f5c6cdebe59c2c8f89b015d7b0/lightning/src/ln/script.rs#L135

smite/src/bolt/shutdown.rs
```diff
@@ -151,4 +215,138 @@ mod tests {
...
+        assert!(is_standard_shutdown_script(&p2pkh, ALL));
+
+        let mut p2sh = vec![OP_HASH160.to_u8(), OP_PUSHBYTES_20.to_u8()];
+        p2sh.extend_from_slice(&[0x22; 20]);
+        p2sh.push(OP_EQUAL.to_u8());
+        assert!(is_standard_shutdown_script(&p2sh, NONE));
+        assert!(is_standard_shutdown_script(&p2sh, ALL));
+
+        let v0 = OP_PUSHBYTES_0.to_u8();
+        let mut p2wkh = vec![v0, OP_PUSHBYTES_20.to_u8()];
```
NishantBansal2003:
```suggestion
        let mut p2wpkh = vec![v0, OP_PUSHBYTES_20.to_u8()];
```

NishantBansal2003 (commented):
Thanks! I was about to add this as a follow-up to https://github.com/lnfuzz/smite/pull/185, but it looks like I don’t have to now

smite/src/bolt/shutdown.rs
ekzyis: Sounds good, I will review #192 and then rebase this PR after merge

smite/src/bolt/shutdown.rs
ekzyis: Ah, nice, thanks! Done in 220b1bf. I noticed I need to use `script.instructions_minimal()` instead of `script.instructions()` to catch non-minimal pushes, see comment + test.

LDK intentionally accepts non-minimal pushes because of Postel's law, see [here](https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/src/commit/384e0d61305b24f5c6cdebe59c2c8f89b015d7b0/lightning/src/ln/script.rs#L339-L343).

smite/src/bolt/shutdown.rs
ekzyis: done in 220b1bf

smite/src/bolt/shutdown.rs
ekzyis: Makes sense, done in 220b1bf

smite/src/bolt/shutdown.rs
```diff
@@ -51,11 +55,77 @@ impl Shutdown {
...
+/// followed by a single minimal push of 6..=80 bytes.
+///
+/// A non-minimal push here would be `OP_PUSHDATA1` used for a payload of fewer than 76 bytes.
+fn is_simple_close_op_return(script: &Script) -> bool {
+    let mut instrs = script.instructions_minimal();
+    if !matches!(instrs.next(), Some(Ok(Instruction::Op(op))) if op == OP_RETURN) {
+        return false;
+    }
+    match instrs.next() {
+        // matches any minimal push of 6..=80 bytes
```
ekzyis: Note for myself: drop the "of 6..=80 bytes" since it will match any amount of bytes and we assert the length ourselves

ekzyis: done in 847c2f1

smite/src/bolt/shutdown.rs
```diff
@@ -51,11 +55,77 @@ impl Shutdown {
...
+    /// program.
+    pub option_shutdown_anysegwit: bool,
+    /// Additionally permits a single-push `OP_RETURN` script.
+    pub option_simple_close: bool,
+}
+
+/// Returns `true` if `spk` is a standard `shutdown` scriptpubkey per BOLT 2: P2WPKH or P2WSH.
+/// Negotiated `features` widen the accepted set.
+///
+/// Legacy P2PKH/P2SH are rejected. A receiver may accept them for backward compatibility, but this
```
NishantBansal2003, line 62: In the oracle, we verify what the peer accepted (and whether it was valid), and then we check our side to ensure that the target is actually sending that.

So, when verifying a target-accepted `open_channel,` I need it to pass even if it contains the legacy script. However, for our received `accept_channel`, I need to ensure that it does not contain legacy scripts

smite/src/bolt/shutdown.rs
```diff
@@ -151,4 +221,142 @@ mod tests {
         assert_eq!(original, decoded);
         assert!(decoded.scriptpubkey.is_empty());
     }
+
+    #[test]
+    #[allow(clippy::similar_names)]
+    fn is_standard_shutdown_script_rejects_legacy_accepts_witness_v0() {
```
NishantBansal2003: nit: I think for testing semantic script types, we should use the `bitcoin` crate to create the scripts, eg: `ScriptBuf::new_p2wpkh(&WPubkeyHash::all_zeros()).into_bytes();`, but for testing non-standard cases, we can construct the raw bytes manually

smite/src/bolt/shutdown.rs
```diff
@@ -151,4 +221,142 @@ mod tests {
...
+        assert!(is_standard_shutdown_script(&p2wpkh, ALL));
+
+        let mut p2wsh = vec![v0, OP_PUSHBYTES_32.to_u8()];
+        p2wsh.extend_from_slice(&[0x44; 32]);
+        assert!(is_standard_shutdown_script(&p2wsh, NONE));
+        assert!(is_standard_shutdown_script(&p2wsh, ALL));
+    }
+
+    #[test]
+    fn is_standard_shutdown_script_accepts_anysegwit() {
```
NishantBansal2003, line 275: nit: spec allows `2..=40 bytes`, so we can test the boundary cases as well

smite/src/bolt/shutdown.rs
```diff
@@ -151,4 +221,142 @@ mod tests {
...
+        trailing.extend_from_slice(&[OP_PUSHBYTES_1.to_u8(), 0xff]);
+        assert!(!is_standard_shutdown_script(&trailing, ALL));
+
+        // OP_PUSHDATA1 with a length above the 80-byte maximum.
+        let mut too_long = vec![OP_RETURN.to_u8(), OP_PUSHDATA1.to_u8(), 81];
+        too_long.extend_from_slice(&[0xab; 81]);
+        assert!(!is_standard_shutdown_script(&too_long, ALL));
+
+        // Non-minimal push: OP_PUSHDATA1 used for less than 76 bytes.
+        let mut non_minimal = vec![OP_RETURN.to_u8(), OP_PUSHDATA1.to_u8(), 10];
```
NishantBansal2003: nit: could test it at the boundary, i.e. 75, to be explicit about the boundary

smite/src/bolt/shutdown.rs
```diff
@@ -151,4 +221,142 @@ mod tests {
...
+        // rejection under any subset.
+        let empty_spk = vec![];
+        assert!(!is_standard_shutdown_script(&empty_spk, ALL));
+
+        let random_spk = vec![0x00, 0x01, 0x02];
+        assert!(!is_standard_shutdown_script(&random_spk, ALL));
+
+        // OP_RETURN with a 1-byte push: below the simple_close minimum of 6.
+        let op_return_spk = vec![OP_RETURN.to_u8(), OP_PUSHBYTES_1.to_u8(), 0xff];
+        assert!(!is_standard_shutdown_script(&op_return_spk, ALL));
```
NishantBansal2003: nit: this case is similar to one of the assertions added in `is_standard_shutdown_script_rejects_invalid_simple_close_op_return`. I think we can follow the same pattern here, remove this `OP_RETURN` assertion, and rename the test to something like `is_standard_shutdown_script_rejects_invalid_segwit` or a better name, since this test mostly contains those cases

NishantBansal2003 (approved):
Mostly, it looks good. Just a few small comments. I'm not sure how this follows the merge with https://github.com/lnfuzz/smite/pull/192, but if it's merged before, I'll add the updated feature check there

ekzyis:
> I'm not sure how this follows the merge with https://github.com/lnfuzz/smite/pull/192

Putting this into draft until #192 is merged, I rebased this and I addressed the feedback

smite/src/bolt/shutdown.rs
ekzyis: Mhh. If we used `bitcoin::ScriptBuf` to construct standard scripts via `ScriptBuf::new_p2wpkh` etc., I think we'd mostly be testing the `bitcoin` crate against itself, since `is_standard_shutdown_script` is implemented with `bitcoin::Script::is_p2wpkh()`.

It would also make it less clear *how* we are testing. I think raw opcodes make it explicit which bytes we are testing. I think tests should be optimized for clear understanding of *what* we are testing *how*. This is why I think it's ok to use `bitcoin::opcodes`.

I think @morehouse mentioned something similar in a review, but I didn't find it.

smite/src/bolt/shutdown.rs
ekzyis: done in 63fed11: added a loop over 2, 20 and 40

smite/src/bolt/shutdown.rs
ekzyis: done in 63fed11

smite/src/bolt/shutdown.rs
ekzyis: Done in 63fed11. There are four rejection tests now: `rejects_legacy`, `rejects_invalid_segwit`. `rejects_invalid_simple_close_op_return` and `rejects_other`.

smite/src/bolt/shutdown.rs
ekzyis: I didn't address this yet in 63fed11. I think we want to handle this in a similar way to `enum Side`.

ekzyis:
I rebased this PR on top of #192 since I don't expect it to change much. I didn't address all feedback yet, see https://github.com/lnfuzz/smite/pull/186#discussion_r3823161402 and TODO in the code + commit message.

---

Update: rebased again on top of #192 (ACK 40f7aa8)

ekzyis:
TODO:

* [x] support legacy shutdown scripts for `open_channel` sent to target (https://github.com/lnfuzz/smite/pull/186#discussion_r3758391851)

morehouse (commented):
Great work!  Needs a rebase now that #192 is merged.  Also would be good to include the additions from #209 before merging this.

ekzyis:
c62c850: rebased on master (87ebab6), added `is_acceptable_shutdown_script` from #192, and updated commit message. I think `is_acceptable_shutdown_script` is simple and enough for our needs; no need for `enum Side` like we did for commitments.

f024116: fixed wrong usage of backticks in `is_standard_shutdown_script` documentation

smite/src/bolt/shutdown.rs
```diff
@@ -51,11 +56,78 @@ impl Shutdown {
...
+/// Returns `true` if `spk` is a shutdown scriptpubkey a receiver may accept per
+/// BOLT 2.
+///
+/// This includes the standard shutdown scriptpubkey forms, as well as legacy
+/// P2PKH/P2SH outputs accepted for backward compatibility.
+#[must_use]
+pub fn is_acceptable_shutdown_script(spk: &[u8], features: &Features) -> bool {
+    let script = Script::from_bytes(spk);
+    is_standard_shutdown_script(spk, features) || script.is_p2pkh() || script.is_p2sh()
+}
```
morehouse, line 84: We should add tests for `is_acceptable_shutdown_script`.

smite/src/bolt/shutdown.rs
```diff
@@ -151,4 +223,191 @@ mod tests {
...
+            assert!(is_standard_shutdown_script(&spk, &with_simple_close()));
+        }
+    }
+
+    #[test]
+    fn is_standard_shutdown_script_rejects_legacy() {
+        // Legacy P2PKH/P2SH are non-standard even with all features negotiated.
+        //
+        // TODO: Oracle verification depends on if we're verifying a target's message or our own
+        // message. Legacy scripts MAY be accepted by receivers, but MUST NOT be sent.
```
morehouse:
```suggestion
```

smite/src/bolt/shutdown.rs
ekzyis: done in ab9cbad

smite/src/bolt/shutdown.rs
ekzyis: done in ab9cbad

ekzyis:
ab9cbad: rebased again on master (d0b2679) + updated commit message

With the new tests, I noticed I could define helpers for p2pkh, p2sh, etc., so there's less to read per test now.

morehouse (approved):
LGTM

PR #187, opened 2026-08-03
smite-scenarios: skip `ping_pong` for parked connections

morehouse:
A bug in CLN causes it to park the connection after receiving a `channel_ready` message with an incorrect `channel_id`.  When this happens, CLN keeps the connection open until the 80s timeout elapses but is completely unresponsive until then.

Check error messages from the target to detect when CLN hits this case, and then skip the `ping_pong` sync to avoid flagging this known issue as a hang.

Ref: https://github.com/ElementsProject/lightning/issues/9369

### DIFF
--- a/smite-scenarios/src/scenarios.rs
+++ b/smite-scenarios/src/scenarios.rs
@@ -16,10 +16,37 @@ use smite::scenarios::ScenarioError;
 use std::time::Duration;
 
 use bitcoin::secp256k1::SecretKey;
-use smite::bolt::{Init, Message, Ping};
+use smite::bolt::{Error, Init, Message, Ping};
 use smite::noise::NoiseConnection;
 
 use crate::targets::Target;
+
+/// Peer `error` messages after which a target is known to "park" the
+/// connection, keeping it open but no longer servicing any messages on it.
+/// When this happens the target will not respond to our pings.
+///
+/// Every entry must describe the behavior justifying it, and be removed once
+/// the upstream fix lands.
+///
+/// - `"Wrong channel id"`: CLN permanently fails the channel without
+///   disconnecting. If a follow-on message arrives while the channel's
+///   subdaemon is still dying, lightningd fails to handle the message properly,
+///   and connectd is left waiting for lightningd's response indefinitely. The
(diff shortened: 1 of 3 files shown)


smite-scenarios/src/scenarios.rs
```diff
@@ -76,14 +112,72 @@ pub fn handshake_with_target<T: Target>(
...
 
     // Read messages until we get a pong
     loop {
         let msg_bytes = conn.recv_message()?;
-        if matches!(Message::decode(&msg_bytes)?, Message::Pong(_)) {
-            return Ok(());
+        match Message::decode(&msg_bytes)? {
+            Message::Pong(_) => return Ok(PingOutcome::Pong),
+            Message::Error(e) if stop_on_known_error && is_known_parked_error(&e) => {
+                let msg = e.message().unwrap_or_default().to_string();
```
NishantBansal2003: nit (optional)
```suggestion
                let msg = e.message().unwrap_or("<non-utf8>").to_string();
```

NishantBansal2003 (approved):
LGTM!

Though I do think patches are a good fit for target specific fixes, since they avoid changing smites behavior for other targets and potentially masking their issues, the downside is that if LN implementation doesn't fix the issue in the next release and updates the patched file, we'll have to redo the patch

smite-scenarios/src/scenarios.rs
morehouse: Done, thanks.

morehouse:
> Though I do think patches are a good fit for target specific fixes, since they avoid changing smites behavior for other targets and potentially masking their issues, the downside is that if LN implementation doesn't fix the issue in the next release and updates the patched file, we'll have to redo the patch

Yeah, unfortunately the [patch](https://github.com/ElementsProject/lightning/pull/9370) for this one is a bit complex and touches many files.  Maintaining that patch in our tree would likely be a burden.  I'm also not confident that it will be merged upstream anytime soon.

I did verify that the error string we're matching against ("Wrong channel id") isn't emitted by any of the other implementations, so this change shouldn't affect them at all.

PR #190, opened 2026-08-06
smitebot:  avoid afl-cmin filename overflow on direct corpus minimize

Ashish-Kumar-Dash:
Follow-up to the corpus command #184 . 

The default path now stages the campaign's queues through the same content-dedup as `merge` into a short-named temp corpus, runs afl-cmin on that, and removes it afterward. A supplied `--input` stays passthrough.

### DIFF
--- a/smitebot/README.md
+++ b/smitebot/README.md
@@ -184,10 +184,10 @@ Removes corpus inputs that do not contribute new coverage, using `afl-cmin` in N
 
 ```bash
 smitebot corpus minimize <campaign-id>
-smitebot corpus minimize <campaign-id> [-i <input>] [-o <output-dir>] [--aflpp-path <path>]
+smitebot corpus minimize <campaign-id> [-i <dir>]... [-o <output-dir>] [--aflpp-path <path>]
 ```
 
 - `<campaign-id>`: directory name under `~/.smitebot/runs`
-- `-i, --input <input>`: input directory or glob pattern passed to `afl-cmin -i`; defaults to `<output_dir>/*/queue/` (all runner queues)
+- `-i, --input <dir>`: One input directory to minimize. If multiple `-i` flags are present, all specified directories are merged before minimizing. If omitted, the campaign's runner queues are merged and minimized.
 - `-o, --output <output-dir>`: output directory; defaults to `~/.smitebot/runs/<id>/corpus-min/`
 - `--aflpp-path <path>`: AFL++ source tree, overriding the `aflpp_path` stored in `state.json` (useful when the checkout has moved)
--- a/smitebot/src/commands/corpus.rs
+++ b/smitebot/src/commands/corpus.rs
@@ -48,10 +48,10 @@ pub struct MergeArgs {
 pub struct MinimizeArgs {
     /// Campaign ID whose sharedir and `aflpp_path` to use.
     campaign_id: String,
-    /// Input directory or glob pattern passed to `afl-cmin -i`
-    /// (default: `<output_dir>/*/queue/`).
+    /// One input directory to minimize. Repeat `-i` to merge several; if
+    /// omitted, the campaign's runner queues are used.
     #[arg(short = 'i', long)]
-    input: Option<String>,
+    input: Vec<PathBuf>,
     /// Output directory (default: `~/.smitebot/runs/<id>/corpus-min/`).
     #[arg(short = 'o', long)]
     output: Option<PathBuf>,
@@ -132,7 +132,20 @@ fn execute_merge(args: &MergeArgs) -> bool {
     true
 }
 
-/// Copies unique queue files from every runner of every state into `output`.
+/// Merges every runner queue of every state into `output`. See [`merge_dirs`].
+fn merge_states(states: &[CampaignState], output: &Path) -> Option<(usize, usize)> {
+    let mut dirs = Vec::new();
+    for state in states {
+        for runner in &state.runners {
+            dirs.push(state.output_dir.join(runner.name()).join("queue"));
+        }
+    }
+    merge_dirs(&dirs, output)
+}
+
+/// Copies unique files (by content hash) from each directory in `sources` into
+/// `output` under short, sequential names. Missing directories are skipped with
+/// a warning.
 ///
 /// Deduplicates by a hash of the file contents rather than the contents
 /// themselves, so a large corpus only costs 8 bytes per unique entry in memory.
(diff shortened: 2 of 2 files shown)


smitebot/src/commands/corpus.rs
```diff
@@ -271,6 +285,38 @@ fn execute_minimize(args: &MinimizeArgs) -> bool {
     true
 }
 
+/// Stages a campaign's runner queues into `stage_dir` with short, sequential
+/// filenames, returning the directory to hand to `afl-cmin -i`.
+///
+/// afl-cmin -X hardlinks each input to `<8hex>_<basename>` (batch mode); smite-ir
```
morehouse: Is this true?  I don't recall seeing such name formats; I thought it just preserved the existing filenames.

smitebot/src/commands/corpus.rs
```diff
@@ -271,6 +285,38 @@ fn execute_minimize(args: &MinimizeArgs) -> bool {
...
+        log::error!(
+            "failed to create staging directory {}: {e}",
+            stage_dir.display()
+        );
+        return None;
+    }
+
+    let (total_in, total_out) = merge_states(std::slice::from_ref(state), stage_dir)?;
+    log::info!("staged {total_out} unique inputs (deduped from {total_in}) for minimization");
+    Some(stage_dir.display().to_string())
```
morehouse: Can we return a `PathBuf` for simplicity?

smitebot/src/commands/corpus.rs
```diff
@@ -254,14 +256,26 @@ fn execute_minimize(args: &MinimizeArgs) -> bool {
...
+        );
+    }
+
+    let status = match status {
         Ok(s) => s,
         Err(e) => {
             log::error!("failed to run {}: {e}", afl_cmin.display());
             return false;
         }
     };
```
morehouse: Nit: I think it would be cleaner to add a `run_afl_cmin` helper function that does everything as a unit, and then we can remove the stage dir after that call.

smitebot/src/commands/corpus.rs
```diff
@@ -271,6 +285,38 @@ fn execute_minimize(args: &MinimizeArgs) -> bool {
...
 }
 
+/// Stages a campaign's runner queues into `stage_dir` with short, sequential
+/// filenames, returning the directory to hand to `afl-cmin -i`.
+///
+/// afl-cmin -X hardlinks each input to `<8hex>_<basename>` (batch mode); smite-ir
+/// queue filenames already approach `NAME_MAX`, so the added prefix overflows
+/// (`ENAMETOOLONG`). Reusing `merge`'s content dedup gives short names (and drops
+/// duplicates, which afl-cmin would discard anyway). Returns `None` on failure,
+/// logged at the site.
```
morehouse: This description is quite wordy and difficult to understand.


```suggestion
/// AFL++ tends to produce long input names near `NAME_MAX`, and afl-cmin can then
/// exceed the max length during minimization. Do an implicit merge step first so that
/// inputs get a short name and afl-cmin is less likely to exceed the max path length.
```

smitebot/README.md
```diff
@@ -188,6 +188,6 @@ smitebot corpus minimize <campaign-id> [-i <input>] [-o <output-dir>] [--aflpp-p
 ```
 
 - `<campaign-id>`: directory name under `~/.smitebot/runs`
-- `-i, --input <input>`: input directory or glob pattern passed to `afl-cmin -i`; defaults to `<output_dir>/*/queue/` (all runner queues)
+- `-i, --input <input>`: input directory passed to `afl-cmin -i`; by default the campaign's runner queues are staged into a short-named temp corpus first (afl-cmin's `<hash>_<name>` staging overflows `NAME_MAX` on long smite-ir queue filenames), so no `--input` is needed. A supplied `--input` is passed through unchanged
```
morehouse: This seems like implementation details the user doesn't need to know.


```suggestion
- `-i, --input <input>`: input directory passed to `afl-cmin -i`; by default the campaign's runner queues are implicitly merged into a single directory that is passed to this flag
```

smitebot/src/commands/corpus.rs
```diff
@@ -243,6 +235,16 @@ fn execute_minimize(args: &MinimizeArgs) -> bool {
...
 
+    // Default = the raw runner queues; stage to short names first (see stage_queues).
+    let stage_dir = runs_dir.join(&args.campaign_id).join(".corpus-stage");
+    let input = match &args.input {
+        Some(i) => i.clone(),
+        None => match stage_queues(&state, &stage_dir) {
+            Some(dir) => dir,
+            None => return false,
+        },
+    };
```
morehouse: Nit:


```suggestion
    let Some(input) = args.input.clone().or_else(|| stage_queues(&state, &stage_dir)) else {
        return false;
    }
```

morehouse (changes requested):
I think we probably want to always do the merge step first, not just in the default-no-input case.  A manually-specified a queue dir as input is going to have the same long-pathname issue.  This probably also means we need the ability to specify multiple `--input` flags.

smitebot/src/commands/corpus.rs
Ashish-Kumar-Dash: Yes. That is afl-cmin's own temp staging, in its batched processing it hardlinks each input into .filelist.<pid>.d/ as [rand_bit function](https://github.com/AFLplusplus/AFLplusplus/blob/stable/afl-cmin.py#L405), which is exactly the path from the error you posted. That's where the length overflows.

smitebot/src/commands/corpus.rs
```diff
@@ -48,10 +48,10 @@ pub struct MergeArgs {
 pub struct MinimizeArgs {
     /// Campaign ID whose sharedir and `aflpp_path` to use.
     campaign_id: String,
-    /// Input directory or glob pattern passed to `afl-cmin -i`
-    /// (default: `<output_dir>/*/queue/`).
+    /// Input directories to minimize (repeatable). Default: the campaign's
+    /// runner queues.
```
NishantBansal2003: This `(repeatable)` is very confusing. I assume you mean that it can take multiple inputs. I think this should be made clearer

smitebot/src/commands/corpus.rs
```diff
@@ -243,18 +245,49 @@ fn execute_minimize(args: &MinimizeArgs) -> bool {
...
+            .runners
+            .iter()
+            .map(|r| state.output_dir.join(r.name()).join("queue"))
+            .collect()
+    } else {
+        args.input.clone()
+    };
+
+    let stage_dir = runs_dir.join(&args.campaign_id).join(".corpus-stage");
+    let Some(input) = stage_inputs(&sources, &stage_dir) else {
```
NishantBansal2003: If staging the inputs fails here, I think we should remove the stage dir, right? Or is that not needed since we always remove it at the start of staging anyway?

smitebot/README.md
```diff
@@ -188,6 +188,6 @@ smitebot corpus minimize <campaign-id> [-i <input>] [-o <output-dir>] [--aflpp-p
 ```
 
 - `<campaign-id>`: directory name under `~/.smitebot/runs`
-- `-i, --input <input>`: input directory or glob pattern passed to `afl-cmin -i`; defaults to `<output_dir>/*/queue/` (all runner queues)
+- `-i, --input <input>`: input directories to minimize (repeatable); these — or, by default, the campaign's runner queues — are implicitly merged into a single directory before minimizing
```
NishantBansal2003: Same applies here as well

smitebot/src/commands/corpus.rs
```diff
@@ -419,6 +483,55 @@ mod tests {
         assert_eq!(dir_contents_sorted(&out), ["shared", "unique"]);
     }
 
+    #[test]
+    fn stage_inputs_shortens_long_input_names() {
+        let dir = tempdir().unwrap();
+        let src = dir.path().join("queue");
+        fs::create_dir_all(&src).unwrap();
+        // A smite-ir queue filename near NAME_MAX (255). afl-cmin -X would prefix
+        // it with "<8hex>_" and overflow (ENAMETOOLONG); staging must rename it.
```
NishantBansal2003, line 495: nit: It would be great to cite code links (here or at `stage_inputs`) justifying this comment as well, to explain why this stage is required in the first place

NishantBansal2003 (approved):
Verified with a campaign: it was giving `OSError: [Errno 36] File name too long:` before this change, and the issue is fixed after this change.

smitebot/README.md
morehouse: The whole sentence is hard to read.  I would eliminate the emdash interruption and just simply describe what the command does.


```suggestion
- `-i, --input <dir>`: One input directory to minimize.  If multiple `-i` flags are present, all specified directories are merged before minimizing.  If omitted, the campaign's runner queues are merged and minimized.
```

smitebot/src/commands/corpus.rs
morehouse: I think it makes sense to clean up on error.

smitebot/src/commands/corpus.rs
```diff
@@ -271,6 +304,37 @@ fn execute_minimize(args: &MinimizeArgs) -> bool {
...
+        log::error!(
+            "failed to create staging directory {}: {e}",
+            stage_dir.display()
+        );
+        return None;
+    }
+
+    let (total_in, total_out) = merge_dirs(sources, stage_dir)?;
+    log::info!("staged {total_out} unique inputs (from {total_in}) for minimization");
+    Some(stage_dir.to_path_buf())
```
morehouse: This just returns a copy of the `stage_dir` the caller already owns.  Looks like we could just return `true` on success instead.

smitebot/README.md
morehouse:
```suggestion
smitebot corpus minimize <campaign-id> [-i <dir>]... [-o <output-dir>] [--aflpp-path <path>]
```

smitebot/src/commands/corpus.rs
```diff
@@ -271,6 +304,40 @@ fn execute_minimize(args: &MinimizeArgs) -> bool {
     true
 }
 
+/// Merges `sources` into `stage_dir` under short, sequential filenames for
+/// `afl-cmin -i`, returning whether staging succeeded.
+///
+/// AFL++ produces long input names near `NAME_MAX`, and afl-cmin can then exceed
+/// the max length during minimization: it stages each input into a temp dir as
+/// `<random-hex>_<basename>`, see
+/// <https://github.com/AFLplusplus/AFLplusplus/blob/stable/afl-cmin.py#L405>.
```
morehouse: The link should be pinned to a specific commit so that it remains stable over time.

morehouse (approved):

PR #192, opened 2026-08-07
smite: use BOLT 9 feature bitfield primitives

### LINKED ISSUE #217: False positive due to non-minimal channel type
Found a false positive while fuzzing LND:

```
Test case failed: invalid accept_channel for temporary_channel_id 0cdb006c68ca6c9d84f83bdc302b272f658ae589bd3481b3bd9fef4270a47420: invalid accept_channel: accept_channel channel_type does not match open_channel
```

The executor sent an `open_channel` with `channel_type = [0x00]`.  LND responded with the same value minimally encoded: `channel_type = []`.  The oracle flagged that they don't match.

### Suggested fix

We should strip leading zeroes before comparing channel types here:

https://github.com/lnfuzz/smite/blob/e3736eadd86078721cf25c53d1acb1c7af6f5f4b/smite/src/oracles/accept_channel.rs#L157-L160

NishantBansal2003:
Depends-on: https://github.com/lnfuzz/smite/pull/185

Adds BOLT 9 feature primitives, which are useful in places where we need to query different features/bits in negotiated features, channel_type, etc, to proceed further. This will also be useful in the future when we add dual-fund support, and is already useful when working on the `accept_channel` oracle, where we need to query supported features/channel_type

This PR also aggregates all the existing feature-related utilities into a single `Features` struct. I've only added the features currently used by smite, we can add more as needed in the future.

Currently, I've only updated the places that actually need to query Features. I haven't changed wire messages that still use `Vec<u8>` for features, to avoid changing a lot of code without any immediate benefit. I think we can use `features.rs` as a utility whenever we need to query or manipulate `Features`, while continuing to use `Vec<u8>` for wire messages where we don't need to inspect the feature bits

Fixes #217.


smite/src/bolt/features.rs
```diff
@@ -0,0 +1,255 @@
+//! BOLT 9 feature bitfield primitives.
+
+/// BOLT 9 feature bit index. Even bits are required; odd bits are optional.
```
ekzyis: nit: We could mention the "it's ok to be odd"-rule here.

smite/src/bolt/features.rs
```diff
@@ -0,0 +1,255 @@
...
+pub type FeatureBit = usize;
```
ekzyis, line 5: If we allow `usize`, I can imagine the fuzzer calling `Features::set_bit` with a bit close to `usize::MAX`. This will then try to allocate up to 2^64 bytes on x86-64.

Since the max message size is 65\_535 bytes, we know that feature bits can't be set higher than 65\_535\*8 - 1=524\_279, so `u32` should be enough (2^32 > 524_279 > 2^16):

```suggestion
pub type FeatureBit = u32;
```

However, since trying to allocate up to 2^32 bytes is still a lot, we could make `FeatureBit` a distinct type instead of only an alias for `u32`. We could then enforce the theoretical known limit as the maximum (524_279) instead of 2^32.

```suggestion
pub struct FeatureBit(u32);
```

Or is this not a reason for concern because we know the fuzzer won't call `set_bit` in the way I imagined?

smite/src/bolt/features.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    /// Consumes the features into their underlying bytes.
+    #[must_use]
+    pub fn into_bytes(self) -> Vec<u8> {
+        self.0
+    }
+
+    /// Sets the bit, extending the features with leading zero bytes if needed.
+    pub fn set_bit(&mut self, bit: FeatureBit) {
```
ekzyis, line 66: Mhh, afaict, if I want to use the constants defined above but want to set the optional bit of a feature, I would need to call this function like this: `f.set_bit(Features::GOSSIP_QUERIES ^ 1)`. I think that's awkward.

Maybe an interface with functions `set_required_bit` and `set_optional_bit` would be more intuitive? I also think reading `set_bit(Features::GOSSIP_QUERIES)` does not make it obvious which bit this is going to set. It requires knowledge of the convention of identifying features by their even bits in the list above.

smite/src/bolt/features.rs
```diff
@@ -0,0 +1,255 @@
+//! BOLT 9 feature bitfield primitives.
+
+/// BOLT 9 feature bit index. Even bits are required; odd bits are optional.
+pub type FeatureBit = usize;
+
+/// BOLT 9 feature bitfield, encoded as big-endian bytes.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
```
ekzyis: `PartialEq` will compare the bytes, so features with the same bits set but with more leading zero bytes will not match:

```rust
// this test fails
#[test]
fn features_equality_ignores_leading_zeroes() {
    assert_eq!(Features::from(vec![0x01]), Features::from(vec![0x00, 0x01]));
}
```


Not sure if that might become a problem. Might be good to add a test to at least document which behavior we're expecting.

ekzyis (commented):
Concept ACK

reviewed only 695d504 so far

smite/src/bolt/features.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// `option_scid_alias` (bits 46/47).
+    pub const OPTION_SCID_ALIAS: FeatureBit = 46;
+    /// `option_zeroconf` (bits 50/51).
+    pub const OPTION_ZEROCONF: FeatureBit = 50;
+    /// `option_simple_taproot` (bits 80/81).
+    pub const OPTION_SIMPLE_TAPROOT: FeatureBit = 80;
+    /// `option_simple_taproot_staging` (bits 180/181).
+    pub const OPTION_SIMPLE_TAPROOT_STAGING: FeatureBit = 180;
+    /// `option_script_enforced_lease` (bits 2022/2023).
+    pub const OPTION_SCRIPT_ENFORCED_LEASE: FeatureBit = 2022;
```
erickcestari, line 36: Not all of these features come from BOLT 9. It could be documented where they come from.

smite/src/bolt/features.rs
```diff
@@ -0,0 +1,255 @@
...
+        self.clear_bit(bit ^ 1);
+    }
+}
+
+impl From<Vec<u8>> for Features {
+    fn from(bytes: Vec<u8>) -> Self {
+        Self(bytes)
+    }
+}
+
```
erickcestari, line 140: We could also implement:

```rust
impl From<&[u8]> for Features {
    fn from(bytes: &[u8]) -> Self {
        Self(bytes.to_vec())
    }
}
```

smite/src/oracles/accept_channel.rs
```diff
@@ -105,7 +105,12 @@ fn verify_accepted_open_channel(open_channel: &OpenChannel) -> Result<(), String
     // Check that the channel type was included.
     // TODO: Check option_channel_type in negotiated features since it is
     // assumed to be supported.
-    let Some(channel_type) = open_channel.tlvs.channel_type.as_deref() else {
+    let Some(channel_type) = open_channel
+        .tlvs
+        .channel_type
+        .as_deref()
+        .map(|channel_type| Features::from(channel_type.to_vec()))
```
erickcestari: nit: If implement the trait `From<&[u8]>`

```suggestion
        .tlvs
        .channel_type
        .as_deref()
        .map(Features::from)
```

smite/src/oracles/accept_channel.rs
```diff
@@ -150,7 +155,12 @@ fn verify_accept_channel(
     open_channel: &OpenChannel,
 ) -> Result<(), String> {
     // Check that the channel type was included.
-    let Some(channel_type) = accept_channel.tlvs.channel_type.as_deref() else {
+    let Some(channel_type) = accept_channel
+        .tlvs
+        .channel_type
+        .as_deref()
+        .map(|channel_type| Features::from(channel_type.to_vec()))
```
erickcestari: nit:
```suggestion
        .tlvs
        .channel_type
        .as_deref()
        .map(Features::from)
```

erickcestari (changes requested):

smite/src/bolt/features.rs
NishantBansal2003: I don't expect the fuzzer to utilize the `Features` type at all. As mentioned above, `Features` is just a utility type and should be used where we need to extract or generate valid information from it. Otherwise, we should use the low-level `Vec<u8>` for the underlying fuzzer.

FYI, LDK also uses `usize`: https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/src/commit/8700bf1d1e9d9df1e7426aeb77d6c148b12656f7/lightning-types/src/features.rs#L1286

smite/src/bolt/features.rs
NishantBansal2003: AFAICT, fuzzer can use the low-level Vec to generate whatever features it wants. If we do want to set a specific feature, I think we should set it as required only, rather than optional, otherwise, setting that feature might not be useful.

Can you tell me in which cases you think we might need to set an optional bit instead of a required one?

smite/src/bolt/features.rs
NishantBansal2003: Thanks, this was a great catch. I verified all implementations, and they compare feature equality based on the bits set, so I fixed that in: https://github.com/lnfuzz/smite/pull/192/commits/455a7aa18701c18e4c7e505656ff407def3eb530

smite/src/bolt/features.rs
ekzyis: > Otherwise, we should use the low-level `Vec<u8>` for the underlying fuzzer.

Ok makes sense thank you!

> FYI, LDK also uses `usize`

Even if the fuzzer isn't going to call this, should we still add a check that the bit is within the theoretical limit? LDK also does this:

https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/src/commit/8700bf1d1e9d9df1e7426aeb77d6c148b12656f7/lightning-types/src/features.rs#L1289

smite/src/bolt/features.rs
ekzyis: > Can you tell me in which cases you think we might need to set an optional bit instead of a required one?

No. We don't need `set_optional_bit` then, but we could still rename `set_bit` to `set_required_bit` to make the function name more self-describing (and rename the rest consistently). 

When I read "set bit of this feature", I don't know which bit is meant.

smite/src/bolt/features.rs
NishantBansal2003: Makes sense!

smite/src/bolt/features.rs
NishantBansal2003: > we could still rename `set_bit` to `set_required_bit`

No, but `set_bit` can set an optional bit as well


> When I read "set bit of this feature", I don't know which bit is meant.

`set_bit` takes the bit as input, so I expect the caller to understand which bit they want to set


Currently, I've only added required feature bits as constants, so `set_bit` can be used for required bits. But if, in the future, a caller needs to set an optional feature bit, I think we can define the optional feature bit as a constant as well, and suffix the constants with `_REQUIRED` and `_OPTIONAL`, respectively.

erickcestari (approved):
LGTM!

smite/src/bolt/features.rs
```diff
@@ -0,0 +1,324 @@
...
+        );
+
+        let mut len = self.0.len();
+        if len <= byte_offset {
+            let new_len = byte_offset + 1;
+            let mut new_features = vec![0u8; new_len];
+            new_features[(new_len - len)..].copy_from_slice(&self.0);
+            self.0 = new_features;
+            len = new_len;
+        }
```
ekzyis: nit: I think this is slightly easier to read but up to you:

```diff
diff --git a/smite/src/bolt/features.rs b/smite/src/bolt/features.rs
index 2e18c36..9fabfb5 100644
--- a/smite/src/bolt/features.rs
+++ b/smite/src/bolt/features.rs
@@ -70,16 +70,14 @@ impl Features {
             "feature bit {bit} exceeds the feature bitfield length limit"
         );
 
-        let mut len = self.0.len();
-        if len <= byte_offset {
-            let new_len = byte_offset + 1;
-            let mut new_features = vec![0u8; new_len];
-            new_features[(new_len - len)..].copy_from_slice(&self.0);
-            self.0 = new_features;
-            len = new_len;
+        let required_len = byte_offset + 1;
+        if self.0.len() < required_len {
+            // prepend enough zeroes
+            self.0.splice(0..0, vec![0u8; required_len - self.0.len()]);
         }
 
         let mask = 1 << (bit % 8);
+        let len = self.0.len();
         self.0[len - 1 - byte_offset] |= mask;
     }
 
```

smite/src/bolt/features.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    #[test]
+    fn equality_with_same_and_different_bits() {
+        let lease = Features::from_bits(&[Features::OPTION_SCRIPT_ENFORCED_LEASE]);
+        assert_eq!(lease, Features::from_bits(&[2022]));
+        assert_eq!(
+            Features::from(vec![0x02, 0x01]),
+            Features::from(vec![0x02, 0x01])
+        );
```
ekzyis, line 308: nit: I initially thought this was unintentional because it looks like a trivial assert, so a comment could be nice:

```suggestion
        // same bytes -> same features
        assert_eq!(
            Features::from(vec![0x02, 0x01]),
            Features::from(vec![0x02, 0x01])
        );
```

smite/src/bolt/features.rs
ekzyis: > But if, in the future, a caller needs to set an optional feature bit, I think we can define the optional feature bit as a constant as well and suffix the constants with `_REQUIRED` and `_OPTIONAL`, respectively.

Makes sense. I think my only issue is that I think the `_REQUIRED` suffix would already be useful. But consider this a nit.

Edit: To clarify further: 

> the `_REQUIRED` suffix would already be useful

I think this is downstream of the inconsistency in how `FeatureBit` is treated as a function argument. Sometimes it's treated as a feature (`supports_feature`); sometimes as a bit (`set_bit`).

Anyway, I'm not going to not ACK this PR because of this.

ekzyis (commented):
partial review, got distracted, will continue review today

ekzyis (changes requested):
43b1bf3 is failing to build:

<details>
<summary>cargo test -p smite</summary>

```
$ cargo test -p smite
   Compiling smite v0.0.0 (/home/ekzyis/prog/smite/smite)
error[E0308]: mismatched types
   --> smite/src/oracles/accept_channel.rs:227:76
    |
227 |     let commitment_cost = CommitmentCost::new(open_channel.feerate_per_kw, channel_type);
    |                           -------------------                              ^^^^^^^^^^^^ expected `&Features`, found `&[u8]`
    |                           |
    |                           arguments to this function are incorrect
    |
    = note: expected reference `&Features`
               found reference `&[u8]`
note: associated function defined here
   --> smite/src/channel_tx/commitment.rs:472:12
    |
472 |     pub fn new(feerate_per_kw: u32, channel_type: &Features) -> CommitmentCost {
    |            ^^^                      -----------------------

For more information about this error, try `rustc --explain E0308`.
error: could not compile `smite` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error[E0308]: mismatched types
   --> smite/src/oracles/accept_channel.rs:227:76
    |
227 |     let commitment_cost = CommitmentCost::new(open_channel.feerate_per_kw, channel_type);
    |                           -------------------                              ^^^^^^^^^^^^ expected `&Features`, found `&[u8]`
    |                           |
    |                           arguments to this function are incorrect
    |
    = note: expected reference `&features::Features`
               found reference `&[u8]`
note: associated function defined here
   --> smite/src/channel_tx/commitment.rs:472:12
    |
472 |     pub fn new(feerate_per_kw: u32, channel_type: &Features) -> CommitmentCost {
    |            ^^^                      -----------------------

error: could not compile `smite` (lib test) due to 1 previous error
```

</details>

I think 74ecd09 should be squashed into 43b1bf3.

Other than that, LGTM! Will ACK after rebase.

smite/src/bolt/features.rs
NishantBansal2003: > I think this is downstream of the inconsistency in how `FeatureBit` is treated as a function argument. Sometimes it's treated as a feature (`supports_feature`); sometimes as a bit (`set_bit`).

Hmm, that's true, and we can definitely fix this, But, I feel like it's a lot of updates to maintain three structs -- `FeatureBit`, `Feature`, and `Features` -- for a relatively small benefit, considering how we're going to operate on `Features`. But I can definitely add them and fix this ambiguity if others also think it's worth updating

CLN also closely follows the current model (https://github.com/ElementsProject/lightning/blob/master/common/features.h), with `OPTIONAL` and `REQUIRED` macros for updating the bit. I think we can keep the const as they are for now, and in the future, add `required` and `optional` func to get the required or optional bit from these defined const. WDYT?

ekzyis (approved):
ACK 40f7aa8

smite/src/bolt/features.rs
ekzyis: > But, I feel like it's a lot of updates to maintain three structs -- `FeatureBit`, `Feature`, and `Features` -- for a relatively small benefit, considering how we're going to operate on `Features`.

I agree. Considering the wider context as you just did, I also don't have a clear suggestion how to improve this, so I think what I mentioned is subjective, and we'd need more opinions to reach a conclusion if/how to improve this.

> I think we can keep the const as they are for now, and in the future, add required and optional func to get the required or optional bit from these defined const. WDYT?

Makes sense! I ACK'ed this PR now.

smite/src/bolt/features.rs
```diff
@@ -0,0 +1,323 @@
...
+impl PartialEq for Features {
+    /// Returns true if this set and other have the same feature bits set.
+    fn eq(&self, other: &Self) -> bool {
+        for bit in 0..(self.0.len().max(other.0.len()) * 8) {
+            if self.is_bit_set(bit) != other.is_bit_set(bit) {
+                return false;
+            }
+        }
+        true
+    }
```
morehouse, line 130: Nit: bit-by-bit comparison is inefficient, can compare bytes directly


```suggestion
    fn eq(&self, other: &Self) -> bool {
        fn trimmed(bytes: &[u8]) -> &[u8] {
            &bytes[bytes.iter().position(|b| *b != 0).unwrap_or(bytes.len())..]
        }
        trimmed(&self.0) == trimmed(&other.0)
    }
```

smite/src/bolt/features.rs
```diff
@@ -0,0 +1,323 @@
...
+        // Required (bit 22), optional (bit 23).
+        assert!(Features::from(vec![0x40, 0x00, 0x00]).supports_feature(Features::OPTION_ANCHORS));
+        assert!(Features::from(vec![0x80, 0x00, 0x00]).supports_feature(Features::OPTION_ANCHORS));
+        // No support.
+        assert!(!Features::from(vec![0x00, 0x00, 0x40]).supports_feature(Features::OPTION_ANCHORS));
+        assert!(!Features::from(vec![0x00, 0x00, 0x80]).supports_feature(Features::OPTION_ANCHORS));
+        assert!(!Features::from(vec![]).supports_feature(Features::OPTION_ANCHORS));
+        assert!(!Features::from(vec![0xff, 0xff]).supports_feature(Features::OPTION_ANCHORS));
+        assert!(!Features::from(vec![0x00, 0x10]).supports_feature(Features::OPTION_ANCHORS));
+    }
```
morehouse, line 259: Nit: test would be clearer if we replaced `OPTION_ANCHORS` with 22 everywhere.

smite/src/bolt/features.rs
```diff
@@ -0,0 +1,323 @@
...
+    #[test]
+    fn supports_feature_matches_either_bit_set() {
+        // Only the required (even) bit set.
+        let required_only = Features::from_bits(&[22]);
+        assert!(required_only.supports_feature(Features::OPTION_ANCHORS));
+        assert!(required_only.supports_feature(23));
+        // Only the optional (odd) bit set.
+        let optional_only = Features::from_bits(&[23]);
+        assert!(optional_only.supports_feature(Features::OPTION_ANCHORS));
+        assert!(optional_only.supports_feature(23));
```
morehouse, line 270: Nit: test would be clearer if we replaced `OPTION_ANCHORS` with 22 everywhere

smite/src/bolt/features.rs
```diff
@@ -0,0 +1,323 @@
...
+    fn clear_feature_clears_both_bits() {
+        // Passing the even bit clears the pair.
+        let mut fv = Features::from_bits(&[22, 23]);
+        fv.clear_feature(Features::OPTION_ANCHORS);
+        assert!(!fv.supports_feature(Features::OPTION_ANCHORS));
+        // Passing the odd bit clears the pair.
+        let mut fv = Features::from_bits(&[22, 23]);
+        fv.clear_feature(23);
+        assert!(!fv.supports_feature(Features::OPTION_ANCHORS));
+    }
```
morehouse, line 285: Nit: test would be clearer if we replaced `OPTION_ANCHORS` with 22 everywhere

smite/src/channel_tx/commitment.rs
```diff
@@ -721,7 +688,7 @@ mod tests {
                 vout: 0,
             },
             funding_satoshis: 10_000_000,
-            channel_type,
+            channel_type: Features::from(channel_type),
```
morehouse: Nit: should we just make the `channel_type` parameter for this function a `Features`?

smite/src/channel_tx/commitment.rs
```diff
@@ -1349,7 +1316,7 @@ mod tests {
                 vout: 0,
             },
             funding_satoshis,
-            channel_type,
+            channel_type: Features::from(channel_type),
```
morehouse: Nit: should we just make the `channel_type` parameter for this function a `Features`?

morehouse (approved):
This is great!  Could we add a tiny commit at the end to fix #217?

morehouse (approved):
LGTM

PR #193, opened 2026-08-08
smite-ir: back `channel_announcement` with a real funding output

devvaansh:
`ChannelAnnouncementGenerator` now creates, broadcasts, and confirms a 2-of-2 P2WSH funding transaction, and announces the `short_channel_id` it lands at along with the bitcoin keys its witness script commits to.  The messages can then pass the on-chain UTXO validation performed by CLN and LND.

#155 resolved the `short_channel_id` half of this, but nothing ever emitted the operation, and the generator picked its bitcoin keys with `generate_fresh`, unrelated to any funding output.  Deriving the funding pubkeys from the announced private keys closes the gap.  No new operations are needed, since `DerivePoint`, `CreateFundingTransaction`, `BroadcastTransaction`, `MineBlocks`, and `LookupShortChannelId` already compose.

`funding_satoshis` and `feerate_per_kw` are emitted directly instead of picked, since `generate_fresh(Amount)` draws from the full u64 range and exceeds `MAX_MONEY` essentially always, which fails coin selection and aborts the program before the announcement is sent.  Only the maximums are bounded: the funding amount is capped at BOLT 2's non-wumbo limit and the feerate is capped so fees don't exhaust the wallet.  Both may still go to zero, since `sign_and_broadcast_tx` already falls back to mining dust outputs and below-min-relay-feerate transactions directly.  `OperationParamMutator` can widen both afterwards.

`FundingFlowGenerator` picks those same two values with `pick_variable` and has the same full-range exposure, which seems worth a follow-up.

Tested with unit tests covering the ordering of the funding sequence, that each announced bitcoin key is the one behind the corresponding funding pubkey, and that the generated amounts stay in range.  Not yet fuzzed against the targets; I'll follow up with coverage once I've run a campaign.

Ref: #71

### DIFF
--- a/smite-ir/src/generators/channel_announcement.rs
+++ b/smite-ir/src/generators/channel_announcement.rs
@@ -1,20 +1,38 @@
 //! Generator for `channel_announcement` gossip message.
 
-use rand::Rng;
+use rand::{Rng, RngExt};
 
 use super::Generator;
 use crate::builder::ProgramBuilder;
 use crate::{Operation, VariableType};
 
-/// Generates an unsolicited `channel_announcement` send.
+/// Generates an unsolicited `channel_announcement` send backed by a real
+/// funding output.
+///
+/// Emits instructions to:
+/// 1. Create, broadcast, and confirm a 2-of-2 P2WSH funding transaction
+/// 2. Look up the `short_channel_id` of the confirmed funding output
+/// 3. Build and send `channel_announcement`
+///
+/// The announced bitcoin keys are the ones committed to by the funding
+/// output's witness script, so the message can pass the on-chain UTXO
+/// validation that lightning implementations perform.
 #[derive(Clone, Copy)]
 pub struct ChannelAnnouncementGenerator;
+
+impl ChannelAnnouncementGenerator {
+    /// BOLT 2 caps `funding_satoshis` at 2^24-1 for non-wumbo channels, and we
+    /// only have so much wallet balance to spend.
+    pub const MAX_FUNDING_SATOSHIS: u64 = 16_777_215;
+
+    /// Bounded so that fees don't exhaust the wallet.
+    pub const MAX_FEERATE_PER_KW: u32 = 10_000;
+}
 
 impl Generator for ChannelAnnouncementGenerator {
     fn generate(&self, builder: &mut ProgramBuilder, rng: &mut impl Rng) {
         let features = builder.pick_variable(VariableType::Features, rng);
         let chain_hash = builder.pick_variable(VariableType::ChainHash, rng);
-        let scid = builder.pick_variable(VariableType::ShortChannelId, rng);
 
         // Use fresh private keys since channel_announcement validation
         // generally requires distinct keys.
@@ -23,12 +41,47 @@ impl Generator for ChannelAnnouncementGenerator {
         let bitcoin_sk_1 = builder.generate_fresh(VariableType::PrivateKey, rng);
         let bitcoin_sk_2 = builder.generate_fresh(VariableType::PrivateKey, rng);
 
+        // The funding output must pay to the same bitcoin keys the message
+        // announces, so derive the funding pubkeys instead of picking them.
+        let funding_pubkey_1 = builder.append(Operation::DerivePoint, &[bitcoin_sk_1]);
+        let funding_pubkey_2 = builder.append(Operation::DerivePoint, &[bitcoin_sk_2]);
+
+        // Load the funding amount and feerate rather than picking them, since a
+        // full-range random amount exceeds Bitcoin's maximum supply and fails
+        // coin selection, aborting the program before the announcement is sent.
+        let funding_satoshis = builder.append(
+            Operation::LoadAmount(rng.random_range(0..=Self::MAX_FUNDING_SATOSHIS)),
+            &[],
+        );
+        let feerate_per_kw = builder.append(
+            Operation::LoadFeeratePerKw(rng.random_range(0..=Self::MAX_FEERATE_PER_KW)),
+            &[],
+        );
+
+        // Create the 2-of-2 P2WSH funding transaction and confirm it.
+        let funding_transaction = builder.append(
+            Operation::CreateFundingTransaction,
+            &[
+                funding_pubkey_1,
+                funding_pubkey_2,
+                funding_satoshis,
+                feerate_per_kw,
+            ],
+        );
+        builder.append(Operation::BroadcastTransaction, &[funding_transaction]);
+        builder.append(Operation::MineBlocks(rng.random_range(1..=16)), &[]);
+
+        // Derive the short_channel_id from the confirmed funding output.
+        let short_channel_id =
+            builder.append(Operation::LookupShortChannelId, &[funding_transaction]);
+
+        // Build and send channel_announcement.
         let msg = builder.append(
             Operation::BuildChannelAnnouncement,
             &[
                 features,
                 chain_hash,
-                scid,
+                short_channel_id,
                 node_sk_1,
                 node_sk_2,
                 bitcoin_sk_1,
--- a/smite-ir/src/generators/channel_announcement.rs
+++ b/smite-ir/src/generators/channel_announcement.rs
@@ -1,20 +1,38 @@
 //! Generator for `channel_announcement` gossip message.
 
-use rand::Rng;
+use rand::{Rng, RngExt};
 
 use super::Generator;
 use crate::builder::ProgramBuilder;
 use crate::{Operation, VariableType};
 
-/// Generates an unsolicited `channel_announcement` send.
+/// Generates an unsolicited `channel_announcement` send backed by a real
+/// funding output.
+///
+/// Emits instructions to:
+/// 1. Create, broadcast, and confirm a 2-of-2 P2WSH funding transaction
+/// 2. Look up the `short_channel_id` of the confirmed funding output
+/// 3. Build and send `channel_announcement`
+///
+/// The announced bitcoin keys are the ones committed to by the funding
+/// output's witness script, so the message can pass the on-chain UTXO
+/// validation performed by CLN and LND.
 #[derive(Clone, Copy)]
 pub struct ChannelAnnouncementGenerator;
+
+impl ChannelAnnouncementGenerator {
+    /// BOLT 2 caps `funding_satoshis` at 2^24-1 for non-wumbo channels, and we
+    /// only have so much wallet balance to spend.
+    pub const MAX_FUNDING_SATOSHIS: u64 = 16_777_215;
+
+    /// Bounded so that fees don't exhaust the wallet.
+    pub const MAX_FEERATE_PER_KW: u32 = 10_000;
+}
 
 impl Generator for ChannelAnnouncementGenerator {
     fn generate(&self, builder: &mut ProgramBuilder, rng: &mut impl Rng) {
         let features = builder.pick_variable(VariableType::Features, rng);
         let chain_hash = builder.pick_variable(VariableType::ChainHash, rng);
-        let scid = builder.pick_variable(VariableType::ShortChannelId, rng);
 
         // Use fresh private keys since channel_announcement validation
         // generally requires distinct keys.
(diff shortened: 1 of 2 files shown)


NishantBansal2003 (changes requested):
I think we should update the existing channel announcement generator to only create, broadcast, and mine the tx, and then use that SCID only. Otherwise, there isn't much difference between the two, and we can always test all the other cases with the current set of mutators. WDYT?

Also, does it make sense to include a full gossip generator that creates a fully funded channel announcement, followed by a node announcement and then a channel update for it?

devvaansh:
>I think we should update the existing channel announcement generator to only create, broadcast, and mine the tx, and >then use that SCID only. Otherwise, there isn't much difference between the two, and we can always test all the other >cases with the current set of mutators. WDYT?

i think Keeping both generators preserves cheap bogus-SCID testing without block-mining overhead.
since in a funded-only program, the `short_channel_id` is structurally immutable - `LookupShortChannelId` has no direct parameters, and `InputSwapMutator`/`InstructionDeleteMutator` bail when no other SCID producer exists in scope.
Relying solely on `GeneratorInsertionMutator` to introduce SCIDs drops arbitrary and boundary SCID exploration.


>Also, does it make sense to include a full gossip generator that creates a fully funded channel announcement, followed >by a node announcement and then a channel update for it?

Yes, worth doing -  `NodeAnnouncementGenerator` and `ChannelUpdateGenerator` pick their signing key via `pick_variable`, and after a channel announcement the most recent key in scope is `bitcoin_sk_2`, so they almost never sign with an actual node key and the messages don't link up.

morehouse:
Concept ACK.  I have no strong feelings on whether we keep the existing `channel_announcement` generator or adapt it into this one -- either approach has tradeoffs.  But if we do experiment with #199, that might benefit from having more generators to choose subsets from.

smite-ir/src/generators/funded_channel_announcement.rs
```diff
@@ -0,0 +1,102 @@
...
+/// ones committed to by the funding output's witness script, so the message can
+/// pass the on-chain UTXO validation performed by CLN and LND.
+#[derive(Clone, Copy)]
+pub struct FundedChannelAnnouncementGenerator;
+
+impl FundedChannelAnnouncementGenerator {
+    /// `funding_satoshis` bounds. The maximum is BOLT 2's limit for non-wumbo
+    /// channels; the minimum keeps change above the dust limit.
+    pub const MIN_FUNDING_SATOSHIS: u64 = 100_000;
+    pub const MAX_FUNDING_SATOSHIS: u64 = 16_777_215;
```
morehouse: I think it makes sense to have a max amount, since we have limited wallet funds to use.  But do we really need a min amount?

smite-ir/src/generators/funded_channel_announcement.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    /// `feerate_per_kw` bounds. The minimum is Bitcoin Core's default minimum
+    /// relay feerate; the maximum keeps fees from exhausting the wallet.
+    pub const MIN_FEERATE_PER_KW: u32 = 253;
+    pub const MAX_FEERATE_PER_KW: u32 = 10_000;
```
morehouse: Likewise here -- do we really need a min feerate?  We already have a mechanism for mining non-standard transactions.

morehouse (changes requested):
After thinking about it more, I think we should merge this generator into `ChannelAnnouncementGenerator` as @NishantBansal2003 suggested, since it's less code overall.  And in general we try to have the generators produce validish flows and then let mutators exercise the invalid flows.  The valid flows are harder to produce by mutators alone; the invalid flows are generally easy.

smite-ir/src/generators/funded_channel_announcement.rs
devvaansh: I'd added the minimum thinking a tiny funding amount would leave dust change or get rejected at broadcast and abort the program. Neither holds --  change is `utxo - funding - fee` so a smaller funding amount makes change bigger, and `sign_and_broadcast_tx` already mines below-dust outputs directly. Dropped it, range is now `0..=MAX_FUNDING_SATOSHIS`; same reasoning for the min feerate since "min relay fee not met" hits the same fallback.

devvaansh:
>After thinking about it more, I think we should merge this generator into ChannelAnnouncementGenerator as @NishantBansal2003 suggested, since it's less code overall. And in general we try to have the generators produce validish flows and then let mutators exercise the invalid flows. The valid flows are harder to produce by mutators alone; the invalid flows are generally easy.


Makes sense - merged it into `ChannelAnnouncementGenerator` and dropped the separate one. Ends up a fair bit smaller since `generators.rs` doesn't need to change at all. One knock-on: `instr_reorder_returns_false_on_single_or_no_act` relied on this generator producing a single Act, so I've pointed it at `NodeAnnouncementGenerator` instead.

smite-ir/src/generators/channel_announcement.rs
```diff
@@ -1,20 +1,38 @@
...
+/// funding output.
+///
+/// Emits instructions to:
+/// 1. Create, broadcast, and confirm a 2-of-2 P2WSH funding transaction
+/// 2. Look up the `short_channel_id` of the confirmed funding output
+/// 3. Build and send `channel_announcement`
+///
+/// The announced bitcoin keys are the ones committed to by the funding
+/// output's witness script, so the message can pass the on-chain UTXO
+/// validation performed by CLN and LND.
```
NishantBansal2003: All lightning implementations performs on-chain UTXO lookup, so we should reference those as well

smite-ir/src/tests.rs
```diff
@@ -1384,6 +1384,102 @@ fn generated_channel_announcement_program_structure() {
...
         "expected exactly one BuildChannelAnnouncement"
     );
+
+    // The funding transaction must be created, broadcast, and confirmed before
+    // its short_channel_id is looked up.
+    let position = |target: &Operation| {
+        ops.iter()
+            .position(|op| std::mem::discriminant(*op) == std::mem::discriminant(target))
+            .unwrap_or_else(|| panic!("expected a {target}"))
+    };
```
NishantBansal2003: This has been repeated many times, I think this could be made into a global helper and used everywhere

smite-ir/src/tests.rs
```diff
@@ -1384,6 +1384,102 @@ fn generated_channel_announcement_program_structure() {
...
+        let build_idx = find(&Operation::BuildChannelAnnouncement);
+
+        let create = &program.instructions[create_idx];
+        let lookup = &program.instructions[lookup_idx];
+        let build = &program.instructions[build_idx];
+
+        // The scid is looked up from the funding transaction just created, and
+        // that scid is what the announcement carries.
+        assert_eq!(lookup.inputs[0], create_idx, "seed {seed}: scid lookup");
+        assert_eq!(build.inputs[2], lookup_idx, "seed {seed}: announced scid");
```
NishantBansal2003: The error message here is not clear about what happened

smite-ir/src/tests.rs
```diff
@@ -1384,6 +1384,102 @@ fn generated_channel_announcement_program_structure() {
...
+
+        // Each announced bitcoin key is the private key behind the
+        // corresponding funding pubkey.
+        for (funding_pos, announced_pos) in [(0, 5), (1, 6)] {
+            let derive = &program.instructions[create.inputs[funding_pos]];
```
NishantBansal2003: nit: a comment would be helpful to explain what these indices refer to, for clarity

smite-ir/src/tests.rs
```diff
@@ -1384,6 +1384,102 @@ fn generated_channel_announcement_program_structure() {
...
+    );
+}
+
+// The bitcoin keys the message announces must be the ones committed to by
+// the funding output's witness script, and the announced scid must come from
+// that same funding transaction. Otherwise the announcement cannot pass
+// on-chain validation.
+#[test]
+fn generated_channel_announcement_commits_to_its_funding_output() {
+    for seed in 0..100 {
```
NishantBansal2003: It's really a generator structure, would there be any benefit to running it 100 times?

smite-ir/src/tests.rs
```diff
@@ -1384,6 +1384,102 @@ fn generated_channel_announcement_program_structure() {
...
+        let create_idx = program
+            .instructions
+            .iter()
+            .position(|i| matches!(i.operation, Operation::CreateFundingTransaction))
+            .expect("expected a CreateFundingTransaction");
+        let create = &program.instructions[create_idx];
+
+        match program.instructions[create.inputs[2]].operation {
+            Operation::LoadAmount(sats) => assert!(
+                (0..=ChannelAnnouncementGenerator::MAX_FUNDING_SATOSHIS).contains(&sats),
```
NishantBansal2003:
```suggestion
                sats <= ChannelAnnouncementGenerator::MAX_FUNDING_SATOSHIS,
```

smite-ir/src/tests.rs
```diff
@@ -1384,6 +1384,102 @@ fn generated_channel_announcement_program_structure() {
...
+                "seed {seed}: implausible funding amount {sats}",
+            ),
+            ref op => panic!("seed {seed}: expected LoadAmount, got {op}"),
+        }
+        match program.instructions[create.inputs[3]].operation {
+            Operation::LoadFeeratePerKw(feerate) => assert!(
+                (0..=ChannelAnnouncementGenerator::MAX_FEERATE_PER_KW).contains(&feerate),
```
NishantBansal2003:
```suggestion
                feerate <= ChannelAnnouncementGenerator::MAX_FEERATE_PER_KW,
```

smite-ir/src/tests.rs
```diff
@@ -1384,6 +1384,102 @@ fn generated_channel_announcement_program_structure() {
...
+    for seed in 0..100 {
+        let program = generate_channel_announcement_program(seed);
+        let create_idx = program
+            .instructions
+            .iter()
+            .position(|i| matches!(i.operation, Operation::CreateFundingTransaction))
+            .expect("expected a CreateFundingTransaction");
+        let create = &program.instructions[create_idx];
+
+        match program.instructions[create.inputs[2]].operation {
```
NishantBansal2003: nit: a comment would be helpful to explain what these indices refer to, for clarity

smite-ir/src/tests.rs
```diff
@@ -1384,6 +1384,102 @@ fn generated_channel_announcement_program_structure() {
...
+                "seed {seed}: bitcoin_key_{} must match the funding output",
+                funding_pos + 1,
+            );
+        }
+    }
+}
+
+// Coin selection fails outright for amounts above Bitcoin's maximum supply, so
+// the generator must emit plausible funding values rather than full-range
+// randoms. Without this the program aborts before sending the announcement.
```
NishantBansal2003:
```suggestion
// randoms. Without this the program exits before sending the announcement.
```

smite-ir/src/tests.rs
devvaansh: Right,  the generator has no `rng-dependent` branches, so every seed produces the same instructions/wiring; only the loaded values differ (already covered by the plausible-values test). Dropped to seed 0.

smite-ir/src/generators/channel_announcement.rs
```diff
@@ -23,12 +41,49 @@ impl Generator for ChannelAnnouncementGenerator {
...
+                funding_satoshis,
+                feerate_per_kw,
+            ],
+        );
+        builder.append(Operation::BroadcastTransaction, &[funding_transaction]);
+        builder.append(Operation::MineBlocks(rng.random_range(1..=16)), &[]);
+
+        // Derive the short_channel_id from the confirmed funding output. If a
+        // mutator drops the MineBlocks above, the executor yields a sentinel
+        // scid and the announcement simply fails on-chain validation.
```
morehouse: Nit: seems out-of-place to mention mutators here in the generator


```suggestion
        // Derive the short_channel_id from the confirmed funding output.
```

smite-ir/src/tests.rs
```diff
@@ -1309,18 +1323,17 @@ fn generated_funding_flow_program_structure() {
...
+    let send_funding_created = find_operation(&program, &Operation::SendFundingCreated);
+    let recv_funding_signed = find_operation(&program, &Operation::RecvFundingSigned);
+    let broadcast_transaction = find_operation(&program, &Operation::BroadcastTransaction);
+    let send_channel_ready = find_operation(
+        &program,
+        &Operation::SendChannelReady {
+            include_alias: false,
+        },
+    );
+    let recv_channel_ready = find_operation(&program, &Operation::RecvChannelReady);
```
morehouse: This is a refactoring unrelated to the new generator.  It should go in a commit by itself.

smite-ir/src/tests.rs
```diff
@@ -34,6 +34,20 @@ fn assert_well_formed(program: &Program) {
     }
 }
 
+/// Returns the index of the first instruction in `program` whose operation is
+/// the same variant as `target`, ignoring the operation's parameters.
+///
+/// # Panics
+///
+/// Panics if `program` contains no such instruction.
+fn find_operation(program: &Program, target: &Operation) -> usize {
```
morehouse: Nit: a macro may be nicer for this since then call-sites don't need to actually create the operation they're matching against

smite-ir/src/tests.rs
```diff
@@ -1384,6 +1397,99 @@ fn generated_channel_announcement_program_structure() {
...
+
+// The bitcoin keys the message announces must be the ones committed to by
+// the funding output's witness script, and the announced scid must come from
+// that same funding transaction. Otherwise the announcement cannot pass
+// on-chain validation.
+//
+// The generator never branches on the rng, so every seed produces the same
+// instructions wired the same way -- only the loaded values differ. One seed
+// is enough to check the wiring; the values are covered by
+// `generated_channel_announcement_uses_plausible_funding_values` below.
```
morehouse: Nit: this comment is unnecessary


```suggestion
```

smite-ir/src/tests.rs
```diff
@@ -1384,6 +1397,99 @@ fn generated_channel_announcement_program_structure() {
...
+            ),
+            ref op => panic!("seed {seed}: expected LoadAmount, got {op}"),
+        }
+        match program.instructions[create.inputs[3]].operation {
+            Operation::LoadFeeratePerKw(feerate) => assert!(
+                feerate <= ChannelAnnouncementGenerator::MAX_FEERATE_PER_KW,
+                "seed {seed}: implausible feerate {feerate}",
+            ),
+            ref op => panic!("seed {seed}: expected LoadFeeratePerKw, got {op}"),
+        }
```
morehouse: Nit: convention in the codebase is to avoid the `ref`


```suggestion
        match &program.instructions[create.inputs[2]].operation {
            Operation::LoadAmount(sats) => assert!(
                *sats <= ChannelAnnouncementGenerator::MAX_FUNDING_SATOSHIS,
                "seed {seed}: implausible funding amount {sats}",
            ),
            op => panic!("seed {seed}: expected LoadAmount, got {op}"),
        }
        match &program.instructions[create.inputs[3]].operation {
            Operation::LoadFeeratePerKw(feerate) => assert!(
                *feerate <= ChannelAnnouncementGenerator::MAX_FEERATE_PER_KW,
                "seed {seed}: implausible feerate {feerate}",
            ),
            op => panic!("seed {seed}: expected LoadFeeratePerKw, got {op}"),
        }
```

morehouse (approved):
LGTM

PR #194, opened 2026-08-09
smite-scenarios: log messages in name(type) format

ekzyis:
I found this quite useful. This changes the logs like this:

```diff
-SendMessage: Some(258), 138 bytes 
+SendMessage: channel_update(258), 138 bytes
```

and

```diff
-expected 33, got 17
+expected accept_channel(33), got error(17)
```

WDYT? Concept ACK?

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -9,8 +9,8 @@ use bitcoin::{OutPoint, ScriptBuf, Txid};
 use smite::bitcoin::{BitcoinCli, TxBlockPosition, Utxo};
 use smite::bolt::{
     AcceptChannel, AnnouncementSignatures, ChannelAnnouncement, ChannelId, ChannelReady,
-    ChannelReadyTlvs, ChannelUpdate, FundingCreated, FundingSigned, Message, NodeAnnouncement,
-    OpenChannel, OpenChannelTlvs, Pong, ShortChannelId, Shutdown, msg_type,
+    ChannelReadyTlvs, ChannelUpdate, FundingCreated, FundingSigned, Message, MessageType,
+    NodeAnnouncement, OpenChannel, OpenChannelTlvs, Pong, ShortChannelId, Shutdown,
 };
 use smite::channel_tx::{
     ChannelConfig, ChannelPartyConfig, ChannelState, FundingTransaction, HolderIdentity, Side,
@@ -203,8 +203,11 @@ pub enum ExecuteError {
     Decode(#[from] smite::bolt::BoltError),
 
     /// Received a different message type than expected.
-    #[error("unexpected message: expected type {expected}, got {got}")]
-    UnexpectedMessage { expected: u16, got: u16 },
+    #[error("unexpected message: expected {expected}, got {got}")]
+    UnexpectedMessage {
+        expected: MessageType,
+        got: MessageType,
+    },
 
     /// The target sent a BOLT `error`.
     #[error("peer error on {:?}: {}", .0.channel_id, .0.message().unwrap_or("<non-utf8>"))]
@@ -1245,7 +1248,7 @@ fn recv_accept_channel(conn: &mut impl Connection) -> Result<AcceptChannel, Exec
     match recv_non_ping(conn, RECV_IDLE_TIMEOUT)? {
         Message::AcceptChannel(ac) => Ok(ac),
         other => Err(ExecuteError::UnexpectedMessage {
-            expected: msg_type::ACCEPT_CHANNEL,
+            expected: MessageType::ACCEPT_CHANNEL,
             got: other.msg_type(),
         }),
     }
@@ -1256,7 +1259,7 @@ fn recv_funding_signed(conn: &mut impl Connection) -> Result<FundingSigned, Exec
     match recv_non_ping(conn, RECV_IDLE_TIMEOUT)? {
         Message::FundingSigned(fs) => Ok(fs),
         other => Err(ExecuteError::UnexpectedMessage {
-            expected: msg_type::FUNDING_SIGNED,
+            expected: MessageType::FUNDING_SIGNED,
             got: other.msg_type(),
         }),
     }
@@ -1280,7 +1283,7 @@ fn recv_channel_ready(
         Message::ChannelReady(cr) => cr,
         other => {
             return Err(ExecuteError::UnexpectedMessage {
-                expected: msg_type::CHANNEL_READY,
+                expected: MessageType::CHANNEL_READY,
                 got: other.msg_type(),
             });
         }
@@ -2448,8 +2451,8 @@ mod tests {
         assert!(matches!(
             err,
             ExecuteError::UnexpectedMessage {
-                expected: msg_type::ACCEPT_CHANNEL,
-                got: msg_type::INIT,
+                expected: MessageType::ACCEPT_CHANNEL,
+                got: MessageType::INIT,
             }
         ));
     }
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -406,10 +406,15 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                 // -- Act operations --
                 Operation::SendMessage => {
                     let bytes = resolve_message(&variables, instr.inputs[0]);
-                    let msg_type = bytes.get(..2).map(|b| u16::from_be_bytes([b[0], b[1]]));
+                    let ty = u16::from_be_bytes(
+                        *bytes
+                            .first_chunk::<2>()
+                            .expect("encoded message has a 2-byte type prefix"),
+                    );
                     log::debug!(
-                        "[{:?}] SendMessage: type {msg_type:?}, {} bytes",
+                        "[{:?}] SendMessage: {}, {} bytes",
                         start.elapsed(),
+                        MessageType::from_u16(ty),
                         bytes.len(),
                     );
                     self.conn.send_message(bytes)?;
@@ -1219,8 +1224,8 @@ fn recv_non_ping(conn: &mut impl Connection, timeout: Duration) -> Result<Messag
                 let pong = Message::Pong(Pong::respond_to(&ping)).encode();
                 conn.send_message(&pong)?;
             }
-            Message::Unknown { msg_type, .. } => {
-                log::debug!("skipping unknown message type {msg_type}");
+            Message::Unknown { .. } => {
+                log::debug!("skipping message {msg}");
             }
             // TODO: Gossip messages are not currently consumed by any scenario,
             // so skip them for now. Revisit this once we want to extract their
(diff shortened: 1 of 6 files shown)


smite-scenarios/src/scenarios/ir.rs
```diff
@@ -76,8 +77,10 @@ impl<T: Target, S: SnapshotSetup<T>> Scenario for IrScenario<T, S> {
                 // Error/Warning when rejecting our input). Usually normal
                 // protocol behavior.
                 log::debug!(
-                    "[{:?}] unexpected message: expected {expected}, got {got}",
+                    "[{:?}] unexpected message: expected {}({expected}), got {}({got})",
                     start.elapsed(),
+                    msg_type::name(expected),
+                    msg_type::name(got),
                 );
```
erickcestari: nit:

```suggestion
Err(e @ ExecuteError::UnexpectedMessage { .. }) => {
                // Target replied with an unexpected message type (often
                // Error/Warning when rejecting our input). Usually normal
                // protocol behavior.
                log::debug!("[{:?}] {e}", start.elapsed());
            }
```

smite-scenarios/src/executor.rs
```diff
@@ -2026,7 +2040,12 @@ mod tests {
...
         let na = match Message::decode(&executor.conn.sent[0]).expect("valid message") {
             Message::NodeAnnouncement(na) => na,
-            other => panic!("expected NodeAnnouncement, got type {}", other.msg_type()),
+            other => panic!(
+                "expected {}({}), got {}",
+                msg_type::name(msg_type::NODE_ANNOUNCEMENT),
+                msg_type::NODE_ANNOUNCEMENT,
+                other
+            ),
         };
```
erickcestari: We could have a type for this log format so that we don't have to go to each place to change the format later.

example:
```rust
/// A message type number that displays as `name(type)`, e.g.
    /// `open_channel(32)`.
    ///
    /// Use this wherever a message type is written into a log line or an error
    /// so the format stays in one place.
    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
    pub struct Named(pub u16);

    impl std::fmt::Display for Named {
        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
            write!(f, "{}({})", name(self.0), self.0)
        }
    }
```
then this would become:

```suggestion
            other => panic!(
                "expected {}, got {other}",
                msg_type::Named(msg_type::NODE_ANNOUNCEMENT)
            ),
};
```

erickcestari (changes requested):
I like the idea to make the logs more self-describing

Concept ACK 667aae873ec56d13208f065237cefe877b623e41

morehouse:
Concept ACK.  I think a much simpler way is to implement the `Display` trait for `MessageType`.

smite-scenarios/src/executor.rs
morehouse: Yes, I was thinking the same thing, though I would prefer to call the struct `MessageType` instead of `Named`.

smite-scenarios/src/scenarios/ir.rs
ekzyis: done in a56bfc7

smite-scenarios/src/executor.rs
ekzyis: done in a56bfc7, named it `MessageType`

ekzyis:
addressed feedback in a56bfc7:

* new `MessageType` struct with `Display` implementation
* use existing `Display` implementation for `UnexpectedMessage` from `#[error(message)]`

ekzyis:
a56bfc7..33a0f32:

* renamed `type_` to `ty` because I learned that `ty` is the Rust convention when your name conflicts with the reserved keyword `type`

smite-scenarios/src/executor.rs
```diff
@@ -1800,16 +1818,19 @@ mod tests {
...
     }
 
     fn decode_open_channel(bytes: &[u8]) -> OpenChannel {
         match Message::decode(bytes).expect("valid message") {
             Message::OpenChannel(oc) => oc,
-            other => panic!("expected OpenChannel, got type {}", other.msg_type()),
+            other => panic!(
+                "expected {}, got {other}",
+                msg_type::MessageType(msg_type::OPEN_CHANNEL)
+            ),
```
morehouse: Nit: we could simplify all these panics

```suggestion
            other => panic!("expected open_channel, got {other}"),
```

smite/src/bolt.rs
```diff
@@ -1139,176 +1209,204 @@ mod tests {
...
+                    payload: vec![],
+                },
+                "unknown",
+                99,
+            ),
+        ];
+        for (msg, name, ty) in cases {
+            assert_eq!(msg.msg_type(), ty, "wrong type number for {name}");
+            assert_eq!(msg.msg_name(), name, "wrong type name for {ty}");
+        }
```
morehouse, line 1419:
```suggestion
        for (msg, name, ty) in cases {
            assert_eq!(msg.msg_type(), ty, "wrong type number for {name}");
            assert_eq!(msg.msg_name(), name, "wrong type name for {ty}");
            assert_eq!(msg.to_string(), format!("{name}({ty})"));
            assert_eq!(msg_type::MessageType(ty).to_string(), format!("{name}({ty})"));
        }
```

smite/src/bolt.rs
```diff
@@ -287,6 +343,13 @@ pub enum Message {
     },
 }
 
+impl std::fmt::Display for Message {
+    /// Formats as `name(type)`, e.g. `open_channel(32)`.
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}({})", self.msg_name(), self.msg_type())
```
morehouse:
```suggestion
        write!(f, "{}", msg_type::MessageType(self.msg_type()))
```

smite/src/bolt.rs
```diff
@@ -287,6 +343,13 @@ pub enum Message {
     },
 }
 
+impl std::fmt::Display for Message {
```
morehouse, line 354: Nit: let's move this to just after the `impl Message` block below.

smite/src/bolt.rs
```diff
@@ -203,6 +203,62 @@ pub mod msg_type {
...
+        }
+    }
+
+    /// A message type number that displays as `name(type)`, e.g.
+    /// `open_channel(32)`.
+    ///
+    /// Use this wherever a message type is written into a log line or an error
+    /// so the format stays in one place.
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub struct MessageType(pub u16);
```
morehouse: Now we're using `MessageType` in some places and `u16` in others.  I think it would be a bit cleaner if we refactored to use `MessageType` everywhere.  For example, the `msg_type::OPEN_CHANNEL` and other constants could all be `MessageType`s, `UnexexpectedMessage` errors can contain `MessageType`s, and `Message::msg_type` can return a `MessageType`.

morehouse:
We should also be able to get rid of the `msg_type` module and move the constants inside an `impl MessageType` block.

erickcestari (commented):
Probably a follow-up PR, but another thing that might be useful for debugging is to log the `warning` and `error` `data` field as ASCII.

smite-scenarios/src/executor.rs
ekzyis: Done in b038b4e. To compare the same format, I changed it to `expected open_channel(32), got {other}`.

smite/src/bolt.rs
ekzyis: done in b038b4e

smite/src/bolt.rs
ekzyis: done in b038b4e

smite/src/bolt.rs
ekzyis: done in b038b4e

smite/src/bolt.rs
ekzyis: Yes, makes sense! Done in 03f13e8, but I moved this PR back into draft, since I couldn't review my own changes yet.

erickcestari (approved):
LGTM!

smite-scenarios/src/scenarios/init.rs
```diff
@@ -59,7 +59,7 @@ impl<T: Target> Scenario for InitScenario<T> {
         );
 
         // Send an init-typed message with fuzz payload.
-        let msg = bolt::message_with_type(bolt::msg_type::INIT, input);
+        let msg = bolt::message_with_type(bolt::MessageType::INIT.as_u16(), input);
```
morehouse: Nit: `message_with_type` can take a `MessageType` parameter

smite/src/bolt.rs
```diff
@@ -287,53 +351,66 @@ pub enum Message {
...
+            Self::UpdateFailHtlc(_) => MessageType::UPDATE_FAIL_HTLC,
+            Self::CommitmentSigned(_) => MessageType::COMMITMENT_SIGNED,
+            Self::RevokeAndAck(_) => MessageType::REVOKE_AND_ACK,
+            Self::UpdateFailMalformedHtlc(_) => MessageType::UPDATE_FAIL_MALFORMED_HTLC,
+            Self::ChannelAnnouncement(_) => MessageType::CHANNEL_ANNOUNCEMENT,
+            Self::NodeAnnouncement(_) => MessageType::NODE_ANNOUNCEMENT,
+            Self::ChannelUpdate(_) => MessageType::CHANNEL_UPDATE,
+            Self::AnnouncementSignatures(_) => MessageType::ANNOUNCEMENT_SIGNATURES,
+            Self::GossipTimestampFilter(_) => MessageType::GOSSIP_TIMESTAMP_FILTER,
+            Self::Unknown { msg_type, .. } => MessageType(*msg_type),
```
morehouse: Perhaps `Unknown.msg_type` should be `MessageType` now?

smite/src/bolt.rs
```diff
@@ -287,53 +351,66 @@ pub enum Message {
...
         }
     }
 
+    /// Returns the message type name (e.g. `open_channel`), or `unknown` for an
+    /// unrecognized type.
+    #[must_use]
+    pub fn msg_name(&self) -> &'static str {
+        self.msg_type().name()
+    }
```
morehouse: This looks unused.  Let's remove it.

smite/src/bolt.rs
```diff
@@ -384,54 +461,62 @@ impl Message {
...
-            msg_type::PING => Ok(Self::Ping(Ping::decode(cursor)?)),
-            msg_type::PONG => Ok(Self::Pong(Pong::decode(cursor)?)),
-            msg_type::OPEN_CHANNEL => Ok(Self::OpenChannel(OpenChannel::decode(cursor)?)),
-            msg_type::ACCEPT_CHANNEL => Ok(Self::AcceptChannel(AcceptChannel::decode(cursor)?)),
-            msg_type::FUNDING_CREATED => Ok(Self::FundingCreated(FundingCreated::decode(cursor)?)),
-            msg_type::FUNDING_SIGNED => Ok(Self::FundingSigned(FundingSigned::decode(cursor)?)),
-            msg_type::CHANNEL_READY => Ok(Self::ChannelReady(ChannelReady::decode(cursor)?)),
-            msg_type::SHUTDOWN => Ok(Self::Shutdown(Shutdown::decode(cursor)?)),
-            msg_type::CLOSING_COMPLETE => {
+        match MessageType(msg_type) {
```
morehouse: Nit


```suggestion
        match MessageType::from_u16(msg_type) {
```

morehouse (approved):

morehouse (approved):
LGTM

Issue #195, opened 2026-08-10
CLN OOM


morehouse:
#169 decreased the default Nyx memory limit from 4 GB to 2 GB, and I've started to run into that limit when fuzzing CLN.

```
==567==ERROR: AddressSanitizer: out of memory: allocator is trying to allocate 0x200028 bytes
```

```
==821==ERROR: AddressSanitizer: out of memory: failed to allocate 0xb09000 (11571200) bytes of FakeStack (error code: 12)
```

The main reason we have this issue with CLN and not the other targets is that ASan causes ~2x memory usage.

Obviously we could just bump the memory limit back up to 4 GB, but I wanted to do some investigation first to see if we're doing things that waste significant memory.  Turns out we are.

### Unused files in Nyx image

Nyx uses tmpfs, which means the entire filesystem is kept in memory.  Any unused files consume memory unnecessarily.  In particular, the `container.tar` archive that contains the Nyx chroot is very large and unused after it is extracted.  Deleting the archive after extraction frees up ~300 MB of memory for CLN (and similar amounts for other targets).

### `syncblocks` overhead

After deleting `container.tar`, CLN ran much better.  But I still ran into OOMs for certain programs.  It turns out all of those programs mined large numbers of blocks (more than 40), and memory spiked by up to 500 MB during mining.

The reason for the memory spike lies in #156, which solved the slow chain sync problem by doing `lightning-cli syncblocks` on every new block mined.  Each ASan-instrumented `lightning-cli` process uses up to 20 MB of memory, and when many blocks are mined at once, there ends up being many concurrent `lightning-cli syncblocks` processes running at once.  The peak I observed was ~25 concurrent `lightning-cli` processes, which explains the 500 MB spike.

Ideas to eliminate this waste:

1. Manually call `lightning-cli syncblocks` *once* after each `MineBlocks` operation, rather than using `-blocknotify` which calls it for *every* block.  This would reduce the number of forked processes by ~8x.
2. Stop using `lightning-cli` to trigger `syncblocks`.  We can do the `syncblocks` RPC with our own minimal code and negligible memory usage.  If we also do (1), this code can even run within the executor (no forked processes at all).
3. Use `flock` to serialize or coalesce concurrent `lightning-cli`s so that only one ever runs at a time.  I'd be concerned about missing blocks though if we choose to coalesce .
4. Exclude `lightning-cli` from ASan instrumentation.  Saves ~50% of the memory spike.

[...]

PR #196, opened 2026-08-10
scripts: stop the guest tmpfs from starving the target

erickcestari:
The Nyx guest has no disk: /tmp is a tmpfs that must hold the container tarball, the tree extracted from it, and every byte the target writes at runtime. Its default cap is half the VM's RAM, so unpacking a 600 MB image needs a 3 GB VM; below that, tar fails with ENOSPC and the agent aborts with an empty message that says nothing about what went wrong.

Partially address #195


morehouse (approved):
LGTM

PR #198, opened 2026-08-10
smite-ir: split determinism out of `has_side_effects`

devvaansh:
Follow-up to #155, where we agreed `LookupShortChannelId` doesn't really have side effects and that CSE wants a separate question answered.

`has_side_effects` was being asked three different questions by three callers: DCE wants to know whether an instruction can be dropped, CSE wants to know whether two instructions can be merged, and `InstructionReorderMutator` wants to know whether an instruction's position matters.  Those coincide for every operation we had until `LookupShortChannelId`, which is read-only but reads chain state, so it was marked as side-effecting to keep CSE away from it.

Add `depends_only_on_inputs`, and `is_pure` on top of it, so each caller asks what it actually means.  DCE keeps using `has_side_effects`, while CSE and the reorder mutator switch to `is_pure`.  `LookupShortChannelId` becomes side-effect free, and it joins `CreateFundingTransaction`, `SendFundingCreated`, and the `Recv` operations in reading state beyond its inputs.

The only behavior change is that DCE can now drop a `LookupShortChannelId` whose result nothing consumes.  I checked every operation, and it is the only one where `!is_pure()` differs from `has_side_effects()`, so CSE's skip set and the reorder mutator's candidate set are unchanged.

Keeping the lookup reorderable matters: moving it across a `MineBlocks` is what turns a confirmed `short_channel_id` into the unconfirmed sentinel, and it's the only way to reach the sentinel in a program whose scid comes from a lookup.  Marking it pure would have silently removed that, so there's a test covering it along with tests for the CSE and DCE behavior.

Ref: #155

### DIFF
--- a/smite-ir/src/operation.rs
+++ b/smite-ir/src/operation.rs
@@ -233,6 +233,7 @@ pub enum Operation {
     /// Sign wallet inputs of the transaction and broadcast it via `bitcoin-cli`.
     /// Input: `FundingTransaction`.
     BroadcastTransaction,
+    // -- Query: read state from outside the program --
     /// Look up the confirmed block position of a broadcast funding transaction
     /// and produce the corresponding BOLT 7 `short_channel_id`.
     ///
@@ -688,7 +689,6 @@ impl fmt::Display for Operation {
             Self::LoadChannelType(v) => write!(f, "LoadChannelType({v})"),
             Self::LoadTargetPubkeyFromContext => write!(f, "LoadTargetPubkeyFromContext()"),
             Self::LoadChainHashFromContext => write!(f, "LoadChainHashFromContext()"),
-            Self::MineBlocks(v) => write!(f, "MineBlocks({v})"),
             // Operations with inputs: parens added by Program::Display.
             Self::DerivePoint => write!(f, "DerivePoint"),
             Self::ExtractAcceptChannel(field) => write!(f, "Extract{field}"),
@@ -713,6 +713,7 @@ impl fmt::Display for Operation {
             Self::RecvAcceptChannel => write!(f, "RecvAcceptChannel"),
             Self::RecvFundingSigned => write!(f, "RecvFundingSigned"),
             Self::RecvChannelReady => write!(f, "RecvChannelReady()"),
+            Self::MineBlocks(v) => write!(f, "MineBlocks({v})"),
             Self::BroadcastTransaction => write!(f, "BroadcastTransaction"),
             Self::LookupShortChannelId => write!(f, "LookupShortChannelId"),
         }
@@ -792,28 +793,6 @@ impl Operation {
                 VariableType::Amount,       // funding_satoshis
                 VariableType::FeeratePerKw, // feerate_per_kw
             ],
-            Self::SendMessage => vec![VariableType::Message],
-            Self::SendOpenChannel => vec![VariableType::OpenChannelMessage],
-            Self::SendFundingCreated => vec![
-                VariableType::FundingTransaction, // funding_transaction
-                VariableType::PrivateKey,         // opener_funding_privkey
-                VariableType::ChannelId,          // temporary_channel_id
-            ],
-            Self::SendChannelReady { .. } => vec![
-                VariableType::ChannelId,      // channel_id
-                VariableType::Point,          // second_per_commitment_point
-                VariableType::ShortChannelId, // short_channel_id (alias)
-            ],
-            Self::SendShutdown => vec![
-                VariableType::ChannelId, // channel_id
-                VariableType::Bytes,     // scriptpubkey
-            ],
-            Self::RecvAcceptChannel => vec![VariableType::SentOpenChannel],
-            Self::RecvFundingSigned => vec![VariableType::SentFundingCreated],
-            Self::BroadcastTransaction | Self::LookupShortChannelId => {
-                vec![VariableType::FundingTransaction]
-            }
-
             Self::BuildOpenChannel => vec![
                 VariableType::ChainHash,    // chain_hash
                 VariableType::ChannelId,    // temporary_channel_id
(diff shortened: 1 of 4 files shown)


smite-ir/src/operation.rs
```diff
@@ -946,9 +946,9 @@ impl Operation {
             | Self::RecvChannelReady
             | Self::MineBlocks(_)
             | Self::CreateFundingTransaction
-            | Self::BroadcastTransaction
-            | Self::LookupShortChannelId => true,
-            Self::LoadAmount(_)
+            | Self::BroadcastTransaction => true,
+            Self::LookupShortChannelId
```
morehouse: Let's keep the match arms in the same order as the operator definitions.  Applies throughout this file.

BTW, the `LookupShortChannelId` definition is currently under the "Act" header comment -- we should move it to its own new section.

smite-ir/src/operation.rs
```diff
@@ -974,6 +974,64 @@ impl Operation {
         }
     }
 
+    /// Returns `true` if this operation always produces the same result given
+    /// the same inputs. A nondeterministic operation cannot be deduplicated by
+    /// CSE even when it is free of side effects, and its position in the
+    /// program is meaningful.
+    #[must_use]
+    pub fn is_deterministic(&self) -> bool {
```
morehouse: I think deterministic actually isn't the right word.  What we're checking is that the operation depends only on its inputs (not some external state).

Maybe `depends_only_on_inputs` would be better.

smite-ir/src/operation.rs
```diff
@@ -974,6 +974,64 @@ impl Operation {
...
+            // contents change as transactions are created and broadcast.
+            // The `Recv` operations read whatever the target sends us.
+            Self::LookupShortChannelId
+            | Self::CreateFundingTransaction
+            | Self::RecvAcceptChannel
+            | Self::RecvFundingSigned
+            | Self::RecvChannelReady => false,
+            Self::SendMessage
+            | Self::SendOpenChannel
+            | Self::SendFundingCreated
```
morehouse: `SendFundingCreated` actually depends on executor state, so it should be in the `false` category.

smite-ir/src/operation.rs
devvaansh: Done -- moved `LookupShortChannelId` under a new // -- Query: read state from outside the program -- section, and fixed the ordering throughout the file.

smite-ir/src/operation.rs
devvaansh: Agreed, renamed to `depends_only_on_inputs`

erickcestari (approved):
LGTM!

smite-ir/src/operation.rs
```diff
@@ -968,12 +1003,39 @@ impl Operation {
...
+            | Self::BuildChannelAnnouncement
             | Self::BuildNodeAnnouncement { .. }
             | Self::BuildChannelUpdate
-            | Self::BuildAnnouncementSignatures => false,
+            | Self::BuildAnnouncementSignatures
+            | Self::SendMessage
+            | Self::SendOpenChannel
+            | Self::SendChannelReady { .. }
+            | Self::SendShutdown
+            | Self::MineBlocks(_)
```
morehouse: `MineBlocks` now can mine from a private mempool, so it should map to false.

smite-ir/src/operation.rs
```diff
@@ -968,12 +1003,39 @@ impl Operation {
...
+            | Self::BroadcastTransaction => true,
```
morehouse: `BroadcastTransaction`'s now dedups against the private mempool, so it should map to false.

smite-ir/src/tests.rs
```diff
@@ -2409,6 +2421,49 @@ fn instr_reorder_preserves_well_formedness() {
     assert_mutator_preserves_well_formedness(&InstructionReorderMutator, &original);
 }
 
+// `LookupShortChannelId` is read-only but nondeterministic, so it stays an
+// `Act` and the mutator may still move it across a `MineBlocks`. That swap is
+// how a confirmed scid becomes the unconfirmed sentinel.
```
morehouse: Nit: It's no longer in the Act category -- it's in the Query category.  We've also stopped using the nondeterminism framing.


```suggestion
// `LookupShortChannelId` is read-only but depends on chain state beyond
// its inputs, so it's still useful for mutators to be able to move it.
```

smite-ir/src/tests.rs
```diff
@@ -2736,6 +2791,80 @@ fn dead_code_idempotent() {
     assert_eq!(once, twice, "elimination is idempotent");
 }
 
+// `LookupShortChannelId` is read-only, so a lookup whose result nothing
+// consumes can be dropped.
+#[test]
+fn dce_drops_unused_lookup_short_channel_id() {
```
morehouse: Nit: naming convention in this file is:


```suggestion
fn dead_code_drops_unused_lookup_short_channel_id() {
```

smite-ir/src/tests.rs
```diff
@@ -2736,6 +2791,80 @@ fn dead_code_idempotent() {
...
+            .instructions
+            .iter()
+            .any(|i| matches!(i.operation, Operation::LookupShortChannelId)),
+        "DCE must drop a LookupShortChannelId whose result is unused",
+    );
+    assert_well_formed(&program);
+}
+
+// A lookup whose result is consumed must survive, now that it no longer
+// relies on being marked as side-effecting to stay in the program.
```
morehouse: Nit:

```suggestion
// A lookup whose result is consumed must not be removed.
```

smite-ir/src/tests.rs
```diff
@@ -2736,6 +2791,80 @@ fn dead_code_idempotent() {
...
+        operation: Operation::LoadChannelId([1u8; 32]),
+        inputs: vec![],
+    });
+    let channel_id = instrs.len() - 1;
+    instrs.push(Instruction {
+        operation: Operation::SendChannelReady {
+            include_alias: false,
+        },
+        inputs: vec![channel_id, point, scid],
+    });
```
morehouse, line 2850: Nit: DCE doesn't check it, but it's confusing to rely on the alias being used with `include_alias: false`.


```suggestion
    instrs.push(Instruction {
        operation: Operation::SendChannelReady {
            include_alias: true,
        },
        inputs: vec![channel_id, point, scid],
    });
```

smite-ir/src/operation.rs
```diff
@@ -968,12 +1003,40 @@ impl Operation {
...
+            // chain state. `MineBlocks` also mines whatever the private
+            // mempool holds, and `BroadcastTransaction` dedups against it.
+            Self::CreateFundingTransaction
+            | Self::SendFundingCreated
+            | Self::RecvAcceptChannel
+            | Self::RecvFundingSigned
+            | Self::RecvChannelReady
+            | Self::LookupShortChannelId
+            | Self::MineBlocks(_)
+            | Self::BroadcastTransaction => false,
```
morehouse: Nit:  this is not enum order any more

morehouse (approved):

PR #200, opened 2026-08-10
fix: use tmpfs for Docker execution

erickcestari:
Docker containers perform disk I/O by default, unlike the NYX VM, which runs from initramfs and is significantly faster. Use a tmpfs for /tmp in local execution, coverage reporting, and crash symbolization to better match the NYX environment.

Also document the tmpfs option and add the ir scenario to the script usage information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,10 @@ When AFL++ finds a crash:
 cp /tmp/smite-out/default/crashes/<crashing-input> ./crash
 
 # Reproduce in local mode (use the matching image and scenario binary)
-docker run --rm -v $PWD/crash:/input.bin -e SMITE_INPUT=/input.bin smite-$TARGET-$SCENARIO /$TARGET-scenario
+docker run --rm --tmpfs /tmp:rw,exec,size=1g -v $PWD/crash:/input.bin -e SMITE_INPUT=/input.bin smite-$TARGET-$SCENARIO /$TARGET-scenario
 ```
+
+The `--tmpfs /tmp:rw,exec,size=1g` option provides a temporary in-memory filesystem for `/tmp`, matching the `tmpfs` environment available inside the Nyx VM. This halves startup time.
 
 ### Coverage Report Mode
 
--- a/README.md
+++ b/README.md
@@ -98,8 +98,10 @@ When AFL++ finds a crash:
 cp /tmp/smite-out/default/crashes/<crashing-input> ./crash
 
 # Reproduce in local mode (use the matching image and scenario binary)
-docker run --rm -v $PWD/crash:/input.bin -e SMITE_INPUT=/input.bin smite-$TARGET-$SCENARIO /$TARGET-scenario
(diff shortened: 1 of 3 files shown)


scripts/coverage-report.sh
```diff
@@ -230,6 +231,7 @@ echo "Merging coverage data and generating report..."
 case "$TARGET" in
     lnd)
         docker run --rm "${DOCKER_USER[@]}" \
+            --tmpfs /tmp:rw,exec,size=1g \
```
ekzyis: just curious: could there be a reason to not want to use `--tmpfs` here?

ekzyis (approved):
utACK 171064c just one question, but can be ignored

scripts/coverage-report.sh
erickcestari: I only see reasons to have, but I could be wrong. On my tests it was able to run 2 times faster than without it.

NishantBansal2003 (approved):
LGTM!

Verified by generating coverage reports across all four implementations, and this change does make it considerably faster than before

scripts/coverage-report.sh
```diff
@@ -135,6 +135,7 @@ run_input() {
         mkdir "$covdir"
 
         docker run --rm "${DOCKER_USER[@]}" \
+            --tmpfs /tmp:rw,exec,size=1g \
```
morehouse: Nit: we should define a single envvar that contains the flag we use everywhere in this script, like we do for `DOCKER_USER`.  That definition should have a comment explaining the purpose of the flag.

Also I tried removing `exec` and it seemed to work for every target except Eclair since the JVM puts some executable stuff in `/tmp`.  It would be good to add a comment explaining why we need `exec`.

README.md
```diff
@@ -98,9 +98,11 @@ When AFL++ finds a crash:
 cp /tmp/smite-out/default/crashes/<crashing-input> ./crash
 
 # Reproduce in local mode (use the matching image and scenario binary)
-docker run --rm -v $PWD/crash:/input.bin -e SMITE_INPUT=/input.bin smite-$TARGET-$SCENARIO /$TARGET-scenario
+docker run --rm --tmpfs /tmp:rw,exec,size=1g -v $PWD/crash:/input.bin -e SMITE_INPUT=/input.bin smite-$TARGET-$SCENARIO /$TARGET-scenario
 ```
 
+The `--tmpfs /tmp:rw,exec,size=1g` option provides a temporary in-memory filesystem for `/tmp`, matching the `tmpfs` environment available inside the Nyx VM. This helps keep the local execution environment consistent with the one used during fuzzing.
```
morehouse: Nit: second sentence repeats info from the first, and doesn't fully explain *why* a user would want the flag


```suggestion
The `--tmpfs /tmp:rw,exec,size=1g` option provides a temporary in-memory filesystem for `/tmp`, matching the `tmpfs` environment available inside the Nyx VM. This halves startup time.
```

morehouse (approved):

PR #202, opened 2026-08-11
smite-scenarios: add support for suppressing known violations

NishantBansal2003:
Targets can now declare known violations that should be
suppressed until they are fixed upstream. IR scenario will
check if a violation is in the known list and skip the failure
if so, logging it as debug instead. This provides the foundation
for handling multiple oracles as they're added in the future,
establishing the standard process for suppressing known violations
until upstream fixes land.

Currently known violations that are suppressed are:

- CLN currently allows `funding_satoshis` values exceeding
total Bitcoin supply during `open_channel` and `accept_channel`
negotiation. These values should be rejected early as they cannot
result in a valid channel. This is tracked upstream and will be
suppressed until fixed.

- LND allows `open_channel` with an omitted `channel_type`,
violating BOLT 2 even though it signals the required
`option_channel_type` feature bit. This will be tracked
upstream and will be suppressed until fixed.

- CLN currently allows `dust_limit_satoshis` values below 354 during
`open_channel` negotiation, violating BOLT 2 which requires rejection
of such values. This will be tracked upstream and will be suppressed
until fixed.

### DIFF
--- a/smite-scenarios/src/scenarios/ir.rs
+++ b/smite-scenarios/src/scenarios/ir.rs
@@ -106,8 +106,16 @@ impl<T: Target, S: SnapshotSetup<T>> Scenario for IrScenario<T, S> {
                 log::debug!("[{:?}] invalid commitment: {e}", start.elapsed());
             }
             Err(ExecuteError::Violation(v)) => {
-                // The target broke a protocol invariant.
-                return ScenarioResult::Fail(v.to_string());
+                let violation = v.to_string();
+                if T::is_violation_known(&violation) {
+                    log::debug!(
+                        "[{:?}] suppressed known violation: {violation}",
+                        start.elapsed(),
+                    );
+                } else {
+                    // The target broke a protocol invariant.
+                    return ScenarioResult::Fail(violation);
+                }
             }
         }
 
--- a/smite-scenarios/src/targets.rs
+++ b/smite-scenarios/src/targets.rs
@@ -77,4 +77,39 @@ pub trait Target: Sized {
     ///
     /// Returns [`TargetError::Crashed`] if the target has crashed.
     fn check_alive(&mut self) -> Result<(), TargetError>;
+
+    /// Known protocol violations for this target that should be suppressed.
+    ///
+    /// Each entry is a sequence of patterns that must appear in the violation
+    /// message in the specified order, without overlapping. Include stable
+    /// context around variable parts to avoid suppressing unrelated violations.
+    ///
+    /// Every entry must describe the behavior justifying it, and be removed
+    /// once the upstream fix lands.
+    #[must_use]
+    fn known_violations() -> &'static [&'static [&'static str]] {
+        &[]
+    }
+
+    /// Check if a violation matches a known pattern for this target.
+    #[must_use]
+    fn is_violation_known(violation: &str) -> bool {
+        Self::known_violations()
+            .iter()
+            .any(|pattern| Self::contains_in_order(violation, pattern))
+    }
+
+    /// Returns true if every fragment in a pattern appears in the violation
+    /// message in the specified order, without overlapping.
+    #[must_use]
+    fn contains_in_order(violation: &str, pattern: &[&str]) -> bool {
+        let mut pos = 0;
+        for fragment in pattern {
+            match violation[pos..].find(fragment) {
+                Some(offset) => pos += offset + fragment.len(),
+                None => return false,
+            }
+        }
+        true
+    }
 }
--- a/smite-scenarios/src/targets/cln.rs
+++ b/smite-scenarios/src/targets/cln.rs
@@ -248,4 +248,17 @@ impl Target for ClnTarget {
         }
         Ok(())
     }
+
+    /// Known violations suppressed until fixed upstream:
+    ///
+    /// - CLN currently allows `funding_satoshis` values above the total Bitcoin
+    ///   supply during `open_channel` and `accept_channel` negotiation. These
+    ///   values should be rejected early, as they cannot result in a valid
+    ///   channel. See: <https://github.com/ElementsProject/lightning/pull/9368>
+    fn known_violations() -> &'static [&'static [&'static str]] {
+        &[&[
+            "accepted invalid open_channel: funding_satoshis",
+            "exceeds maximum funding of 2100000000000000 sat",
+        ]]
+    }
 }
--- a/smite-scenarios/src/targets/cln.rs
+++ b/smite-scenarios/src/targets/cln.rs
@@ -255,10 +255,21 @@ impl Target for ClnTarget {
     ///   supply during `open_channel` and `accept_channel` negotiation. These
     ///   values should be rejected early, as they cannot result in a valid
     ///   channel. See: <https://github.com/ElementsProject/lightning/pull/9368>
+    ///
+    /// - CLN currently allows `dust_limit_satoshis` values below 354 during
+    ///   `open_channel` negotiation. These values should be rejected early, as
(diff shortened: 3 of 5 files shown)


morehouse:
Here's another one we'll need an exception for: https://github.com/ElementsProject/lightning/issues/9403

erickcestari (approved):
LGTM! In the future, we could have the option to disable Oracle entirely if it starts crashing frequently. That way, we can continue fuzzing without needing to change the code.

smite-scenarios/src/targets/lnd.rs
```diff
@@ -316,4 +316,13 @@ impl Target for LndTarget {
         }
         Ok(())
     }
+
+    /// Known violations reported as failures:
+    ///
+    /// - `open_channel does not include a channel_type`: LND allows `open_channel`
+    ///   with an omitted `channel_type`, violating BOLT 2 even though it signals
+    ///   the required `option_channel_type` feature.
```
morehouse: It would be good to have issue links for each suppression, so the future reader can easily check whether this has been fixed upstream.

smite-scenarios/src/targets/cln.rs
```diff
@@ -248,4 +248,21 @@ impl Target for ClnTarget {
...
+    ///   `open_channel` and `accept_channel` negotiation. These values should
+    ///   be rejected early, as they cannot result in a valid channel.
+    ///
+    /// - `is below the minimum of 354 sat`: CLN currently allows `dust_limit_satoshis`
+    ///   values below 354 during `open_channel` negotiation. These values
+    ///   should be rejected early, as required by BOLT 2.
+    fn known_violations() -> &'static [&'static str] {
+        &[
+            "exceeds maximum funding of 2100000000000000 sat",
+            "is below the minimum of 354 sat",
```
morehouse: This suppresses two cases:

1. CLN accepted our undersized dust limit.  (This is the one we've observed).
2. CLN chose an undersized dust limit for itself.

The second case has not been observed and is a serious vulnerability if we ever do see it.  So we should make sure we don't suppress it.

smite-scenarios/src/targets/cln.rs
```diff
@@ -248,4 +248,21 @@ impl Target for ClnTarget {
         }
         Ok(())
     }
+
+    /// Known violations reported as failures:
```
morehouse:
```suggestion
    /// Known violations suppressed until fixed upstream:
```

smite-scenarios/src/targets/lnd.rs
```diff
@@ -316,4 +316,13 @@ impl Target for LndTarget {
         }
         Ok(())
     }
+
+    /// Known violations reported as failures:
```
morehouse:
```suggestion
    /// Known violations suppressed until reported upstream:
```

smite-scenarios/src/scenarios/ir.rs
```diff
@@ -106,8 +106,17 @@ impl<T: Target, S: SnapshotSetup<T>> Scenario for IrScenario<T, S> {
                 log::debug!("[{:?}] invalid commitment: {e}", start.elapsed());
             }
             Err(ExecuteError::Violation(v)) => {
-                // The target broke a protocol invariant.
-                return ScenarioResult::Fail(v.to_string());
+                let violation = v.to_string();
+                if T::is_violation_known(&violation) {
+                    log::debug!(
+                        "[{:?}] Suppressed known violation: {}",
```
morehouse: Nit

```suggestion
                        "[{:?}] suppressed known violation: {violation}",
```

morehouse (changes requested):

morehouse:
Also two more exceptions:

- https://github.com/ACINQ/eclair/issues/3349
- https://github.com/ACINQ/eclair/issues/3350

smite-scenarios/src/targets/cln.rs
NishantBansal2003: Hmm, it's bad. I added a regex dependency for this so we can always match the complete pattern instead of a suffix or prefix, which might be the same for some symmetrical cases and could cause us to miss it again in future cases

smite-scenarios/src/targets.rs
```diff
@@ -77,4 +77,27 @@ pub trait Target: Sized {
...
+    fn known_violations() -> &'static [&'static str] {
+        &[]
+    }
+
+    /// Check if a violation matches a known pattern for this target.
+    #[must_use]
+    fn is_violation_known(violation: &str) -> bool {
+        Self::known_violations().iter().any(|pattern| {
+            regex::Regex::new(pattern)
+                .ok()
```
morehouse: A bad regex is silently skipped here.


```suggestion
            regex::Regex::new(pattern)
                .expect("valid regex")
```

morehouse (approved):

smite-scenarios/Cargo.toml
```diff
@@ -26,3 +26,4 @@ libc = "0.2"
 serde.workspace = true
 serde_json.workspace = true
 tempfile = "3"
+regex = "1"
```
morehouse: Ah sorry, I just checked into this dependency and its huge.  The scenario binary sizes approximately doubled with this.

At the very least we should slim it down to just what we use:

```suggestion
regex = { version = "1", default-features = false, features = ["std"] }
```

Even better would be to solve this without the dependency at all.  Something like this:

```rust
fn known_violations() -> &'static [&'static [&'static str]] { &[] }

fn is_violation_known(violation: &str) -> bool {
  Self::known_violations().iter().any(|pattern| contains_in_order(violation, pattern))
}

// Returns true if every fragment of pattern appears in violation, in order
// and without overlapping.
fn contains_in_order(violation: &str, pattern: &[&str]) -> bool { ... }
```

Then we would have patterns like:

```rust
&[
  "accepted invalid open_channel: funding_satoshis ",
  " exceeds maximum funding of",
],
```

morehouse (changes requested):

smite-scenarios/Cargo.toml
NishantBansal2003: Yup, I wanted to do that in the first place, but then I thought we could simplify the logic with something built-in. But I think if the regex dependency is huge, we should drop it

Done in: https://github.com/lnfuzz/smite/pull/202/commits/9170600fb3e0c235533345a2bf321805df56568a

smite-scenarios/src/targets.rs
```diff
@@ -77,4 +77,39 @@ pub trait Target: Sized {
...
+    fn is_violation_known(violation: &str) -> bool {
+        Self::known_violations()
+            .iter()
+            .any(|pattern| Self::contains_in_order(violation, pattern))
+    }
+
+    /// Returns true if every fragment in a pattern appears in the violation
+    /// message in the specified order, without overlapping.
+    #[must_use]
+    fn contains_in_order(violation: &str, pattern: &[&str]) -> bool {
```
morehouse, line 105: Nit: doesn't belong in the trait, could be a free func

smite-scenarios/src/targets.rs
```diff
@@ -77,4 +77,39 @@ pub trait Target: Sized {
...
+        let mut pos = 0;
+        for fragment in pattern {
+            match violation[pos..].find(fragment) {
+                Some(offset) => pos += offset + fragment.len(),
+                None => return false,
+            }
+        }
+        true
```
morehouse, line 113: The only tricky thing is that if we ever accidentally leave an empty pattern `&[]`, then everything gets suppressed.  Maybe we should assert pattern isn't empty?

morehouse (approved):

Issue #205, opened 2026-08-12
Add configurable AFL++ timeout to orchestration script


erickcestari:
I also noticed that the orchestration script doesn't have an option to configure the `-t` value and instead uses AFL++'s default. Running without an increased timeout can cause the fuzzer to time out on inputs, especially the longer ones that are capable of opening channels.

```
  -t msec       - timeout for each run (auto-scaled, default 1000 ms). Add a '+'
                  to auto-calculate the timeout, the value being the maximum.
```

_Originally posted by @erickcestari in https://github.com/lnfuzz/smite/issues/135#issuecomment-5270988578_

morehouse:
Added by https://github.com/lnfuzz/smite/pull/206.

PR #206, opened 2026-08-12
scripts: Add explicit timeouts and record commit metadata in reports

Chand-ra:
This PR introduces two improvements to the Smite coverage-based effectiveness evaluation pipeline:

- Explicit fuzzer timeouts: Add explicit execution and hang timeout parameters to the fuzzer invocation in `smite-orchestrator.py`. This prevents false-positive hangs and timeouts for IR programs.

- Evaluation data lineage: Record the latest Git commit hash and date for each configuration worktree during campaign startup, displaying this metadata at the top of the generated evaluation reports. This ensures strict data provenance for a rigorous evaluation.

### DIFF
--- a/scripts/smite-orchestrator.py
+++ b/scripts/smite-orchestrator.py
@@ -874,6 +874,31 @@ def ensure_seed_dir(args, console: Console):
     )
 
 
+def save_commit_metadata(out_dir: Path, smite_dirs: dict, console: Console):
+    """Extracts the latest git commit hash and date, saving them to the output directory."""
+    for label, smite_dir in smite_dirs.items():
+        config_out = out_dir / label
+        config_out.mkdir(parents=True, exist_ok=True)
+
+        try:
+            # %h = abbreviated hash, %cd = commit date
+            res = subprocess.run(
+                ["git", "log", "-1", "--format=%h (%cd)", "--date=short"],
+                cwd=smite_dir,
+                capture_output=True,
+                text=True,
+                check=True,
+            )
+            commit_info = res.stdout.strip()
+        except Exception:
+            commit_info = "Unknown commit"
+            console.print(
+                f"[yellow]Warning: Could not get git info for '{label}' in {smite_dir}[/]"
+            )
+
+        (config_out / "latest_commit.txt").write_text(commit_info + "\n")
+
+
 def parse_args():
     """Parse CLI args and resolve all filesystem paths to absolute up front."""
     p = argparse.ArgumentParser(
@@ -930,6 +955,8 @@ def main():
             smite_dirs[l] = Path(d.strip()).expanduser().resolve()
     except ValueError:
         sys.exit("ERROR: --configs must use 'label:smite_dir' format")
+
+    save_commit_metadata(args.out_dir, smite_dirs, console)
 
     EnvironmentManager.validate(args.afl_dir, smite_dirs, console)
     EnvironmentManager.validate_paths(args.afl_dir, smite_dirs, console)
--- a/scripts/smite-orchestrator.py
+++ b/scripts/smite-orchestrator.py
@@ -47,6 +47,8 @@ Usage:
     --afl-dir AFL_DIR \
     [--trials N | --trial-ids ID[,ID...]] \
     [--timeout SECONDS] \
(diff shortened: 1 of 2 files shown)


scripts/smite-orchestrator.py
```diff
@@ -189,6 +189,8 @@ def build_afl_cmd(self) -> list[str]:
         """The exact afl-fuzz invocation for a standalone runner."""
         # Fixed power schedule for every standalone trial.
         POWER_SCHEDULE = "explore"
+        # Fixed timeout for a single input.
+        EXEC_TIMEOUT = 2000  # 2 seconds
```
erickcestari: This value could be configured by a flag and also have a bigger default as `5000`.

Chand-ra:
Addresses #205.

erickcestari (changes requested):
Nice, really quick PR, and the commit hash is a great addition.

This should fix the issue I've recently opened #205.

scripts/smite-orchestrator.py
Chand-ra: The default value was decided upon after an offline discussion with @morehouse.

However, if you've actually encountered a non-hanging/non-timeout input take 5 seconds to execute on your machine, I agree that it is a good reason to increase the current value.

erickcestari: I didn't find any hang with 5000 so far. But I agree that 2 seconds should already be enough.

morehouse: I agree it would be good to make this a flag.

scripts/smite-orchestrator.py
```diff
@@ -215,6 +219,7 @@ def build_afl_env(self) -> dict:
                 "AFL_NO_UI": "1",
                 "AFL_NO_COLOR": "1",
                 "AFL_FORKSRV_INIT_TMOUT": "1800000",
+                "AFL_HANG_TIMEOUT": "4000",  # 4 seconds
```
morehouse: Wrong env var name.  We need to use `AFL_HANG_TMOUT`.

Would also be good to make this a flag.

scripts/smite-orchestrator.py
```diff
@@ -865,6 +870,31 @@ def ensure_seed_dir(args, console: Console):
...
+                check=True,
+            )
+            commit_info = res.stdout.strip()
+        except Exception:
+            commit_info = "Unknown commit"
+            console.print(
+                f"[yellow]Warning: Could not get git info for '{label}' in {smite_dir}[/]"
+            )
+
+        (config_out / "latest_commit.txt").write_text(commit_info + "\n")
```
morehouse, line 899: This breaks the evaluation script for me.

```
[*] Detected Targets: ['cln', 'latest_commit.txt', 'lnd']
    - baseline/cln: found 3 valid trials
NotADirectoryError: [Errno 20] Not a directory: 'camp/baseline/latest_commit.txt'
```

morehouse (changes requested):
How was this tested?  It broke for me on the first try.

Chand-ra:
Rebased on top of the latest `master` and verified with a 10 minutes fuzzing campaign on LDK. Here's the abbreviated evaluation report:
```
# Fuzzing Evaluation Report

**Configuration A (Baseline):** `control` *(Commit: adcba47 (2026-08-12))*
**Configuration B (Experimental):** `test` *(Commit: adcba47 (2026-08-12))*
...
```

scripts/smite-orchestrator.py
```diff
@@ -215,6 +221,7 @@ def build_afl_env(self) -> dict:
                 "AFL_NO_UI": "1",
                 "AFL_NO_COLOR": "1",
                 "AFL_FORKSRV_INIT_TMOUT": "1800000",
+                "AFL_HANG_TIMEOUT": str(self.hang_timeout),
```
morehouse: This is still the wrong env var.

morehouse (changes requested):

scripts/smite-orchestrator.py
Chand-ra: Stupid mistake. Fixed now.

morehouse (approved):
LGTM

PR #207, opened 2026-08-13
smitebot: implement `reproduce` command

Ashish-Kumar-Dash:
## Description

  ### What

Adds `smitebot reproduce <campaign-id> -i <input> [--attach]`: replays a single input against a campaign's target by running its Docker image once, the same way `scripts/coverage-report.sh` feeds inputs to `LocalRunner`. The target's output streams to the terminal; the command interprets nothing as the logs are the deliverable.

  ```bash
  smitebot reproduce <campaign-id> -i path/to/crash
  smitebot reproduce <campaign-id> -i path/to/crash --attach   # interactive TTY
```
  ### Mechanism

  Resolves image + target from the campaign's state.json (no live campaign required), then runs:

``` bash
  docker run --rm [-it] -v <input-parent>:/corpus:ro \
  -e SMITE_INPUT=/corpus/<basename> <image> /<target>-scenario
```
  - The input's parent directory is bind-mounted read-only and the basename is passed via SMITE_INPUT, so colons in AFL crash names (id:000000,sig:06,...) never land in the colon-delimited -v spec. This mirrors [coverage-report.sh](https://github.com/lnfuzz/smite/blob/master/scripts/coverage-report.sh#L137).
  - --attach adds -it to watch the run live.
  - execute returns `true` once the container runs to completion regardless of its exit code; false only on operational failure (unknown campaign, missing input, missing image, docker spawn error). Errors name the offending path/value and the fix, fore.g. a missing image points at smitebot build .  

 ### Scope for v1

  Local-Docker reproduction only. Deliberately deferred for later PR:
  - Nyx-mode reproduction - a later phase.
  
Ref. #70


smitebot/README.md
```diff
@@ -244,3 +244,18 @@ smitebot corpus minimize <campaign-id> [-i <dir>]... [-o <output-dir>] [--aflpp-
 - `-i, --input <dir>`: One input directory to minimize. If multiple `-i` flags are present, all specified directories are merged before minimizing. If omitted, the campaign's runner queues are merged and minimized.
 - `-o, --output <output-dir>`: output directory; defaults to `~/.smitebot/runs/<id>/corpus-min/`
 - `--aflpp-path <path>`: AFL++ source tree, overriding the `aflpp_path` stored in `state.json` (useful when the checkout has moved)
+
+### smitebot reproduce
+
+Replays a single input against a campaign's target by running its Docker image once, with the input fed to the target the same way the fuzzer's `LocalRunner` does. The target's output is streamed to the terminal; `reproduce` interprets nothing — read the logs to see what happened. Resolves the image and target from the campaign's `state.json`; no live campaign required.
```
NishantBansal2003: nit: could remove the implementation details here, just like the semantics followed above

smitebot/src/commands/reproduce.rs
```diff
@@ -0,0 +1,207 @@
...
+            Err(e) => {
+                log::error!("{e}");
+                log::error!(
+                    "campaign '{}' not found; list campaigns with: ls {}",
+                    args.campaign_id,
+                    runs_dir.display()
+                );
+                return false;
+            }
+        };
```
NishantBansal2003, line 56: This part of the code is repeated across many commands, I think we could refactor it into a helper in a follow-up PR

smitebot/src/commands/reproduce.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        if !docker_image_exists(&state.image) {
+            log::error!(
+                "Docker image '{}' not found; build it with: smitebot build --target {} --scenario {}",
+                state.image,
+                state.target,
+                state.scenario
+            );
+            return false;
+        }
```
NishantBansal2003: I'm not sure about others, but I often run a campaign with a specific docker image and then continue development alongside it, creating new docker images with the same name. So, if this command runs a docker image referenced by the same name, I might get false results without realizing it

I think we should compare the found docker image's digest with the one stored in `image_digest` in `state.json` and log a warning if they don't match, saying that the results might differ from what was observed during the campaign. We could also log the smite git hash used for the campaign, so the user can rebuild the image from that commit to reproduce the same results.

smitebot/src/commands/reproduce.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        let run_args = docker_run_args(&state.image, state.target, parent, &basename, args.attach);
+        log::info!(
```
NishantBansal2003, line 105: This log should come before `let run_args...`

smitebot/src/commands/reproduce.rs
```diff
@@ -0,0 +1,207 @@
...
+        // docker run [flags] <image> <cmd> — image must come after all flags.
+        let parent = PathBuf::from("/tmp");
+        let args = docker_run_args("myimage", Target::Ldk, &parent, "in", false);
+        let image_pos = args.iter().position(|a| a == "myimage").unwrap();
+        let v_pos = args.iter().position(|a| a == "-v").unwrap();
+        let e_pos = args.iter().position(|a| a == "-e").unwrap();
+        assert!(v_pos < image_pos, "-v must appear before image");
+        assert!(e_pos < image_pos, "-e must appear before image");
+        assert_eq!(args.last().unwrap(), "/ldk-scenario");
+        assert!(args.contains(&"--rm".to_string()));
```
NishantBansal2003: nit: We could define flags upfront, like `["--rm", "-v", "-e"]`, and then iterate over them to get their positions in `args` and assert their positions relative to `image_pos`, this would also allow us to add the attach flag and test both cases

smitebot/src/commands/reproduce.rs
```diff
@@ -0,0 +1,207 @@
...
+#[derive(Debug, Args)]
+pub struct ReproduceArgs {
+    /// Campaign ID whose Docker image and target to reproduce against.
+    campaign_id: String,
+    /// Input file (e.g. a crash from a runner's `crashes/`) to replay.
+    #[arg(short = 'i', long)]
+    input: PathBuf,
+    /// Allocate an interactive TTY so the run can be watched live.
+    #[arg(long)]
+    attach: bool,
```
NishantBansal2003: I don't think we currently have any observable behavior with the `-it` option? It might be useful in the future, but I don't see any use for it currently. If we do need it later, I think that change should be paired with the change that introduces the relevant use case

NishantBansal2003 (changes requested):

smitebot/src/commands/reproduce.rs
```diff
@@ -0,0 +1,210 @@
...
+pub struct ReproduceArgs {
+    /// Campaign ID whose Docker image and target to reproduce against.
+    campaign_id: String,
+    /// Input file (e.g. a crash from a runner's `crashes/`) to replay.
+    #[arg(short = 'i', long)]
+    input: PathBuf,
+}
+
+impl ReproduceCommand {
+    /// Replays `input` against the campaign's Docker target.
```
NishantBansal2003:
```suggestion
    /// Replays `input` against the campaign's target in Docker.
```

smitebot/src/commands/reproduce.rs
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+}
+
+/// Returns the local Docker image's ID (content digest), or `None` if the image
+/// is not present.
+///
+/// Mirrors `start`'s `docker_image_id` (same `--format={{.Id}}`), so the value
+/// compares directly against the `image_digest` stored in `state.json`; both
+/// should fold into a shared helper (see the state-loading refactor follow-up).
```
NishantBansal2003: We should reuse it now instead of defining it again and removing it later

smitebot/src/commands/reproduce.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+/// Builds the `docker run` argument list for a single reproduction.
+///
+/// Mirrors `scripts/coverage-report.sh`: mount the input's parent directory
+/// read-only at `/corpus` and point `SMITE_INPUT` at `/corpus/<basename>`; the
+/// per-target entrypoint is `/<target>-scenario`.
+fn docker_run_args(image: &str, target: Target, parent: &Path, basename: &str) -> Vec<String> {
+    // ponytail: paths under ~/.smitebot are UTF-8 in practice; `display()`
+    // lossiness would only bite a non-UTF-8 corpus path, a non-goal for v1.
```
NishantBansal2003: This comment is not very clear to me, what does `ponytail` refer to?

Ashish-Kumar-Dash: Apologies, that was leftover jargon from a local tool and shouldn't have shipped.

smitebot/src/commands/reproduce.rs
```diff
@@ -0,0 +1,192 @@
...
+            );
+            return false;
+        };
+        // A campaign's image can be rebuilt under the same tag during development,
+        // so a matching name is not a matching image. Warn on a digest mismatch and
+        // point at the campaign's smite commit for an exact rebuild.
+        if image_id != state.image_digest {
+            log::warn!(
+                "Docker image '{}' has digest {} but the campaign ran with {}; results may differ. \
+                 Rebuild from smite commit {} to reproduce the original run.",
```
NishantBansal2003: nit: Even if we rebuild the image from the same git hash, the docker image digest may vary, so we could falsely warn again, even if the user follows the same steps we logged here. So, I'm not entirely sure what the better approach is, but it's probably fine to have a false warning log?

NishantBansal2003 (approved):

smitebot/README.md
```diff
@@ -244,3 +244,16 @@ smitebot corpus minimize <campaign-id> [-i <dir>]... [-o <output-dir>] [--aflpp-
 - `-i, --input <dir>`: One input directory to minimize. If multiple `-i` flags are present, all specified directories are merged before minimizing. If omitted, the campaign's runner queues are merged and minimized.
 - `-o, --output <output-dir>`: output directory; defaults to `~/.smitebot/runs/<id>/corpus-min/`
 - `--aflpp-path <path>`: AFL++ source tree, overriding the `aflpp_path` stored in `state.json` (useful when the checkout has moved)
+
+### smitebot reproduce
+
+Replays a single input against a campaign's target and streams the target's output to the terminal; `reproduce` interprets nothing — read the logs to see what happened. Resolves the image and target from the campaign's `state.json`; no live campaign required.
```
morehouse: Seems wordy and scatter-brained.

```suggestion
Replays a single input against a campaign's target and scenario, logging output to the terminal.
```

smitebot/src/commands/reproduce.rs
```diff
@@ -0,0 +1,190 @@
+//! Crash reproduction: replay a single input against the campaign's target in
+//! local Docker.
+//!
+//! Runs the campaign's Docker image once with `SMITE_INPUT` pointed at the given
+//! input (the same `LocalRunner` path `scripts/coverage-report.sh` uses), streams
+//! the target's output to the terminal, and interprets nothing — the logs are the
+//! deliverable. Nyx-mode reproduction is a later phase.
```
morehouse: Nit: I don't think it's helpful here to describe things this *doesn't* do


```suggestion
//! Runs the campaign's Docker image once with `SMITE_INPUT` pointed at the given
//! input (the same `LocalRunner` path `scripts/coverage-report.sh` uses), streaming
//! the target's output to the terminal.
```

smitebot/src/commands/start.rs
```diff
@@ -592,15 +592,6 @@ fn build_ir_mutator(smite_dir: &Path) -> bool {
 }
 
 /// Runs a command and returns its trimmed stdout on success.
```
morehouse:
```suggestion
```

smitebot/src/commands/reproduce.rs
```diff
@@ -0,0 +1,190 @@
...
+            log::error!("input is not a regular file: {}", input.display());
+            return false;
+        }
+        // canonicalize yields an absolute path ending in the file's name.
+        let parent = input.parent().expect("canonical input path has a parent");
+        let basename = input
+            .file_name()
+            .expect("canonical input path has a file name")
+            .to_string_lossy();
+
```
morehouse, line 83: Nit: probably better to return an error if the filename is non-utf8

morehouse (approved):
LGTM

PR #208, opened 2026-08-15
bitcoind: check createwallet status

ekzyis:
When I use `SMITE_DATA_DIR` to mount the bitcoin datadir, the wallet can already exist, and then `createwallet` fails.

With this PR, the error message will mention `failed to create wallet` instead of `failed to generate initial blocks`.

### DIFF
--- a/smite-scenarios/src/targets/bitcoind.rs
+++ b/smite-scenarios/src/targets/bitcoind.rs
@@ -133,13 +133,20 @@ pub fn start(
 /// Creates wallet and generates initial blocks.
 fn setup_wallet(cli: &BitcoinCli) -> Result<(), TargetError> {
     // Create wallet
-    let _ = cli
+    let status = cli
         .run()
         .arg("createwallet")
         .arg("default")
         .stdout(Stdio::null())
         .stderr(Stdio::null())
-        .status();
+        .status()?;
+
+    // command fails if wallet already exists (i.e. SMITE_DATA_DIR was mounted)
+    if !status.success() {
+        return Err(TargetError::StartFailed(
+            "failed to create wallet (does it already exist?)".into(),
+        ));
+    }
 
     // Generate initial blocks
     let status = cli


smite-scenarios/src/targets/bitcoind.rs
```diff
@@ -133,13 +133,17 @@ pub fn start(
...
+    let status = cli
         .run()
         .arg("createwallet")
         .arg("default")
         .stdout(Stdio::null())
         .stderr(Stdio::null())
-        .status();
+        .status()?;
+
+    if !status.success() {
```
Chand-ra, line 145: Maybe we should match on the return code of the underlying command? The CLI can fail for reasons other than a failed `createwallet` command, and we might want to distinguish between the two.

smite-scenarios/src/targets/bitcoind.rs
```diff
@@ -133,13 +133,17 @@ pub fn start(
...
+        return Err(TargetError::StartFailed("failed to create wallet".into()));
```
NishantBansal2003: I think this error case only happens when we reused the `bitcoind` directory, where a wallet already exists. So I think it would be better to add a comment explaining the reason for this error

smite-scenarios/src/targets/bitcoind.rs
ekzyis: >  The CLI can fail for reasons other than a failed `createwallet` command, and we might want to distinguish between the two.

You are right, but so far, this has reliably worked when there wasn't already a wallet, so I think we're safe to assume in the error message that it failed because the wallet already exists (`createwallet` RPC failed) and don't need error handling based on exit codes here. Seems like @NishantBansal2003 has a similar point of view in https://github.com/lnfuzz/smite/pull/208#discussion_r3816485000.

smite-scenarios/src/targets/bitcoind.rs
ekzyis: done in 4cd8583

NishantBansal2003 (approved):
LGTM!

morehouse (approved):
LGTM

PR #209, opened 2026-08-19
smite: validate negotiated features and channel types in `accept_channel` oracle

NishantBansal2003:
Depends-on: https://github.com/lnfuzz/smite/pull/192
Depends-on: https://github.com/lnfuzz/smite/pull/186 (added some changes to it, which will be removed once that PR is finalized)
Opening this as a draft since it depends on two PRs. Otherwise, the changes are ready for review.

I tried to add all the definitive checks for `open_channel` and `accept_channel` mentioned in the BOLT 2, including ones that aren't currently possible, such as `option_dual_fund`, since we haven't negotiated it. I think having parity with the BOLTs is good, especially for the future when we extend smite to support more protocols

Some more missing oracles that I will add in follow-up PR are:

#### `open_channel`

- [ ] `chain_hash != regtest`
- [ ] `channel_reserve_satoshis >= funding_satoshis`
- [ ] `dust_limit_satoshis > 10_000` sat
- [ ] `to_self_delay > 2016`
- [ ] `feerate_per_kw < 253` for non 0FC channels

#### `accept_channel`

- [ ]  `dust_limit_satoshis > 10_000` sat
- [ ] `htlc_minimum_msat > max_htlc_value_in_flight_msat`
- [ ] `htlc_minimum_msat > open_channel.funding_satoshis`
- [ ] `to_self_delay < 144`
- [ ] `max_accepted_htlcs == 0`
- [ ] pubkey match any pubkeys from `open_channel` and `accept_channel`
- [ ] `per_commitment_point` should not be reused

Known violations found by this that need to be suppressed are:
- https://github.com/ElementsProject/lightning/issues/9444
- https://github.com/ACINQ/eclair/issues/3374


smite/src/bolt/features.rs
```diff
@@ -120,6 +126,18 @@ impl Features {
         self.clear_bit(bit);
         self.clear_bit(bit ^ 1);
     }
+
+    /// Returns whether every bit set here is supported by `other`, where the
+    /// feature's required (even) or optional (odd) bit both count as support.
+    #[must_use]
+    pub fn is_supported(&self, other: &Features) -> bool {
```
ekzyis: nit: Maybe `is_compatible` would be a better name?

smite/src/oracles/accept_channel.rs
```diff
@@ -269,11 +351,31 @@ fn verify_initial_commitment(
     Ok(())
 }
 
+/// Returns the maximum funding amount allowed by the negotiated features.
+pub fn max_funding_satoshis(negotiated_features: &Features) -> u64 {
```
ekzyis:
```suggestion
fn max_funding_satoshis(negotiated_features: &Features) -> u64 {
```

smite/src/oracles/accept_channel.rs
```diff
@@ -269,11 +351,31 @@ fn verify_initial_commitment(
...
+    if negotiated_features.supports_feature(Features::OPTION_SUPPORT_LARGE_CHANNEL) {
+        Amount::MAX_MONEY.to_sat()
+    } else {
+        MAX_FUNDING_SATOSHIS_NO_WUMBO
+    }
+}
+
+/// Returns the maximum number of inbound HTLCs allowed by the channel type.
+pub fn max_accepted_htlcs_limit(channel_type: &Features) -> u16 {
```
ekzyis:
```suggestion
fn max_accepted_htlcs_limit(channel_type: &Features) -> u16 {
```

smite-scenarios/src/scenarios/setup.rs
```diff
@@ -86,7 +86,11 @@ impl<T: Target> SnapshotSetup<T> for PostInitSetup {
             // this is the floor. Dynamic per-target queries can replace it
             // later.
             block_height: u32::try_from(INITIAL_BLOCKS).expect("fits in u32"),
-            target_features: target_init.features,
+            // Since we echo the same features the target sent, but strip both
+            // required and optional bits to exercise only the single funded
+            // flow and avoid unrelated noise, negotiated features are just the
+            // features we sent in our init.
+            negotiated_features: Features::from(our_init.features),
```
ekzyis, line 93: Shouldn't this be a bitwise-or of `our_init.globalfeatures` and `our_init.features`?

> - MUST combine (bitwise OR) the two feature bitmaps into one logical `features` map.

https://github.com/lightning/bolts/blob/152897261850d93c4f4597f39cf22d7d22d6ede6/01-messaging.md?plain=1#L318

NishantBansal2003: Spec-wise, yes, but I think `globalfeatures` is a legacy feature. From what I can see, the spec doesn't require the two vectors to be disjoint, and in practice, every target duplicates rather than partitions them: LDK sends a masked copy of bits 0-13 from its own features, CLN and LND send only `var_onion`/`static_remotekey`, which they also set in features, and Eclair sends it empty. `globalfeatures` is a subset of `features` for all four, so there shouldn't be any issues if we just use `features` instead of OR-ing them

see:
- https://github.com/ACINQ/eclair/blob/232f721b39398b4a20ca896b6e24b64c53df069e/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala#L33
- https://github.com/lightningnetwork/lnd/blob/1049d3010a4eb07bbf000f2af15f2bb3862365fb/lnwire/init_message.go#L13-L16

smite/src/bolt/features.rs
NishantBansal2003: Renamed to `is_supported_by`, I think that would be clearer when checking whether something is a subset of a feature

smite-scenarios/src/executor.rs
```diff
@@ -136,8 +136,8 @@ pub struct ProgramContext {
     pub chain_hash: [u8; 32],
     /// Current block height at snapshot time.
     pub block_height: u32,
-    /// Target's advertised feature bits from init message.
-    pub target_features: Vec<u8>,
+    /// Features negotiated between the target node and Smite.
+    pub negotiated_features: Features,
```
ekzyis: nit: what confuses me a little here is that since these are already _negotiated_ features, I'm not sure what would be the difference between an option vs required bit to be set for a feature.

For the rest of the code, does it matter if a negotiated feature was optional from both peers etc.?

ekzyis, line 140: What confused me a little here is that since these are already _negotiated_ features, I'm not sure what would be the difference between an odd vs even bit for a feature.

`is_supported_by` checks both bits, so for the rest of the code, it doesn't matter which bit of a feature you set afaict. If I could tell correctly, I think this would be nice to clarify:


```suggestion
    /// Features negotiated between the target node and Smite. A feature may
    /// appear as an even or odd bit; the distinction carries no meaning here.
    pub negotiated_features: Features,
```

smite/src/oracles/accept_channel.rs
```diff
@@ -102,9 +121,24 @@ fn verify_accepted_open_channel(open_channel: &OpenChannel) -> Result<(), String
...
+        } else {
+            return Err("open_channel does not include upfront_shutdown_script".to_string());
+        }
+    }
+
+    // Check option_channel_type in negotiated features since it is assumed to
+    // be supported.
+    if !negotiated_features.supports_feature(Features::OPTION_CHANNEL_TYPE) {
+        return Err("option_channel_type is not supported".to_string());
+    }
```
ekzyis, line 139: I thought `ASSUMED` and "can safely be ignored" in BOLT-9 means that the peer doesn't need to set any bit for this feature, but we simply assume it's enabled. So I'm not sure if this check here makes sense.

smite-scenarios/src/executor/tests/harness.rs
```diff
@@ -139,7 +139,11 @@ pub fn sample_context() -> ProgramContext {
         target_pubkey: sample_pubkey(1),
         chain_hash: [0xcc; 32],
         block_height: 800_000,
-        target_features: vec![],
+        negotiated_features: Features::from_bits(&[
+            Features::OPTION_STATIC_REMOTEKEY,
+            Features::OPTION_ANCHORS,
+            Features::OPTION_CHANNEL_TYPE,
+        ]),
```
ekzyis, line 146: Related to my other comment about removing the check for `OPTION_CHANNEL_TYPE`: I interpret "can be safely ignored" as "don't check if the feature is enabled; assume it is", so a peer who didn't set this bit in `init` isn't wrong.

I think it could make sense to distinguish which bits we set here because we assume them to be negotiated (`OPTION_STATIC_REMOTEKEY` and `OPTION_CHANNEL_TYPE`) vs. for other reasons (`OPTION_ANCHORS`). So instead of adding assumed feature bits manually here, we could have a list of assumed features, and then include them here for our sample `ProgramContext` in the tests and in `SnapshotSetup` during fuzzing.

However, this is related to the question of whether we even need to include assumed bits in our context in the first place if we can assume them to be set, so all checks based on these features are unconditional.

We could also postpone this discussion for an `InitOracle`, because it would be more concerned with the meaning of bits, like that only one bit should be set for a feature.

WDYT?

smite/src/oracles/accept_channel.rs
ekzyis, line 270: Unrelated to this PR: Shouldn't this be this?

```diff
-if accept_channel.dust_limit_satoshis > accept_channel.channel_reserve_satoshis
+if accept_channel.dust_limit_satoshis > open_channel.channel_reserve_satoshis
```

smite/src/oracles/accept_channel.rs
```diff
@@ -114,13 +148,40 @@ fn verify_accepted_open_channel(open_channel: &OpenChannel) -> Result<(), String
...
+        && open_channel.feerate_per_kw != 0
+    {
+        return Err(format!(
+            "zero_fee_commitments requires feerate_per_kw to be 0, but got {}",
+            open_channel.feerate_per_kw,
+        ));
+    }
+
+    // Check that option_scid_alias is only negotiated for private channels.
+    let announce_channel = (open_channel.channel_flags & 1) == 1;
```
ekzyis, line 175: nit: could define a constant `ANNOUNCE_CHANNEL_FLAG` so it's clear why we use 1 here without looking up the spec. It's kind of easy to miss in the spec:

https://github.com/lightning/bolts/blob/152897261850d93c4f4597f39cf22d7d22d6ede6/02-peer-protocol.md?plain=1#L769-L772

smite/src/oracles/accept_channel.rs
```diff
@@ -83,13 +95,20 @@ impl Oracle<AcceptChannelContext<'_>> for AcceptChannelOracle {
...
 ///   such as LDK accept zero channel reserves on the receiving side, so we do
 ///   not enforce this check on the target's receiving side.
-fn verify_accepted_open_channel(open_channel: &OpenChannel) -> Result<(), String> {
+fn verify_accepted_open_channel(
+    open_channel: &OpenChannel,
+    negotiated_features: &Features,
+) -> Result<(), String> {
+    // Check that option_dual_fund has not been negotiated.
+    if negotiated_features.supports_feature(Features::OPTION_DUAL_FUND) {
+        return Err("option_dual_fund has been negotiated".to_string());
```
ekzyis, line 104: nit: We use `supports_feature` here and at line 138, but the error message uses "negotiated" vs "supported". I think we should use one verb consistently in the error messages. I have a slight preference towards "negotiated".

smite/src/oracles/accept_channel.rs
```diff
@@ -102,9 +121,24 @@ fn verify_accepted_open_channel(open_channel: &OpenChannel) -> Result<(), String
...
+            }
+        } else {
+            return Err("open_channel does not include upfront_shutdown_script".to_string());
+        }
+    }
+
+    // Check option_channel_type in negotiated features since it is assumed to
+    // be supported.
+    if !negotiated_features.supports_feature(Features::OPTION_CHANNEL_TYPE) {
+        return Err("option_channel_type is not supported".to_string());
```
ekzyis, line 138: nit: see other comment

```suggestion
        return Err("option_channel_type has not been negotiated".to_string());
```

but would also need replacement in tests

smite/src/bolt/features.rs
```diff
@@ -301,6 +319,111 @@ mod tests {
...
+
+        // A feature advertised in neither parity is still not negotiated.
+        let anchors = Features::from_bits(&[Features::OPTION_ANCHORS]);
+        assert!(!anchors.is_supported_by(&negotiated));
+    }
+
+    #[test]
+    fn is_supported_by_accepts_required_bit_for_optional_bit() {
+        // The pairing is symmetric: an optional bit on the left is satisfied
+        // by the required bit on the right.
```
ekzyis, line 419: nit:


```suggestion
        // The pairing is symmetric: an optional bit on the left is satisfied
        // by the required bit on the right and vice versa.
```

smite/src/oracles/accept_channel.rs
```diff
@@ -102,9 +121,24 @@ fn verify_accepted_open_channel(open_channel: &OpenChannel) -> Result<(), String
...
     }
 
+    // Check that the upfront shutdown script is present and valid when negotiated.
+    if negotiated_features.supports_feature(Features::OPTION_UPFRONT_SHUTDOWN_SCRIPT) {
+        if let Some(script) = &open_channel.tlvs.upfront_shutdown_script {
+            if !script.is_empty() && !is_acceptable_shutdown_script(script, negotiated_features) {
+                return Err("upfront_shutdown_script is not valid".to_string());
+            }
+        } else {
+            return Err("open_channel does not include upfront_shutdown_script".to_string());
```
ekzyis, line 131: nit: I like how we use only `upfront_shutdown_script` here, but we use `a channel_type` in other places. Probably not worth fixing, but I thought it was worth mentioning.

smite/src/oracles/accept_channel.rs
```diff
@@ -102,9 +121,24 @@ fn verify_accepted_open_channel(open_channel: &OpenChannel) -> Result<(), String
...
+        }
```
ekzyis, line 132: refactor suggestion to avoid triple nesting:

```diff
diff --git a/smite/src/oracles/accept_channel.rs b/smite/src/oracles/accept_channel.rs
index c07c378..a80d53b 100644
--- a/smite/src/oracles/accept_channel.rs
+++ b/smite/src/oracles/accept_channel.rs
@@ -123,12 +123,11 @@ fn verify_accepted_open_channel(
 
     // Check that the upfront shutdown script is present and valid when negotiated.
     if negotiated_features.supports_feature(Features::OPTION_UPFRONT_SHUTDOWN_SCRIPT) {
-        if let Some(script) = &open_channel.tlvs.upfront_shutdown_script {
-            if !script.is_empty() && !is_acceptable_shutdown_script(script, negotiated_features) {
-                return Err("upfront_shutdown_script is not valid".to_string());
-            }
-        } else {
+        let Some(script) = &open_channel.tlvs.upfront_shutdown_script else {
             return Err("open_channel does not include upfront_shutdown_script".to_string());
+        };
+        if !script.is_empty() && !is_acceptable_shutdown_script(script, negotiated_features) {
+            return Err("upfront_shutdown_script is not valid".to_string());
         }
     }
 
@@ -218,12 +217,11 @@ fn verify_accept_channel(
 ) -> Result<(), String> {
     // Check that the upfront shutdown script is present and valid when negotiated.
     if negotiated_features.supports_feature(Features::OPTION_UPFRONT_SHUTDOWN_SCRIPT) {
-        if let Some(script) = &accept_channel.tlvs.upfront_shutdown_script {
-            if !script.is_empty() && !is_standard_shutdown_script(script, negotiated_features) {
-                return Err("upfront_shutdown_script is not valid".to_string());
-            }
-        } else {
+        let Some(script) = &accept_channel.tlvs.upfront_shutdown_script else {
             return Err("accept_channel does not include upfront_shutdown_script".to_string());
+        };
+        if !script.is_empty() && !is_standard_shutdown_script(script, negotiated_features) {
+            return Err("upfront_shutdown_script is not valid".to_string());
         }
     }
```

smite/src/oracles/accept_channel.rs
```diff
@@ -83,13 +95,20 @@ impl Oracle<AcceptChannelContext<'_>> for AcceptChannelOracle {
 ///   be less than or equal to the channel reserve. However, implementations
 ///   such as LDK accept zero channel reserves on the receiving side, so we do
 ///   not enforce this check on the target's receiving side.
```
ekzyis, line 97: The following checks appear to be unintentionally missing:

1. unexpected `chain_hash`
2. unreasonably large `to_self_delay`
3. `feerate_per_kw` that is too small or unreasonably large

I think we could add 1. by adding `expected_chain_hash` to `AcceptChannelContext`. For 2. and 3., we would need to decide which values are too small or unreasonably large. I think we've done this before for other values, i.e. bitcoin amounts > 21m don't make sense.

smite/src/oracles/accept_channel.rs
```diff
@@ -10,7 +13,9 @@ use bitcoin::Amount;
 
 // Constants from the BOLT 2 `open_channel` and `accept_channel` requirements:
 // https://github.com/lightning/bolts/blob/master/02-peer-protocol.md#requirements-8
-const MAX_ACCEPTED_HTLCS_LIMIT: u16 = 483;
+const MAX_FUNDING_SATOSHIS_NO_WUMBO: u64 = (1 << 24) - 1;
+const MAX_ACCEPTED_HTLCS_ZERO_FEE_COMMITMENTS: u16 = 114;
+const MAX_ACCEPTED_HTLCS_DEFAULT: u16 = 483;
```
ekzyis, line 18: nit: could rename `MAX_FUNDING_SATOSHIS_NO_WUMBO` to `MAX_FUNDING_SATOSHIS_DEFAULT` to be consistent with `MAX_ACCEPTED_HTLCS_DEFAULT`

ekzyis (commented):
Left some more nits, but LGTM! Will probably ACK tomorrow

smite/src/oracles/accept_channel.rs
```diff
@@ -114,13 +115,35 @@ fn verify_accepted_open_channel(open_channel: &OpenChannel) -> Result<(), String
...
+            "zero_fee_commitments requires feerate_per_kw to be 0, but got {}",
+            open_channel.feerate_per_kw,
+        ));
+    }
+
+    // Check that option_scid_alias is only negotiated for private channels.
+    let announce_channel = (open_channel.channel_flags & 1) == 1;
+    if announce_channel && channel_type.supports_feature(Features::OPTION_SCID_ALIAS) {
+        return Err("option_scid_alias requires the channel to be private".to_string());
+    }
```
morehouse, line 140: I see the spec sender requirement for this but not the receiver one.  Should we update the spec with it?

smite/src/oracles/accept_channel.rs
morehouse: See https://github.com/lnfuzz/smite/pull/185#discussion_r3754034464.

smite/src/oracles/accept_channel.rs
```diff
@@ -582,6 +731,31 @@ mod tests {
...
+    #[test]
+    fn commitment_validation_skipped_for_option_simple_taproot() {
+        let mut oc = open_channel();
+        oc.tlvs.channel_type = Some(ChannelTypeVariant::SimpleTaproot.encode());
+        oc.push_msat = oc.funding_satoshis * 1000;
+        let mut ac = accept_channel();
+        ac.tlvs.channel_type = Some(ChannelTypeVariant::SimpleTaproot.encode());
+
+        assert_pass(&ac, Some(&pending_negotiation(oc)));
+    }
```
morehouse, line 757: IIUC, the first test is for an acceptable behavior (i.e. the opener should be able to push the full channel balance out for 0FC since they don't have to pay for the anchor or commitment fee).

But I don't think that's true for taproot channels -- the opener still needs to afford the anchor output and commitment fees.  So the second test may actually be something we should eventually reject...

smite-scenarios/src/executor/tests/harness.rs
```diff
@@ -139,7 +139,7 @@ pub fn sample_context() -> ProgramContext {
         target_pubkey: sample_pubkey(1),
         chain_hash: [0xcc; 32],
         block_height: 800_000,
-        target_features: vec![],
+        negotiated_features: Features::from(vec![0x40, 0x10, 0x00]),
```
morehouse: Nit: we should use the feature bit constants in this commit, rather than switching to them in the final commit

smite/src/bolt/features.rs
```diff
@@ -120,6 +126,18 @@ impl Features {
         self.clear_bit(bit);
         self.clear_bit(bit ^ 1);
     }
+
+    /// Returns whether every bit set here is supported by `other`, where the
+    /// feature's required (even) or optional (odd) bit both count as support.
+    #[must_use]
+    pub fn is_supported_by(&self, other: &Features) -> bool {
```
morehouse, line 133: Nit: I think it's more consistent with the existing methods to call this `supports_features` and swap the usage of `self` and `other`

smite/src/oracles/accept_channel.rs
```diff
@@ -423,22 +667,101 @@ mod tests {
...
+            &sample_negotiated_features(),
             "invalid open_channel: open_channel does not include a channel_type",
         );
     }
 
     #[test]
-    fn open_channel_max_accepted_htlcs_above_the_limit() {
+    fn open_channel_channel_type_contains_non_negotiated_features() {
+        let mut oc = open_channel();
+        oc.tlvs.channel_type = Some(vec![0x10, 0x00]);
```
morehouse, line 678: Can we use the existing constants here for clarity?

smite/src/oracles/accept_channel.rs
```diff
@@ -423,22 +667,101 @@ mod tests {
...
+            "invalid open_channel: channel_type contains features that were not negotiated",
+        );
+    }
+
+    #[test]
+    fn open_channel_with_unknown_channel_type_variant() {
+        let mut oc = open_channel();
+        oc.tlvs.channel_type = Some(vec![0x40, 0x40, 0x10, 0x00]);
+
+        let mut negotiated_features = Features::from(vec![0x40, 0x40, 0x10, 0x00]);
```
morehouse, line 698: Can we use constants here for clarity?

smite/src/oracles/accept_channel.rs
morehouse: Agree.  I don't think it's a spec violation to omit the feature bit -- the idea behind making it ASSUMED was to allow reusing the feature bit eventually with a different meaning.

smite/src/oracles/accept_channel.rs
```diff
@@ -102,9 +121,24 @@ fn verify_accepted_open_channel(open_channel: &OpenChannel) -> Result<(), String
         ));
     }
 
+    // Check that the upfront shutdown script is present and valid when negotiated.
+    if negotiated_features.supports_feature(Features::OPTION_UPFRONT_SHUTDOWN_SCRIPT) {
```
morehouse, line 125: I couldn't find anywhere in the spec that requires the receiver to validate the sender's shutdown script, and I'm not sure it's needed.  IIUC the worst thing that happens in this case is that the sender uses an invalid/unspendable script, which harms themselves, not the receiver.

morehouse (changes requested):

smite-scenarios/src/executor/tests/harness.rs
NishantBansal2003: Will remove the check for `OPTION_CHANNEL_TYPE` being present in the negotiated features, so I think we can safely remove `OPTION_CHANNEL_TYPE` from the negotiated features as well

smite/src/oracles/accept_channel.rs
NishantBansal2003: Removed this check. See: https://github.com/lnfuzz/smite/pull/209#discussion_r4013270793

smite/src/oracles/accept_channel.rs
NishantBansal2003: 1. Planned as a follow-up, see the PR description
2. Default Max for all is 2016, so maybe we should also check `to_self_delay <= 2016` in `open_channel` only
3. The cap on small is already planned, see the PR description. For the large feerate, I don't think it affects any target, since it would only be set by the opener to pay

smite/src/oracles/accept_channel.rs
NishantBansal2003: Would prefer `MAX_FUNDING_SATOSHIS_NO_WUMBO` only, since `MAX_FUNDING_SATOSHIS_DEFAULT` could be confused with the Max Bitcoin supply

smite/src/oracles/accept_channel.rs
NishantBansal2003: This could be clarified in the spec, and it would be great if the receiver rejected these. Since HTLCs are rejected immediately rather than held until expiry, the reputation loss would not be significant, though it might eventually affect the target. I think this would be beneficial for the network and nodes, but should not be required

I will update the spec and see what others think about it.

smite/src/oracles/accept_channel.rs
NishantBansal2003: - The first test is currently passing by chance and will also pass once we add support for 0FC, so it is fine to keep it as is
- The second test fails without a barrier and will also fail once we update the commitment fee for taproot channels, It should be updated once support for it is added

Will add comments to both of these tests to make it clear

smite/src/oracles/accept_channel.rs
NishantBansal2003: I think if the sender script is invalid, a cooperative close is impossible for the channel's lifetime, and if the target accepted it without validating it, the only option to close the channel is a force-close, leaving its `to_local` output delayed

smite/src/oracles/accept_channel.rs
```diff
@@ -423,22 +650,102 @@ mod tests {
...
+            Some(&pending_negotiation(oc)),
+            &sample_negotiated_features(),
+            "invalid open_channel: zero_fee_commitments requires feerate_per_kw to be 0",
+        );
+    }
+
+    #[test]
+    fn open_channel_option_scid_alias_for_public_channel() {
+        let mut oc = open_channel();
+        oc.channel_flags = 1;
```
morehouse: Nit:


```suggestion
        oc.channel_flags = ANNOUNCE_CHANNEL_FLAG;
```

or alternatively, it looks like the announce bit is already set in `open_channel()`:


```suggestion
```

morehouse (approved):
LGTM pending squash

smite/src/oracles/accept_channel.rs
ekzyis: Sorry, forgot to read the PR description. Makes sense!

PR #211, opened 2026-08-19
smite-scenarios: sync target chain view via RPC after mining

NishantBansal2003:
ref: #195 

Targets (CLN, LDK) previously learned about new blocks through bitcoind's `blocknotify`, which fires asynchronously. When multiple blocks are mined, this can queue up many sync calls, adding unnecessary load. Instead, explicitly sync the target once after each `MineBlocks` operation.
Introduce a `TargetRpc` trait with a `chain_sync()` method, called by the executor immediately after `MineBlocks`. CLN and LDK sync explicitly via RPC and signals, respectively, while LND and Eclair use ZMQ and remain no-ops.
Dropping `-blocknotify` also removes the burst of asynchronous `SIGUSR1`s during initial block generation, so LDK no longer needs the `pre_exec` `SIG_IGN`

The reason I went with an RPC trait is that, in the future, we'll need access to the target's CLI/RPC to add payment hash and preimage information to the target so that it can send an `update_fulfill_htlc` message.
Also, for now, I only added the CLI as a trait implementation for each target to keep the changes contained and concise. In the future, we can definitely add direct RPC calls over the socket as well to save overhead.

### DIFF
--- a/smite-scenarios/src/targets/cln.rs
+++ b/smite-scenarios/src/targets/cln.rs
@@ -9,7 +9,9 @@
 //! This means checking lightningd's liveness is sufficient for crash detection.
 
 use std::fs;
+use std::io::{self, ErrorKind};
 use std::net::SocketAddr;
+use std::os::unix::net::UnixStream;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 use std::time::Duration;
@@ -20,7 +22,7 @@ use smite::bitcoin::BitcoinCli;
 use smite::process::ManagedProcess;
 
 use super::bitcoind;
-use super::{Target, TargetError, check_crash_log};
+use super::{Target, TargetError, TargetRpc, check_crash_log};
 
 /// Configuration for the CLN target.
 pub struct ClnConfig {
@@ -43,15 +45,120 @@ impl Default for ClnConfig {
 }
 
 impl ClnConfig {
-    fn bitcoind_config(&self, data_dir: &Path) -> bitcoind::BitcoindConfig {
+    fn bitcoind_config(&self) -> bitcoind::BitcoindConfig {
         bitcoind::BitcoindConfig {
             rpc_port: self.bitcoind_rpc_port,
             p2p_port: self.bitcoind_p2p_port,
-            extra_args: vec![format!(
-                "-blocknotify=lightning-cli --lightning-dir='{}' --network=regtest syncblocks",
-                data_dir.join("cln").display()
-            )],
             ..bitcoind::BitcoindConfig::default()
+        }
+    }
+}
+
+/// RPC handle for interacting with CLN node target.
+#[derive(Debug, Clone)]
+pub struct ClnRpc {
+    /// Path to the CLN node's unix RPC socket.
+    rpc_socket: PathBuf,
+}
+
+impl ClnRpc {
+    /// Bound RPC socket I/O so a stalled lightningd cannot block indefinitely.
+    const RPC_IO_TIMEOUT: Duration = Duration::from_secs(1);
+
+    /// Sends a JSON-RPC request to CLN over its Unix RPC socket and returns the
+    /// `result` of the response.
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`io::Error`] only when lightningd is unreachable: the socket
+    /// cannot be connected to, or the reply times out, is cut short or never
+    /// arrives. That is a symptom of an earlier crash/hang rather than a fault
+    /// in the call.
+    ///
+    /// # Panics
+    ///
+    /// If `params` cannot be serialized, the reply is not valid JSON-RPC, or
+    /// CLN answers with a JSON-RPC `error` object.
+    fn run(&self, method: &str, params: impl serde::Serialize) -> io::Result<serde_json::Value> {
+        /// A JSON-RPC response from lightningd. `jsonrpc` and `id` are echoed
+        /// back but unused, and exactly one of `result` or `error` is set.
+        #[derive(Deserialize)]
+        struct RpcResponse {
+            #[serde(default)]
+            result: serde_json::Value,
+            error: Option<serde_json::Value>,
+        }
+
+        let response = (|| -> io::Result<RpcResponse> {
+            // Connect to lightningd's RPC socket.
+            let mut sock = UnixStream::connect(&self.rpc_socket)?;
+            sock.set_read_timeout(Some(Self::RPC_IO_TIMEOUT))
+                .expect("valid timeout");
+            sock.set_write_timeout(Some(Self::RPC_IO_TIMEOUT))
+                .expect("valid timeout");
+
+            // Write the request to the socket.
+            let request = serde_json::json!({
+                "jsonrpc": "2.0",
+                "id": "smite",
+                "method": method,
+                "params": params,
+            });
+            serde_json::to_writer(&mut sock, &request)?;
+
+            // Read the response.
+            serde_json::Deserializer::from_reader(&mut sock)
+                .into_iter()
+                .next()
+                .transpose()?
+                .ok_or_else(|| {
+                    io::Error::new(
+                        io::ErrorKind::UnexpectedEof,
+                        format!("lightningd closed the socket without answering {method}"),
+                    )
+                })
+        })();
+
+        // Return errors that may indicate the target crashed or hung so
+        // `check_alive/ping-pong` can classify them later. Any other error
+        // indicates a malformed request or a complete response that is not
+        // valid JSON-RPC.
+        let response = match response {
+            Ok(response) => response,
+            Err(e) => match e.kind() {
+                ErrorKind::ConnectionRefused
+                | ErrorKind::ConnectionReset
+                | ErrorKind::ConnectionAborted
+                | ErrorKind::BrokenPipe
+                | ErrorKind::WouldBlock
+                | ErrorKind::TimedOut
+                | ErrorKind::UnexpectedEof => return Err(e),
+                _ => panic!("malformed {method} request or response: {e}"),
+            },
+        };
+
+        if let Some(error) = response.error {
+            panic!("lightningd rejected {method}: {error}");
+        }
+
+        Ok(response.result)
+    }
+}
+
+impl TargetRpc for ClnRpc {
+    /// RPC to make CLN poll for new blocks immediately instead of waiting for
+    /// its regular poll interval, allowing it to sync faster.
+    ///
+    /// # Panics
+    ///
+    /// - If lightningd answers with an error, which means the call itself is at
+    ///   fault rather than the target having crashed/hanged.
+    fn chain_sync(&mut self) {
+        if let Err(e) = self.run("syncblocks", serde_json::json!({})) {
+            // lightningd is unreachable, indicating that CLN has already
+            // crashed/hanged, check_alive/ping-pong will report the crash/hang
+            // at the end.
+            log::warn!("syncblocks could not reach lightningd: {e}");
         }
     }
 }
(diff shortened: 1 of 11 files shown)


smite-scenarios/src/targets/ldk.rs
```diff
@@ -41,14 +41,40 @@ impl LdkConfig {
...
+            .arg("-USR1")
+            .arg("-f")
+            .arg("^ldk-node-wrapper")
+            .output()
+            .expect("pkill -USR1 ldk-node-wrapper should not fail");
+        assert!(
+            out.status.success(),
+            "pkill -USR1 ldk-node-wrapper failed: {}",
+            String::from_utf8_lossy(&out.stderr)
+        );
```
erickcestari: Should we have an assert here? Imagine the executor crashes the target while executing the input. It would probably fail to sync and panic with a `pkill` error instead of failing with a `Violation::Crashed `error.

smite-scenarios/src/targets/ldk.rs
```diff
@@ -90,28 +116,6 @@ impl LdkTarget {
             cmd.env("LD_PRELOAD", handler);
         }
 
-        // Ignore SIGUSR1 for the window between exec and the wrapper blocking
```
erickcestari, line 93: Some comments at `workloads/ldk/src/main.rs` also need to be updated after the remove of `SIG_IGN`.

erickcestari (changes requested):
Good idea! It's much better that it only calls the sync signal once per `MineBlock` instead of calling it for every block mined.

smite-scenarios/src/targets/ldk.rs
NishantBansal2003: Hmm, great point, I think `pkill` or `lightning-cli` should always succeed in any case, and they should only error out when the target has already crashed. So, I think we could check whether the target has crashed before asserting this. If it has crashed, we can simply log the failure here, otherwise, we should assert it. I don't think there is any other case where RPC commands should be unresponsive, so I think it's better to be stricter. WDYT?

erickcestari: Yes, I agree! I also think logging is a good option.

smite-scenarios/src/targets/cln.rs
```diff
@@ -43,19 +45,87 @@ impl Default for ClnConfig {
...
-                data_dir.join("cln").display()
-            )],
             ..bitcoind::BitcoindConfig::default()
         }
     }
 }
 
+/// RPC handle for interacting with CLN node target.
+#[derive(Debug, Clone)]
+pub struct ClnCli {
```
NishantBansal2003: I realized this should now be renamed to `ClnRpc`. I'll update it

smite-scenarios/src/targets/ldk.rs
```diff
@@ -41,14 +41,49 @@ impl LdkConfig {
...
+            Some(0) => {}
+            // pkill exits 1 when nothing matches, so LDK is gone, check_alive
+            // will report the crash at the end.
+            Some(1) => log::warn!("ldk-node-wrapper is not running, skipping chain sync"),
+            _ => panic!(
+                "pkill -USR1 ldk-node-wrapper failed: {}",
+                String::from_utf8_lossy(&out.stderr)
+            ),
+        }
+    }
```
erickcestari, line 76: We could use the ldk `ManagedProcess` to get the `PID` and use the syscall `kill` directly instead of creating a `pkill` process to do this for us.

erickcestari (changes requested):
LGTM! I think we can optimize this a bit by avoiding spawning the `pkill` process, consequently reducing the memory overhead and the amount of memory dirtied by process creation.

smite-scenarios/src/targets/ldk.rs
NishantBansal2003: Done and Rebased!

smite-scenarios/src/targets/cln.rs
```diff
@@ -43,19 +45,87 @@ impl Default for ClnConfig {
...
+            .next()
+            .unwrap_or_else(|| panic!("lightningd closed the socket without answering {method}"))
+            .unwrap_or_else(|e| panic!("failed to read {method} response from lightningd: {e}"));
+        assert!(
+            response.get("error").is_none(),
+            "lightningd rejected {method}: {}",
+            response["error"]
+        );
+
+        Ok(response["result"].take())
```
morehouse: Should we define an `RpcResponse` struct like we do for all other RPCs?  It would make it clearer what we expect a response to contain.

smite-scenarios/src/targets/cln.rs
```diff
@@ -43,19 +45,87 @@ impl Default for ClnConfig {
...
             ..bitcoind::BitcoindConfig::default()
         }
     }
 }
 
+/// RPC handle for interacting with CLN node target.
+#[derive(Debug, Clone)]
+pub struct ClnRpc {
+    /// Path to the CLN node's unix RPC socket.
+    pub rpc_socket: PathBuf,
```
morehouse:
```suggestion
    rpc_socket: PathBuf,
```

smite-scenarios/src/targets/cln.rs
```diff
@@ -43,19 +45,87 @@ impl Default for ClnConfig {
...
+}
+
+impl ClnRpc {
+    // Bound RPC socket I/O so a stalled lightningd cannot block indefinitely.
```
morehouse:
```suggestion
    /// Bound RPC socket I/O so a stalled lightningd cannot block indefinitely.
```

smite-scenarios/src/targets/cln.rs
```diff
@@ -43,19 +45,87 @@ impl Default for ClnConfig {
...
+            "params": params,
+        });
+        serde_json::to_writer(&mut sock, &request)
+            .unwrap_or_else(|e| panic!("failed to send {method} to lightningd: {e}"));
+
+        let mut response: serde_json::Value = serde_json::Deserializer::from_reader(&mut sock)
+            .into_iter()
+            .next()
+            .unwrap_or_else(|| panic!("lightningd closed the socket without answering {method}"))
+            .unwrap_or_else(|e| panic!("failed to read {method} response from lightningd: {e}"));
```
morehouse: This code is confusing -- what is being unwrapped twice?  Also, the second panic triggers on timeouts too, which seems like it should also be "warn-only" since a crash/hang will be detected later.

Perhaps we should match on the response for clarity and to let us distinguish the timeout error types from other types.

workloads/ldk/src/main.rs
```diff
@@ -32,12 +32,12 @@ fn install_panic_hook() {
...
-/// moment it is delivered, while a *blocked* one stays pending until `sigwait()`
-/// consumes it, regardless of its disposition. So this call is what makes
-/// bitcoind's `-blocknotify` SIGUSR1 observable, and why nothing here calls
-/// `sigaction`: the wait loop below is the only consumer these signals need.
+/// defaults to terminating the process. That distinction is the whole point: an
+/// *ignored* signal is discarded the moment it is delivered, while a *blocked*
+/// one stays pending until `sigwait()` consumes it, regardless of its
+/// disposition. So this call is what makes target's SIGUSR1 observable, and
+/// why nothing here calls `sigaction`: the wait loop below is the only consumer
+/// these signals need.
```
morehouse: No need to explain the difference between signal-blocking and `SIG_IGN` anymore.

```suggestion
/// defaults to terminating the process. A blocked SIGUSR1
/// stays pending until `sigwait()` consumes it, regardless of its
/// disposition. So this call is what makes target's SIGUSR1 observable, and
/// why nothing here calls `sigaction`: the wait loop below is the only consumer
/// these signals need.
```

smite-scenarios/src/targets/ldk.rs
```diff
@@ -41,14 +41,41 @@ impl LdkConfig {
...
+    /// Panics if the signal cannot be sent, which means the call itself is at
+    /// fault rather than the target having crashed.
+    fn chain_sync(&mut self) {
+        // A crashed wrapper remains an unreaped zombie, so the signal is sent
+        // but discarded. check_alive will report the crash at the end.
+        if let Err(e) = send_sigusr1(self.pid) {
+            panic!(
+                "failed to send SIGUSR1 to ldk-node-wrapper (pid {}): {e}",
+                self.pid
+            );
```
morehouse, line 76: Why do we panic for LDK but warn for CLN in the same situation?

smite-scenarios/src/targets/lnd.rs
```diff
@@ -81,6 +81,16 @@ impl CoveragePipes {
...
 }
 
+/// RPC handle for interacting with LND node target.
+#[derive(Debug, Clone)]
+pub struct LndRpc;
+
+impl TargetRpc for LndRpc {
+    /// LND receives new blocks directly from bitcoind over ZMQ, so no manual
+    /// chain synchronization is required.
+    fn chain_sync(&mut self) {}
```
ekzyis: nit: could it make sense to assert the block height here, since after calling this function, I think the caller expects the chain to be synced? But then we would need to pass the expected block height.

ekzyis, line 91: nit: could it make sense to assert the block height here, since after calling this function, I think the caller expects the chain to be synced? But then we would need to pass the expected block height somehow.

smite-scenarios/src/targets/ldk.rs
NishantBansal2003: Because bad RPC calls already panic inside `self.run` for CLN, while cases where CLN has already crashed or hung just return `Err` so the caller can log them. In the future, I think we'll mostly remove this pattern, and will return `Err` directly from `chain_sync` or other RPC calls and instead let `ir.rs` log the RPC error itself, since we may also run commands like `addinvoice` and want their responses to proceed further

For LDK, the signal-based approach is essentially fire-and-forget. Even if LDK has crashed or hung, its PID remains until the parent reaps it, so the signal is delivered normally (though it won't be responsive) and the call returns `Ok`. So, any other error is unexpected, so we panic on it

smite-scenarios/src/targets/cln.rs
NishantBansal2003: I updated the error handling to cover possible cases where the target is hung or crashed, whether before the call, during the write, or during the read, and also made it a bit clearer. I also verified some crash cases by crashing CLN before the run, after the write, and after the read, and the handling covers all of them

Wasn't fully sure how to test the hang cases, but I think `TimedOut` or `UnexpectedEof` should cover those cases as well

smite-scenarios/src/targets/lnd.rs
NishantBansal2003: Yeah, I mean that is the defensive way, but the current sync model is essentially fire-and-forget. To actually get the updated block height for each implementation, we would need to wait for some time and then make another RPC call, and I'm not sure how long we should wait or how many times we should retry. In the worst case, we might make several RPC calls, which could hurt the fuzzing itself

So I believe a cheaper way to check whether the blocks have been synced is to run a full funding flow and see if we receive `channel_ready` from the peer

smite-scenarios/src/targets/cln.rs
```diff
@@ -58,54 +58,90 @@ impl ClnConfig {
...
+        let response = match response {
+            Ok(response) => response,
+            Err(e) => match e.kind() {
+                ErrorKind::ConnectionRefused
+                | ErrorKind::ConnectionReset
+                | ErrorKind::ConnectionAborted
+                | ErrorKind::BrokenPipe
+                | ErrorKind::WouldBlock
+                | ErrorKind::TimedOut
+                | ErrorKind::UnexpectedEof => return Err(e),
```
morehouse, line 135: Should we also return an error for `NotFound` and `NotConnected`?

morehouse (approved):
Ready to squash and rebase

smite-scenarios/src/targets/cln.rs
NishantBansal2003: IDTS, `NotFound` can't happen because the socket file is never unlinked after startup and is known to exist even if CLN has crashed. And `NotConnected` can't happen because the stream is connected by construction. So even if the target crashes afterward, we get a pipe error or some other connection error, but not `NotConnected`

PR #216, opened 2026-08-27
smite-scenarios: fix channel state on temporary_channel_id reuse

### LINKED ISSUE #215: Channel state desync on `temporary_channel_id` collision
The following sequence of events leads to false positive reports due to the executor's channel state getting desynced with the target's:

1. Complete full funding flow with `temporary_channel_id` `T1` and funding outpoint `F1`.
2. Send `funding_created` reusing `T1` and a new outpoint `F2`.
3. Complete full funding flow with a new `temporary_channel_id` `T3` and reuse outpoint `F2`.

During step 1, the executor sets the `funding_built` flag on the negotiation for `T1` and creates a `channel_states` entry for `F1`.

During step 2 the executor doesn't check `funding_built` at all and happily creates a new `channel_states` entry for `F2`.  The target sees the sent `funding_created` message as invalid since there's no matching open/accept negotiation currently live for `T1`, and rejects it.  CLN and LDK reject by sending an `error`, which lets the executor bail out and avoid this false positive.  But LND and Eclair just *ignore* the `funding_created`, so we continue to step 3.

During step 3 the executor skips creating a `channel_states` entry for `F1` since one already exists.  The executor ends up keeping the invalid channel state from step 2 while the target saves the valid state from step 3.  As a result, our signature check for the received `funding_signed` fails.

### Suggested fix

I think the fix is pretty simple: only create a new `channel_state` entry if there's a valid negotiation active that does not have `funding_built` set.

NishantBansal2003:
Closes: #215 

Completing a funding flow for `temporary_channel_id` and then sending another `funding_created` reusing `temporary_channel_id` with a different funding outpoint could desync our channel state from the target's. The second message may be ignored by targets such as LND and Eclair, while we still recorded its channel state. If a later funding flow legitimately uses the same outpoint, we may not update that state and instead surface the target's valid `funding_signed` signature as invalid.

Only insert a `channel_states` entry while the negotiation has not built a `funding_created` yet.

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -961,17 +961,25 @@ fn build_funding_created(
         open_channel.funding_satoshis,
     );
 
-    // Building the same message again must not clobber a channel whose state
-    // has already been established (and possibly advanced).
-    channel_states.entry(channel_id).or_insert_with(|| {
-        ChannelState::new(
-            config,
-            holder,
-            state,
-            is_funding_outpoint_valid,
-            mined_txids.contains(&funding_outpoint.txid),
-        )
-    });
+    // Only track a new channel when this negotiation has not built a
+    // `funding_created` yet. If it has, we are likely resending one for the
+    // same `temporary_channel_id` with a different outpoint, which the target
+    // may ignore (LND and Eclair currently do), leaving us tracking a channel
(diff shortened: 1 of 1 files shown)


morehouse (approved):
LGTM

PR #220, opened 2026-08-27
smite-scenarios: take the opener funding pubkey from the wire

morehouse:
When we save the channel config in `build_funding_created`, we currently use the negotiated values for everything except `funding_pubkey`, which gets derived from the local signing key instead.  In practice this doesn't really matter -- usually the negotiated and local keys are the same, and if they differ the target should reject our `funding_created` anyway.

But there are a few small reasons to use the negotiated funding_pubkey instead:
- consistency with the other channel config values
- we avoid a `Secp256k1` context and pubkey derivation
- it's what is intended to be saved anyway


NishantBansal2003 (approved):
LGTM

PR #226, opened 2026-09-02
smite: name the temporary_channel_id with a type alias

erickcestari:
`open_channel` and its replies carry a `temporary_channel_id` that is a plain `ChannelId` on the wire, so every field and map keyed by it read as just another channel id. In particular, a signature such as `HashMap<ChannelId, PendingChannel>` did not say which of the two ids it was keyed by.

Add `TemporaryChannelId` as an alias rather than a newtype: the wire format is identical and the compiler still treats both as one type, so this only documents which id a value holds, at no cost to the callers.

### DIFF
--- a/smite/src/bolt.rs
+++ b/smite/src/bolt.rs
@@ -76,7 +76,7 @@ pub use tx_remove_output::TxRemoveOutput;
 pub use types::{
     BigSize, CHANNEL_ID_SIZE, COMPACT_SIGNATURE_SIZE, ChannelId, MAX_MESSAGE_SIZE,
     PAYMENT_ONION_PACKET_SIZE, PER_COMMITMENT_SECRET_SIZE, PUBLIC_KEY_SIZE, SHA256_HASH_SIZE,
-    SHORT_CHANNEL_ID_SIZE, ShortChannelId, TXID_SIZE, Tu32, Tu64,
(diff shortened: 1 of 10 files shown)


ekzyis (approved):
utACK 6a92930

smite/src/bolt.rs
```diff
@@ -76,7 +76,7 @@ pub use tx_remove_output::TxRemoveOutput;
 pub use types::{
     BigSize, CHANNEL_ID_SIZE, COMPACT_SIGNATURE_SIZE, ChannelId, MAX_MESSAGE_SIZE,
     PAYMENT_ONION_PACKET_SIZE, PER_COMMITMENT_SECRET_SIZE, PUBLIC_KEY_SIZE, SHA256_HASH_SIZE,
-    SHORT_CHANNEL_ID_SIZE, ShortChannelId, TXID_SIZE, Tu32, Tu64,
+    SHORT_CHANNEL_ID_SIZE, ShortChannelId, TXID_SIZE, TemporaryChannelId, Tu32, Tu64,
```
morehouse, line 79: The import is added, but none of the tests were actually converted.

morehouse (approved):
We also should convert `Violation::InvalidAcceptChannel`.

smite/src/bolt.rs
erickcestari: I've also updated the tests at `accept_channel.rs` oracle.

morehouse (approved):
LGTM

PR #229, opened 2026-09-03
smite-scenarios: add executor test fixture

morehouse:
The test fixture eliminates lots of boilerplate code and makes it easier to read and understand what each test is doing.

Additionally, all the hex-string magic constants are given names and moved to `harness.rs` to further improve readability.


smite-scenarios/src/executor/tests/harness.rs
```diff
@@ -111,6 +107,158 @@ impl BitcoinRpc for MockBitcoinCli {
...
+    const TYPE: MessageType;
+
+    /// Returns the extracted BOLT message if `msg`'s type matches, `None`
+    /// otherwise.
+    fn from_message(msg: Message) -> Option<Self>;
+}
+
+/// Implements [`FromMessage`] for BOLT messages whose [`Message`] variant has
+/// the same name.
+macro_rules! impl_from_message {
```
NishantBansal2003: I think this should go in `smite::bolt`, since this logic is also duplicated in the actual connection receive paths, e.g. `recv_accept_channel` and `recv_funding_signed`

smite-scenarios/src/executor/tests.rs
```diff
@@ -1476,65 +1162,29 @@ fn execute_create_funding_transaction_insufficient_funds() {
...
-        .execute(
-            &Program {
-                instructions: send_funding_created_and_recv_funding_signed_instructions(),
-            },
-            std::time::Instant::now(),
-        )
-        .unwrap();
+    let mut fx = Fixture::new()
+        .with_negotiation(sample_funding_negotiation())
+        .queue(&funding_signed_reply(channel_id));
```
NishantBansal2003: nit: This part is repeated 3 times with `channel_id = funding_channel_id()`. I think we could define a `recv_funding_signed_fixture`, similar to `recv_channel_ready_fixture`, and use it in all relevant places

NishantBansal2003 (commented):
Would be great if we could next refactor the instruction-related code in the tests. There's a lot of duplication, and some parts are not idiomatic, e.g. using `instrs.pop()` to drop `RecvFundingSigned` or `send_open_channel_instructions()` + `RecvAcceptChannel` repeated many times

smite-scenarios/src/executor/tests/harness.rs
morehouse: Done in 2e67eba508eacb3555aa6c61e3d648a6ee1b1d4f.

smite-scenarios/src/executor/tests.rs
morehouse: Done in 1f280f1fa6377d24558a862557e3994d66dfba6f.

NishantBansal2003 (approved):
LGTM!

PR #230, opened 2026-09-03
smite-scenarios: log warning and tx_abort data on receive

erickcestari:
`warning` and `tx_abort` are, with `error`, the only wire messages that carry a human-readable `data` field. `recv_non_ping` already surfaces `error` through `ExecuteError::PeerError`, but a `warning` or `tx_abort` was handed back to the caller as-is, and callers only inspect the message type, so the text the target sent was lost.

Log the channel id and data at warn level before returning the message. The bytes are printed with a lossy UTF-8 conversion rather than filtered to printable ASCII: targets are cooperating implementations under test, so readability is the only concern.

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -1246,6 +1246,24 @@ fn recv_non_ping(conn: &mut impl Connection, timeout: Duration) -> Result<Messag
             }
             // Surface the received error message.
             Message::Error(e) => return Err(ExecuteError::PeerError(e)),
+            // Log the human-readable data before handing the message to the
+            // caller, which typically only inspects the message type.
+            Message::Warning(ref w) => {
+                log::debug!(
+                    "received warning on {}: {}",
+                    w.channel_id,
+                    String::from_utf8_lossy(&w.data)
+                );
+                return Ok(msg);
+            }
+            Message::TxAbort(ref a) => {
+                log::debug!(
+                    "received tx_abort on {}: {}",
+                    a.channel_id,
+                    String::from_utf8_lossy(&a.data)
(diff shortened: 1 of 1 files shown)


smite-scenarios/src/executor.rs
```diff
@@ -1246,6 +1246,24 @@ fn recv_non_ping(conn: &mut impl Connection, timeout: Duration) -> Result<Messag
             }
             // Surface the received error message.
             Message::Error(e) => return Err(ExecuteError::PeerError(e)),
+            // Log the human-readable data before handing the message to the
+            // caller, which typically only inspects the message type.
+            Message::Warning(ref w) => {
+                log::warn!(
```
morehouse: Let's make these debug-level like all our other extra logging
```suggestion
                log::debug!(
```

smite-scenarios/src/executor.rs
```diff
@@ -1246,6 +1246,24 @@ fn recv_non_ping(conn: &mut impl Connection, timeout: Duration) -> Result<Messag
...
+                    "received warning on {:?}: {}",
+                    w.channel_id,
+                    String::from_utf8_lossy(&w.data)
+                );
+                return Ok(msg);
+            }
+            Message::TxAbort(ref a) => {
+                log::warn!(
+                    "received tx_abort on {:?}: {}",
```
morehouse: `ChannelId` has its own display override we should use to get the hex values.


```suggestion
                    "received tx_abort on {}: {}",
```

morehouse (commented):
Did you run into a situation where this logging is helpful?

morehouse (approved):
LGTM

PR #232, opened 2026-09-04
smite-scenarios: deduplicate executor variable resolvers

erickcestari:
The 17 `resolve_*` functions differed only in the matched variant, and the two `consume_sent_*` functions only in the affine type they voided.

Generate the resolvers with a `define_resolver!` macro that shares one `type_mismatch` panic, and replace the consumers with a single `consume_affine` that takes the expected `VariableType`.

### DIFF
--- a/smite-scenarios/src/executor.rs
+++ b/smite-scenarios/src/executor.rs
@@ -22,7 +22,7 @@ use smite::oracles::{AcceptChannelContext, AcceptChannelOracle, Oracle};
 use smite::pending_channel::PendingChannel;
 use smite::violation::Violation;
 use smite_ir::operation::AcceptChannelField;
-use smite_ir::{Operation, Program, Variable};
+use smite_ir::{Operation, Program, Variable, VariableType};
 use std::collections::{HashMap, HashSet};
 use std::time::Duration;
 
@@ -483,7 +483,11 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                 }
 
                 Operation::RecvAcceptChannel => {
-                    consume_sent_open_channel(&mut variables, instr.inputs[0]);
+                    consume_affine(
+                        &mut variables,
+                        instr.inputs[0],
+                        instr.operation.input_types()[0],
+                    );
                     log::debug!("[{:?}] RecvAcceptChannel: waiting", start.elapsed());
                     let ac = recv_accept_channel(&mut self.conn)?;
                     log::debug!("[{:?}] RecvAcceptChannel: received", start.elapsed());
@@ -496,7 +500,11 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                 }
 
                 Operation::RecvFundingSigned => {
-                    consume_sent_funding_created(&mut variables, instr.inputs[0]);
+                    consume_affine(
+                        &mut variables,
+                        instr.inputs[0],
+                        instr.operation.input_types()[0],
+                    );
                     log::debug!("[{:?}] RecvFundingSigned: waiting", start.elapsed());
                     let fs = recv_funding_signed(&mut self.conn)?;
                     log::debug!("[{:?}] RecvFundingSigned: received", start.elapsed());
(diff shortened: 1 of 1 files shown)


smite-scenarios/src/executor.rs
```diff
@@ -600,197 +608,68 @@ fn resolve(variables: &[Option<Variable>], index: usize) -> &Variable {
...
+    &OpenChannel
+);
+define_resolver!(resolve_accept_channel, AcceptChannel, &AcceptChannel);
+define_resolver!(
+    resolve_funding_transaction,
+    FundingTransaction,
+    &FundingTransaction
+);
+
+/// Consumes an affine variable, leaving its slot void so it cannot be used
```
Chand-ra, line 665: Nit: Rust doesn't have "voids". What we're doing is replacing the used variable with `None`.

Something like this would perhaps be better?
```
Consumes an affine variable, replacing it with `None` to prevent double-use.
```

Chand-ra (commented):
LGTM!

smite-scenarios/src/executor.rs
erickcestari: We are using the term `void` in the codebase to refer to a `None` `Variable`.

For example: https://github.com/erickcestari/smite/blob/7ab8dfbc37ec56a86c06ab1619f1b82637579d71/smite-ir/src/tests.rs#L1999

smite-scenarios/src/executor.rs
```diff
@@ -600,197 +608,68 @@ fn resolve(variables: &[Option<Variable>], index: usize) -> &Variable {
...
-
-fn resolve_channel_id(variables: &[Option<Variable>], index: usize) -> ChannelId {
-    match resolve(variables, index) {
-        Variable::ChannelId(v) => *v,
-        other => panic!(
-            "variable {index}: expected ChannelId, got {:?}",
-            other.var_type(),
-        ),
-    }
+fn type_mismatch(index: usize, expected: VariableType, actual: &Variable) -> ! {
```
NishantBansal2003: nit: could add a rustdoc comment about its usage

smite-scenarios/src/executor.rs
```diff
@@ -600,197 +608,68 @@ fn resolve(variables: &[Option<Variable>], index: usize) -> &Variable {
...
-    }
-}
-
-fn consume_sent_open_channel(variables: &mut [Option<Variable>], index: usize) {
-    match resolve(variables, index) {
-        Variable::SentOpenChannel => {
-            // Consume the affine `SentOpenChannel`.
-            variables[index] = None;
+/// Defines a resolver for a single [`Variable`] variant. A `&T` return type
+/// borrows the payload from the variable slice; any other type copies it out.
```
NishantBansal2003: nit
```suggestion
/// Defines a resolver for a single [`Variable`] variant, panicking on a type
/// mismatch. A `&T` return type borrows the payload from the variable slice;
/// any other return type copies it out.
```

smite-scenarios/src/executor.rs
```diff
@@ -600,197 +608,68 @@ fn resolve(variables: &[Option<Variable>], index: usize) -> &Variable {
...
+define_resolver!(resolve_accept_channel, AcceptChannel, &AcceptChannel);
+define_resolver!(
+    resolve_funding_transaction,
+    FundingTransaction,
+    &FundingTransaction
+);
+
+/// Consumes an affine variable, leaving its slot void so it cannot be used
+/// again.
+fn consume_affine(variables: &mut [Option<Variable>], index: usize, expected: VariableType) {
```
NishantBansal2003, line 667: I think we should add an assert here to panic if `expected` is not an affine type. Otherwise, calling this on a non-affine variable would invalidate every later instruction that reads it

NishantBansal2003 (changes requested):

smite-scenarios/src/executor.rs
erickcestari: Added

smite-scenarios/src/executor.rs
Chand-ra: You're right. A [comment](https://github.com/erickcestari/smite/blob/7ab8dfbc37ec56a86c06ab1619f1b82637579d71/smite-ir/src/tests.rs#L2004) there says:
```
// SendMessage is last and has void output.
```
which uses `void` in the same breadth as you do here. I don't like it there either, but I guess it is a colloquial for the project atp and hence fine.

smite-scenarios/src/executor.rs
```diff
@@ -600,197 +608,75 @@ fn resolve(variables: &[Option<Variable>], index: usize) -> &Variable {
...
-    match resolve(variables, index) {
-        Variable::FundingTransaction(v) => v,
-        other => panic!(
-            "variable {index}: expected FundingTransaction, got {:?}",
-            other.var_type(),
-        ),
-    }
+/// Panics with a diagnostic naming the variable slot, the [`VariableType`] an
+/// operation required, and the type actually found there.
+fn type_mismatch(index: usize, expected: VariableType, actual: &Variable) -> ! {
```
morehouse: Nit: looks like passing the actual variable type is a bit cleaner


```suggestion
fn type_mismatch(index: usize, expected: VariableType, actual: VariableType) -> ! {
```

smite-scenarios/src/executor.rs
```diff
@@ -483,7 +483,11 @@ impl<C: Connection, B: BitcoinRpc> Executor<C, B> {
                 }
 
                 Operation::RecvAcceptChannel => {
-                    consume_sent_open_channel(&mut variables, instr.inputs[0]);
+                    consume_affine(
+                        &mut variables,
+                        instr.inputs[0],
+                        VariableType::SentOpenChannel,
```
morehouse: Nit: we could use `instr.input_types()` to avoid hardcoding the type

```suggestion
                        instr.input_types()[0],
```

Also applies below.

morehouse (approved):

Chand-ra (approved):

smite-scenarios/src/executor.rs
erickcestari: Done!

smite-scenarios/src/executor.rs
erickcestari: Done!

PR #233, opened 2026-09-04
smite: check if upfront shutdown script is standard if negotiated

ekzyis:
Adds the standardness check for upfront shutdown scripts to `AcceptChannelOracle` if negotiated.

This is needed for #163 because `ShutdownOracle` does not check standardness again.


morehouse:
Superseded by #209.

PR #234, opened 2026-09-04
smite: move ChannelTypeVariant to bolt::types

NishantBansal2003:
The `accept_channel` oracle needs to verify that `channel_type` uses the smallest possible bitmap and represents a defined channel type. Keeping `ChannelTypeVariant` in `bolt::types` gives the oracle a canonical set of channel types to validate against and avoids duplicating the definition across modules.

This commit was part of https://github.com/lnfuzz/smite/pull/209, but I thought of separating some of the code-move changes to keep the diff smaller. So this is mostly a code-move change with a few small code changes


smite-ir/src/operation.rs
```diff
@@ -14,7 +14,8 @@ use std::fmt::Write;
 use bitcoin::{opcodes::all as opcodes, script::Builder, script::PushBytes};
 use rand::{Rng, RngExt};
 use serde::{Deserialize, Serialize};
-use smite::bolt::{FeatureBit, Features, ShortChannelId};
+pub use smite::bolt::ChannelTypeVariant;
```
morehouse: I think we should update the other imports rather than exporting this one.

morehouse (changes requested):
Also should move the `ChannelTypeVariant` tests to `types.rs`.

morehouse (approved):
LGTM

Issue #178, opened 2026-07-29
Semantic field malformation


morehouse:
### Problem

Currently there are several message fields that the IR can only produce (mostly) valid values for:

| Message | Field | Why it's always valid |
|---|---|---|
| `funding_created` | `funding_txid` | derived from the constructed funding outpoint |
| `funding_created` | `funding_output_index` | derived from the constructed funding outpoint (always 0) |
| `funding_created` | `signature` | computed |
| `channel_announcement` | all 4 signatures | computed |
| `node_announcement` | `signature` | computed |
| `announcement_signatures` | both signatures | computed |

For these fields, smite is never able to produce invalid values such as all zeroes, all `0xff` bytes, or random data.

This currently prevents smite from finding bugs like the [CLN `funding_txid==0` assertion crash](https://delvingbitcoin.org/t/vulnerability-disclosure-assertion-dos-in-core-lightning/2507).

### Solution

We can add the ability to selectively malform the bytes of messages before we send them to the target.

## Design

New module `smite-ir/src/malform.rs`.

```rust
/// Overwrites one field of an encoded message with the given bytes.
pub struct Malformation {
    /// Byte offset into the encoded message, including the 2-byte type prefix.
    pub offset: u16,
    /// Replacement bytes.
    pub bytes: Vec<u8>,
}

impl Malformation {
    /// Applies the edit in place. Returns `true` if the bytes changed.
    pub fn apply(&self, msg: &mut Vec<u8>) -> bool;
}
```

The relevant `Send*` operations (i.e. `SendFundingCreated`, `SendMessage`) get a new `Option<Malformation>` parameter, and the executor applies the malformation if present before sending the corresponding message.

## Choosing Malformations

Malformations are limited to the fields listed above since other fields can already be freely modified by mutators.  In theory, generators *could* apply malformations themselves, but in general I think it's better to leave it up to `OperationParamMutator`.

`OperationParamMutator` can randomly pick an allowlisted field for the current message type and then generate a sequence of bytes to replace that field with -- either random bytes or a repeated byte biased toward the existing `INTERESTING_U8` values.  Smaller field replacements could also draw directly from the `INTERESTING_U*` lists (encoding them in big endian).

## Handling Oracles

Currently the executor maintains various oracle state to validate target responses against.  If we malform the messages we send to the target, our oracle state can drift from the target's state and lead to false positive `Violation`s.

[...]

PR #235, opened 2026-09-06
smite-ir: add malformation parameter to funding_created

NishantBansal2003:
ref: #178 

Depends-on: https://github.com/lnfuzz/smite/pull/212

Add a malformation parameter to `SendFundingCreated` that overwrites an allowlisted field of `funding_created` before sending it to the target. Some fields currently only contain valid values, so this allows the IR to exercise their invalid values as well.

Oracle state is updated based on the message actually sent to the target, after applying the malformation. If the malformed signature cannot be decoded, we send only the invalid signature without marking the state as desynced, since the target should simply reject it as an invalid signature rather than behave differently.

The generator produces only non-malformed `funding_created` messages, `OperationParamMutator` adds malformations using repeated bytes, random bytes, or interesting integers matching the field's exact width.

I verified it, and smite is now able to detect the bug disclosed in [CLN `funding_txid==0` assertion crash](https://delvingbitcoin.org/t/vulnerability-disclosure-assertion-dos-in-core-lightning/2507)


smite/src/bolt.rs
```diff
@@ -138,6 +138,25 @@ pub enum BoltError {
...
+/// into the encoded message.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct MalformableField {
+    /// Field name to overwrite.
+    pub name: &'static str,
+    /// Byte offset in the encoded message, including the message type prefix.
+    pub offset: u16,
+    /// Field length in bytes.
+    pub len: u16,
+}
```
erickcestari, line 158: AFIK we haven't mentioned the `IR` in `bolts` crate until this PR. I think it would be better if  `bolts` had no IR knowledge.

```suggestion
/// Location of a fixed-offset field within an encoded message.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub struct FieldSpan {
    /// Field name as written in the BOLT spec.
    pub name: &'static str,
    /// Byte offset in the encoded message, including the message type prefix.
    pub offset: u16,
    /// Field length in bytes.
    pub len: u16,
}
```

smite/src/bolt/funding_created.rs
```diff
@@ -24,6 +24,31 @@ pub struct FundingCreated {
...
+    pub const SIGNATURE_FIELD: MalformableField = MalformableField {
+        name: "signature",
+        offset: 68,
+        len: 64,
+    };
+    pub const MALFORMABLE_FIELDS: &'static [MalformableField] = &[
+        Self::FUNDING_TXID_FIELD,
+        Self::FUNDING_OUTPUT_INDEX_FIELD,
+        Self::SIGNATURE_FIELD,
+    ];
```
erickcestari, line 50: This should live inside the `malform.rs`, since  the fields that are unreachable depends on the IR's operation, not on the codecs.

```suggestion
```

smite-scenarios/src/executor/tests.rs
```diff
@@ -996,6 +997,105 @@ fn execute_send_funding_created_no_accept_channel() {
...
+    // Everything before the signature is unchanged, so the channel is tracked
+    // under the resolved outpoint with only the signature flagged.
+    assert_eq!(fx.channel_states().len(), 1);
+
+    let state = fx.channel_state(&funding_channel_id());
+    assert_eq!(state.config.funding_outpoint, funding_outpoint());
+    assert!(state.is_funding_outpoint_valid);
+    assert!(state.sent_invalid_signature);
+}
+
```
erickcestari, line 1098: These two returns aren't being covered in the tests if they're also applying correctly the malformation. 

```rust
// Without both the recorded `open_channel` and the peer's `accept_channel`
    // we cannot build the commitment to sign, so fall back to an unsigned
    // `funding_created` and leave `channel_states` untouched.
    let Some(pending) = negotiations.get(&temporary_channel_id) else {
        return Ok(encoded);
    };
    let open_channel = &pending.open_channel;
    let Some(accept_channel) = pending.accept_channel.as_ref() else {
        return Ok(encoded);
    };
```

erickcestari (changes requested):

PR #239, opened 2026-09-08
Channel establishment v2 (dual funding)

erickcestari:
Adds fuzzing support for the BOLT 2 v2 channel establishment flow: open_channel2, interactive transaction construction, v2 commitment signatures and tx_signatures.

Protocol and transaction primitives

- bolt: derive v2 channel ids from the revocation basepoints.
- bitcoin: add raw transaction and sign-only RPCs.
- channel_tx: add interactive transaction construction and compute the initiator's collaborative transaction fee.

IR and generators

- smite-ir: add open_channel2, interactive transaction construction, v2 commitment and tx_signatures operations.
- smite-ir: add the dual funding flow generator.
- smite-ir-mutator: select the generator set per campaign.

Scenarios

- Negotiate option_dual_fund with the peer.
- Give the wallet more than one spendable UTXO.
- Count the interactive tx replies the peer owes and stop receiving once the exchange concludes.

Depends on #222 , #221 and #240

For follow ups:

- Enable dual funding for LDK (ldk-node doesn't expose an API to accept dual funding channels)
- Add oracles
- Add tx_remove_input and  tx_remove_output to interactive-tx
- Add interactive-tx rbf flow
- Add splicing
- Send tx_abort mid-negotiation

PR #240, opened 2026-09-09
Interactive transaction construction for channel establishment v2

erickcestari:
First slice of #239, split out so the dual funding work can be reviewed incrementally. This PR contains only library code in the `smite` crate: no IR, executor, or scenario changes.

- The v2 channel id derivation (BOLT 2: SHA256 of the lexicographically sorted revocation basepoints), plus the temporary id built from the zero basepoint.
- bitcoind RPC wrappers for `createrawtransaction`, `decoderawtransaction`, and `signrawtransactionwithwallet` so the wallet can sign only its own inputs of a collaboratively built funding transaction.
- `channel_tx::interactive_tx`: builds the shared funding transaction from both sides' inputs and outputs, ordered by serial id per BOLT 2, and computes the fee the initiator owes for the common fields.

Depends on https://github.com/lnfuzz/smite/pull/222 and https://github.com/lnfuzz/smite/pull/221

### DIFF
--- a/smite/src/bitcoin.rs
+++ b/smite/src/bitcoin.rs
@@ -6,7 +6,7 @@ use std::path::PathBuf;
 use std::process::Command;
 use std::str::FromStr;
 
-use bitcoin::consensus::encode::serialize_hex;
+use bitcoin::consensus::encode::{deserialize, serialize_hex};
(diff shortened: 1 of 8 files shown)


erickcestari:
I'll close this one since it depends of other 2 PRs. And it's also outdated with the #239

PR #241, opened 2026-09-09
workloads: bump target versions

morehouse:
```
LDK v0.2.4 -> v0.2.6
LND v0.21.1 -> v0.21.3
Eclair v0.14.1 -> v0.14.2
```

Smoke tested with ~1h IR fuzzing campaigns against all three targets.  No issues encountered.


NishantBansal2003 (approved):
Maybe we could bump LDK to 0.2.6, which was just released yesterday? https://crates.io/crates/lightning

PR #242, opened 2026-09-09
smite-scenarios: suppress known oracle violations

morehouse:
I went through all the known issues and added suppressions for those that won't be fixed after bumping the target versions in #241.


NishantBansal2003 (approved):
LGTM!

erickcestari (approved):
LGTM!

PR #243, opened 2026-09-09
Bump dependencies

morehouse:
Update to the latest stable versions for all dependencies.

chacha20poly1305 v0.11 deprecates `Nonce::from_slice`, so we use `nonce_bytes.into()` instead.

Smoke tested by fuzzing LDK with the IR scenario for several hours.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,12 +4,12 @@ version = 4
 
 [[package]]
 name = "aead"
-version = "0.5.2"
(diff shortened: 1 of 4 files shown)


NishantBansal2003 (approved):
LGTM!

> I think this is the right approach to bumping dependencies unless there's a specific feature we need from a minor release -- then we would need to pin the latest version.

Makes sense!

PR #247, opened 2026-09-11
ir: weight generator selection

erickcestari:
Add Generator::weight (default 10) and AnyGenerator::choose, which picks by weight instead of uniformly. The AFL mutator uses it for fresh programs and generator insertion.

This lets future generators such as error, warning or pong be picked less often, since they stall the peer rather than advance the protocol. All existing generators keep the default, so selection is unchanged.

rand needs the alloc feature for choose_weighted.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ warnings = "deny"
 [workspace.dependencies]
 libc = "0.2"
 log = "0.4"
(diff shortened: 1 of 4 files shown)


NishantBansal2003 (approved):
This will be really useful in the future, with https://github.com/lnfuzz/smite/pull/219 as well

PR #250, opened 2026-09-13
smite-ir: implement AnnouncementSignaturesGenerator

devvaansh:
Generates programs that open, fund, and confirm an announced channel, then sign and send `announcement_signatures` for it.

`BuildAnnouncementSignatures` has had no producer since #124, so no generated program could reach it. The message only means anything on a channel the target has already opened with us, so the generator emits the funding flow itself rather than relying on a preceding one: the channel id, the funding keys, and the `short_channel_id` all have to come from the same channel, and `pick_variable` cannot promise that.

The signatures cover the `channel_announcement` body the target rebuilds for itself, so every field has to match what it already knows. That is why the features are empty and the scid is looked up from the confirmed funding output rather than loaded. It is also why the node secret now comes from the context: the Noise static key is our node id on the wire and is what targets verify `node_signature` against, but it was private to `smite-scenarios` and absent from `ProgramContext`, so a `LoadPrivateKey` literal could never produce a valid signature. `LoadLocalNodeSecretFromContext` is not param-mutable, since `OperationParamMutator` can already invalidate a signature through the message body.

`announcement_signatures` is only legal on an announced channel, so `append_open_channel` takes an `announce` flag. Announcing also restricts the channel type, since LDK and LND reject an announced channel that negotiates `option_scid_alias` or `option_zeroconf`; a test pins the excluded list to that property so a new variant fails rather than silently changing what announced channels negotiate. Both existing call sites pass `false`, so the flows they generate are unchanged.

Tested with unit tests covering the ordering of the flow, that the channel is announced with an announceable type and enough confirmations, and that each signed field is tied to the channel being announced: the scid to the funding output, `bitcoin_key_1` to the key that signed `funding_created`, `bitcoin_key_2` to the funding pubkey from `accept_channel`, and the node signature to the context key. Not yet fuzzed against the targets; I'll follow up with coverage once I've run a campaign.

Ref: #71


smite-scenarios/src/executor/tests/harness.rs
```diff
@@ -257,6 +257,9 @@ pub fn sample_context() -> ProgramContext {
         chain_hash: [0xcc; 32],
         block_height: 800_000,
         target_features: vec![],
+        // Deliberately not the real Noise static key, so a test can tell the
+        // executor read the context rather than a constant.
+        local_node_secret: [0xee; 32],
```
erickcestari: nit:

```suggestion
        local_node_secret: [0xee; 32],
```

smite-ir/src/tests.rs
```diff
@@ -1209,11 +1213,10 @@ fn assert_open_channel_params_are_bounded(program: &Program, seed: u64) {
...
         OpenChannelGenerator::MAX_MAX_ACCEPTED_HTLCS,
     );
+    let undefined_bits = channel_flags & !u64::from(OpenChannelGenerator::ANNOUNCE_CHANNEL_FLAG);
     assert_eq!(
-        channel_flags,
-        u64::from(OpenChannelGenerator::CHANNEL_FLAGS),
-        "seed {seed}: channel_flags should be {} but got {channel_flags}",
-        OpenChannelGenerator::CHANNEL_FLAGS,
+        undefined_bits, 0,
+        "seed {seed}: channel_flags should leave undefined bits clear but got {channel_flags}",
```
erickcestari: We could verify if it's correctly announcing or not the channel.

```suggestion
let expected_flags = if announce {
        u64::from(OpenChannelGenerator::ANNOUNCE_CHANNEL_FLAG)
    } else {
        0
    };
    assert_eq!(
        channel_flags, expected_flags,
        "seed {seed}: channel_flags should be {expected_flags} but got {channel_flags}",
    );
```

erickcestari (changes requested):
I don't have a strong opinion on whether we should fold it into `funding_flow`. But if we choose to create a new generator for it, I think having an `append_funding_flow()` would be worthwhile.

PR #252, opened 2026-09-15
smitebot: deduplicate campaign load into CampaignState::load_campaign

Ashish-Kumar-Dash:
Follows [#207](https://github.com/lnfuzz/smite/pull/207#discussion_r3815239494)
`stop`, `status`, and `reproduce` each contained the same 18-line block:resolve `runs_dir`, build the state path, call `CampaignState::load`, and log a not-found hint on failure.

Extract it into `CampaignState::load_campaign(id) -> Option<(Self, PathBuf)>`,which handles all three error paths internally and returns `None` on any failure.


smitebot/src/state.rs
```diff
@@ -174,6 +174,32 @@ impl CampaignState {
...
     }
+
+    /// Loads state for campaign `id` from `~/.smitebot/runs/<id>/state.json`,
+    /// logging errors on failure.
+    ///
+    /// Returns the state and its path (the path is needed by callers that
+    /// mutate and re-save state, e.g. `stop`). Returns `None` if the home
+    /// directory cannot be determined or if the state file is missing or
+    /// corrupt.
+    pub fn load_campaign(id: &str) -> Option<(Self, PathBuf)> {
```
NishantBansal2003: The convention used in `CampaignState` is to return a `StateError` instead of logging the error

NishantBansal2003 (changes requested):
Also needs to be updated: https://github.com/lnfuzz/smite/blob/71696a77b43d76c51db5e5b67f546554300256ae/smitebot/src/commands/corpus.rs#L74

smitebot/src/state.rs
```diff
@@ -174,6 +176,17 @@ impl CampaignState {
...
+    /// Loads state for campaign `id` from `~/.smitebot/runs/<id>/state.json`.
+    ///
+    /// Returns the state and its path (the path is needed by callers that
+    /// mutate and re-save state, e.g. `stop`).
+    pub fn load_campaign(id: &str) -> Result<(Self, PathBuf), StateError> {
+        let runs_dir = Self::runs_dir().ok_or(StateError::HomeDir)?;
+        let state_path = runs_dir.join(id).join("state.json");
+        let state = Self::load(&state_path)?;
+        Ok((state, state_path))
+    }
```
NishantBansal2003: I think a cleaner approach would be to separate `load_campaign` from the `save` logic, so we can define them separately. This looks cleaner as well:

```suggestion
    /// Saves the campaign state to `~/.smitebot/runs/<id>/state.json`.
    pub fn save_campaign(&self) -> Result<(), StateError> {
        let runs_dir = Self::runs_dir().ok_or(StateError::HomeDir)?;
        self.save(&runs_dir.join(&self.id).join("state.json"))
    }

    /// Loads state for campaign `id` from `~/.smitebot/runs/<id>/state.json`.
    pub fn load_campaign(id: &str) -> Result<Self, StateError> {
        let runs_dir = Self::runs_dir().ok_or(StateError::HomeDir)?;
        Self::load(&runs_dir.join(id).join("state.json"))
    }
```

smitebot/src/state.rs
NishantBansal2003: I think with this approach (https://github.com/lnfuzz/smite/pull/252#discussion_r4054489585), we can make both `load` and `save` private and clean up some other parts of the code to use the public functions

NishantBansal2003 (changes requested):

PR #259, opened 2026-09-17
smite: extend accept_channel oracle with sanity and key reuse checks

NishantBansal2003:
ref: https://github.com/lnfuzz/smite/pull/209#issue-5193013315

Some more oracles added in this PR:

#### `open_channel`

- [x] `chain_hash != regtest`
- [x] `channel_reserve_satoshis >= funding_satoshis`
- [x] `dust_limit_satoshis > 10_000` sat
- [x] `to_self_delay > 2016`
- [x] `feerate_per_kw == 0` for non 0FC channels

#### `accept_channel`

- [x]  `dust_limit_satoshis > 10_000` sat
- [x] `htlc_minimum_msat > max_htlc_value_in_flight_msat`
- [x]  `htlc_minimum_msat > open_channel.funding_satoshis`
- [x] `to_self_delay == 0`
- [x] `max_accepted_htlcs == 0`
- [x] pubkey match any pubkeys from `open_channel` and `accept_channel`
- [x] `per_commitment_point` should not be reused

#### Notes

- CLN defaults `to_self_delay` to 6 in `accept_channel` on regtest and 144 on mainnet, so I think we should only check that `to_self_delay != 0`
- CLN defaults the `feerate_per_kw` floor to 1 on regtest and 253 on mainnet when accepting the `open_channel` message, so I think we should only check that `feerate_per_kw != 0`


smite/src/oracles/accept_channel.rs
```diff
@@ -158,13 +194,24 @@ fn verify_accepted_open_channel(
...
+    // Check that feerate_per_kw is 0 when `zero_fee_commitments` is negotiated,
+    // and that it is non-zero otherwise.
+    if channel_type.supports_feature(Features::ZERO_FEE_COMMITMENTS) {
+        if open_channel.feerate_per_kw != 0 {
+            return Err(format!(
+                "zero_fee_commitments requires feerate_per_kw to be 0, but got {}",
+                open_channel.feerate_per_kw,
+            ));
+        }
+    } else if open_channel.feerate_per_kw == 0 {
```
morehouse, line 206: Why is this zero instead of less than 253?

smite/src/oracles/accept_channel.rs
```diff
@@ -270,22 +325,69 @@ fn verify_accept_channel(
...
+    if !accept_channel
+        .pubkeys()
+        .into_iter()
+        .all(|pubkey| pubkeys.insert(pubkey))
+    {
+        return Err("accept_channel reuses a pubkey from the negotiation".to_string());
+    }
+
+    // Check that the first_per_commitment_point is not reused from an earlier negotiation.
+    if per_commitment_points.contains(&accept_channel.first_per_commitment_point) {
```
morehouse, line 385: This check comes before the temp-cid-reuse check, which would make us misclassify a repeated `accept_channel` message as "reused pcp" instead of "reused-temp-cid".

smite-scenarios/src/executor.rs
```diff
@@ -1241,21 +1253,47 @@ fn is_channel_ready_expected(
...
 /// Records a sent `open_channel`, keyed by `temporary_channel_id`, so the
 /// funding flow can build commitments from the values actually put on the wire.
 ///
 /// If a negotiation for the same `temporary_channel_id` is still in progress,
 /// it is left untouched, preserving the first `open_channel`. Once a
 /// `funding_created` has been built, it is overwritten, allowing the
 /// `temporary_channel_id` to be reused for a new negotiation.
+///
+/// Whenever a negotiation is recorded, its `first_per_commitment_point` is
+/// added to `per_commitment_points` as revealed by us.
```
morehouse, line 1293: Shouldn't we record the sent PCP regardless of whether a negotiation exists?  If the target reuses either one its a vuln.

smite-scenarios/src/executor.rs
```diff
@@ -930,12 +935,15 @@ fn build_channel_ready(
...
     // and make us reject a valid received commitment signature as invalid.
+    //
+    // The same point is added to `per_commitment_points` as revealed by us.
     if let Some(state) = channel_states.get_mut(&channel_id)
         && state.commitment.commitment_number == 0
     {
         let next_point = state.next_holder_per_commitment_point_mut();
         if next_point.is_none() {
             *next_point = Some(second_per_commitment_point);
+            per_commitment_points.insert(second_per_commitment_point);
```
morehouse, line 946: Shouldn't we record this PCP regardless of whether a valid `channel_states` entry exists, since either way the target can't reuse it?

smite/src/oracles/accept_channel.rs
```diff
@@ -962,6 +1143,113 @@ mod tests {
...
+        }) {
+            Err(Violation::InvalidAcceptChannel(chan_id, reason)) => {
+                assert_eq!(ac.temporary_channel_id, chan_id);
+                assert!(
+                reason.contains(&format!(
+                    "invalid accept_channel: first_per_commitment_point {} was reused from an earlier negotiation",
+                    ac.first_per_commitment_point,
+                )),
+                "unexpected failure reason: {reason}"
+            );
```
morehouse, line 1247: Indentation is off.

smite-scenarios/src/executor/tests.rs
```diff
@@ -498,18 +514,57 @@ fn execute_recv_accept_channel_rejects_reuse_before_funding() {
...
+    negotiate_channel(&mut b, &second_open_channel);
+
+    // Use the first `accept_channel`'s `first_per_commitment_point` so the second
+    // `accept_channel` is otherwise valid, with only its
+    // `first_per_commitment_point` reused.
+    let earlier_point = sample_accept_channel().first_per_commitment_point;
+    let second_accept_channel = AcceptChannel {
+        temporary_channel_id,
+        ..sample_accept_channel()
+    };
```
morehouse, line 536: Actually it looks like all the points are reused.

smite-scenarios/src/executor/tests/harness.rs
```diff
@@ -296,7 +301,7 @@ pub fn sample_accept_channel() -> AcceptChannel {
         minimum_depth: 6,
         to_self_delay: 144,
         max_accepted_htlcs: 483,
-        funding_pubkey: sample_pubkey(1),
+        funding_pubkey: sample_pubkey(7),
```
morehouse, line 304: `SampleOpenChannel::new` still reuses pubkeys.  Should we make those distinct as well?

smite/src/oracles/accept_channel.rs
```diff
@@ -270,22 +325,69 @@ fn verify_accept_channel(
...
     let htlc_limit = max_accepted_htlcs_limit(&channel_type);
     if accept_channel.max_accepted_htlcs > htlc_limit {
         return Err(format!(
             "max_accepted_htlcs {} exceeds the limit of {htlc_limit}",
             accept_channel.max_accepted_htlcs,
         ));
     }
+    if accept_channel.max_accepted_htlcs == 0 {
+        return Err("max_accepted_htlcs 0 leaves the channel unable to carry HTLCs".to_string());
+    }
```
morehouse, line 338: Nit: `max_accepted_htlcs` is tested against max then min, while `dust_limit_satoshis` is tested against min then max.  Would be nice to be consistent.

smite/src/oracles/accept_channel.rs
```diff
@@ -31,6 +43,8 @@ pub struct AcceptChannelContext<'a> {
     pub negotiation: Option<&'a PendingChannel>,
     /// Features negotiated between the target node and Smite.
     pub negotiated_features: &'a Features,
+    /// Per-commitment points revealed by either us or the target.
+    pub per_commitment_points: &'a HashSet<PublicKey>,
```
morehouse, line 47: We could find more bugs if we expand this map to include all pubkeys that shouldn't be repeated by the target (i.e. open pubkeys, accept pubkeys, PCPs).

To provide better error reports, we could include some data in the map values:

```rust
struct KeyOrigin {
  side: Side,
  channel: TemporaryChannelId,
  field: &'static str,  // "funding_pubkey", "first_per_commitment_point", etc.
}
revealed_pubkeys: HashMap<PublicKey, KeyOrigin>
```

smite/src/oracles/accept_channel.rs
NishantBansal2003: > * CLN defaults the `feerate_per_kw` floor to 1 on regtest and 253 on mainnet when accepting the `open_channel` message, so I think we should only check that `feerate_per_kw != 0`

If we want to check with 253, we need to either override this with a config or patch it

smite-scenarios/src/executor/tests/harness.rs
NishantBansal2003: We can send reused pubkeys, but we shouldn't receive one

PR #260, opened 2026-09-17
smite: add channel_ready oracle

NishantBansal2003:
Depends-on: https://github.com/lnfuzz/smite/pull/212 and https://github.com/lnfuzz/smite/pull/259

Add `ChannelReadyOracle` and consolidate all previously scattered checks relevant to the `channel_ready` negotiation into it, including unknown `channel_id`, cases where the target omits the `short_channel_id alias` required by `option_scid_alias`, and cases where the target reuses a per-commitment point from an earlier negotiation.

#### Deferred oracle checks

- `short_channel_id` alias collisions: BOLT 2 requires aliases not to collide with any of the target's real `short_channel_ids`. Checking this requires fetching the `short_channel_id` for all channels we have with the target over RPC. Bulk fetching adds RPC overhead that reduces fuzzing throughput, while lazy lookups can still miss collisions. Until this can be checked more efficiently, it is not worthwhile for this narrow surface.


erickcestari (approved):
The last 2 commits LGTM!

smite/src/oracles/channel_ready.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    /// Funded channel state for testing.
+    fn channel_state() -> ChannelState {
+        let key = pubkey(1);
+        let party = || ChannelPartyConfig {
+            funding_pubkey: key,
+            payment_basepoint: key,
+            revocation_basepoint: key,
+            delayed_payment_basepoint: key,
```
morehouse, line 114: Should these be distinct keys?

smite/src/oracles/channel_ready.rs
```diff
@@ -0,0 +1,269 @@
...
+            false,
+        )
+    }
+
+    #[track_caller]
+    fn assert_pass(
+        channel_ready: &ChannelReady,
+        channel: Option<&ChannelState>,
+        negotiated_features: &Features,
+        per_commitment_points: &HashSet<PublicKey>,
```
morehouse, line 155: Maybe we should consider adding this same parameter for the `accept_channel` oracle tests.  It might make some of the tests a bit cleaner.

smite/src/oracles/channel_ready.rs
```diff
@@ -0,0 +1,269 @@
...
+/// establishment requirements.
+///
+/// # Deferred oracle checks
+///
+/// - `short_channel_id` alias collisions: BOLT 2 requires aliases not to collide
+///   with any of the target's real `short_channel_ids`. Checking this requires
+///   fetching the `short_channel_id` for all channels we have with the target
+///   over RPC. Bulk fetching adds RPC overhead that reduces fuzzing throughput,
+///   while lazy lookups can still miss collisions. Until this can be checked
+///   more efficiently, it is not worthwhile for this narrow surface.
```
morehouse, line 35: Even without the lookups we could at least check that aliases aren't reused by keeping an in-memory set like we do for PCPs.

smite/src/oracles/channel_ready.rs
```diff
@@ -0,0 +1,269 @@
...
+                context.channel_ready.channel_id,
+                "unknown channel_id: no channel was funded for it".to_string(),
+            ));
+        }
+
+        // Check that an alias is set when option_scid_alias was negotiated.
+        if context
+            .negotiated_features
+            .supports_feature(Features::OPTION_SCID_ALIAS)
+            && context.channel_ready.tlvs.short_channel_id.is_none()
```
morehouse, line 52: Is it the alias feature from `init` that matters, or the one from `channel_type`?  BOLT 2 seems to say both at different places:

- https://github.com/lightning/bolts/blob/152897261850d93c4f4597f39cf22d7d22d6ede6/02-peer-protocol.md?plain=1#L1077
- https://github.com/lightning/bolts/blob/152897261850d93c4f4597f39cf22d7d22d6ede6/02-peer-protocol.md?plain=1#L1089

smite/src/oracles/channel_ready.rs
```diff
@@ -0,0 +1,269 @@
...
+        {
+            return Err(Violation::InvalidChannelReady(
+                context.channel_ready.channel_id,
+                format!(
+                    "second_per_commitment_point {} was reused from an earlier negotiation",
+                    context.channel_ready.second_per_commitment_point,
+                ),
+            ));
+        }
+
```
morehouse, line 74: Should we also check that `minimum_depth` was satisfied?

I know it's unlikely to even run this oracle if the depth isn't reached, since we turn `RecvChannelReady` into a no-op.  But I think it's still possible if there's multiple negotiations in flight.

Also it will be easier to actually receive early `channel_ready`s if we later implement the "implicit receive" logic from https://github.com/lnfuzz/smite/issues/111#issuecomment-4708842848.  Then we could be attempting to receive a different message and get an unexpected `channel_ready` instead, allowing us to run this oracle.

smite/src/oracles/channel_ready.rs
```diff
@@ -0,0 +1,269 @@
...
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
morehouse, line 80: Might be worth some executor tests as well like we have for the `accept_channel` oracle checks.

smite/src/oracles/channel_ready.rs
```diff
@@ -0,0 +1,269 @@
...
+                context.channel_ready.channel_id,
+                "option_scid_alias negotiated but short_channel_id alias is missing".to_string(),
+            ));
+        }
+
+        // Check that the second_per_commitment_point is not reused from an
+        // earlier negotiation.
+        if context
+            .per_commitment_points
+            .contains(&context.channel_ready.second_per_commitment_point)
```
morehouse, line 64: The spec does explicitly allow sending multiple `channel_ready` messages with the same PCPs:

https://github.com/lightning/bolts/blob/152897261850d93c4f4597f39cf22d7d22d6ede6/02-peer-protocol.md?plain=1#L1091

I'm not sure if anyone implements that, but I think we should probably be more careful with this check.  Probably we should allow reuse if the first use was for the same channel (i.e. `channel.next_counterparty_per_commitment_point` exists and matches), otherwise report a violation.

morehouse (changes requested):

PR #261, opened 2026-09-18
workloads/lnd: map coverage counters directly onto AFL's map

erickcestari:
Remap Go's libfuzzer counter section onto the AFL shared memory instead of copying it on every sync. Startup coverage no longer pollutes every map, and crashing or timed-out inputs now report coverage.

The trigger/ack pipes stay as a liveness handshake: try_wait still sees a just-crashed LND as running.

I didn't yet run an evaluation to see the coverage increasing, but this should align the coverage report with the others implementation. Where the startup coverage don't pollute every map, since each input will be started from a zeroed map.


erickcestari:
### Target: lnd

image

#### Distribution Comparisons

|:---:|:---:|
| image | image |

